PR #49, opened 2015-01-06
standardize runit service names. if bin/launch is an executable, the run...

mpuncel:
...it service will now be called `<podname>__<launchablename>__launch` whereas it used to be called `<podname>__<launchablename>`

@anthonybishopric @petertseng

### DIFF
--- a/pkg/pods/hoist_launchable.go
+++ b/pkg/pods/hoist_launchable.go
@@ -223,7 +223,7 @@ func (hoistLaunchable *HoistLaunchable) Executables(serviceBuilder *runit.Servic
 	// ideally a launchable will have just one launch script someday (can't be
 	// a dir)
 	if !(binLaunchInfo.IsDir()) {
-		serviceName := hoistLaunchable.Id // use the ID of the launchable as its unique Runit service name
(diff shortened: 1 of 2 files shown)


petertseng:
ogad the formatting messed the thing all up

pkg/pods/launch_script_only_test_hoist_launchable/current
```diff
@@ -0,0 +1 @@
+/Users/mpuncel/Development/go/src/github.com/square/p2/pkg/pods/launch_script_only_test_hoist_launchable/installs/testLaunchable
```
anthonybishopric: rm, otherewise lgtm

petertseng: are you ok with this having paths specific to your machine

anthonybishopric: just better to not pollute if possible

mpuncel: i'm not sure what those current symlinks are about, i don't remember putting them there, ill make sure tests still pass without them

mpuncel: oh they get created by the test, i'll just not check it in

PR #50, opened 2015-01-06
standardize bin names

anthonybishopric:
Specifically renames `bootstrap` to `p2-bootstrap` and similarly for `preparer`.

@mpuncel @petertseng


mpuncel:
:+1:

PR #51, opened 2015-01-07
write all servicebuilder information for a pod in one file. This meant m...

mpuncel:
...oving funcionality for writing servicebuilder files out of HoistLaunchable and into Pod

@anthonybishopric @petertseng

### DIFF
--- a/pkg/pods/hoist_launchable.go
+++ b/pkg/pods/hoist_launchable.go
@@ -130,12 +130,6 @@ func (hoistLaunchable *HoistLaunchable) Stop(serviceBuilder *runit.ServiceBuilde
 // All services will attempt to be started.
 func (hoistLaunchable *HoistLaunchable) Start(serviceBuilder *runit.ServiceBuilder, sv *runit.SV) error {
 
-	// if the service is new, building the runit services also starts them, making the sv start superfluous but harmless
-	err := hoistLaunchable.BuildRunitServices(serviceBuilder)
-	if err != nil {
-		return err
-	}
-
 	executables, err := hoistLaunchable.Executables(serviceBuilder)
 	if err != nil {
 		return err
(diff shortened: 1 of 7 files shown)


pkg/pods/pod.go
```diff
@@ -27,20 +27,19 @@ func PodPath(manifestId string) string {
 }
 
 type Pod struct {
+	Id             string
```
mpuncel: This seemed like the lesser of the evils between extracting Id from the path or from the manifest

pkg/pods/hoist_launchable.go
```diff
@@ -130,19 +130,13 @@ func (hoistLaunchable *HoistLaunchable) Stop(serviceBuilder *runit.ServiceBuilde
...
-	}
-
 	executables, err := hoistLaunchable.Executables(serviceBuilder)
 	if err != nil {
 		return err
 	}
 
 	for _, executable := range executables {
-		_, err := sv.Restart(&executable.Service)
+		_, err := sv.Start(&executable.Service)
```
anthonybishopric: I feel like I've changed this line back and forth a bunch of times. Are we sure that we want 'Start'?

mpuncel: why would restart be preferable? the service should always be stopped already at this point

anthonybishopric: We don't stop the service before this point. We rely on restart to atomically stop and start - this is required to do the self-hosted preparer. Otherwise, you'd run `sv stop` on the preparer, which would kill the launch process.

mpuncel: hmm okay. So if we remove a launchable from a pod we never explicitly stop it, it gets stop by rewriting the servicebuilder file and running servicebuilder.

On a slightly unrelated note, this seems weird for the preparer because even calling restart we're basically halting execution of any code after this line

PR #53, opened 2015-01-12
Add service registration in pkg/intent

tummychow:
Now when you intent.SetPod, p2 will also register a Consul service. The ID of the service is that of the manifest itself. p2 also associates a health check with the new service, which is basically a curl on its `/_status` endpoint.

I'm going to do some integration tests by hand because I want to see if this code actually works, but look it over and let me know what's missing.

### DIFF
--- a/pkg/intent/intent.go
+++ b/pkg/intent/intent.go
@@ -120,10 +120,10 @@ func (i *Store) SetPod(node string, manifest pods.PodManifest) (time.Duration, e
 	podService := &consulapi.AgentServiceRegistration{
 		Name: manifest.ID(),
 	}
-	if manifest.Port != 0 {
+	if manifest.StatusPort != 0 {
 		podService.Check = &consulapi.AgentServiceCheck{
 			// prints the HTTP response on stderr, while exiting 0 if the status code is 200
-			Script:   fmt.Sprintf(`if [[ $(curl 0.0.0.0:%v/_status -s -o /dev/stderr -w "%%{http_code}") == "200" ]] ; then exit 0 ; else exit 2; fi`, manifest.Port),
+			Script:   fmt.Sprintf(`if [[ $(curl 0.0.0.0:%v/_status -s -o /dev/stderr -w "%%{http_code}") == "200" ]] ; then exit 0 ; else exit 2; fi`, manifest.StatusPort),
 			Interval: "5s", // magic number alert
 		}
 	}
--- a/pkg/intent/intent.go
+++ b/pkg/intent/intent.go
@@ -18,6 +18,7 @@ const INTENT_TREE string = "/intent"
 
 type ConsulClient interface {
 	KV() *consulapi.KV
+	Agent() *consulapi.Agent
 }
 
 type Options struct {
--- a/pkg/intent/intent.go
+++ b/pkg/intent/intent.go
@@ -116,11 +116,24 @@ func (i *Store) SetPod(node string, manifest pods.PodManifest) (time.Duration, e
 	if err != nil {
 		return 0, err
 	}
+
+	podService := &consulapi.AgentServiceRegistration{
+		Name: manifest.ID(),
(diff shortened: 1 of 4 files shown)


pkg/pods/manifest_parser.go
```diff
@@ -27,6 +27,7 @@ type PodManifest struct {
 	Id                string                      `yaml:"id"` // public for yaml marshaling access. Use ID() instead.
 	LaunchableStanzas map[string]LaunchableStanza `yaml:"launchables"`
 	Config            map[string]interface{}      `yaml:"config"`
+	Port              int                         `yaml:"port"`
```
anthonybishopric: Lets name this `StatusPort` to clarify that it's not the only port that an app can expose.

tummychow:
Renamed Port as suggested. I'll probably split that commit up and squash it back into the other three before merging.

pkg/intent/intent.go
```diff
@@ -115,11 +116,26 @@ func (i *Store) SetPod(node string, manifest pods.PodManifest) (time.Duration, e
...
 		return 0, err
 	}
+
+	podService := &consulapi.AgentServiceRegistration{
+		Name: manifest.ID(),
+	}
+	if manifest.StatusPort != 0 {
+		podService.Check = &consulapi.AgentServiceCheck{
+			// prints the HTTP response on stderr, while exiting 0 if the status code is 200
+			Script:   fmt.Sprintf(`if [[ $(curl 0.0.0.0:%v/_status -s -o /dev/stderr -w "%%{http_code}") == "200" ]] ; then exit 0 ; else exit 2; fi`, manifest.StatusPort),
```
petertseng: Sure it should be 0.0.0.0 ? Which machine runs the health check? I recall we wanted to assume the checks run on machines other than the app hosts.

Basically asking "should a VIP go here instead"

petertseng: no vip (wouldn't check the right host). interesting question of what goes here

tummychow: I see one example in consul docs where they are referring to localhost in a
service check: http://www.consul.io/intro/getting-started/checks.html

But I see nothing about who's actually running the check. I'm going to
spelunk in consul's code a bit, I don't think there's any documentation for
this.

On Mon, Jan 12, 2015 at 4:00 PM, petertseng <notifications@github.com>
wrote:

> In pkg/intent/intent.go
> <https://github.com/square/p2/pull/53#discussion-diff-22833893>:
>
> > @@ -115,11 +116,26 @@ func (i *Store) SetPod(node string, manifest pods.PodManifest) (time.Duration, e
> >  	if err != nil {
> >  		return 0, err
> >  	}
> > +
> > +	podService := &consulapi.AgentServiceRegistration{
> > +		Name: manifest.ID(),
> > +	}
> > +	if manifest.StatusPort != 0 {
> > +		podService.Check = &consulapi.AgentServiceCheck{
> > +			// prints the HTTP response on stderr, while exiting 0 if the status code is 200
> > +			Script:   fmt.Sprintf(`if [[ $(curl 0.0.0.0:%v/_status -s -o /dev/stderr -w "%%{http_code}") == "200" ]] ; then exit 0 ; else exit 2; fi`, manifest.StatusPort),
>
> no vip (wouldn't check the right host). interesting question of what goes
> here
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/square/p2/pull/53/files#r22833893>.
>



-- 
Stephen

tummychow:
Yeah, this is no good. Service registration is bound to the agent that services the request. The intent store doesn't actually know which agent it's talking to (out of the various p2 nodes in the cluster), so it can't possibly be sure what IP the target service will end up running on.

It looks like we have to move this code into the preparer. The preparer is always talking to a consul agent (whether server or client) on localhost, so it can guarantee that the node answering its service request is also the node that can execute checks against localhost. I'll have another PR implementing this once I sort out what the API should look like.

PR #54, opened 2015-01-13
Add the reality store

petertseng:
This is pretty much just copy the intent store, change its name, remove
the watch, add a lookup. I lose some DRY points for doing ths.


tummychow:
LGTM. Yeah it's kinda un-DRY when you put it next to the intent store, but I can't think of a way to combine those two into one without confusing people.

PR #55, opened 2015-01-13
Remove leading slash on intent path

tummychow:
moral of the story: path.Join is correct, string concatenation is incorrect https://github.com/armon/consul-api/blob/master/kv.go#L171

I looked at all the references to INTENT_TREE and they all look safe to me with the leading slash removed, but I've already screwed up twice on this change, so some extra eyes would be welcome.


anthonybishopric:
lgtm

PR #56, opened 2015-01-13
Remove leading slash on reality path

petertseng:
Because consul API is using concat instead of path.Join


anthonybishopric:
oh yeah

PR #57, opened 2015-01-13
Add service registration to preparer

tummychow:
Mostly need some feedback on the placement of this code right now. Tests are broken on this branch.

e: to be exact, the last commit is the one that bothers me; I think the other three are harmless but they can be tweaked too if needed.

### DIFF
--- a/pkg/intent/intent.go
+++ b/pkg/intent/intent.go
@@ -18,6 +18,7 @@ const INTENT_TREE string = "intent"
 
 type ConsulClient interface {
 	KV() *consulapi.KV
+	Agent() *consulapi.Agent
 }
 
 type Options struct {
--- a/pkg/intent/intent.go
+++ b/pkg/intent/intent.go
@@ -127,3 +127,20 @@ func (i *Store) SetPod(node string, manifest pods.PodManifest) (time.Duration, e
 	}
 	return writeMeta.RequestTime, err
 }
+
+func (i *Store) RegisterPodService(manifest pods.PodManifest) error {
+	podService := &consulapi.AgentServiceRegistration{
+		Name: manifest.ID(),
+	}
+
+	if manifest.StatusPort != 0 {
+		podService.Port = manifest.StatusPort
+		podService.Check = &consulapi.AgentServiceCheck{
+			// prints the HTTP response on stderr, while exiting 0 if the status code is 200
+			Script:   fmt.Sprintf(`if [[ $(curl 0.0.0.0:%v/_status -s -o /dev/stderr -w "%%{http_code}") == "200" ]] ; then exit 0 ; else exit 2; fi`, manifest.StatusPort),
+			Interval: "5s", // magic number alert
+		}
+	}
+
+	return i.client.Agent().ServiceRegister(podService)
+}
--- a/pkg/preparer/orchestrate.go
+++ b/pkg/preparer/orchestrate.go
@@ -178,6 +178,7 @@ func (p *Preparer) installAndLaunchPod(newManifest *pods.PodManifest, pod Pod, l
 		err := p.iStore.RegisterPodService(*newManifest)
 		if err != nil {
 			logger.WithField("err", err).Errorln("Service registration failed")
+			return false
 		}
 		ok, err := pod.Launch(newManifest)
 		if err != nil {
--- a/pkg/preparer/orchestrate.go
+++ b/pkg/preparer/orchestrate.go
@@ -175,6 +175,10 @@ func (p *Preparer) installAndLaunchPod(newManifest *pods.PodManifest, pod Pod, l
 	}
 
 	if newOrDifferent || problemReadingCurrentManifest {
+		err := p.iStore.RegisterPodService(*newManifest)
+		if err != nil {
+			logger.WithField("err", err).Errorln("Service registration failed")
+		}
 		ok, err := pod.Launch(newManifest)
 		if err != nil {
 			logger.WithFields(logrus.Fields{
--- a/pkg/intent/intent_test.go
+++ b/pkg/intent/intent_test.go
@@ -18,6 +18,11 @@ func (f fakeClient) KV() *consulapi.KV {
 	return client.KV()
 }
 
+func (f fakeClient) Agent() *consulapi.Agent {
+	client, _ := consulapi.NewClient(consulapi.DefaultConfig())
+	return client.Agent()
+}
+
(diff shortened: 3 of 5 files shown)


pkg/preparer/orchestrate.go
```diff
@@ -68,7 +68,7 @@ func WatchForPodManifestsForNode(nodeName string, consulAddress string, hooksDir
 
 // no return value, no output channels. This should do everything it needs to do
 // without outside intervention (other than being signalled to quit)
-func handlePods(hooks *pods.HookDir, podChan <-chan pods.PodManifest, quit <-chan struct{}, logger logging.Logger) {
+func handlePods(hooks *pods.HookDir, podChan <-chan pods.PodManifest, quit <-chan struct{}, logger logging.Logger, store *intent.Store) {
```
anthonybishopric: this arg list is getting out of hand pretty fast. next opportunity lets change this to a function on a struct with these things instead.

pkg/preparer/orchestrate.go
```diff
@@ -151,6 +151,12 @@ func installAndLaunchPod(newManifest *pods.PodManifest, pod Pod, logger logging.
 	}
 
 	if newOrDifferent || problemReadingCurrentManifest {
+		err := store.RegisterPodService(*newManifest)
+		if err != nil {
+			logger.WithFields(logrus.Fields{
```
anthonybishopric: if you're just sending the error, use `WithField('err', err).Errorln`

pkg/intent/intent.go
```diff
@@ -126,3 +127,16 @@ func (i *Store) SetPod(node string, manifest pods.PodManifest) (time.Duration, e
 	}
 	return writeMeta.RequestTime, err
 }
+
+func (i *Store) RegisterPodService(manifest pods.PodManifest) error {
+	podService := &consulapi.AgentServiceRegistration{
```
anthonybishopric: This this registration idempotent?

pkg/preparer/orchestrate.go
tummychow: I was actually considering that refactor as part of this commit. It'll also be useful for the reality store. @petertseng thoughts? I don't want to block your work on this refactor so we could delay it until then.

pkg/preparer/orchestrate.go
tummychow: Will fix

pkg/intent/intent.go
tummychow: It should be. If you register two services with the same ID on the same node, consul overwrites the older one with the newer one (which is desirable behavior for us, I think). Unfortunately this behavior is not documented, but I found it somewhere in their codebase. Let me find a link.

Repeatedly registering a service over and over might have an effect on the timing of its health checks (ie maybe the health check gets fired every time you register) but I don't think that's a significant concern.

anthonybishopric: As long as it's not registering duplicate checks we're probably good.

tummychow:
Oh, and don't forget about the broken tests... installAndLaunchPod depends on a consul instance now (and it didn't before, which is why it was so easy to test). I haven't fixed them yet. pkg/orchestrate depends directly on pkg/intent.Store right now, so unbreaking the tests will probably require more refactoring.

tummychow:
This will be rebased onto #58 (once that's merged), which makes the argument passing much nicer.

tummychow:
Rebase onto #58 complete. Tests will fail due to nil pointer dereference. We can either fix that here, or wait for the reality store update PR to be merged (then fix them in a third PR).

pkg/preparer/orchestrate.go
```diff
@@ -175,6 +175,10 @@ func (p *Preparer) installAndLaunchPod(newManifest *pods.PodManifest, pod Pod, l
 	}
 
 	if newOrDifferent || problemReadingCurrentManifest {
+		err := p.iStore.RegisterPodService(*newManifest)
+		if err != nil {
+			logger.WithField("err", err).Errorln("Service registration failed")
```
anthonybishopric: What are the error cases when the intent store registration fails? Just want to hear the thinking on not bailing out / continuing onto `Launch`

pkg/preparer/orchestrate.go
```diff
@@ -175,6 +175,10 @@ func (p *Preparer) installAndLaunchPod(newManifest *pods.PodManifest, pod Pod, l
...
+		}
 		ok, err := pod.Launch(newManifest)
```
anthonybishopric: ... especially since we won't see the err if it gets masked by a successful launch here.

tummychow:
Errors in service registration should be strictly http related (consul host being unreachable for whatever reason). From the perspective of the consul server, I don't think service registration should fail unless the JSON body is malformed, but the API package makes that impossible.

anthonybishopric:
If we can't register a service check on that host even after it launches it will have implications for the replicator (because the host will never be either healthy or current). We should try to make the preparer aggressive in trying to recover from this failure somehow, especially if it's a transient issue.

tummychow:
Do you want to add some retry logic, maybe? That's the only failure recovery mode I can think of. It would be enough to deal with a transient connectivity error, but if Consul was returning a persistent error (no documentation indicating what these might be), the retries would be pointless.

The other option is to bail on the launch if the service registration fails, ie successful service registration is considered an inextricable part of pod preparation.

anthonybishopric:
The logic of the preparer is to retry as long as `ok == false` - so, it may not be transient in the sense that it disappears on its own (but that can absolutely happen) but it could be solved by human intervention. For example - if the machine reboots and Consul, for whatever reason, can't be started by runit, but human comes along and starts it, we don't want that human to now also have to restart any pending installations. It improves the resilience of the system.

So yeah, I'd call it what you called it - an inextricable part of pod preparation. It really has to happen for anything else to work, unlike our policy towards launchables which are not controlled by us.

tummychow:
Okay, so here's what I'm thinking: if we try to register the service (in `installAndLaunchPod`) and it fails, we should return the error and not attempt to `pod.Launch`.

Personally I like that approach, because it's really really simple and prevents pods from being "partially prepared" (ie installed on the system but not registered as a service).

anthonybishopric:
Yep sounds good to me.

What's up with the build?

tummychow:
Build failures are due to that nil pointer dereference that we talked about on the refactor PR #58. Once this PR and #59 are merged, I'll work on fixing all the tests together (because they're both failing for the same reason).

Last commit will be up in a moment

PR #58, opened 2015-01-14
Refactor shared arguments of preparer into struct

tummychow:
I was working pretty fast on this one so there might be stupid mistakes. I'll rebase #53 onto this once it's merged.

### DIFF
--- a/pkg/preparer/orchestrate_test.go
+++ b/pkg/preparer/orchestrate_test.go
@@ -64,7 +64,9 @@ func TestPreparerLaunchesNewPodsThatArentInstalledYet(t *testing.T) {
 		launchSuccess: true,
 	}
 	newManifest := testManifest(t)
-	success := installAndLaunchPod(newManifest, testPod, logging.DefaultLogger)
+
+	var p *Preparer
+	success := p.installAndLaunchPod(newManifest, testPod, logging.DefaultLogger)
 
 	Assert(t).IsTrue(success, "should have succeeded")
 	Assert(t).IsTrue(testPod.launched, "Should have launched")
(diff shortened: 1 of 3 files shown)


pkg/preparer/orchestrate_test.go
```diff
@@ -82,7 +84,9 @@ func TestPreparerLaunchesPodsThatHaveDifferentSHAs(t *testing.T) {
 		currentManifest: existing,
 	}
 	newManifest := testManifest(t)
-	success := installAndLaunchPod(newManifest, testPod, logging.DefaultLogger)
+
+	var p *Preparer
```
anthonybishopric: How doesn't this cause a panic? Declaring the pointer without initializing it should result in a dereferencing error or am I missing something

tummychow: It's never dereferenced anywhere in installAndLaunchPod. Once @petertseng and I add our code, we'll need to actually declare something there.

anthonybishopric: Ok I understand - it just happens to work because the function is on `*Preparer` not `Preparer`, and nothing is dereferenced about that pointer specifically. Seems like a timebomb for tests.

tummychow: Agreed, it is a time bomb. We should be able to fix it sometime tomorrow, though.

After the reality store update and service registration get merged, the tests will fail regardless of the null pointer deref (because the requests to Consul will fail in installAndLaunchPod). I can fix the tests then. I mainly want to get the refactor into the repo quicker so that we can rebase our code onto it.

anthonybishopric:
Looks good to me, lets fix the pointer issue in a follow up

tummychow:
Tested on a universe node and the preparer didn't explode or anything. Merging

PR #59, opened 2015-01-14
orchestrate: SetPod in reality after launching

petertseng:


petertseng:
too simple?!

anthonybishopric:
lgtm

tummychow:
Code lgtm but the corresponding GetPod also needs to be implemented (it will probably replace the pod.CurrentManifest on line 154). Might as well do that in this PR

tummychow:
Discovered bug while deploying to universe cluster. In `reality.Pod`, a 404 will cause a nil pointer dereference. (this bug was actually introduced in #54 but the code path wasn't being exercised until now)

I originally guessed that the consul-api package would propagate a 404 back to the caller if the key didn't exist. Unfortunately I was [wrong](https://github.com/armon/consul-api/blob/master/kv.go#L42). If the key doesn't exist, then the returned KVPair and error are *both* nil. Then we try to dereference the nil pointer on line 60 (`kvPair.Value`) and that causes an explosion.

Correct behavior would be to detect `kvPair == nil` and return a nil pod manifest.

pkg/reality/reality.go
```diff
@@ -57,6 +57,9 @@ func (r *Store) Pod(node, podId string) (*pods.PodManifest, error) {
 	if err != nil {
 		return nil, err
 	}
+	if kvPair == nil {
+		return nil, pods.NoCurrentManifest
```
anthonybishopric: :+1:

tummychow:
lgtm, also tested on universe and verified

anthonybishopric:
lg

PR #60, opened 2015-01-14
Fix tests in pkg/preparer

tummychow:
We should probably wait to merge this until after #59 (I'll rebase it back onto master once that's done), but let me know what you think of the code so far.

This is also an opportunity to investigate writing more complex tests for pkg/preparer (eg testing handlePods or WatchForPodManifestsForNode).

### DIFF
--- a/pkg/preparer/orchestrate.go
+++ b/pkg/preparer/orchestrate.go
@@ -14,7 +14,6 @@ import (
 type Pod interface {
 	Launch(*pods.PodManifest) (bool, error)
 	Install(*pods.PodManifest) error
-	CurrentManifest() (*pods.PodManifest, error)
 	Halt() (bool, error)
 }
 
--- a/pkg/preparer/orchestrate.go
+++ b/pkg/preparer/orchestrate.go
@@ -18,10 +18,19 @@ type Pod interface {
 	Halt() (bool, error)
 }
 
+type RealityStore interface {
+	Pod(string, string) (*pods.PodManifest, error)
+	SetPod(string, pods.PodManifest) (time.Duration, error)
+}
+type IntentStore interface {
+	RegisterPodService(pods.PodManifest) error
+	WatchPods(string, <-chan struct{}, chan<- error, chan<- pods.PodManifest) error
+}
+
 type Preparer struct {
 	node   string
-	iStore *intent.Store
-	rStore *reality.Store
+	iStore IntentStore
+	rStore RealityStore
 	hooks  *pods.HookDir
 	Logger logging.Logger
 }
--- a/pkg/preparer/orchestrate_test.go
+++ b/pkg/preparer/orchestrate_test.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"runtime"
 	"testing"
+	"time"
 
 	. "github.com/anthonybishopric/gotcha"
 	"github.com/square/p2/pkg/logging"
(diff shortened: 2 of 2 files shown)


pkg/preparer/orchestrate.go
```diff
@@ -18,10 +18,19 @@ type Pod interface {
...
+type IntentStore interface {
+	RegisterPodService(pods.PodManifest) error
+	WatchPods(string, <-chan struct{}, chan<- error, chan<- pods.PodManifest) error
+}
+
 type Preparer struct {
 	node   string
-	iStore *intent.Store
-	rStore *reality.Store
+	IStore IntentStore
```
anthonybishopric: You don't need to make these public to make them accessible in `orchestrate_test` - was that your goal?

tummychow: Oh yeah I'll fix that. At home my tests usually use the `foo_test` package so they can only access exported identifiers.

On a related note, any opinions about which fields should be public and which ones shouldn't? I wasn't really sure when I made the refactor, so I leaned towards not exporting them.

anthonybishopric: Defaulting to "don't export" is the right approach. Basically if you don't have a reason to export it, don't export it.

anthonybishopric:
looks good

pkg/preparer/orchestrate_test.go
```diff
@@ -59,13 +60,34 @@ func testManifest(t *testing.T) *pods.PodManifest {
...
+}
+
 func TestPreparerLaunchesNewPodsThatArentInstalledYet(t *testing.T) {
 	testPod := &TestPod{
 		launchSuccess: true,
 	}
 	newManifest := testManifest(t)
 
-	var p *Preparer
+	p, _ := New("hostname", "0.0.0.0", "/hooks/dir", logging.DefaultLogger)
```
petertseng: This and the three subsequent lines keep getting repeated. Worth putting a small function in the test file that returns a test preparer?

tummychow: will fix

pkg/preparer/orchestrate_test.go
```diff
@@ -113,25 +135,10 @@ func TestPreparerWillNotLaunchIfSHAIsTheSame(t *testing.T) {
...
-	var p *Preparer
+	p := testPreparer(testManifest)
 	success := p.installAndLaunchPod(testManifest, testPod, logging.DefaultLogger)
 
 	Assert(t).IsTrue(success, "Should have been a success to prevent retries")
 	Assert(t).IsFalse(testPod.launched, "Should not have attempted to launch")
 	Assert(t).IsTrue(testPod.installed, "Should have installed")
 }
-
-func TestInstallReturnsFalseIfManifestErrsOnRead(t *testing.T) {
```
petertseng: Hmm, we could totally keep this -- you see how you had `func (f *FakeStore) Pod(string, string) (*pods.PodManifest, error)` return a manifest of your choosing. So you could make it return an error of your choosing as well.

tummychow: good point - I'll amend the commit to change the test instead of deleting it.

anthonybishopric: Agreed, that this is a good test to keep - this ensures that the reality store being corrupt for whatever reason doesn't stop a scheduled install.

pkg/preparer/orchestrate_test.go
```diff
@@ -121,14 +150,15 @@ func TestPreparerWillNotLaunchIfSHAIsTheSame(t *testing.T) {
...
-		currentManifestError: fmt.Errorf("it erred"),
-		launchSuccess:        true,
+		launchSuccess: true,
 	}
 
-	var p *Preparer
+	p := testPreparer(&FakeStore{
+		currentManifestError: fmt.Errorf("it erred"),
+	})
 	success := p.installAndLaunchPod(testManifest, testPod, logging.DefaultLogger)
```
anthonybishopric: :+1:

petertseng:
rad

PR #61, opened 2015-01-14
Block on net.Listen in logger socket test

tummychow:
Sometimes this test fails because the scheduler doesn't execute the
net.Listen right away. Then the logger fires the hook expecting the
socket to be there, and it isn't yet (because nobody is listening).

To eliminate these sporadic errors, we can just move the Listen call
out of the goroutine, so that the test is blocked until the socket is
ready.


anthonybishopric:
:+1:

PR #62, opened 2015-01-15
Introduces the hooks listener

anthonybishopric:
This is the draft of the thing that will listen to Consul for hooks as they are added to the system.

More tests are needed / integration with preparer is coming soon


bin/p2-bootstrap/bootstrap.go
```diff
@@ -37,6 +37,7 @@ func main() {
 			log.Fatalf("Could not get consul manifest: %s", err)
 		}
 		consulPod := pods.PodFromManifestId(consulManifest.ID())
+		consulPod.RunAs = "root"
```
anthonybishopric: I had to make the `RunAs` for a pod configurable, which had the nice effect of removing the hardcoded pod names that were present in the pod package.

bin/p2-watch/main.go
```diff
@@ -35,13 +34,13 @@ func main() {
 
 	quit := make(chan struct{})
 	errChan := make(chan error)
-	podCh := make(chan pods.PodManifest)
+	podCh := make(chan intent.ManifestResult)
```
anthonybishopric: I switched this to a wrapper object so I could see what the path in Consul was when we read the pod.

pkg/hooks/listener.go
```diff
@@ -0,0 +1,182 @@
...
+	if err != nil {
+		return util.Errorf("Couldn't make event exec dir %s", eventExecDir)
+	}
+	launchables, err := hookPod.GetLaunchables(manifest)
+	if err != nil {
+		return err
+	}
+
+	// First remove any pre-existing hooks for that pod.
+	podHookPattern := path.Join(eventExecDir, fmt.Sprintf("%s__*", hookPod.Id))
```
anthonybishopric: Here I'm kind of nastily globbing every symlink from the previous version of the pod and removing them. Notice two bad things - one that there isn't a better grouping mechanism (like a sub folder) for hooks created by a particular pod - and two, that I just happen to know that `{podId}__*` is the wildcard for the name of hooks assigned to a pod. 

To do: refactor the hooks library to execute hooks in directories recursively.

tummychow:
Other than those comments, lgtm, but that commit was pretty big so I can't be sure I didn't miss anything

PR #63, opened 2015-01-15
only run hooks if sha is different

anthonybishopric:


pkg/hooks/hooks.go
```diff
@@ -0,0 +1,101 @@
+package hooks
```
anthonybishopric: This is just a move from the pods package

pkg/preparer/orchestrate_test.go
```diff
@@ -127,28 +169,32 @@ func TestPreparerFailsIfInstallFails(t *testing.T) {
...
+	p, hooks := testPreparer(&FakeStore{
 		currentManifest: testManifest,
 	})
 	success := p.installAndLaunchPod(testManifest, testPod, logging.DefaultLogger)
 
 	Assert(t).IsTrue(success, "Should have been a success to prevent retries")
+	Assert(t).IsFalse(hooks.ranBeforeInstall, "Should not have run hooks prior to install")
+	Assert(t).IsFalse(testPod.installed, "Should not have installed")
 	Assert(t).IsFalse(testPod.launched, "Should not have attempted to launch")
-	Assert(t).IsTrue(testPod.installed, "Should have installed")
```
anthonybishopric: This is the only test that changed semantically - Installs should not happen if the SHA is different.

petertseng: I think your code is correct but this comment is not correct. You mean to say "Installs should not happen if the SHA is the same" or "Installs should only happen if the SHA is different", instead of what you actually said.

pkg/preparer/orchestrate_test.go
```diff
@@ -127,28 +169,32 @@ func TestPreparerFailsIfInstallFails(t *testing.T) {
...
+	Assert(t).IsFalse(hooks.ranAfterLaunch, "Should not have launched")
```
petertseng: maybe "Should not have run after_launch hooks"?

petertseng:
other than comments I have no problems

pkg/preparer/orchestrate_test.go
anthonybishopric: yep good catch

pkg/preparer/orchestrate_test.go
anthonybishopric: Wow yeah. It is opposite day apparently.

tummychow:
Code lgtm. Now that we've fixed all the weird restart errors, I'm going to check out this branch on a new universe and see if it works properly.

tummychow:
Verified on universe:

- we no longer get into an endless restart loop
- we do not repeatedly attempt to invoke hooks for pods that have not changed
- pods launch properly
- preparer restarts cleanly when it finds its own sha has changed
- after restart, preparer is aware of other pods that it launched in a previous incarnation
- p2-bootstrap works with existing consul URLs
- reality store updates are now working properly

Only one outstanding issue, which we knew about:

- when restarting self, the launch of our own pod times out because we refuse to exit until all our other goroutines have exited. So that creates an error in the logs. No discernable impact on the behavior of the preparer, so I'd say this is one for the icebox

overall :+1: :shipit:

anthonybishopric:
Not sure what's up with the EOF in the Hoist launchable expansion. Going to merge for now, will take a look at what's going on with that in a bit.

PR #68, opened 2015-01-17
Refactor intent and reality into combined package

tummychow:
This branch introduces pkg/p2consul (sorry I couldn't come up with a good name, suggestions needed), which provides a generalized version of the functionality in pkg/intent and pkg/reality. This package can read, write, list and watch manifests at any path in consul, eliminating the primary distinction between the two original packages (ie the path trees they used).

In addition, this branch migrates three components (notably including the preparer!) onto the new package. Some of our code, eg pkg/hooks, is moving a bit too fast for me to migrate that since it would block other work. Once this PR is merged, we can start refactoring code to use this package as we implement other features.

Tests on universe confirmed that all three affected components (preparer, bootstrap, inspect) continue to work as expected.


tummychow:
I'll also add some tests for this akin to what we have for pkg/intent

tummychow:
Incorporating `pkg/hooks` into `pkg/preparer` (#67) caused a dramatic merge conflict with this PR. To fix the conflict, I had to port `pkg/hooks` onto `pkg/p2consul` as well. Luckily it went pretty smoothly. We're closer to migrating "everything" than I had originally expected.

Verified on universe that all seven binaries still worked. I have not tested hooks in universe, so I don't know if that functionality was affected... we can discuss that on Tuesday. I'm sure we won't be merging this PR over the long weekend anyway because it simply touches too much code.

bin/p2-bootstrap/bootstrap.go
```diff
@@ -5,7 +5,7 @@ import (
 	"os"
 	"time"
 
-	"github.com/square/p2/pkg/intent"
+	"github.com/square/p2/pkg/p2consul"
```
anthonybishopric: My main problem with the package name is the collision with the `p2-consul` artifact we have internally. It's also redundant since it already says 'p2' in the path. Suggest renaming to `consul` or `store`.

tummychow: I like `store`. After resolving any other comments, I'll do a big rebase to rename the package and all the references to its old name.

bin/p2-bootstrap/bootstrap.go
```diff
@@ -72,15 +72,12 @@ func InstallConsul(consulPod *pods.Pod, consulManifest *pods.PodManifest) error
...
-	if err != nil {
-		return err
-	}
+	store := p2consul.NewStore(p2consul.Options{})
 	hostname, err := os.Hostname()
 	if err != nil {
 		return err
 	}
-	_, err = store.SetPod(hostname, *agentManifest)
+	_, err = store.SetPod(p2consul.Slash(p2consul.INTENT_TREE, hostname, agentManifest.ID()), *agentManifest)
```
anthonybishopric: We should wrap this join in a method specific to the different stores we have instead of forcing all callers to do the `Join` themselves.

bin/p2-bootstrap/bootstrap.go
tummychow: wait, `store` would probably collide with the local variable `store := p2consul.NewStore(...)`. Maybe something like `kv`?

bin/p2-bootstrap/bootstrap.go
tummychow: I'll do that instead of adding the `Slash` function

pkg/p2consul/kv.go
```diff
@@ -0,0 +1,138 @@
...
+	"strings"
+	"time"
+
+	"github.com/armon/consul-api"
+	"github.com/square/p2/pkg/pods"
+	"github.com/square/p2/pkg/util"
+)
+
+type ManifestResult struct {
+	Manifest *pods.PodManifest
```
anthonybishopric: This was a value, not a pointer, previously - as a best practice we are trying to avoid sending pointers over channels to mitigate potential race conditions if either side manipulates the value.

tummychow: will fix

bin/p2-bootstrap/bootstrap.go
anthonybishopric: Could call it `kp` for "key-pod"

pkg/p2consul/kv.go
```diff
@@ -0,0 +1,138 @@
...
+	if err != nil {
+		return 0, err
+	}
+	keyPair := &consulapi.KVPair{
+		Key:   key,
+		Value: buf.Bytes(),
+	}
+
+	writeMeta, err := s.client.KV().Put(keyPair, nil)
+	if writeMeta == nil {
```
anthonybishopric: Shouldn't this be checking if `err != nil`?

tummychow: the error gets returned anyway, so I don't think it matters. The bigger issue there is that, if there was an error, writemeta might be nil so derefing it causes a panic.

pkg/p2consul/kv.go
```diff
@@ -0,0 +1,138 @@
...
+
+// WatchPods watches the key-value store for any changes under the given key
+// prefix. The resulting manifests are emitted on podChan. WatchPods does not
+// return in the event of an error, but it will emit the error on errChan. To
+// terminate WatchPods, emit on quitChan.
+//
+// All the values under the given key prefix must be pod manifests. Emitted
+// manifests might be unchanged from the last time they were read. It is the
+// caller's responsibility to filter out unchanged manifests.
+func (s *Store) WatchPods(keyPrefix string, quitChan <-chan struct{}, errChan chan<- error, podChan chan<- ManifestResult) {
```
anthonybishopric: This is a really nice simplification.

anthonybishopric:
Still looks like there's room for some automated testing on the new consolidated package. Otherwise, looks good

tummychow:
The great rebase is complete:

- fixed package name (`kp`) across all commits. If you see any mentions of `p2consul` please call them out
- instead of the `Slash` function, added `IntentPath`, `RealityPath`, `HookPath` (see 4096b52)
- in manifestresult, use manifest value instead of manifest pointer

pkg/kp/kv.go
```diff
@@ -0,0 +1,132 @@
...
+type Store struct {
+	client *consulapi.Client
+}
+
+func NewStore(opts Options) *Store {
+	conf := consulapi.DefaultConfig()
+	conf.Address = opts.Address
+	conf.Token = opts.Token
+
+	// the error is always nil
```
petertseng: I wonder if one day consul will change the behavior of NewClient such that error could be nil. We would be a bit unprepared in that case

anthonybishopric:
:+1:

pkg/kp/kv.go
tummychow: The package we rely on is deprecated (in favor of `github.com/hashicorp/consul/api`) so it's unlikely to change. The newer version of the package also returns a nil error every time (at least right now).

I don't like ignoring errors, but that makes superfluous errors all the more wasteful. Personally I find it really dumb to push the error down to every client of this package when it's always going to be nil... if we move to the new package we can consider this more carefully, but I don't think there should be an error return on consulapi.NewClient in the first place.

anthonybishopric: Low probability but yeah, it is a land-mine on some level.

PR #69, opened 2015-01-21
p2-watch/schedule features

tummychow:
Some new features for p2-watch and p2-schedule. I'm still thinking about how to test `pkg/kp` so I probably won't have a PR for that today.


bin/p2-watch/main.go
```diff
@@ -28,6 +29,9 @@ func main() {
 	}
 
 	path := kp.IntentPath(*nodeName)
```
anthonybishopric: Lets add `--hooks` too :)

anthonybishopric:
:+1:

bin/p2-watch/main.go
tummychow: good suggestion, will add

PR #71, opened 2015-01-22
Implement post-activate

tummychow:
also fixes a minor bug regarding our tarball code

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,6 +3,10 @@ language: go
 go:
   - 1.4
 
+before_install:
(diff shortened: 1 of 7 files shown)


tummychow:
whoops, I was busy hacking at web errors and I forgot to fix some tests. They're failing now because chpst doesn't exist, so invokeBinScripts will fail with `os.IsNotExist`, and that error gets silenced. I'll investigate a solution for this. It would be ideal if we just replaced chpst with our own code based on tianon/gosu or something.

PR #72, opened 2015-01-22
utilities for hooks

anthonybishopric:
Useful things for hooks to use.


tummychow:
I added fake_chpst in my PR; is it needed here? otherwise lgtm

anthonybishopric:
Will remove fake_chpst

PR #73, opened 2015-01-23
Add hooktype as prefix to hook scheduling, watching.

anthonybishopric:
This had the side-effect of causing a circular dependency
between the hooks package and kp. I've moved the Listener to the
preparer package to distinguish hook execution from how they're
deployed (via the intent store).


tummychow:
lgtm

petertseng:
lgtm

PR #75, opened 2015-01-23
POD_HOME exports in p2

tummychow:
The fake_chpst fix was because, for some reason, there were unreadable files in my $TMPDIR. The test was setting ConfigDir to $TMPDIR and fake_chpst was trying to read them, then throwing exceptions.

### DIFF
--- a/pkg/pods/pod.go
+++ b/pkg/pods/pod.go
@@ -302,7 +302,7 @@ func (pod *Pod) Install(manifest *PodManifest) error {
 
 	// we may need to write config files to a unique directory per pod version, depending on restart semantics. Need
 	// to think about this more.
-	err = setupConfig(pod.EnvDir(), pod.ConfigDir(), manifest)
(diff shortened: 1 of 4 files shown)


pkg/pods/pod.go
```diff
@@ -360,6 +360,10 @@ func setupConfig(envDir string, configDir string, podManifest *PodManifest) erro
 	if err != nil {
 		return err
 	}
+	err = writeEnvFile(envDir, "POD_HOME", PodPath(podManifest.ID()))
```
anthonybishopric: This is not guaranteed to be correct. `PodPath` provides a default pod path if you only have the manifest ID. You should make `setupConfig` a member function on `Pod` and then use `pod.Path()` instead.

tummychow: will fix

anthonybishopric:
:+1:

PR #76, opened 2015-01-23
Create folders one by one during tar unpack

tummychow:
Fixes the bug with folders being owned by root: they only contained other folders, so after mkdirAll, the chown would only affect the bottom folder in the chain.


anthonybishopric:
:+1:

PR #78, opened 2015-01-27
Test hooks in the integration test

anthonybishopric:
This commit replaces the use of user.CreateUser with a
simple create_user shell script that is launched as a before_install
hook pod. This has the effect of verifying that the hook is
found in the intent store, installed correctly, and launched at
the correct time during the install / launch.

This commit includes a number of small fixes and changes to facilitate
that goal.

### DIFF
--- a/integration/single-node-slug-deploy/check.go
+++ b/integration/single-node-slug-deploy/check.go
@@ -13,9 +13,10 @@ import (
 	"runtime"
 	"time"
 
+	"github.com/square/p2/pkg/hooks"
 	"github.com/square/p2/pkg/intent"
+	"github.com/square/p2/pkg/kp"
 	"github.com/square/p2/pkg/pods"
-	"github.com/square/p2/pkg/user"
 	"github.com/square/p2/pkg/util"
 )
 
@@ -43,9 +44,13 @@ func main() {
 	if err != nil {
 		log.Fatalf("Could not execute bootstrap: %s", err)
 	}
-	_, err = user.CreateUser("hello", pods.PodPath("hello"))
-	if err != nil && err != user.AlreadyExists {
-		log.Fatalf("Could not create user: %s", err)
+	// _, err = user.CreateUser("hello", pods.PodPath("hello"))
+	// if err != nil && err != user.AlreadyExists {
+	// 	log.Fatalf("Could not create user: %s", err)
+	// }
+	err = scheduleUserCreationHook(tempdir)
+	if err != nil {
+		log.Fatalf("Couldn't schedule the user creation hook: %s", err)
 	}
 	err = postHelloManifest(tempdir)
 	if err != nil {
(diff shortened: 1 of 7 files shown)


integration/single-node-slug-deploy/check.go
```diff
@@ -43,9 +44,13 @@ func main() {
 	if err != nil {
 		log.Fatalf("Could not execute bootstrap: %s", err)
 	}
-	_, err = user.CreateUser("hello", pods.PodPath("hello"))
-	if err != nil && err != user.AlreadyExists {
-		log.Fatalf("Could not create user: %s", err)
+	// _, err = user.CreateUser("hello", pods.PodPath("hello"))
```
anthonybishopric: will delete commented lines

pkg/hooks/hooks.go
```diff
@@ -64,15 +65,22 @@ func runDirectory(dirpath string, environment []string, logger logging.Logger) e
 			continue
 		}
 		cmd := exec.Command(fullpath)
-		cmd.Stdout = os.Stdout
-		cmd.Stderr = os.Stderr
+		hookOut := &bytes.Buffer{}
+		cmd.Stdout = hookOut
+		cmd.Stderr = hookOut
```
anthonybishopric: Send the output of hooks to logs as a field, instead of raw to stdout / stderr.

pkg/kp/kv.go
```diff
@@ -109,24 +109,24 @@ func (s *Store) WatchPods(keyPrefix string, quitChan <-chan struct{}, errChan ch
...
-		if err != nil {
-			errChan <- err
-		} else {
-			for _, pair := range pairs {
-				manifest, err := pods.PodManifestFromBytes(pair.Value)
-				if err != nil {
-					errChan <- util.Errorf("Could not parse pod manifest at %s: %s. Content follows: \n%s", pair.Key, err, pair.Value)
-				} else {
-					podChan <- ManifestResult{*manifest, pair.Key}
+		case <-time.After(1 * time.Second):
```
anthonybishopric: Don't DDOS ourselves by constantly trying to read from the intent store. Without this `time.After`, it's really easy for the preparer to write hundreds of megabytes of errors to the log.

pkg/kp/kv.go
```diff
@@ -109,24 +109,24 @@ func (s *Store) WatchPods(keyPrefix string, quitChan <-chan struct{}, errChan ch
...
+			pairs, meta, err := s.client.KV().List(keyPrefix, &consulapi.QueryOptions{
+				WaitIndex: curIndex,
+			})
+			if err != nil {
+				errChan <- err
+			} else {
+				curIndex = meta.LastIndex
```
anthonybishopric: moved `curIndex =...` under the `else`

tummychow:
lgtm

PR #80, opened 2015-01-27
run after_install hooks

anthonybishopric:


pkg/preparer/orchestrate.go
```diff
@@ -213,6 +213,14 @@ func (p *Preparer) installAndLaunchPod(newManifest *pods.PodManifest, pod Pod, l
 			return false
 		}
 
+		err = p.hooks.RunHookType(hooks.AFTER_INSTALL, pod, newManifest)
+		if err != nil {
+			logger.WithFields(logrus.Fields{
+				"err":   err,
+				"hooks": hooks.AFTER_LAUNCH,
```
anthonybishopric: woops, fixing

pkg/preparer/orchestrate.go
```diff
@@ -200,7 +200,7 @@ func (p *Preparer) installAndLaunchPod(newManifest *pods.PodManifest, pod Pod, l
 		if err != nil {
 			logger.WithFields(logrus.Fields{
 				"err":   err,
-				"hooks": "before_install",
+				"hooks": hooks.BEFORE_INSTALL,
```
petertseng: you are changing this but not the "after_launch"?

anthonybishopric: just fixed that, good catch

petertseng:
:+1: 
(Y)

petertseng:
That's weird that the CI build fails. It works locally!

PR #81, opened 2015-01-27
VERSION should be a variable, not a constant

tummychow:
Variables can be set at link time with the -X flag, where as constants are fixed at compile time and therefore cannot be reconfigured later. I plan to put this in my universe script because I usually check out a new branch every time I build a new universe.

Use case:

```bash
$ touch pkg/version/version.go
$ godep go install -ldflags "-X github.com/square/p2/pkg/version.VERSION $(git describe --tags)" ./bin/...
$ p2-inspect --version
0.0.01-47-g123abcd
```

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -36,7 +36,7 @@ end
 
 desc 'Install all built binaries'
 task :install => :godep_check do
-  e "godep go install ./..."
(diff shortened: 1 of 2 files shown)


anthonybishopric:
Interesting. This would be good for `rake install` to do as well. Mind throwing it in there?

tummychow:
I'll add that

anthonybishopric:
:+1:

PR #82, opened 2015-01-27
reality: Get version info from reality store.

petertseng:


tummychow:
lgtm in general, but it's been a while since I looked at the semantics of this package. You might want to take a look at this as well @anthonybishopric

pkg/health/health.go
```diff
@@ -66,18 +67,31 @@ func (s *ServiceNodeStatus) IsCurrentVersion(version string) bool {
...
 		Health: consulHealth,
+		Store:  store,
 	}
 }
 
-func (s *ConsulHealthChecker) toNodeStatus(entry consulapi.ServiceEntry) *ServiceNodeStatus {
+func (s *ConsulHealthChecker) toNodeStatus(serviceID string, entry consulapi.ServiceEntry) *ServiceNodeStatus {
+
+	version := "unknown"
+	manifest, _, _ := s.Store.Pod(kp.RealityPath(entry.Node.Node, serviceID))
```
anthonybishopric: Can't there be an error here?

pkg/health/health.go
```diff
@@ -66,18 +67,31 @@ func (s *ServiceNodeStatus) IsCurrentVersion(version string) bool {
...
+	if manifest != nil {
+		var err error
+		version, err = manifest.SHA()
+		if err != nil {
+			version = "unknown"
```
anthonybishopric: make a const

pkg/health/health.go
petertseng: the manifest would be nil in that case, wouldn't it? But I'll check the err as well.

pkg/health/health.go
petertseng: hmm is 4ead91a what you are looking for

pkg/health/health.go
```diff
@@ -66,18 +69,30 @@ func (s *ServiceNodeStatus) IsCurrentVersion(version string) bool {
...
+		Store:  store,
 	}
 }
 
-func (s *ConsulHealthChecker) toNodeStatus(entry consulapi.ServiceEntry) *ServiceNodeStatus {
+func (s *ConsulHealthChecker) toNodeStatus(serviceID string, entry consulapi.ServiceEntry) *ServiceNodeStatus {
+
+	version := UnknownSHA
+	manifest, _, err := s.Store.Pod(kp.RealityPath(entry.Node.Node, serviceID))
+	if manifest != nil && err == nil {
```
anthonybishopric: well, more specifically, we should probably know if this fails, either by returning the error or something smarter.

petertseng:
Now returning an error from toNodeStatus

pkg/health/health.go
```diff
@@ -66,25 +67,41 @@ func (s *ServiceNodeStatus) IsCurrentVersion(version string) bool {
...
+		Store:  store,
 	}
 }
 
-func (s *ConsulHealthChecker) toNodeStatus(entry consulapi.ServiceEntry) *ServiceNodeStatus {
+func (s *ConsulHealthChecker) toNodeStatus(serviceID string, entry consulapi.ServiceEntry) (*ServiceNodeStatus, error) {
+
+	version := ""
+	manifest, _, err := s.Store.Pod(kp.RealityPath(entry.Node.Node, serviceID))
+	if err != nil {
```
petertseng: Wondering whether I should handle `pods.NoCurrentManifest` here specially, such as `err != nil && err != pods.NoCurrentManifest`

petertseng: yeah doing it

pkg/health/health.go
```diff
@@ -96,12 +114,18 @@ func (s *ConsulHealthChecker) LookupHealth(serviceID string) (*ServiceStatus, er
...
 		Statuses: make(map[string]*ServiceNodeStatus),
 	}
+
+	var firstErr error = nil
 	for _, entry := range entries {
 		if serviceID == entry.Service.ID {
-			serviceStatus.Statuses[entry.Node.Node] = s.toNodeStatus(*entry)
+			status, err := s.toNodeStatus(serviceID, *entry)
+			if err != nil && firstErr == nil {
+				firstErr = err
```
anthonybishopric: I'm still feeling the tingle of error suppression but this will at least signal that things are messed up.

anthonybishopric:
:+1:

PR #83, opened 2015-01-27
replication: use SetPod on the intent path, not node name

petertseng:


tummychow:
combine this into the previous PR? code looks fine to me

PR #85, opened 2015-01-28
Add p2-replicate

petertseng:


bin/p2-replicate/main.go
```diff
@@ -0,0 +1,46 @@
...
+
+	conf := consulapi.DefaultConfig()
+	conf.Address = *consulUrl
+
+	// the error is always nil
+	client, _ := consulapi.NewClient(conf)
+
+	healthChecker := health.NewConsulHealthChecker(*store, client.Health())
+
+	manifest, err := pods.PodManifestFromPath(*manifestUri)
```
petertseng: Actually I should actually support URIs like p2-launch. As written, this only supports local paths.

bin/p2-replicate/main.go
```diff
@@ -0,0 +1,46 @@
...
+	"gopkg.in/alecthomas/kingpin.v1"
+)
+
+var (
+	manifestUri = kingpin.Arg("manifest", "a path or url to a pod manifest that will be replicated.").String()
+	hosts       = kingpin.Arg("hosts", "Hosts to replicate to").Strings()
+	consulUrl   = kingpin.Flag("consul", "The hostname and port of a consul agent in the p2 cluster. Defaults to 0.0.0.0:8500.").String()
+)
+
+func main() {
```
anthonybishopric: As with the other "major" binaries, please consider adding copious user documentation to the --help

bin/p2-replicate/main.go
```diff
@@ -0,0 +1,46 @@
...
+	kingpin.Version("0.0.1")
```
anthonybishopric: We're using `github.com/square/p2/pkg/version` to get the right version string

petertseng:
Oops test fails because this is dependent on #82 .

anthonybishopric:
pretty straightforward - i like it.

anthonybishopric:
Oops, broken CI

bin/p2-replicate/main.go
```diff
@@ -0,0 +1,61 @@
...
+
+	conf := consulapi.DefaultConfig()
+	conf.Address = *consulUrl
+
+	// the error is always nil
+	client, _ := consulapi.NewClient(conf)
+
+	healthChecker := health.NewConsulHealthChecker(*store, client.Health())
+
+	// Fetch manifest (could be URI) into temp file
```
petertseng: Copy-pasta from p2-launch - consider extract to `pod.PodManifestFromURI` or something similar?!

petertseng:
TODO: Allow control of minimum number of nodes

bin/p2-replicate/main.go
anthonybishopric: I like that idea - I'd even suggest replacing `PodManifestFromPath` with that function.

bin/p2-replicate/main.go
```diff
@@ -0,0 +1,68 @@
...
+
+var (
+	replicate = kingpin.New("p2-replicate", `p2-replicate uses the replication package to schedule deployment of a pod across multiple nodes. See the replication package's README and godoc for more information.
+
+	Example invocation: p2-replicate helloworld.yaml aws{1,2,3}.example.com
+
+	`)
+	manifestUri = replicate.Arg("manifest", "a path or url to a pod manifest that will be replicated.").Required().String()
+	hosts       = replicate.Arg("hosts", "Hosts to replicate to").Required().Strings()
+	minNodes    = replicate.Flag("minnodes", "The minimum number of healthy nodes that must remain up while replicating.").Default("1").Short('m').Int()
```
anthonybishopric: Change to `min-nodes`

bin/p2-replicate/main.go
```diff
@@ -0,0 +1,68 @@
...
+	"github.com/square/p2/pkg/replication"
+	"github.com/square/p2/pkg/uri"
+	"github.com/square/p2/pkg/version"
+	"gopkg.in/alecthomas/kingpin.v1"
+)
+
+var (
+	replicate = kingpin.New("p2-replicate", `p2-replicate uses the replication package to schedule deployment of a pod across multiple nodes. See the replication package's README and godoc for more information.
+
+	Example invocation: p2-replicate helloworld.yaml aws{1,2,3}.example.com
```
anthonybishopric: Suggestion: change it to
```
p2-replicate --min-nodes 2 helloworld.yaml aws{1,2,3}.example.com
```
And add an explanatory comment about exactly what the replication package would do for this command invocation.

PR #86, opened 2015-01-28
Don't panic if we are missing health

petertseng:


pkg/replication/replication.go
```diff
@@ -98,7 +98,7 @@ func (repl *Replicator) Enact(store IntentStore, serviceChecker ServiceChecker,
 				}
 				status, err := serviceStatus.ForNode(node.Name)
 
-				healthy, current := repl.isHealthyAndCurrent(status, err)
+				healthy, current := repl.isHealthyAndCurrent(node.Name, status, err)
```
anthonybishopric: Shouldn't the right thing be to add ensure that `node.Name` is properly assigned to the resulting status, even if there's an error? Otherwise we're altering the `isHealthyAndCurrent` function just to support a log line.

petertseng: That can be arranged. d145bb7d64b2

anthonybishopric:
:+1:

PR #87, opened 2015-01-28
Hooks should run with their pod config as well

anthonybishopric:


pkg/runit/runit_service.go
```diff
@@ -67,3 +68,7 @@ func convertToErr(msg string, original error) (string, error) {
 		return msg, original
 	}
 }
+
+func (service *Service) RunScript() string {
```
anthonybishopric: This is unused - will remove

tummychow:
I'm going to roll this branch out in a universe first, but code lgtm.

tummychow:
Oh, one thing I forgot to mention but just remembered: hook_env.go should be changed to take advantage of the new environment variables.

tummychow:
okay, did another pass. Code lgtm except for the three tasks as I mentioned, but to summarize:

- hook environment appending to existing environment vs creating a new one (should probably be the latter so that the preparer's environment is not inherited)
- hook_env.go should have functions for the new variables, and existing functions should be ported onto the new variables if they haven't been already
- we have `pkg/config.LoadFromEnvironment` for `CONFIG_PATH`, we should have something similar for `HOOKED_CONFIG_PATH` (could be in pkg/hooks or pkg/config, whichever one makes more sense to you)

Functionality also lgtm. Once the comments are addressed I will take a third pass and probably a fourth to be sure I didn't miss anything.

tummychow:
code lgtm, gonna take it for another spin to confirm functionality again

anthonybishopric:
The travis failure is real this time. Looking into it.

pkg/pods/hoist_executable_test.go
```diff
@@ -0,0 +1,47 @@
...
+	"strings"
+	"testing"
+
+	. "github.com/anthonybishopric/gotcha"
+)
+
+func TestExecutableWritesValidScript(t *testing.T) {
+	user, err := user.Current()
+	Assert(t).IsNil(err, "test setup failure - should not have failed to get current user")
+	envdir, err := ioutil.TempDir("", "envdir")
```
petertseng: the error from this line is never checked because the following line immediately overwrites it

petertseng: eh I guess it'll be fine because the WriteFile into the directory will fail

anthonybishopric: I'm also making a (maybe dumb) assumption that the next line, which is also a temp directory creation, would almost definitely have an error if the previous line had an error.

PR #88, opened 2015-01-29
Prevent races when downloading the same artifact

anthonybishopric:
This ensures that a new temporary directory serves as the download
destination for Hoist artifacts, even when installing artifacts
of the same name and verison. In practice this won't affect day-to-day
operations of the preparer, but does have a meaningful impact on
test cases.

### DIFF
--- a/pkg/pods/hoist_launchable.go
+++ b/pkg/pods/hoist_launchable.go
@@ -220,6 +220,7 @@ func (hoistLaunchable *HoistLaunchable) Install() error {
 	}
 
 	outDir, err := ioutil.TempDir("", hoistLaunchable.Version())
+	defer os.RemoveAll(outDir)
 	if err != nil {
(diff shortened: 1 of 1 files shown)


pkg/pods/hoist_launchable.go
```diff
@@ -219,9 +219,14 @@ func (hoistLaunchable *HoistLaunchable) Install() error {
 		return nil
 	}
 
-	outPath := path.Join(os.TempDir(), hoistLaunchable.Version())
+	outDir, err := ioutil.TempDir("", hoistLaunchable.Version())
```
petertseng: is it ok that we are not removing  the tempdir

anthonybishopric: just fixed, good catch

petertseng:
GREAT

PR #89, opened 2015-02-02
Schedule hooks without node name

tummychow:


bin/p2-schedule/main.go
```diff
@@ -49,7 +49,7 @@ func main() {
 			if err != nil {
 				log.Fatalln(err)
 			}
-			path = kp.HookPath(hookType, *nodeName, manifest.ID())
+			path = kp.HookPath(hookType, manifest.ID())
```
anthonybishopric: Consider refactoring as struct - we need more flexibility with better safety guarantees

anthonybishopric:
Looks good for now, with caveat about refactoring

PR #90, opened 2015-02-02
p2-replicate: Use text logger

petertseng:


bin/p2-replicate/main.go
```diff
@@ -68,6 +69,7 @@ func main() {
 	allocated := allocation.NewAllocation(*hosts...)
 
 	replicator := replication.NewReplicator(*manifest, allocated)
+	replicator.Logger.Logger.Formatter = new(logrus.TextFormatter)
```
petertseng: logger logger

anthonybishopric: Yeah, a little weird. Not sure it's worth doing anything about

petertseng:
Tested in universe, looking nice and colorful.

anthonybishopric:
:+1:

PR #91, opened 2015-02-04
Remove nareix/curl, update deps

anthonybishopric:


tummychow:
other than that comment it lgtm

PR #92, opened 2015-02-05
Chown pod dirs correctly

tummychow:
This addresses three of the four main tasks I mentioned on the trello card:

- create /data/pods explicitly
- add mkdirchownall
- use mkdirchownall in all pod-related contexts.

The missing first task (parameterizing /data/pods) will probably need a bit more work. Implicit use of pods.DEFAULT_PATH is threaded through our code and refactoring that might get messy, but for now this functionality is all we really care about, and it simplifies deploys of some existing applications.

### DIFF
--- a/pkg/util/mkdir.go
+++ b/pkg/util/mkdir.go
@@ -1,0 +1,49 @@
+package util
+
+import (
+	"os"
+	"syscall"
+)
+
+func MkdirChownAll(path string, uid, gid int, perm os.FileMode) error {
+	// Fast path: if we can tell whether path is a directory or file, stop with success or error.
+	dir, err := os.Stat(path)
+	if err == nil {
+		if dir.IsDir() {
+			return nil
+		}
+		return &os.PathError{"mkdir", path, syscall.ENOTDIR}
+	}
+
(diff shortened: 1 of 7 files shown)


pkg/util/mkdir.go
```diff
@@ -0,0 +1,49 @@
+package util
+
+import (
+	"os"
+	"syscall"
+)
+
+func MkdirChownAll(path string, uid, gid int, perm os.FileMode) error {
```
anthonybishopric: This deserves a unit test suite

anthonybishopric:
looks good, would like test coverage on MkdirChownAll. Getting that kind of function correct on a first try would be impressive ;)

tummychow:
well considering I copied nearly all the code for that function from the standard library, it "probably" works... but I agree tests would be good. I'll add some

tummychow:
whoops forgot to chown some of the files being created in those dirs, i'll add that as well

tummychow:
confirmed in universe that all the files are now chowned to the correct user (and that it plays nice with a user-creating hook). going to add those mkdirAll tests now

anthonybishopric:
:+1:

PR #93, opened 2015-02-06
switch existing-url to existing-pod, fix integration test

anthonybishopric:


bin/p2-bootstrap/bootstrap.go
```diff
@@ -42,11 +44,24 @@ func main() {
...
+		if err != nil {
+			log.Fatalf("The existing consul pod is invalid: %s", err)
+		}
+		consulManifest, err = consulPod.CurrentManifest()
+		if err != nil {
+			log.Fatalf("Cannot get the current consul manifest: %s", err)
+		}
+	}
+
+	err = ScheduleForThisHost(consulManifest)
```
anthonybishopric: We're now also scheduling Consul itself

tummychow:
A few things that came to mind about this design (that I should have mentioned to you earlier but forgot to):

- I have yet to restart a consul agent in an integration environment; I don't know if it automatically rejoins its fellows. If you raise a cluster in this way, the consul nodes will all enter the intent store but not the reality store, and the preparer will respond by restarting it. (We already see this when the preparer bootstraps and immediately restarts itself). Will a consul cluster recover if it loses server quorum due to preparer-triggered restarts?

- future consideration: we should add retry logic to p2-bootstrap. If you launch the consul pod using p2-bootstrap, there may be some time delay (I have seen it in practice) before consul starts listening to 8500, and probably even more delay before it joins other nodes. I currently hack around this by polling consul for readiness before running p2-bootstrap. It might be useful to just move the polling into p2-bootstrap altogether.

- we should remove the hack for `intent` as root in this commit. As we've confirmed elsewhere, it is no longer necessary.

- I actually think there might be another way for us to structure these flags. In this PR, if you want to launch a new consul agent, you `p2-bootstrap --agent-pod <manifest>`. If you already have one running, you `p2-bootstrap --existing-agent-pod <current manifest of installed pod>`.

  I propose an alternative flag structure, `--consul-pod` (mandatory) and `--no-launch`. If you want a new consul instance, you do the same thing as you do now. But if you already have one running, you would `p2-bootstrap --consul-pod <current manifest> --no-launch`. I think this is a bit easier to understand than the current flags, but you might disagree because you just finished implementing them. I'm not sure if this style is better but maybe something to think about.

anthonybishopric:
Good comments, will address in order:

On Consul restarts - this is a good point, and it does beg a question about resilience. We haven't given a ton of thought to how this thing will withstand Consul restarts in general, let alone during the server bootstrapping process. I can't answer this confidently without expanding our test automation to include a multi-node setup. That deserves its own pull request and work, which is out of scope, but absolutely necessary. I'll throw it in the icebox. In the meantime, I'll remove the Consul scheduling part of this as we need to flesh out that use case.

I will add a loop to check for Consul availability before launching the preparer. 

I'll also remove the intent hack if we do in fact believe that it's not useful anymore. That will certainly improve the security profile of Consul in p2.

I think we'll need to talk more about the flag model in a separate discussion. `--consul-pod` and `--agent-pod` are not the same semantic flag. `--agent-pod` is the flag for establishing the preparer, not Consul. Perhaps it'd be more useful to see comparisons of the commands with your version of the flags with the existing ones - as it stands, they seem roughly the same.

tummychow:
ah dammit, I said agent pod when I meant consul pod

anthonybishopric:
Ok now I understand the idea of `--no-launch`. I'd like to keep using `--existing-pod` specifically so that we can enable the future case of bootstrapping hosts and adding the consul pod to the intent store. If we go with `--no-launch`, we have to implicitly find the pod using the default path and default pod ID, which is less preferable.

tummychow:
fair enough, no other comments. lgtm

bin/p2-bootstrap/bootstrap.go
```diff
@@ -68,17 +89,37 @@ func InstallConsul(consulPod *pods.Pod, consulManifest *pods.PodManifest) error
...
+			time.Sleep(200 * time.Millisecond)
+			err := store.Ping()
+			if err == nil {
+				consulIsUp <- struct{}{}
+				return
+			}
+		}
+	}()
+	select {
+	case <-time.After(10 * time.Second):
```
tummychow: Let me take this for a spin in universe, I'm really eager to see if it eliminates my awful "ping consul in ruby using raw net::http" code.

As an aside for future improvements, it might be useful to configure the timeout or set it to zero to skip it altogether.

tummychow:
worked on the first try, :+1:

anthonybishopric:
awesome :dancer:

PR #94, opened 2015-02-06
Add kp Ping

tummychow:
Ping the consul agent that a Store is connected to. If the store is reachable and has a leader, it'll respond 200 to the request, and the special header `X-Consul-Knownleader` will be true, therefore this method will return nil. If the agent has no leader, the agent will respond 500 with the special header set to false. This method will propagate the 500 error back to the caller. Finally, if the agent is outright unreachable, there'll probably be a tcp dial error that also gets returned to the caller.


anthonybishopric:
:+1:

PR #96, opened 2015-02-09
Execute disable

tummychow:


pkg/pods/hoist_launchable.go
```diff
@@ -53,9 +53,14 @@ func (hoistLaunchable *HoistLaunchable) Halt(serviceBuilder *runit.ServiceBuilde
 }
 
 func (hoistLaunchable *HoistLaunchable) Launch(serviceBuilder *runit.ServiceBuilder, sv *runit.SV) error {
+	_, err := hoistLaunchable.Disable()
```
anthonybishopric: This should be extracted to `orchestrate.go` as `pod.Disable(oldManifest)`. As written this would disable the new manifest, as well as assuming that there _is_ a previous install to disable.

pkg/preparer/orchestrate.go
```diff
@@ -226,6 +227,12 @@ func (p *Preparer) installAndLaunchPod(newManifest *pods.PodManifest, pod Pod, l
 			logger.WithField("err", err).Errorln("Service registration failed")
 			return false
 		}
+
+		if currentManifest != nil {
+			// the exit code of disable is always ignored
+			pod.Disable(currentManifest)
```
anthonybishopric: On the topic of ordering and formal constraints on what we're doing, lets take a look at how we can improve orchestrate.go. We have a task to establish the failure cases concretely, it might help to structure this file better and write some automation to enforce our expectations a bit more tightly.

anthonybishopric:
lgtm :+1:

pkg/preparer/orchestrate.go
tummychow: agreed, I'll add that to the card

PR #97, opened 2015-02-09
Bootstrap timeout

tummychow:
This feature is useful when bootstrapping in universe, where sometimes one node takes forever to `go get github.com/tools/godep` and it ends up 15-20 seconds behind all the others. We can use this flag to extend the timeout and smooth over those delays.

### DIFF
--- a/bin/p2-bootstrap/bootstrap.go
+++ b/bin/p2-bootstrap/bootstrap.go
@@ -17,6 +17,7 @@ var (
 	existingConsul          = kingpin.Flag("existing-consul-pod", "A path to an existing Consul pod that will be supplied to the base agent's configuration.").ExistingDir()
 	agentManifestPath       = kingpin.Flag("agent-pod", "A path to the manifest that will used to boot the base agent.").ExistingFile()
 	additionalManifestsPath = kingpin.Flag("additional-pods", "(Optional) a directory of additional pods that will be launched and added to the intent store immediately").ExistingDir()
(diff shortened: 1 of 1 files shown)


bin/p2-bootstrap/bootstrap.go
```diff
@@ -92,7 +93,15 @@ func InstallConsul(consulPod *pods.Pod, consulManifest *pods.PodManifest) error
 	return nil
 }
 
-func VerifyConsulUp() error {
+func VerifyConsulUp(timeout string) error {
+	timeoutDur, err := time.ParseDuration(timeout)
+	if err != nil {
```
anthonybishopric: Return this error instead of silencing, otherwise :+1:

bin/p2-bootstrap/bootstrap.go
```diff
@@ -92,7 +93,15 @@ func InstallConsul(consulPod *pods.Pod, consulManifest *pods.PodManifest) error
...
+		timeoutDur = 10 * time.Second
+	}
+	if timeoutDur == 0 {
+		return nil
```
anthonybishopric: If this is going to be the behavior, we should probably be clearer about the argument name. `--timeout` could express a constraint on the entire bootstrapping process - it's not obvious that it's just applied to the consul ping. Or, we make `timeout=0` mean try infinitely and add a different argument `--skip-consul-ping` to do what you want here.

bin/p2-bootstrap/bootstrap.go
tummychow: fixed

tummychow:
I tweaked the flag description a bit, what do you think now?

PR #98, opened 2015-02-09
Add p2-run-hooks

anthonybishopric:
This a debugging utility to make running hooks for apps a little
easier. This also fixes the integration test, which changed when
users were given ownership of their pod directory.

### DIFF
--- a/bin/p2-preparer/main.go
+++ b/bin/p2-preparer/main.go
@@ -7,6 +7,7 @@ import (
 	"syscall"
 
 	"github.com/Sirupsen/logrus"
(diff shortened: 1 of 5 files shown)


tummychow:
code lgtm

This shouldn't block the merge, but I'm curious about the use cases for this. It would be a bit weird to run hooks on an application that had already been installed and launched. Any examples of where this particular tool would be useful?

anthonybishopric:
Just now I used it to debug the integration tests (`could not execute useradd: cannot create directory`) and I needed to replay that hook with the right environment, etc.

PR #100, opened 2015-02-10
Lower log level for "new manifest received"

tummychow:
This is definitely the least useful and noisiest of our log messages


anthonybishopric:
:+1:

PR #101, opened 2015-02-12
PGP signed manifests

tummychow:
Apologies for the titanic diff; that's what happens when you import two new packages and one of them is really big. My changes are way down at the bottom.

Anyway this code "should" implement our security model:

- manifests can be part of a PGP clearsigned message (ie manifest in plain text with a signature attached to the end, generated with gpg `--clearsign` flag)
- the preparer boots up with an optional keyring somewhere on the filesystem
- if the keyring was provided, the preparer will reject manifests that were not signed by a member of that keyring
- the public key of a signed manifest is logged (debug level), while unsigned or unknown-signer manifests will raise warnings
- if no keyring was provided, the preparer ignores signatures altogether and installs any manifest, as it did before (they wouldn't be meaningful anyway, because the signature can't be identified without a keyring)

I'll test it in universe after messing with gpg a bit more. For now please review code.

Implementation notes:

- I know we originally planned to have another struct that wrapped around PodManifest, but in the end it was **much** simpler to add the `Signer` method to PodManifest directly. A ton of type signatures would have changed if I had added a new structure.

- Signature validation is delayed for as long as possible - only when you invoke `Signer` with the keyring in question. This allows the vast majority of our code to ignore the existence of a signature altogether (otherwise you would have to copy the keyring all the way down to the `kp.Store` that generated the manifests).

  In exchange, you have to preserve the original message forever, to avoid remarshaling (which might result in a slightly different text that invalidates the signature). In addition, if you mutate the public fields of the PodManifest, this API still behaves as if the pod manifest is signed.

  There's no way to fix either of these problems unless we remove the four exported fields from PodManifest and, continuing from there, make the entire structure immutable. I actually really wanted to make this refactor happen (while I'm at it, I'd convert every `*PodManifest` to a `PodManifest`), but it would have taken a few days, where as this took a few hours. We should come back to it next month or something.


pkg/pods/manifest_parser.go
```diff
@@ -74,6 +97,14 @@ func (manifest *PodManifest) Write(out io.Writer) error {
 }
 
 func (manifest *PodManifest) Bytes() ([]byte, error) {
+	if manifest.signature != nil {
+		// if it's signed, we must recycle the original content to preserve the
+		// signature's validity. remarshaling it might change the exact text of
```
xaviershay: remarshaling should be deterministic right? Why isn't it?

xaviershay: oh, maybe hash ordering is deterministic in go?

tummychow: No, hash ordering is random. The reason that we can't remarshal is because the originally signed YAML might be different from the marshalled YAML.

For example I compose a manifest with this exact text:
```yaml
---
# author: sjung
id: web
launchables:
    web:
        launchable_type: hoist
        launchable_id: web
        location: /home/sjung/web.tar.gz
```

Note that the indentation is 4 spaces (YAML standard indentation is 2) and there's also a comment. These are superficial features. If I remarshal the YAML all of them will disappear. The indentation will go back to 2 spaces and there will be no comments, because they aren't preserved when you parse the YAML.

Now I sign the manifest above. My signature is tied to the exact text that I had when I made the signature (which is the whole point). If I remarshal the YAML, then the text will be slightly different, and the signature will no longer match against the resulting manifest. So the original manifest has to be preserved for later signature validation.

tummychow:
Found a minor bug in integration (you're supposed to verify signature against the Block.Bytes, not the Block.Plaintext), will address comments and push up some fixed commits.

It'll take some time to come up with good tests for this, but I'll think about how to approach it. It might be time to separate manifest from pkg/pods as part of test-writing.

pkg/preparer/orchestrate.go
```diff
@@ -165,6 +169,27 @@ func (p *Preparer) handlePods(podChan <-chan pods.PodManifest, quit <-chan struc
 	}
 }
 
+// check if a manifest satisfies the signature requirement of this preparer
+func (p *Preparer) verifySignature(manifest pods.PodManifest, logger logging.Logger) bool {
+	// do not remove the logger argument, it's not the same as p.Logger
+	if p.keyring == nil {
```
dgalling: This fails open if there's no keyring?

tummychow: If there is no keyring, then the preparer is operating in unauthenticated mode, and the signature of the manifest (if any) is irrelevant. Therefore the manifest should be accepted.

dgalling: What does it mean for the preparer to operate in unauthenticated mode?

tummychow: ie manifest signatures will be ignored.

If the preparer is "authenticated" (ie it has a pgp keyring), then every manifest it receives will have to be signed by one of the keys in that ring, or otherwise the preparer will refuse to deploy it. "unauthenticated" mode is the opposite - signatures will be ignored, unsigned manifests will also be allowed, etc.

dgalling: Right. What is a use case for the unauthenticated mode?

tummychow:
To be clear: the keyring is not mandatory. If you don't configure one, then the preparer will start without it, and it will ignore manifest signatures. This is useful in development or in universe environments where the keyring doesn't exist, or where it's inconvenient to sign manifests as part of testing. Note that p2 is trivial to deploy in an empty environment, unlike hoist, so supporting this case is important for our workflow.

I would also like to mention that an unauthenticated preparer can seamlessly be upgraded to authentication, because it can redeploy itself. You can add the keyring to a later pod manifest for the preparer and deploy it again. From there on, the preparer will require signed manifests as expected. You cannot downgrade the preparer (ie remove the keyring) unless you sign the manifest that removes that configuration.

If you aren't familiar with our design plan here please come by my desk and discuss it with us. This was all covered in a document that I'm pretty sure we sent to infosec already.

pkg/preparer/orchestrate.go
dgalling: Per our discussion: let's remove the unauthenticated code path and provide some dummy keys that are used in development/testing environments. I'm happy to help with this.

tummychow:
ping @alokmenghrajani @dgalling @anthonybishopric please review the last few commits, this is waiting for merge

anthonybishopric:
:+1:

pkg/pods/manifest_parser.go
```diff
@@ -89,19 +123,20 @@ func (manifest *PodManifest) WriteConfig(out io.Writer) error {
 	return nil
 }
 
-// SHA() returns a string containing a hex encoded SHA-1
-// checksum of the manifest's contents
+// SHA() returns a string containing a hex encoded SHA-1 checksum of the
+// manifest's contents. The contents are normalized, such that all equivalent
+// YAML structures have the same SHA (despite differences in comments,
+// indentation, etc).
 func (manifest *PodManifest) SHA() (string, error) {
```
dgalling: What is this checksum used for?

tummychow: This is to determine if the data content of two manifests are equivalent.

PR #102, opened 2015-02-18
ACL token support for all p2 consul-interacting binaries

tummychow:
mostly self-explanatory

### DIFF
--- a/bin/p2-bootstrap/bootstrap.go
+++ b/bin/p2-bootstrap/bootstrap.go
@@ -18,6 +18,7 @@ var (
 	agentManifestPath       = kingpin.Flag("agent-pod", "A path to the manifest that will used to boot the base agent.").ExistingFile()
 	additionalManifestsPath = kingpin.Flag("additional-pods", "(Optional) a directory of additional pods that will be launched and added to the intent store immediately").ExistingDir()
 	timeout                 = kingpin.Flag("consul-timeout", "How long to wait for consul to begin serving. 0 will skip the consul check altogether.").Default("10s").String()
+	consulToken             = kingpin.Flag("consul-token", "The ACL token to pass to consul when registering the bootstrapped pods").String()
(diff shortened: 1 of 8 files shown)


bin/p2-bootstrap/bootstrap.go
```diff
@@ -102,7 +103,9 @@ func VerifyConsulUp(timeout string) error {
 		return nil
 	}
 
-	store := kp.NewStore(kp.Options{})
+	store := kp.NewStore(kp.Options{
+		Token: *consulToken, // not actually necessary because this endpoint is unauthenticated
```
anthonybishopric: The comment isn't necessarily true - different organizations may point to an existing consul setup with strict reader ACLs.

tummychow: No the catalog endpoints actually cannot be authenticated, consul doesn't have that feature. kp contains primarily kv endpoints (which support ACLs) but it also contains service endpoints for service registration (no ACL support) and catalog endpoints for pinging (no ACL support).

tummychow:
Took it for a spin in universe (with a consul cluster that defaulted to deny) and it works fine. As we discussed, bootstrapping a new node is still a bit of an open question (have to generate an ACL for its permissions and pass it to the preparer somehow) but we don't need to implement that in p2 itself.

anthonybishopric:
The wiki page for setting up a cluster should include a note about this, since pgp is not currently mentioned anywhere on there and will surely screw up anyone who tries to set up a cluster now. As long as the documentation makes it clear how to bootstrap a preparer with the `consul_token_path`, this doesn't need to be solved by p2 directly. :+1:

tummychow:
I added PGP notes to the wiki. ACLs are optional because consul defaults to allowing all un-ACLed requests, so new users shouldn't have to worry about this functionality.

PR #103, opened 2015-02-19
Travis

tummychow:
no comments needed atm, I just want travis to build it

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,4 +14,6 @@ install:
   - export PATH=$PATH:$GOBINS
   - rake install
 
-script: rake test
+script:
+  - rake test
+  - integration/travis.sh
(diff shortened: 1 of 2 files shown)


tummychow:
okay I need to spin up a travis VM to test this some more, still no comment required

tummychow:
okay so: these RunSV errors are coming from hoistLaunchable.Start. It's trying to invoke sv start in the service directory, and sv start bails out because runsv isn't running yet. In a travis VM (running on my macbook) I can also see that there's a clear delay between the dir being created, and runsvdir spinning off a runsv instance for it. Setting aside this error, the test does actually work.

This never happens in our real environments so the next step is to find out why this error is occurring here and what we can do to prevent it.

tummychow:
Going to close this for now because this problem doesn't come up in our real hosts, and getting travis to do more for us is not a priority atm. We can come back to this unit of work in the future

PR #104, opened 2015-02-20
GPG in integration test

tummychow:
vagrant integration tests should pass on this branch.


integration/single-node-slug-deploy/check.go
```diff
@@ -62,6 +163,29 @@ func main() {
 	}
 }
 
+func loadGPG(workdir string) error {
+	importpub := exec.Command("gpg", "--no-default-keyring", "--keyring", path.Join(workdir, "pubring.gpg"), "--secret-keyring", path.Join(workdir, "secring.gpg"), "--import")
```
anthonybishopric: Why not just create the pubring upfront and save it in this test directory? You can use `util.From` to make it a little easier to locate.

integration/single-node-slug-deploy/check.go
```diff
@@ -173,14 +316,25 @@ func postHelloManifest(dir string) error {
...
-	store, err := intent.LookupStore(intent.Options{})
+	manifestPath, err = signManifest(manifestPath, dir)
 	if err != nil {
 		return err
 	}
-	hostname, _ := os.Hostname()
-	_, err = store.SetPod(hostname, *manifest)
-	return err
+
+	return exec.Command("p2-schedule", manifestPath).Run()
```
anthonybishopric: :star: nice expansion of test scope + refactoring

anthonybishopric:
LGTM with the one nit about saving the pubring out of band

integration/single-node-slug-deploy/check.go
tummychow: good point, I'll add the rings to our repo instead of creating them at test time

PR #105, opened 2015-02-20
Handle symlinks in a hoist artifact tarball

tummychow:
We weren't reacting to the symlink type before, so we just created an empty file and left it that way.

In the future, any types we don't recognize should result in errors rather than silent coercion. This clearly indicates we can't handle the tarball and we can then implement a proper response for that type.

### DIFF
--- a/pkg/pods/hoist_launchable.go
+++ b/pkg/pods/hoist_launchable.go
@@ -335,7 +335,12 @@ func (hoistLaunchable *HoistLaunchable) extractTarGz(fp *os.File, dest string) (
 		}
 		fpath := path.Join(dest, hdr.Name)
 
-		if hdr.Typeflag == tar.TypeSymlink {
+		if hdr.Typeflag == tar.TypeSymlink, tar.TypeLink {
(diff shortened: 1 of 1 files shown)


pkg/pods/hoist_launchable.go
```diff
@@ -335,7 +335,18 @@ func (hoistLaunchable *HoistLaunchable) extractTarGz(fp *os.File, dest string) (
...
+			// hard links are intentionally treated as symlinks, because the
+			// target of a hardlink might be later in the tarball and not
+			// unpacked yet
+			// creating a hardlink in that situation would raise an error
+			// because the hardlink target doesn't exist
+			err = os.Symlink(hdr.Linkname, fpath)
+			if err != nil {
+				return util.Errorf("Unable to create destination symlink %s (to %s) when unpacking %s: %s", fpath, hdr.Linkname, fp.Name(), err)
+			}
+		case tar.TypeDir:
```
anthonybishopric: We may want to extract this as a standalone utility at some point. This function is really getting gnarly.

anthonybishopric:
:+1:

PR #106, opened 2015-02-20
Allowed deployers

petertseng:
Original:

OK, this has the replicator check the intent store for allowed_deployers before writing to the intent store.

This works, but we also need to do this for p2-schedule, don't we?

I've a mind to write tests for this, if I hear that this is the direction we want to take.

Now it's the simplified version for preparer only.

### DIFF
--- a/pkg/preparer/orchestrate.go
+++ b/pkg/preparer/orchestrate.go
@@ -157,7 +157,23 @@ func (p *Preparer) verifySignature(manifest pods.PodManifest, logger logging.Log
 
 	signer, err := manifest.Signer(p.keyring)
 	if signer != nil {
-		logger.WithField("signer_key", signer.PrimaryKey.KeyIdShortString()).Debugln("Resolved manifest signature")
+		signerId := signer.PrimaryKey.KeyIdShortString()
+		logger.WithField("signer_key", signerId).Debugln("Resolved manifest signature")
+
+		// Hmm, some hacks here.
+		if manifest.Id == "p2-preparer" {
+			foundAuthorized := false
+			for _, authorized := range p.authorizedDeployers {
+				if authorized == signerId {
+					foundAuthorized = true
+				}
+			}
+			if !foundAuthorized {
+				logger.WithField("signer_key", signerId).Errorln("Not an authorized deployer of the preparer")
+				return false
+			}
+		}
+
 		return true
 	}
 
--- a/pkg/preparer/orchestrate.go
+++ b/pkg/preparer/orchestrate.go
@@ -31,13 +31,14 @@ type Store interface {
 }
 
 type Preparer struct {
-	node         string
-	store        Store
-	hooks        Hooks
-	hookListener HookListener
-	Logger       logging.Logger
-	keyring      openpgp.KeyRing
-	podRoot      string
+	node                string
+	store               Store
+	hooks               Hooks
+	hookListener        HookListener
+	Logger              logging.Logger
+	keyring             openpgp.KeyRing
+	podRoot             string
+	authorizedDeployers []string
 }
 
 func (p *Preparer) WatchForHooks(quit chan struct{}) {
--- a/pkg/preparer/orchestrate.go
+++ b/pkg/preparer/orchestrate.go
@@ -11,6 +11,10 @@ import (
 	"golang.org/x/crypto/openpgp"
 )
 
+// The Pod ID of the preparer.
+// Used because the preparer special-cases itself in a few places.
+const POD_ID = "p2-preparer"
+
 type Pod interface {
 	hooks.Pod
 	Launch(*pods.PodManifest) (bool, error)
@@ -133,7 +137,7 @@ func (p *Preparer) handlePods(podChan <-chan pods.PodManifest, quit <-chan struc
 				// HACK ZONE. When we have better authz, rewrite.
 				// Still need to ensure that preparer launches correctly
 				// as root
-				if pod.Id == "p2-preparer" {
+				if pod.Id == POD_ID {
 					pod.RunAs = "root"
 				}
 
@@ -161,7 +165,7 @@ func (p *Preparer) verifySignature(manifest pods.PodManifest, logger logging.Log
 		logger.WithField("signer_key", signerId).Debugln("Resolved manifest signature")
 
 		// Hmm, some hacks here.
-		if manifest.Id == "p2-preparer" {
+		if manifest.Id == POD_ID {
 			foundAuthorized := false
 			for _, authorized := range p.authorizedDeployers {
 				if authorized == signerId {
--- a/pkg/preparer/orchestrate.go
+++ b/pkg/preparer/orchestrate.go
@@ -165,7 +165,7 @@ func (p *Preparer) verifySignature(manifest pods.PodManifest, logger logging.Log
 		logger.WithField("signer_key", signerId).Debugln("Resolved manifest signature")
 
 		// Hmm, some hacks here.
-		if manifest.Id == POD_ID {
+		if manifest.Id == POD_ID && len(p.authorizedDeployers) > 0 {
 			foundAuthorized := false
 			for _, authorized := range p.authorizedDeployers {
 				if authorized == signerId {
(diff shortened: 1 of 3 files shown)


pkg/replication/replication.go
```diff
@@ -111,8 +114,49 @@ func (repl *Replicator) Enact(store IntentStore, serviceChecker ServiceChecker,
 					// node is out of date, update the store with the new manifest.
 					// this will almost definitely result in duplicate Sets in the
 					// preparer, which is expected to ignore duplicate updates.
+					intentPath := kp.IntentPath(node.Name, repl.Manifest.ID())
+					existingManifest, dur, err := store.Pod(intentPath)
```
tummychow: should this be reading from the reality store instead?

tummychow:
I'm not a big fan of authz at schedule time. Like xavier said in slack, it creates awkward coupling between deploying a manifest and updating authz controls. We should do a design doc next week for how this feature needs to look; until then hold off on merge.

Assuming this is the design we want, my only comments are the one above, and to add the Pod method to the fake store in replication tests.

pkg/replication/replication.go
```diff
@@ -111,8 +114,49 @@ func (repl *Replicator) Enact(store IntentStore, serviceChecker ServiceChecker,
...
+						repl.Logger.WithFields(logrus.Fields{
+							"err":      err,
+							"node":     node.Name,
+							"duration": dur,
+						}).Errorln("Could not read current manifest")
+						continue
+					}
+
+					// If manifest has an AllowedDeployers list, we need to check it against the signature.
+					if err != pods.NoCurrentManifest && existingManifest.AllowedDeployers != nil {
```
anthonybishopric: This is a substantially complex new addition to this method and deserves its own modularization + tests.

anthonybishopric:
I would rather get some protection on the preparer using its own config than try to come up with a perfect authz solution for all artifacts and their deployers right now. I have two suggestions:

* Move this check into the preparer, not into schedule/replicate. You aren't actually creating any security guarantees by checking client-side (consider: a malicious entity bypasses Consul controls and schedules a manifest). 
* Move this check into the preparer's config for now and remove from the pod manifest specification. There are other concerns with this approach, aside from potential awkwardness of deploys. The list allowed deployers may `!=` the list of people allowed to change the list of deployers.

petertseng:
I am scaling back and only considering the preparer. Will deal with generalization later.

pkg/preparer/orchestrate_test.go
```diff
@@ -248,27 +274,55 @@ func TestPreparerWillRequireSignatureWithKeyring(t *testing.T) {
...
+	Assert(t).IsFalse(p.verifySignature(*manifest, logging.DefaultLogger), "expected preparer to reject manifest (no authorized deployer)")
+}
 
-	sigWriter.Write(manifestBytes)
-	sigWriter.Close()
+func TestPreparerWillRejectUnauthorizedSignatureForPreparer(t *testing.T) {
+	manifest, fakeSigner := testSignedManifest(t, func(m *pods.PodManifest, _ *openpgp.Entity) {
+		m.Id = "p2-preparer"
+		m.Config = map[string]interface{}{
+			"authorized_deployers": []string{"nobodylol"},
```
petertseng: Need halp here. Go does not believe this is actually a []string. Instead, it thinks:

```
--- FAIL: TestPreparerWillRejectUnauthorizedSignatureForPreparer (1.13s)
panic: interface conversion: interface is []interface {}, not []string [recovered]
	panic: interface conversion: interface is []interface {}, not []string
```

pkg/preparer/orchestrate.go
```diff
@@ -155,7 +155,39 @@ func (p *Preparer) verifySignature(manifest pods.PodManifest, logger logging.Log
...
+		// Hmm, some hacks here.
+		if manifest.Id == "p2-preparer" {
+			raw, ok := manifest.Config["authorized_deployers"]
+
+			if !ok {
+				logger.NoFields().Errorln("authorized_deployers not found in preparer's config")
+				return false
+			}
+
+			authorizedDeployers, ok := raw.([]string)
```
petertseng: so, perhaps this needs to change into `raw.([]interface{})` and then I need to do `allowed.(string)` below? seems a bit icky.

tummychow: I have definitely had bad experiences type-asserting through slices in the past. lemme review the laws of reflection and take another look

pkg/preparer/orchestrate_test.go
petertseng: dealt with it by consulting http://stackoverflow.com/questions/10478641/can-i-type-assert-a-slice-of-interface-values on the right way to do things.

7c856f6ea969395de30656a494815db1e77d4f37 deals with it.

petertseng:
Y'all, I think I did this terribly wrong. This code is checking the new manifest instead of the old. That means anyone can deploy a p2-preparer simply by making a new manifest and including yourself in the authorized_deployers.

It really needs to check the old manifest. No merge on this.

pkg/preparer/orchestrate.go
```diff
@@ -155,7 +155,39 @@ func (p *Preparer) verifySignature(manifest pods.PodManifest, logger logging.Log
 
 	signer, err := manifest.Signer(p.keyring)
 	if signer != nil {
-		logger.WithField("signer_key", signer.PrimaryKey.KeyIdShortString()).Debugln("Resolved manifest signature")
+		signerId := signer.PrimaryKey.KeyIdShortString()
+		logger.WithField("signer_key", signerId).Debugln("Resolved manifest signature")
+
+		// Hmm, some hacks here.
+		if manifest.Id == "p2-preparer" {
+			raw, ok := manifest.Config["authorized_deployers"]
```
petertseng: Nooooooo!!!

This code says "when deploying p2-preparer, make sure the new manifest is signed by someone that the new manifest lists as authorized". That's no good.

The preparer should, on startup, read the authorized_deployers from its config, and store it somewhere.
Then it should check any p2-preparer manifests against that stored set of authorized_deployers.
That mechanism implements the desired behavior of "when deploying p2-preparer, make sure the new manifest is signed by someone that the OLD manifest lists as authorized"

petertseng:
Fixed to check config.

pkg/preparer/orchestrate_test.go
```diff
@@ -248,27 +274,53 @@ func TestPreparerWillRequireSignatureWithKeyring(t *testing.T) {
...
+	p.authorizedDeployers = []string{"nobodylol"}
 
-	Assert(t).IsTrue(p.verifySignature(*manifest, logging.DefaultLogger), "should have accepted signed manifest")
+	Assert(t).IsFalse(p.verifySignature(*manifest, logging.DefaultLogger), "expected preparer to reject manifest (unauthorized deployer)")
+}
+
+func TestPreparerWillAcceptAuthorizedSignatureForPreparer(t *testing.T) {
+	sig := ""
+	manifest, fakeSigner := testSignedManifest(t, func(m *pods.PodManifest, e *openpgp.Entity) {
+		m.Id = "p2-preparer"
```
anthonybishopric: Can we make a `pods.PreparerID` instead of referring to this in places? That way when we generalize the concept (when we have time) it'll be easy to go through and remove references.

anthonybishopric:
:+1: looks good except for one variable extraction nit.

pkg/preparer/orchestrate_test.go
petertseng: Propose `preparer.PodId` instead. Done.

tummychow:
code looks good to me with one food-for-thought item:

Right now the mandatory keyring is only enforced in `preparer.New` (if you don't specify a keyring path then this function errors out). But in lower layers of the preparer code, we support keyring==nil cases. In other words, authn can be disabled at any layer below `preparer.New`. I'm wondering if we should have the same behavior for our authz.

I see that right now the preparer will reject a self-deploy if the authorized deployers array is nil/empty. Should we support the nil case as a way to skip authz? If the slice is nil in `preparer.New` we would return an error there.

anthonybishopric:
Keyring enforcement actually only happens in `p2-preparer/main.go` right now, fwiw

tummychow:
even better!

anthonybishopric:
Please make sure your new cached signer approach doesn't cause issues when run with `go test -race`

petertseng:
I am slightly surprised. `go test -race` has not complained.

anthonybishopric:
:+1:

PR #107, opened 2015-02-23
Refactoring

anthonybishopric:
These are three big commits that deserve their own reviews, but because Git, here is one PR with all of them.

The commits are:
* Move the `hoist_launchable` to its own package.
* Rename `manifest_parser` to `pod_manifest`.
* Move preparer initialization code to `pkg/preparer/setup.go`.


tummychow:
:+1: might need to rebase onto master after the tarball symlinking PR

anthonybishopric:
Looks like Git is smart enough to merge the `git mv` and your changes to `extract`.

tummychow:
fyi there are still a few hardcoded instances of pods.DEFAULT_PATH, particularly in PodFromManifestId, that will need to be refactored before we can really say we've "parameterized /data/pods".

PR #108, opened 2015-02-24
actually parameterize data/pods

anthonybishopric:


tummychow:
:+1: but we should take it for a spin with a nonstandard pod root to prove that it really works. I'll pick it up for a universe in a bit to test it.

PR #109, opened 2015-02-25
HTTPS health checking

tummychow:
This commit adds support for https health checks. Pods have to opt into https in their manifests. The CA for the HTTPS endpoint is in consul's `$SECRETS_PATH`.

pretty icky code to be putting in an open source project but we can improve the usability later

### DIFF
--- a/pkg/kp/service.go
+++ b/pkg/kp/service.go
@@ -10,13 +10,13 @@ import (
 var (
 	// prints the HTTP response on stderr, while exiting 0 if the status code is 200
 	httpStatusCheck  = `if [[ $(curl http://$(hostname):%v/_status -s -o /dev/stderr -w "%%{http_code}") == "200" ]] ; then exit 0 ; else exit 2; fi`
-	httpsStatusCheck = `if [[ $(curl https://$(hostname):%v/_status -s -o /dev/stderr -w "%%{http_code}" --cacert ${SECRETS_PATH}/service2service.ca.pem) == "200" ]] ; then exit 0 ; else exit 2; fi`
+	httpsStatusCheck = `if [[ $(curl https://$(hostname):%v/_status -s -o /dev/stderr -w "%%{http_code}" --cacert '%s') == "200" ]] ; then exit 0 ; else exit 2; fi`
 )
 
 // RegisterService creates a consul service for the given pod manifest. If the
 // manifest specifies a status port, the resulting consul service will also
 // include a health check for that port.
-func (s *Store) RegisterService(manifest pods.PodManifest) error {
+func (s *Store) RegisterService(manifest pods.PodManifest, caPath string) error {
 	podService := &consulapi.AgentServiceRegistration{
 		Name: manifest.ID(),
 	}
@@ -30,7 +30,7 @@ func (s *Store) RegisterService(manifest pods.PodManifest) error {
 		if manifest.StatusHTTP {
 			podService.Check.Script = fmt.Sprintf(httpStatusCheck, manifest.StatusPort)
 		} else {
-			podService.Check.Script = fmt.Sprintf(httpsStatusCheck, manifest.StatusPort)
+			podService.Check.Script = fmt.Sprintf(httpsStatusCheck, manifest.StatusPort, caPath)
 		}
 	}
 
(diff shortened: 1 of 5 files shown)


pkg/pods/pod_manifest.go
```diff
@@ -30,6 +30,7 @@ type PodManifest struct {
 	LaunchableStanzas map[string]LaunchableStanza `yaml:"launchables"`
 	Config            map[string]interface{}      `yaml:"config"`
 	StatusPort        int                         `yaml:"status_port,omitempty"`
+	StatusHTTPS       bool                        `yaml:"status_https,omitempty"`
```
anthonybishopric: Switch to `StatusHTTP`, reverse the logic. Encourage movement to HTTPS

tummychow:
switched the default to HTTPS just as you were commenting

pkg/kp/service.go
```diff
@@ -7,6 +7,12 @@ import (
 	"github.com/square/p2/pkg/pods"
 )
 
+var (
+	// prints the HTTP response on stderr, while exiting 0 if the status code is 200
+	httpStatusCheck  = `if [[ $(curl http://$(hostname):%v/_status -s -o /dev/stderr -w "%%{http_code}") == "200" ]] ; then exit 0 ; else exit 2; fi`
+	httpsStatusCheck = `if [[ $(curl https://$(hostname):%v/_status -s -o /dev/stderr -w "%%{http_code}" --cacert ${SECRETS_PATH}/service2service.ca.pem) == "200" ]] ; then exit 0 ; else exit 2; fi`
```
petertseng: is it too us-specific to have `service2service.ca.pem` in here?

tummychow: That's exactly what I was worried about, but at this point I don't think it's worth the code to find a better, more general solution (I am totally open to ideas though). Especially because consul has to have access to whatever certificate you specify. If the pod manifest gets to name the cert to use, then consul has to have read rights on it... gets out of hand pretty quickly for arbitrary certificates.

anthonybishopric: This is pretty nasty as is though. Right now, p2 users will run p2-bootstrap believing that they have set up a functioning cluster but will immediately crash on finding that they 1) haven't exported a SECRETS_PATH and 2) don't have any need for a custom certificate authority.

Ideally Consul would offer an option to provide a CA path when doing http checks. Looking at the Consul source this seems relatively easy to do. 

In the meantime I'd prefer to make this path configurable in the preparer instead.

tummychow: The problem with putting it in the preparer config is that you have to set it to `/data/pods/consul/secrets/service2service.ca.pem`, which is awful for a variety of reasons:

1. writing down the path to secrets
2. writing down the id of a pod...
3. and the pod id you're referencing isn't even your own

You cannot use the preparer's s2s cert because it's consul that ends up executing the curl. They are the same cert yes, but Consul does not have read permissions on that particular copy. I agree that the hardcoding is gross, but this configuration option does not look much better to me.

Also, you won't be affected by this at bootstrap time because neither consul nor the preparer have health checks. http health checks will also be unaffected because they do not attempt to set a `--ca-cert`. You won't actually notice the missing cert at all unless you attempt to schedule a pod manifest that contains an https health check.

anthonybishopric: > 1. writing down the path to secrets
We're doing that anyway, but in code instead of in configuration.

> 2. writing down the ID of a pod...
Yes, but it's not a pod ID we'll be imposing on anyone else. We just need to be aware of it in our own bootstrapping process.

> 3. and the pod id you're referencing isn't even your own
Weird yes, but I don't think this is significantly worse than 1. or 2.

> It's still gross, it just moves the grossness into a pod manifest.
At least we can document it so that people writing manifests for their preparer have a path to addressing it. As we've done, we've required people to at the very least 1) write a hook that exposes `$SECRETS_PATH` 2) symlink or otherwise recreate their CA to our PEM name `service2service.ca.pem`. 

> You won't actually notice the missing cert at all unless you attempt to schedule a pod manifest that contains an https health check.
Which we just made the default. As soon as something is scheduled, the healthcheck is going to start failing.

tummychow: HTTPS is the default if you have a health check, but health checks aren't turned on by default. You would still have to schedule a manifest that contained a `status_port` before you encountered the problem.

I will move it to a config option but I don't think it's an improvement

tummychow:
ah yes peter merged his pr, one sec while I rebase

anthonybishopric:
:+1:

PR #110, opened 2015-02-27
don't run consul as root after bootstrapping

anthonybishopric:


tummychow:
:+1:, but this won't have much actual impact because you have to be root to run p2-bootstrap

anthonybishopric:
It will because it will extract the tar as the consul user - the issue i was observing was that the consul directory is owned by root. Updates to the consul binary via `p2-launch` crash due to permission denied.

tummychow:
ah good point, but now you need a `consul` user to bootstrap (which we have, but we'll probably have to write that down somewhere)

PR #111, opened 2015-02-27
Create integration test consul user

tummychow:
Fix is needed after bootstrap stopped running consul as root (aa071cd)


petertseng:
Confirm that integration test is broken without this commit, and fixed with it.

PR #112, opened 2015-02-27
Add AFTER_AUTH_FAIL hook type

tummychow:
as requested. I'm gonna go test it but take a look at the code.

One thought I just had: it would be nice if we could pass more information to hooks of this type, specifically to communicate how the auth failure occurred. We don't really have a mechanism for this right now (the hook basically has to parse the manifest itself and check the signature by somehow reading the preparer's public keyring) but it would be cool.

### DIFF
--- a/pkg/preparer/orchestrate.go
+++ b/pkg/preparer/orchestrate.go
@@ -131,6 +131,15 @@ func (p *Preparer) handlePods(podChan <-chan pods.PodManifest, quit <-chan struc
 			manifestLogger.NoFields().Debugln("New manifest received")
 
 			working = p.verifySignature(manifestToLaunch, manifestLogger)
+			if !working {
+				err = p.hooks.RunHookType(hooks.AFTER_AUTH_FAIL, pods.NewPod(manifestToLaunch.ID(), pods.PodPath(p.podRoot, manifestToLaunch.ID())), &manifestToLaunch)
+				if err != nil {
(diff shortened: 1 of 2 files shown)


pkg/preparer/orchestrate.go
```diff
@@ -131,6 +131,15 @@ func (p *Preparer) handlePods(podChan <-chan pods.PodManifest, quit <-chan struc
 			manifestLogger.NoFields().Debugln("New manifest received")
 
 			working = p.verifySignature(manifestToLaunch, manifestLogger)
+			if !working {
+				err = p.hooks.RunHookType(hooks.AFTER_AUTH_FAIL, pods.NewPod(manifestToLaunch.ID(), pods.PodPath(p.podRoot, manifestToLaunch.ID())), &manifestToLaunch)
+				if err != nil {
```
anthonybishopric: Previously there were two methods `RunBeforeHooks` and `RunAfterHooks`. I removed them in favor of the general `RunHookType`, but it does make specializing the data we export difficult. I'm open to revisiting that pattern.

tummychow:
I just confirmed that it works

As for tweaking the API, I think we can leave it for later. It's a question of how strongly coupled hooks should be to the way they're scheduled (eg the contents of their manifests, what hook type they are - btw hooks cannot introspect that right now except by reading their own directory path) and executed.

anthonybishopric:
:+1:

PR #113, opened 2015-03-02
Cgroup limiting

tummychow:
Introduces a dependency on our [contain](https://github.com/square/prodeng/tree/master/cgroup-container) script. I was planning to use libcontainer for this, but there are too many moving parts in their top-level API, and their internal packages are clearly not intended for third-party consumption.

I kind of dislike the code surrounding "insert this binary into the runit script for this hoist executable". This has me thinking about a better API for that area, but it can wait for now.

### DIFF
--- a/pkg/hoist/hoist_launchable.go
+++ b/pkg/hoist/hoist_launchable.go
@@ -29,6 +29,7 @@ type HoistLaunchable struct {
 	FetchToFile Fetcher // Callback that downloads the file from the remote location.
 	RootDir     string  // The root directory of the launchable, containing N:N>=1 installs.
 	Chpst       string  // The path to chpst
+	Contain     string  // The path to contain
 }
 
 func DefaultFetcher() Fetcher {
@@ -171,12 +172,14 @@ func (h *HoistLaunchable) Executables(serviceBuilder *runit.ServiceBuilder) ([]H
 		servicePath := path.Join(serviceBuilder.RunitRoot, serviceName)
 		runitService := &runit.Service{servicePath, serviceName}
 		executable := &HoistExecutable{
-			Service:   *runitService,
-			ExecPath:  binLaunchPath,
-			Chpst:     h.Chpst,
-			Nolimit:   "/usr/bin/nolimit",
-			RunAs:     h.RunAs,
-			ConfigDir: h.ConfigDir,
+			Service:       *runitService,
+			ExecPath:      binLaunchPath,
+			Chpst:         h.Chpst,
+			Contain:       h.Contain,
+			ContainerName: h.Id,
+			Nolimit:       "/usr/bin/nolimit",
+			RunAs:         h.RunAs,
+			ConfigDir:     h.ConfigDir,
 		}
 
 		return []HoistExecutable{*executable}, nil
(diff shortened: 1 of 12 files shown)


pkg/hoist/hoist_launchable.go
```diff
@@ -22,13 +22,15 @@ type Fetcher func(string, string) error
...
-	RootDir     string  // The root directory of the launchable, containing N:N>=1 installs.
-	Chpst       string  // The path to chpst
+	Location      string  // A URL where we can download the artifact from.
+	Id            string  // A unique identifier for this launchable, used when creating runit services
+	RunAs         string  // The user to assume when launching the executable
+	ConfigDir     string  // The value for chpst -e. See http://smarden.org/runit/chpst.8.html
+	FetchToFile   Fetcher // Callback that downloads the file from the remote location.
+	RootDir       string  // The root directory of the launchable, containing N:N>=1 installs.
+	Chpst         string  // The path to chpst
+	Contain       string
```
anthonybishopric: docs plz

pkg/runit/servicebuilder.go
```diff
@@ -74,6 +74,7 @@ var DefaultBuilder = &ServiceBuilder{
 }
 
 var DefaultChpst = "/usr/bin/chpst"
+var DefaultContain = "/usr/bin/contain"
```
anthonybishopric: This shouldn't be in the `runit` package. I'd move it to wherever `DefaultNolimit` is right now

pkg/hoist/hoist_launchable.go
tummychow: yeah this structure is accumulating a lot of fields... i'll add doc comments

pkg/runit/servicebuilder.go
tummychow: we don't actually have a DefaultNolimit right now, so what package did you have in mind? Personally I want to take all three of these things (nolimit, chpst, contain) and factor them into something "prettier", but I don't know what it should look like right now.

anthonybishopric:
Please modify https://github.com/square/p2#dependencies and https://github.com/square/p2/wiki/Setting-Up-A-Cluster to reflect the new dependency.

pkg/runit/servicebuilder.go
anthonybishopric: Apparently I hallucinated that we had extracted that path to a constant. Yeah, ideally we allow flat configuration (clients may list out these dependencies) but that their defaults are kept isolated to their packages to maintain high cohesion (only `runit` should "know" about chpst, only XXX should no about `nolimit` and `contain`, etc)

tummychow:
wait I forgot about a bug in this, the app name being passed to contain is wrong and containers would not be shared across a launchable. will fix momentarily

pkg/pods/pod_test.go
```diff
@@ -203,41 +203,60 @@ func TestBuildRunitServices(t *testing.T) {
...
 
 	Assert(t).IsNil(err, "Got an unexpected error when attempting to start runit services")
 
 	pod.BuildRunitServices([]hoist.HoistLaunchable{*hoistLaunchable})
-	expected := fmt.Sprintf(`%s:
+	expected := fmt.Sprintf(`%[1]s:
   run:
   - /usr/bin/nolimit
-  - %s
+  - %[6]s
```
petertseng: I wish there were a way to do this that is easier to figure out than using the indices, but I couldn't find one.

anthonybishopric: The right way to do this is http://golang.org/pkg/text/template/ but I was lazy when I wrote the tests and now they are index spaghetti :sob:

tummychow: I don't think it's that bad. template is more powerful but the API is also much less convenient to use. Explicit indices are enough to avoid duplicating arguments, and it's obvious that the indices are significant so we can avoid tripping over them in the future.

tummychow:
one more commit incoming, I need to add this feature to p2-bin2pod since it's now a mandatory key under the launchable stanza

anthonybishopric:
Per discussion in person, we should probably just make container sizing optional. It's not always applicable to all apps / environments.

anthonybishopric:
:+1:

PR #114, opened 2015-03-03
HookListener.sync: Don't run hooks as root

petertseng:


pkg/preparer/listener.go
```diff
@@ -79,13 +78,6 @@ func (l *HookListener) Sync(quit <-chan struct{}, errCh chan<- error) {
 			}
 
 			hookPod := pods.NewPod(result.Manifest.ID(), path.Join(l.DestinationDir, event, result.Manifest.ID()))
-			// install hooks as current user, not the pod ID. TODO: authorization checks for the pod
-			// to ensure they are permitted.
-			curUser, err := user.Current()
-			if err != nil {
-				sub.WithField("err", err).Errorln("could not get current user")
-			}
-			hookPod.RunAs = curUser.Username
```
anthonybishopric: You will probably need to run the integration test, which attempts to do an `adduser` execution. This might fail now that it's no longer running as root.

tummychow:
probably no longer necessary now that #118 is merged, closing

PR #116, opened 2015-03-04
Fix tar hardlinks

tummychow:
Inside a tarball, the target of a hard link is encoded as a path relative to the tarball root, rather than from the hard link itself. The original code I wrote to handle hardlinks assumed they were encoded like symlinks when in fact that was wrong, so this is the fix.


anthonybishopric:
:+1:

PR #117, opened 2015-03-05
Use SHA256 in pod manifests

tummychow:
as we mentioned before


petertseng:
Cool. with `git grep sha` I think we may like to change `bin/p2-inspect/README.md` ? Otherwise I think everything else is covered.

tummychow:
Good catch, I'll make those shas in the readme 64chars long.

petertseng:
:+1:

anthonybishopric:
:+1:

PR #118, opened 2015-03-05
hooks may select a user to run as

anthonybishopric:


pkg/pods/pod_manifest.go
```diff
@@ -28,7 +28,7 @@ type LaunchableStanza struct {
 type PodManifest struct {
 	Id                string                      `yaml:"id"` // public for yaml marshaling access. Use ID() instead.
 	LaunchableStanzas map[string]LaunchableStanza `yaml:"launchables"`
-	Config            map[string]interface{}      `yaml:"config"`
+	Config            map[interface{}]interface{} `yaml:"config"`
```
tummychow: was this change actually necessary to get the rest of this to work?

anthonybishopric: yes, because of [this line](#diff-bfa12c9e8482818801371a2b62106430R162)

tummychow: Yeah but why can't LoadFromUnpacked just take a `map[string]interface{}` instead?

anthonybishopric: There's nothing wrong with an application specifying an integer key in their own config, so I'd rather not coerce it to a string.

tummychow: In general I don't like using interfaces as keys; if they're uncomparable you get a panic at time of indexing. It doesn't really help that YAML is a weird format and might allow unusual key types (I read the YAML spec front to back once and still don't understand it). The impact is small here and the risk is unclear, so I won't block the PR on this, but personally I would rather enforce string keys everywhere.

tummychow:
looks like some typos in the commit message, "may select"? otherwise :+1:

petertseng:
I think you may want to remove https://github.com/square/p2/blob/master/pkg/preparer/listener.go#L88 (this line, far as I can tell, comes after you set `runAs` at https://github.com/square/p2/pull/118/files#diff-bfa12c9e8482818801371a2b62106430R199)

pkg/pods/pod_manifest.go
anthonybishopric: YAML is confusing enough, us adding a new constraint on the otherwise freeform YAML `config` stanza is going to be deeply unintuitive. Agreed that we should discourage it though, it *is* weird.

anthonybishopric:
Wow that commit message is really messed up huh

anthonybishopric:
@petertseng good catch, will change.

pkg/preparer/listener.go
```diff
@@ -79,13 +79,23 @@ func (l *HookListener) Sync(quit <-chan struct{}, errCh chan<- error) {
...
-			curUser, err := user.Current()
+
+			// Hooks may specify a user to run as - by default they will run as their pod ID.
+			topLevelConfig := config.LoadFromUnpacked(result.Manifest.Config)
+			hookConfig, err := topLevelConfig.ReadMap("hook")
+			if err != nil {
+				sub.WithField("err", err).Errorln("Hook configuration is invalid")
+				break
+			}
+			runAs, err := hookConfig.ReadString("run_as")
```
stouset: Seems like it'd be useful to be able to specify nested paths in one call.

anthonybishopric: We had talked about a dot syntax for that at one point

`config.ReadString("foo.bar.baz")`

May come back to that in the future

PR #119, opened 2015-03-06
allow hook in integration test to run as root

anthonybishopric:
since we now default hooks to run as the pod ID, we need to allow this user creation hook to run as root


tummychow:
:+1: change looks fine, but maybe we should just vendor the entire hook as part of our repo instead of generating programmatically? Up to you

anthonybishopric:
Good call, there is precedent for vendoring hooks like this. It's actually a decent example hook overall. Will do that in a follow-up request

PR #121, opened 2015-03-10
Forbidden Pod IDs

petertseng:
Like root, for example

### DIFF
--- a/pkg/preparer/orchestrate.go
+++ b/pkg/preparer/orchestrate.go
@@ -215,6 +215,13 @@ func (p *Preparer) installAndLaunchPod(newManifest *pods.PodManifest, pod Pod, l
 
 	if p.podIdForbidden(newManifest.ID()) {
 		logger.WithField("manifest", newManifest.ID()).Errorln("Cannot use this pod ID")
+		err := p.hooks.RunHookType(hooks.AFTER_AUTH_FAIL, pod, newManifest)
+		if err != nil {
+			logger.WithFields(logrus.Fields{
+				"err":   err,
+				"hooks": hooks.AFTER_AUTH_FAIL,
+			}).Warnln("Could not run hooks")
+		}
 		return false
 	}
 
--- a/pkg/preparer/orchestrate.go
+++ b/pkg/preparer/orchestrate.go
@@ -109,6 +109,15 @@ func (p *Preparer) WatchForPodManifestsForNode(quitAndAck chan struct{}) {
 	}
 }
 
+func (p *Preparer) podIdForbidden(id string) bool {
+	for _, forbidden := range p.forbiddenPodIds {
+		if forbidden == id {
+			return true
+		}
+	}
+	return false
+}
+
 // no return value, no output channels. This should do everything it needs to do
 // without outside intervention (other than being signalled to quit)
 func (p *Preparer) handlePods(podChan <-chan pods.PodManifest, quit <-chan struct{}) {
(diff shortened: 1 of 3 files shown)


pkg/preparer/orchestrate.go
```diff
@@ -88,6 +89,12 @@ func (p *Preparer) WatchForPodManifestsForNode(quitAndAck chan struct{}) {
 			}).Errorln("there was an error reading the manifest")
 		case result := <-podChan:
 			podId := result.Manifest.ID()
+
+			if p.podIdForbidden(podId) {
+				p.Logger.WithField("pod", podId).Errorln("Cannot use this pod ID")
```
anthonybishopric: This needs to trigger an authz failure hook invocation.

pkg/preparer/orchestrate.go
```diff
@@ -203,6 +219,11 @@ func (p *Preparer) verifySignature(manifest pods.PodManifest, logger logging.Log
 func (p *Preparer) installAndLaunchPod(newManifest *pods.PodManifest, pod Pod, logger logging.Logger) bool {
 	// do not remove the logger argument, it's not the same as p.Logger
 
+	if p.podIdForbidden(newManifest.ID()) {
```
anthonybishopric: Why call this here as well?

petertseng: This is interesting. Normally I would put this only in `WatchForPodManifestsForNode`, since it seems wisest to check this as early as possible and not create extra goroutines for the forbidden IDs. I also wanted a test on this, and our tests call `installAndLaunchPod` instead. Perhaps I should figure out some way for tests on `WatchForPodManifestsForNode`?

It would be better to have the check in only one place if it triggers authz failure hooks

pkg/preparer/orchestrate.go
```diff
@@ -108,6 +115,15 @@ func (p *Preparer) WatchForPodManifestsForNode(quitAndAck chan struct{}) {
 	}
 }
 
+func (p *Preparer) podIdForbidden(id string) bool {
+	for _, forbidden := range p.forbiddenPodIds {
```
petertseng: It occurs to me that instead of doing a linear search through array, should probably look for its presence in a map instead.

pkg/preparer/orchestrate.go
tummychow: personally I'd rather have one nooping goroutine floating around than putting this logic in a place that's difficult to test. installAndLaunchPod seems like the better location to me

petertseng: ok now it is the only place

pkg/preparer/orchestrate.go
petertseng: ok e4e2bf6f8161cab458de22f53f8df1f7437584c0

pkg/preparer/orchestrate.go
anthonybishopric: One other small code etiquette thing because this function is now >100 lines :oncoming_police_car: :ambulance: lets move this into a separate function we invoke.

petertseng: My first nomination for the thing that needs to be moved into a separate function, actually... 
RunHookType, or warn if unable

anthonybishopric: That seems good too.

petertseng: Yeah IMO 3b62b7df7ae688b419688486eb080a728cf666a8 was long overdue

PR #122, opened 2015-03-11
cgroups round 2

tummychow:
Reimplementation of cgroups to eliminate the dependency on `contain` and pull the cgroup-manipulating functionality into our own code. I evaluated libcontainer but ultimately the API was too confusing and offered too much functionality to be worth integrating.


pkg/cgroups/cgroups.go
```diff
@@ -0,0 +1,155 @@
...
+	"strconv"
+	"strings"
+)
+
+// map of cgroup subsystems to their respective paths
+type Cgroups map[string]string
+
+// Find retrieves the mount points for all cgroup subsystems on the host. The
+// result of this operation should be cached if possible.
+func Find() (Cgroups, error) {
```
tummychow: stolen from https://github.com/docker/libcontainer/blob/master/cgroups/utils.go#L18 https://github.com/docker/libcontainer/blob/master/cgroups/utils.go#L101 https://github.com/measure/os/blob/master/misc/misc.go#L74

not yet tested, parsing all those files is probably expensive and the cgroup path is pretty predictable in our system so I didn't bother using this.

pkg/cgroups/cgroups.go
```diff
@@ -0,0 +1,155 @@
...
+
+	_, err = writeIfChanged(filepath.Join(cg["memory"], name, "memory.memsw.limit_in_bytes"), []byte(strconv.Itoa(hardLimit)+"\n"), 0)
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func writeIfChanged(filename string, data []byte, perm os.FileMode) (bool, error) {
```
tummychow: stolen from my own reimplementation of servicebuilder

pkg/pods/pod.go
```diff
@@ -367,17 +368,26 @@ func (pod *Pod) Install(manifest *PodManifest) error {
...
 		if err != nil {
 			pod.logLaunchableError(launchable.Id, err, "Unable to install launchable")
 			return err
 		}
+
+		err = cgroups.Default.SetCPU(launchable.Id, launchableStanza.CgroupConfig.Cpus)
+		if err != nil {
+			pod.logLaunchableError(launchable.Id, err, "Unable to set cgroup cpu limits")
+		}
+		err = cgroups.Default.SetMemory(launchable.Id, launchableStanza.CgroupConfig.Memory)
```
tummychow: We always generate cgroups at install time, but the default value will leave those groups completely unrestricted. Because the code is kinda scattered right now (cgroup generation in pod, cgexec invocation in hoistexecutable), it's easier than making the cgroup instantiation conditional

pkg/cgroups/cgroups.go
```diff
@@ -0,0 +1,155 @@
...
+
+// map of cgroup subsystems to their respective paths
+type Cgroups map[string]string
+
+// Find retrieves the mount points for all cgroup subsystems on the host. The
+// result of this operation should be cached if possible.
+func Find() (Cgroups, error) {
+	// refer to `man proc` or
+	// https://www.kernel.org/doc/Documentation/filesystems/proc.txt section 3.5
+	mountInfo, err := os.Open("/proc/self/mountinfo")
```
anthonybishopric: I'd like to have a specific package error (`cgroups.NotSupported` or similar) to signify that the cgroup feature should be ignored / warned to deployers.

pkg/cgroups/cgroups.go
```diff
@@ -0,0 +1,155 @@
...
+
+var Default Cgroups = map[string]string{
+	"cpu":    "/cgroup/cpu",
+	"memory": "/cgroup/memory",
+}
+
+// set the number of logical CPUs in a given cgroup, 0 to unrestrict
+func (cg Cgroups) SetCPU(name string, cpus int) error {
+	period := 1000000 // one million microseconds
+	quota := cpus * 1000000
```
anthonybishopric: should this just be `cpus * period`?

pkg/cgroups/cgroups.go
```diff
@@ -0,0 +1,155 @@
...
+		return err
+	}
+
+	return nil
+}
+
+// set the memory limit on a cgroup, 0 to unrestrict
+func (cg Cgroups) SetMemory(name string, bytes int) error {
+	softLimit := bytes
+	hardLimit := bytes * 2
```
anthonybishopric: This is from `contain` right? I think Xavier was talking about killing the distinction between the hard and soft limits.

pkg/cgroups/cgroups.go
```diff
@@ -0,0 +1,155 @@
...
+	}
+
+	err := os.Mkdir(filepath.Join(cg["memory"], name), 0755)
+	if err != nil && !os.IsExist(err) {
+		return err
+	}
+
+	// the hard memory limit must be set BEFORE the mem+swap limit
+	// so we must clear the swap limit at the start
+	err = ioutil.WriteFile(filepath.Join(cg["memory"], name, "memory.memsw.limit_in_bytes"), []byte("-1\n"), 0)
```
anthonybishopric: Hardcoding field names like `memory.memsw.limit_in_bytes` is fine for now since I assume this is just copied code, but we should provide a better package-level API for manipulating cgroups at some point.

pkg/cgroups/cgroups.go
xaviershay: yeah I don't want to expose two limits to apps. They should only see "softlimit". hard limit is an implementation detail.

pkg/hoist/hoist_executable.go
```diff
@@ -9,31 +9,27 @@ import (
...
-			e.Contain,
-			"-a",
-			e.ContainerName,
-			"-s",
-			e.Container,
-			"-v",
-			"--",
-		)
-	}
+	ret = append(ret,
```
anthonybishopric: Ideally we'd provide the arguments to `cgexec` from the `cgroups` package, and probably do similar for `runit`.

pkg/cgroups/cgroups.go
tummychow: ie they should be set to the same number?

pkg/cgroups/cgroups.go
tummychow: At the extreme end is the API provided by libcontainer (which I was considering mimicking), but I think it's overkill for this very simple use case. If we end up doing more complex things like out-of-band cgroup mutation (eg as syam was suggesting yesterday) then it might be useful to offer an API that is more powerful.

pkg/cgroups/cgroups.go
tummychow: yeah good catch

pkg/cgroups/cgroups.go
tummychow: The possibility of running on a system without cgroup support is interesting, I didn't think about it much at the time of writing this. For starters you probably wouldn't have cgexec on system so you'd be dead from that... is this worth supporting?

pkg/cgroups/cgroups.go
```diff
@@ -0,0 +1,155 @@
...
+				}
+			}
+		}
+	}
+
+	return ret, nil
+}
+
+var Default Cgroups = map[string]string{
+	"cpu":    "/cgroup/cpu",
```
anthonybishopric: I don't really understand why this is a `map[string]string` and not just a struct with cpu and memory fields.

pkg/hoist/hoist_executable.go
tummychow: I like this suggestion a lot but the change surface will be significantly larger. Let me think about the API and we can discuss it in a few hours, it's definitely possible without a giant refactor

pkg/cgroups/cgroups.go
tummychow: There are a lot of subsystems, just we happen to only use these two... I was considering a refactor to struct though and it's probably the right call. Will change

pkg/cgroups/cgroups.go
anthonybishopric: I think it's worth supporting given that right now the only reason you can't use `p2` on a non-Linux machine is due to its dependency on `runit`. Having the `NotSupported` error at least makes it more approachable to work on other platforms later if future authors know that this package will warn them appropriately. It's also fairly easy to do, which is why I suggest it.

tummychow:
Code should be good to go, but we had some concerns about changing the cgroup of the running application before invoking `sv restart`. If an application was scheduled with a smaller cgroup, it might be choked for resources and be unable to cleanup before being killed by runit. We're going to pursue another change that separates the stop and start, and if it works then we can come back and complete this PR.

On hold for now.

tummychow:
Actually the new branch is really different from this one (significant refactors) so I'm just gonna close this and roll a new PR.

PR #124, opened 2015-03-12
Stop and (re)start a launchable instead of waiting and then restarting

tummychow:


tummychow:
whoops, launchable.Halt already invokes disable, let me move the tweaks a little bit to reflect that behavior

tummychow:
Alright this should be good to go. We're coming towards a nice API design of `Install(new)`, `Halt(old)` and `Launch(new)` as our supported, external API for `pkg/pods.Pod`. `Verify(new)` is a bit of an outlier... The interaction and callgraph from pod to launchable is also (still) kinda messy, but this is a good improvement.

anthonybishopric:
Nice symmetry in the Hoist launchable API now. :+1:

PR #125, opened 2015-03-12
Cgroup reimplementation

tummychow:
I recommend you ignore the last two commits, those are exclusively renames. Peruse the diffs if you want, but they're almost all automatically generated.

This PR reimplements cgroup functionality while removing the dependency on contain. It incorporates all of the feedback from the previous PR, and uses the stop/start refactor to install cgroup configuration without affecting a live application instance.


tummychow:
Locate packages that import `path` with:

```bash
$ go list -f '{{ $pkg := . }}{{ range $elem := .Imports }}{{ if eq $elem "path" }}{{ $pkg.ImportPath }}{{ end }}{{ end }}' ./...
```

(replace `.Imports` with `.TestImports` to find packages that use it in tests)

We should refactor these packages to use `path/filepath` instead. Its API is a strict superset and it's OS-path-separator aware.

pkg/cgroups/cgroups.go
```diff
@@ -0,0 +1,172 @@
...
+			}
+		}
+	}
+
+	return ret, nil
+}
+
+var DefaultCgexec = "/bin/cgexec"
+
+func FakeCgexec() string {
```
anthonybishopric: We should move this to the test file or a test helper.

pkg/cgroups/cgroups.go
```diff
@@ -0,0 +1,172 @@
...
+
+// set the memory limit on a cgroup, 0 to unrestrict
+// https://www.kernel.org/doc/Documentation/cgroups/memory.txt
+func (subsys Subsystems) SetMemory(name string, bytes int) error {
+	if subsys.Memory == "" {
+		return UnsupportedError("memory")
+	}
+
+	softLimit := bytes
+	hardLimit := bytes * 11 / 10
```
anthonybishopric: Hopefully nobody writes anything with a limit of 1 byte.

pkg/cgroups/config.go
```diff
@@ -0,0 +1,29 @@
...
+	"fmt"
+)
+
+type Config struct {
+	Name   string `yaml:"-"`                // The name of the cgroup in cgroupfs
+	CPUs   int    `yaml:"cpus,omitempty"`   // The number of logical CPUs
+	Memory int    `yaml:"memory,omitempty"` // The number of bytes of memory
+}
+
+func (subsys Subsystems) Write(config Config) error {
```
anthonybishopric: Lets move this to the file where `Subsystems` is defined

pkg/hoist/hoist_executable.go
```diff
@@ -45,7 +36,7 @@ func (e HoistExecutable) SBEntry() []string {
 	return ret
 }
 
-func (e HoistExecutable) WriteExecutor(writer io.Writer) error {
+func (e Executable) WriteExecutor(writer io.Writer) error {
```
anthonybishopric: Good rename :star:

pkg/hoist/hoist_launchable.go
```diff
@@ -22,37 +22,36 @@ import (
...
 	// probably want to do something with output at some point
-	_, err := hoistLaunchable.Disable()
+	_, err := hl.disable()
 	if err != nil {
 		return err
 	}
 
 	// probably want to do something with output at some point
-	err = hoistLaunchable.Stop(serviceBuilder, sv)
+	err = hl.stop(serviceBuilder, sv)
```
anthonybishopric: I suspect we'll need to change the access level for `disable` and `stop` back to public at some point.

tummychow: Possibly, but we should try to stick to the high-level APIs (install/halt/launch) wherever possible. Unexporting the lower-level APIs emphasizes that using them is generally not correct. That level of granularity makes it too easy to violate our ordering guarantees. We should not have situations where a pod iterates over its launchables and invokes Stop or Start on all of them... and if we do, then we should refactor that before it gets merged.

anthonybishopric: Generally agreed, I believe there will be features in deployment that people will miss in p2 and we'll need to either educate them on how to do it the "p2 way" or implement the feature in p2 itself. For example, restarting a service without updating the version of the service on all hosts.

anthonybishopric:
:+1:

anthonybishopric:
Pending my suggestions, looks good

tummychow:
Resolved those comments (unfortunately they applied to the root commit of the branch so the interactive rebase splashed all 11 commits).

tummychow:
tests pass locally, travis is just being slow

PR #127, opened 2015-03-12
Rename PodManifest -> Manifest

tummychow:
Programmatic rename of PodManifest -> Manifest plus various other abbreviations that follow therefrom


anthonybishopric:
:+1:

PR #128, opened 2015-03-12
Fix cgroups hooks

tummychow:
The special-case behavior of hooks vs other pods definitely deserves a closer look in the future.

### DIFF
--- a/pkg/preparer/listener.go
+++ b/pkg/preparer/listener.go
@@ -9,6 +9,7 @@ import (
 	"strings"
 
 	"github.com/Sirupsen/logrus"
+	"github.com/square/p2/pkg/cgroups"
 	"github.com/square/p2/pkg/config"
 	"github.com/square/p2/pkg/kp"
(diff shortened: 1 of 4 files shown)


pkg/preparer/listener.go
```diff
@@ -191,13 +197,14 @@ func (l *HookListener) writeHook(event string, hookPod *pods.Pod, manifest *pods
 		}
 
 		for _, executable := range executables {
+			executable.Cgexec = "" // do not cgexec in a hook
```
anthonybishopric: As an aid to hook deployers, perhaps warn if the value of `Cgexec` is `!=""`

anthonybishopric:
:+1: with one minor nit

pkg/preparer/listener.go
tummychow: I'm kind of confused by this comment but I realized there's a good point in there regarding where we should set `cgexec=""`, will fix in a second

PR #129, opened 2015-03-13
Allow degraded health checks in replication rollouts

tummychow:
Verified in universe - I set the wrong status port so all the health checks would be critical, then I set the replicate threshold to critical, and the rollout completed successfully.

### DIFF
--- a/bin/p2-replicate/main.go
+++ b/bin/p2-replicate/main.go
@@ -35,6 +35,7 @@ var (
 	minNodes    = replicate.Flag("min-nodes", "The minimum number of healthy nodes that must remain up while replicating.").Default("1").Short('m').Int()
 	consulUrl   = replicate.Flag("consul", "The hostname and port of a consul agent in the p2 cluster. Defaults to 0.0.0.0:8500.").String()
 	consulToken = replicate.Flag("token", "The ACL token to use for consul").String()
+	threshold   = replicate.Flag("threshold", "The minimum health level to treat as healthy. One of (in order) passing, warning, unknown, critical.").String()
 )
 
 func main() {
@@ -53,7 +54,14 @@ func main() {
 	// the error is always nil
 	client, _ := api.NewClient(conf)
 
-	healthChecker := health.NewConsulHealthChecker(*store, client.Health())
+	inner := health.NewConsulHealthChecker(*store, client.Health())
+	var healthChecker replication.ServiceChecker = inner
+	if *threshold != "" {
+		healthChecker = replication.ServiceUpgrader{
+			Inner:     inner,
+			Threshold: health.ToHealthState(*threshold),
+		}
+	}
 
 	// Fetch manifest (could be URI) into temp file
 	localMan, err := ioutil.TempFile("", "tempmanifest")
--- a/pkg/replication/rollout_order.go
+++ b/pkg/replication/rollout_order.go
@@ -23,14 +23,6 @@ func (c compare) less() bool {
 	return c == lessThan
 }
 
-func compareFromBool(b bool) compare {
-	if b {
-		return lessThan
-	} else {
-		return greaterThan
-	}
-}
-
 type rolloutOrder struct {
 	nodes           []allocation.Node
 	referenceStatus *health.ServiceStatus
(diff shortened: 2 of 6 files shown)


pkg/replication/rollout_order.go
```diff
@@ -88,19 +80,19 @@ func (r *rolloutOrder) compareErrors(iErr, jErr error) compare {
...
-	if iHealth.Healthy && jHealth.Healthy {
-		return compareFromBool(iNode.Name < jNode.Name)
-	}
-	if !iHealth.Healthy && jHealth.Healthy {
-		return lessThan
-	}
-	if !iHealth.Healthy && !jHealth.Healthy {
-		return compareFromBool(iNode.Name < jNode.Name)
+	comp := health.Compare(iHealth.Health, jHealth.Health)
+	if comp == 0 {
```
anthonybishopric: `comp == equal`

anthonybishopric:
So now the other half is getting Consul to read the `P2-Status` header. Good update to the sorting algorithm for `p2-replicate`

pkg/replication/rollout_order.go
tummychow: That will fail because `equal` is of type compare and `comp` is of type int. I guess you can cast it. Personally I don't see the value in having an extra type just for these three constant integers.

tummychow:
I'm not so sure about the ordering in this PR now... obviously passing>warning>critical but I don't know if unknown can even be compared to other statuses. This has impact on the next part of the PR (do not replicate to unhealthy nodes). Needs more thought

bin/p2-replicate/main.go
```diff
@@ -77,6 +87,27 @@ func main() {
...
+		status, err := healthChecker.LookupHealth(manifest.ID())
+		if err != nil {
+			replicator.Logger.WithField("err", err).Errorln("Could not retrieve health checks to resolve skipped nodes")
+			os.Exit(1)
+		}
+
+		var newHosts []string
+		for _, host := range *hosts {
+			thisHealth, _ := status.ForNode(host)
+			if thisHealth.Health != health.Passing && thisHealth.Health != health.Unknown {
```
anthonybishopric: Shouldn't this if statement be skipping the `append` if it evaluates to true? Also, shouldn't the levels it checks against be the one specified by the `ServiceUpgrader`?

bin/p2-replicate/main.go
```diff
@@ -77,6 +87,27 @@ func main() {
...
+				replicator.Logger.WithFields(logrus.Fields{
+					"host":   host,
+					"health": thisHealth.Health,
+				}).Infoln("Skipping unhealthy host")
+			}
+			newHosts = append(newHosts, host)
+		}
+		replicator.Allocation = allocation.NewAllocation(newHosts...)
```
anthonybishopric: I'm having a somewhat difficult time wrapping my head around the `skip` semantics. Is skip intended to ignore hosts that are as or less healthy than the `threshold`? I imagine you'd not want to skip the nodes if they are out of date anyway.

bin/p2-replicate/main.go
anthonybishopric: Ignore the second comment, I realize that `health.Passing` is assigned to nodes depending on the threshold.

tummychow: That's a good question, now I'm not sure how it even worked. Will fix.

bin/p2-replicate/main.go
tummychow: Skip ignores nodes that are strictly less than the threshold (ie the default value of the threshold is passing, so skip ignores warn/critical). I agree the semantics are rather confusing because of how the two flags interact... this was the most logical approach I could come up with.

I don't really see the point of having a skip at all if you would still update out-of-date unhealthy nodes. What's the intended use case for skipping unhealthy nodes? My impression was that we'd probably use this flag if a node was so damaged that it could not take a deploy, and the deployer doesn't want it to block a deployment to the other healthy nodes. Were there other ideas in mind?

bin/p2-replicate/main.go
tummychow: fixed and verified

tummychow:
discarded skip functionality as we discussed. I'd still like another review because I'm kinda iffy on the ordering semantics of passing>warning>unknown>critical

PR #130, opened 2015-03-16
Move to github.com/hashicorp/consul/api

tummychow:
The repo we were using has been deprecated and won't receive updates for consul v0.5.0 API features. The change surface is very very small so we might as well migrate now. No impact in godep tests, integration tests or universe


anthonybishopric:
:+1:

PR #131, opened 2015-03-16
Custom HTTP headers

tummychow:
Like so:

`p2-schedule --header X-Foo=bar --header X-Qux=baz manifest.yaml`

### DIFF
--- a/bin/p2-inspect/main.go
+++ b/bin/p2-inspect/main.go
@@ -9,6 +9,7 @@ import (
 
 	"github.com/hashicorp/consul/api"
 	"github.com/square/p2/pkg/kp"
+	"github.com/square/p2/pkg/util/net"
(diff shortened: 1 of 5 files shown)


pkg/kp/client.go
```diff
@@ -0,0 +1,36 @@
+package kp
```
anthonybishopric: This is a useful file but maybe it should be in a `pkg/util/netutil`?

anthonybishopric: or just `pkg/netutil`

anthonybishopric:
:+1: very useful for the various headers exposed by Consul like `X-Consul-LastContact`

pkg/kp/client.go
tummychow: I'll move it to `pkg/util/net`

PR #132, opened 2015-03-17
Schedule consul in intent and reality during bootstrap

tummychow:
First step in making consul scheduleable. Confirmed that preparer does not attempt to restart consul immediately after bootstrapping.


anthonybishopric:
woot :tada:

PR #133, opened 2015-03-17
Refactor inspect to use pkg/health

tummychow:
Allows inspect to gracefully react to multiple health checks on a single service, choosing the poorest of them.

### DIFF
--- a/bin/p2-inspect/main.go
+++ b/bin/p2-inspect/main.go
@@ -8,6 +8,7 @@ import (
 	"strings"
 
 	"github.com/hashicorp/consul/api"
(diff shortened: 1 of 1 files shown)


anthonybishopric:
:+1: but we'll probably eventually want some way to quickly distinguish what healtcheck is failing (node-wide or service)

tummychow:
Yeah that's something I was worried about. There must be a point at which curling consul directly makes more sense than using p2-inspect, but I'm not really sure where that line can be drawn. It would probably also lead to some changes in pkg/health.

PR #134, opened 2015-03-18
Fix broken tests on master

tummychow:
1. Apparently travis VMs are so entropy-starved that they can't always generate the first RSA keypair that we cached for all subsequent tests. Switch to an insecure RNG instead to avoid blocking on reads to /dev/urandom.

2. our fake hooks implementation didn't support AFTER_AUTH_FAIL, which was getting triggered after the read to /dev/urandom timed out and the manifest did not get signed. We don't really exercise this right now but maybe we should in the future. Might as well fix it.


anthonybishopric:
:+1:

tummychow:
Okay so crypto/rand.Prime is trying to read the random number generator until it can find a prime of the desired size, and the insecure PRNG is taking so long to return a prime that travis is still timing out. Way to go travis. I'll hardcode the entity keypairs instead of generating them here and see how that goes instead.

PR #135, opened 2015-03-18
Do not return errors directly from consul API

tummychow:
Prevents ACL leakage. Unfortunately consul does not return typed errors, so there's really no way to get at the contents of the error value and take out the things we don't want. We just have to throw the whole thing away.

### DIFF
--- a/pkg/kp/kv.go
+++ b/pkg/kp/kv.go
@@ -4,6 +4,7 @@ package kp
 
 import (
 	"bytes"
+	"fmt"
 	"net/http"
 	"time"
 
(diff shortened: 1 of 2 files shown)


pkg/kp/kv.go
```diff
@@ -37,6 +38,18 @@ func NewStore(opts Options) *Store {
 	return &Store{client: client}
 }
 
+// KVError encapsulates an error in a Store operation. Errors returned from the
+// Consul API cannot be exposed because they may contain the URL of the request,
+// which includes an ACL token as a query parameter.
```
xaviershay: this is so derpy. File a bug against consul for it?

tummychow: already done, went on the mailing list and armon opened https://github.com/hashicorp/consul/issues/791. But it might be a while before they get it fixed. If I was the one who had to fix it I'd probably do so grudgingly while grumbling about special cases for *that one* query parameter, so I can't blame them.

anthonybishopric: I wish we could at least extract the HTTP status code, but I think I'm okay without it so long as we can still execute `consul monitor`. Eventually though this really needs to get addressed

tummychow: Yeah you can't really figure out the contents of the error without reflection. Consul returns an `fmt.Errorf` for bad response codes so you'd actually have to parse a string to find it. I wish more go libraries were disciplined about the use of custom error values or types, but unfortunately this is not one of those cases. *glances nervously at our own `util.Errorf`*

PR #136, opened 2015-03-19
Sort inspect locations

tummychow:
when scrolling through a huge list of p2-inspect output for a given pod, it's much easier to figure out which location you care about if they're in a consistent order.


anthonybishopric:
:+1:

PR #137, opened 2015-03-22
Verify reality store during bootstrap

anthonybishopric:
Verifies that the Preparer and Consul both make it into the reality store. If ACLs are misconfigured for the preparer, this step will catch such an issue.


tummychow:
:+1:

PR #139, opened 2015-03-23
HTTPS in p2 binaries

tummychow:
I was originally going to add it as a scheme (eg `127.0.0.1:8500` vs `https://127.0.0.1:8500`) but it turns out that doesn't really work. Consul supports unix sockets in the address field (ie instead of `127.0.0.1` you can say `unix:///var/run/consul.sock`), and the unix socket scheme kinda gets in the way of the https part. So you have to pass it as a boolean.

The changes to health.NewConsulHealthChecker will be useful if we refactor the reality-checking out of pkg/health later on.


petertseng:
OK looking good. Any interest in adding to p2-watch as well?

tummychow:
I've really been neglecting p2-watch cause I'm not sure if it's a good fit for remote use, but fair point. I'll add the various flags (token, headers, https) to it in another PR.

PR #140, opened 2015-03-25
Check reality store before replicating

tummychow:
don't replicate to hosts that don't have a preparer. (btw not yet tested, gonna test tomorrow in universe. I think they've been down for most of the day)

### DIFF
--- a/bin/p2-replicate/main.go
+++ b/bin/p2-replicate/main.go
@@ -77,6 +77,13 @@ func main() {
 		log.Fatalf("Invalid manifest: %s", err)
 	}
 
+	for _, host := range *hosts {
+		_, _, err := store.Pod(kp.RealityPath(host, "p2-preparer"))
+		if err != nil {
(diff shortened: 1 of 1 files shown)


bin/p2-replicate/main.go
```diff
@@ -77,6 +77,13 @@ func main() {
 		log.Fatalf("Invalid manifest: %s", err)
 	}
 
+	for _, host := range *hosts {
+		_, _, err := store.Pod(kp.RealityPath(host, "p2-preparer"))
+		if err != nil {
+			log.Fatalf("Could not verify preparer on %s: %s", host, err)
```
anthonybishopric: Most people won't understand what this means - I'd suggest `"p2 does not appear to be running on host %s"`

tummychow:
tested in universe, unknown nodes are rejected and otherwise replication proceeds. merging

PR #141, opened 2015-03-27
Register consul service during bootstrap

tummychow:
Consul server nodes register a service for themselves automatically. Consul client nodes do not, but we end up registering one for them anyway as part of the reality-updating process. So we should probably do that in bootstrap too.


anthonybishopric:
Looks good, check that the integration test still works

PR #142, opened 2015-03-27
Hide error from SetPod when writeMeta != nil

tummychow:
This error comes from the consul API package, so it cannot be exposed to the caller


anthonybishopric:
:+1:

PR #143, opened 2015-03-27
ensure pod home before running hooks

anthonybishopric:


tummychow:
:+1:

petertseng:
LGTM

anthonybishopric:
Unfortunately this breaks the case when the `before_install` hook is used to create the user, because `EnsureHome` creates the home directory with the correct permissions.

PR #144, opened 2015-03-28
Add artifact cacher

anthonybishopric:
To be integrated, this will allow us to reduce unnecessarily download times when deploying shared artifacts.

### DIFF
--- a/pkg/hoist/artifact_cache.go
+++ b/pkg/hoist/artifact_cache.go
@@ -1,0 +1,117 @@
+package hoist
+
+import (
+	"io/ioutil"
+	"os"
+	"path/filepath"
+	"sort"
+	"time"
+
+	"github.com/Sirupsen/logrus"
+	"github.com/square/p2/pkg/logging"
+	"github.com/square/p2/pkg/uri"
+	"github.com/square/p2/pkg/util"
+)
+
+// ArtifactCacher keeps a directory of Hoist artifacts that can be reused for future downloads.
+// This allows artifacts to be cheaply shared amongst different pods, especially in cases where
+// the artifacts contain base libraries and runtimes. For example, a pod may include a Java 8
+// artifact that is reused across other pods.
+//
+// The cache assumes that Hoist artifact URIs are unique at the basename, ie. comply with the
+// convention of {artifact_name}_{artifact_SHA}.tar.gz pattern. No attempt is made to distinguish
+// artifacts that are named the same but contain different content. (Checksumming is presumed to
+// be done by the Launchable itself)
+//
+// Once a cached copy of an artifact is created, it will be hardlinked into the final destination
+// URI provided by the caller. This allows the ArtifactCacher to safely prune without worrying about
+// whether the artifact is still in use by a particular Launchable (say, as the Last link). However,
+// as a result of this strategy, pruning the cache dir will not be sufficient to free system disk
+// resources as other links to the same tars will also need to be removed.
+type ArtifactCacher struct {
+	cacheDir      string
+	nestedFetcher Fetcher
+	Logger        *logging.Logger
+}
+
+// Complies with the Fetcher interface
+func (a *ArtifactCacher) Fetch(fromURI, toURI string) error {
+	base := filepath.Base(toURI)
+	if base == "." || base == string(filepath.Separator) {
+		return util.Errorf("%s is not a valid toURI", toURI)
+	}
+	cachePath := filepath.Join(a.cacheDir, base)
+	if _, err := os.Stat(cachePath); os.IsNotExist(err) {
+		a.nestedFetcher(fromURI, cachePath)
+	} else if err != nil {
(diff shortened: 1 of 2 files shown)


pkg/hoist/artifact_cache.go
```diff
@@ -0,0 +1,117 @@
...
+// resources as other links to the same tars will also need to be removed.
+type ArtifactCacher struct {
+	cacheDir      string
+	nestedFetcher Fetcher
+	Logger        *logging.Logger
+}
+
+// Complies with the Fetcher interface
+func (a *ArtifactCacher) Fetch(fromURI, toURI string) error {
+	base := filepath.Base(toURI)
```
tummychow: `toURI` should probably be `toPath`? That would match the signature of the existing uri.URICopy.

pkg/hoist/artifact_cache.go
```diff
@@ -0,0 +1,117 @@
...
+		err = a.touch(cachePath)
+		if err != nil {
+			a.Logger.WithField("err", err).Warnln("Error when updating mtime on cached artifact")
+		}
+	}
+	// hardlink so we can safely prune without breaking Last/Current links
+	linkErr := os.Link(cachePath, toURI)
+	if linkErr != nil {
+		// we may be on different mounts, just copy
+		err := uri.URICopy(cachePath, toURI)
```
tummychow: cachePath does not have a scheme. will URICopy work with a schemeless fromURI?

pkg/hoist/artifact_cache.go
```diff
@@ -0,0 +1,117 @@
...
+func (b byMtime) Swap(i, j int)      { b[i], b[j] = b[j], b[i] }
+func (b byMtime) Less(i, j int) bool { return b[i].ModTime().Before(b[j].ModTime()) }
+
+// Prune will remove all but `keep` # of artifacts
+// from the cache dir. Prune will favor more recently modified artifacts
+func (a *ArtifactCacher) Prune(keep int) {
+	infos, err := ioutil.ReadDir(a.cacheDir)
+	if err != nil {
+		err = util.Errorf("cache dir unlistable: %s", err)
+		a.Logger.WithField("err", err).Errorln("Pruning failed")
```
tummychow: why not just return these errors from `Prune` and delegate logging to the caller?

pkg/hoist/artifact_cache.go
```diff
@@ -0,0 +1,117 @@
...
+
+type byMtime []os.FileInfo
+
+func (b byMtime) Len() int           { return len(b) }
+func (b byMtime) Swap(i, j int)      { b[i], b[j] = b[j], b[i] }
+func (b byMtime) Less(i, j int) bool { return b[i].ModTime().Before(b[j].ModTime()) }
+
+// Prune will remove all but `keep` # of artifacts
+// from the cache dir. Prune will favor more recently modified artifacts
+func (a *ArtifactCacher) Prune(keep int) {
```
tummychow: General thought: our pruning hook uses a different heuristic of "prune artifacts older than a certain age", instead of "keep all but a certain number". Should we use the age-based heuristic here as well?

pkg/hoist/artifact_cache_test.go
```diff
@@ -0,0 +1,108 @@
...
+func TestArtifactCacheWillTouchExistingArtifactsOnAccess(t *testing.T) {
+	cacher := tempCacher(t)
+	defer cleanupCacher(cacher)
+	preCopiedPath := filepath.Join(cacher.cacheDir, filepath.Base(helloTar()))
+	err := uri.URICopy(helloTar(), preCopiedPath)
+	Assert(t).IsNil(err, "test setup - couldn't prewarm cache")
+	aYearAgo := time.Now().Add(-365 * 24 * time.Hour)
+	anHourAgo := time.Now().Add(-time.Hour)
+
+	fmt.Println(aYearAgo)
```
petertseng: suspect you don't want to keep these

pkg/hoist/artifact_cache.go
anthonybishopric: It uses both age and count, by sorting by the mtime before removing

tummychow: I mean "keep an arbitrary number of artifacts as long as they're younger than a certain age", that's what the prune hook does right now.

pkg/hoist/artifact_cache.go
anthonybishopric: Yes, it assumes a scheme-less URI is a path. Also, there is no `os.Copy`, which makes this a bit easier.

pkg/hoist/artifact_cache.go
anthonybishopric: good catch

pkg/hoist/artifact_cache_test.go
anthonybishopric: yep, correct

PR #145, opened 2015-04-01
Restart all executables

tummychow:
If the error was nil, it would be returned after the first executable, and the others would still be down. (Probably wasn't as obvious of an issue before because we only recently started stopping AND starting)


anthonybishopric:
:+1: doh

PR #147, opened 2015-04-02
cleanup some logging statements

anthonybishopric:


mpuncel:
:+1:

tummychow:
this will probably silence the warning about "couldn't run hooks" when none exist, right? good change :+1:

PR #148, opened 2015-04-08
p2-exec

tummychow:


bin/p2-exec/main.go
```diff
@@ -0,0 +1,163 @@
...
+	"gopkg.in/yaml.v2"
+)
+
+// usage without cgroups:
+//   p2-exec username envdir launchable:cgname myexecutable myargs
+// there must be a PLATFORM_CONFIG_PATH envar (possibly in the envdir) to
+// specify the cgroup parameters for the <launchable>, which is created under
+// <cgname>
+func main() {
+	username := os.Args[1]
```
anthonybishopric: Reason for straying from kingpin? It supports args and I'd like to see a usage when executed, even if this is executed automatically most of the time.

bin/p2-exec/main_darwin.go
```diff
@@ -0,0 +1,77 @@
...
+	}
+	_, err = C.setrlimit(C.RLIMIT_RSS, unlimit)
+	if err != nil {
+		return err
+	}
+	return nil
+}
+
+func changeUser(username string) error {
+	uid, gid, err := user.IDs(username)
```
anthonybishopric: We are probably going to want to pass users and groups separately in the future. An ask has come from infosec that we execute applications with files owned by p2 but the group id being the user. This makes writing new files at install time difficult and thus discourages tampering.

tummychow: This sets the user's primary group (GID) and supplementary groups correctly. Since the GID is set, the app would be able to read files owned by its group even if the owning uid was different. Are there actually other groups we need to set?

bin/p2-exec/main.go
tummychow: mostly a matter of laziness and the fact that this program is really human-unfriendly anyway, but kingpin would make it easier to separate the launchablename and cgroupname so I'll implement that. Gonna cause a lot of churn in the PR though so review the rest of it and I'll amend the commits afterwards

bin/p2-exec/main.go
```diff
@@ -0,0 +1,163 @@
...
+	if err != nil {
+		return err
+	}
+
+	for _, envFile := range envFiles {
+		if envFile.IsDir() {
+			continue
+		}
+		if strings.IndexByte(envFile.Name(), '=') != -1 {
+			return util.Errorf("env file %q cannot contain equals sign in its filename", filepath.Join(dir, envFile.Name()))
```
anthonybishopric: Are there other chars that need to be prevented?

tummychow: No, just the equals sign because it separates the key from the value in the representation of the environment (which is a list of null-separated strings where each string is a key, an equals sign and a value). Other characters (eg non-printing control chars, unicode) would be strange, but from a technical standpoint, they would work fine.

bin/p2-exec/main_darwin.go
anthonybishopric: Ah sorry, I am thinking of artifact unpacking. The process running as `username:username` probably won't change.

integration/common-provision.sh
```diff
@@ -24,13 +24,11 @@ go get github.com/tools/godep
 # Build p2.
 cd $GOPATH/src/github.com/square/p2
 godep go install ./...
+cp $GOPATH/bin/p2-exec /usr/local/bin
 
 # Install ruby + rake
 sudo yum install -y ruby rubygem-rake
 
-# Install cgexec
-sudo yum install -y libcgroup libcgroup-tools
```
anthonybishopric: you don't still need `libcgroup`?

tummychow: it was only needed as a dependency of cgexec (libcgroup-tools) and in fact probably we never had to explicitly put it there at all. (I don't know how yum works so idk)

bin/p2-exec/main_darwin.go
```diff
@@ -0,0 +1,77 @@
...
+	// #include <pwd.h>
+	// #include <grp.h>
+	"C"
+	"unsafe"
+
+	"github.com/square/p2/pkg/user"
+)
+
+func nolimit() error {
+	maxFDs := C.OPEN_MAX
```
eam: OPEN_MAX is usually set to some unreasonably low value. Try something like this instead: https://github.com/square/prodeng/blob/master/nolimit/nolimit.c#L65-L75

bin/p2-exec/main.go
```diff
@@ -0,0 +1,163 @@
...
+		enval = strings.TrimRight(enval, " \t")
+		// replace nulls with newlines
+		enval = strings.Map(func(c rune) rune {
+			if c == 0 {
+				return '\n'
+			}
+			return c
+		}, enval)
+
+		err = os.Setenv(envFile.Name(), enval)
```
anthonybishopric: per @eam's recommendation, we should probably explicitly clear the environment prior to setting these. While runit's environment will probably be clean most of the time, we probably want to avoid surprises in the future and we have an opportunity to do that here.

Granted, if we do that we will also need to add all the environment variables hooks use to the hook's env dir, but that seems like a fairly simple change.

tummychow: as we discussed, this behavior has been implemented in p2-exec as a flag, but p2 itself will not opt into it yet

PR #150, opened 2015-04-10
Use full 20-byte GPG fingerprints

tummychow:
GPG short IDs are only 4 bytes long and can risk collisions. Using the entire fingerprint (20 bytes) is more predictable.

http://security.stackexchange.com/questions/74009/what-is-an-openpgp-key-id-collision


mpuncel:
LGTM

PR #151, opened 2015-04-11
Move setuid/setgid/exec into cgo

tummychow:
tested a bit on universe, also works on macbook, I have to add back the functionality that makes the flags optional but please review. cc @eam because my C is terrible and could probably use some criticism. there's probably a dumb memory leak in there or something.

ps: shoutout to https://groups.google.com/forum/#!topic/golang-nuts/pQueMFdY0mk

### DIFF
--- a/bin/p2-exec/main_darwin.go
+++ b/bin/p2-exec/main_darwin.go
@@ -2,76 +2,72 @@ package main
 
 import (
 	// #include <stdlib.h>
+	// #include <stdio.h>
 	// #include <unistd.h>
 	// #include <sys/resource.h>
 	// #include <sys/types.h>
-	// #include <sys/syslimits.h>
 	// #include <pwd.h>
 	// #include <grp.h>
+	// #include <sys/syslimits.h>
+	//
+	// char** stringSlice(int size) {
+	//   char** ret = calloc(sizeof(char*), size+1);
+	//   ret[size] = NULL;
+	//   return ret;
+	// }
+	// void setStringSlice(char **a, char *s, int n) {
+	//   a[n] = s;
+	// }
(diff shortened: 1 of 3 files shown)


bin/p2-exec/main_darwin.go
```diff
@@ -2,76 +2,72 @@ package main
...
+	//   if (setrlimit(RLIMIT_CPU, &lim)) { return "Could not set RLIMIT_CPU"; }
+	//   if (setrlimit(RLIMIT_DATA, &lim)) { return "Could not set RLIMIT_DATA"; }
+	//   if (setrlimit(RLIMIT_FSIZE, &lim)) { return "Could not set RLIMIT_FSIZE"; }
+	//   if (setrlimit(RLIMIT_MEMLOCK, &lim)) { return "Could not set RLIMIT_MEMLOCK"; }
+	//   if (setrlimit(RLIMIT_NPROC, &lim)) { return "Could not set RLIMIT_NPROC"; }
+	//   if (setrlimit(RLIMIT_RSS, &lim)) { return "Could not set RLIMIT_RSS"; }
+	//
+	//   pw = getpwnam(username);
+	//   if (!pw) { return "Could not getpwnam"; }
+	//   if (initgroups(username, pw->pw_gid)) { return "Could not initgroups"; }
```
anthonybishopric: It'd be nice to emit the errno here, if it does come up. I'm still unclear on how this fixes the race condition - is this reliant on Go's own mutex it places around CGo invocations?

anthonybishopric: Nevermind, I understand that it assigns to `errno` anyway

tummychow:
discarding in favor of #152

PR #152, opened 2015-04-11
Handle return values in p2-exec before checking errno

tummychow:
plus various refactors. Thanks @syamp. cc @anthonybishopric @eam

### DIFF
--- a/bin/p2-exec/main.go
+++ b/bin/p2-exec/main.go
@@ -1,6 +1,9 @@
 package main
 
 import (
+	// #include <sys/resource.h>
+	"C"
+
 	"io/ioutil"
 	"log"
 	"os"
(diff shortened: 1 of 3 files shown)


anthonybishopric:
:+1:

syamp:
You want to check ret value first and then in the sub block retrieve err if present.

anthonybishopric:
@syamp how is that different from short circuiting the `ret !=0` part of the if statement?

anthonybishopric:
oh i suppose `ret` may be `!= 0` but error could be nil, in which case this would falsely say it was successful

syamp:
The idea is not bother about checking for errno unless return value is not as expected.  return
value should be your only signal. In this particular case, it might not matter but is a wrong pattern imo.

From manpage of errno:
----
 Its value is significant only
       when the return value of the call indicated an error (i.e., -1 from
       most system calls; -1 or NULL from most library functions); a
       function that succeeds is allowed to change errno.

PR #153, opened 2015-04-13
Mpuncel/run as

mpuncel:
@anthonybishopric @petertseng @tummychow

### DIFF
--- a/pkg/preparer/listener.go
+++ b/pkg/preparer/listener.go
@@ -10,7 +10,6 @@ import (
 
 	"github.com/Sirupsen/logrus"
 	"github.com/square/p2/pkg/cgroups"
-	"github.com/square/p2/pkg/config"
 	"github.com/square/p2/pkg/kp"
 	"github.com/square/p2/pkg/logging"
 	"github.com/square/p2/pkg/pods"
@@ -80,23 +79,6 @@ func (l *HookListener) Sync(quit <-chan struct{}, errCh chan<- error) {
 			}
 
 			hookPod := pods.NewPod(result.Manifest.ID(), path.Join(l.DestinationDir, event, result.Manifest.ID()))
-
-			// Hooks may specify a user to run as - by default they will run as their pod ID.
-			topLevelConfig := config.LoadFromUnpacked(result.Manifest.Config)
-			hookConfig, err := topLevelConfig.ReadMap("hook")
-			if err != nil {
-				sub.WithField("err", err).Errorln("Hook configuration is invalid")
-				break
-			}
-			runAs, err := hookConfig.ReadString("run_as")
-			if err != nil {
-				sub.WithField("err", err).Errorln("Hook configuration is invalid")
-				break
-			}
-			if runAs == "" {
-				runAs = hookPod.Id
-			}
-			hookPod.RunAs = runAs
 
 			// Figure out if we even need to install anything.
 			// Hooks aren't running services and so there isn't a need
--- a/pkg/preparer/orchestrate.go
+++ b/pkg/preparer/orchestrate.go
@@ -155,13 +155,6 @@ func (p *Preparer) handlePods(podChan <-chan pods.Manifest, quit <-chan struct{}
 			if working {
 				pod := pods.NewPod(manifestToLaunch.ID(), pods.PodPath(p.podRoot, manifestToLaunch.ID()))
 
-				// HACK ZONE. When we have better authz, rewrite.
-				// Still need to ensure that preparer launches correctly
-				// as root
-				if pod.Id == POD_ID {
-					pod.RunAs = "root"
-				}
-
 				ok := p.installAndLaunchPod(&manifestToLaunch, pod, manifestLogger)
 				if ok {
 					manifestToLaunch = pods.Manifest{}
@@ -213,8 +206,11 @@ func (p *Preparer) verifySignature(manifest pods.Manifest, logger logging.Logger
 func (p *Preparer) installAndLaunchPod(newManifest *pods.Manifest, pod Pod, logger logging.Logger) bool {
 	// do not remove the logger argument, it's not the same as p.Logger
 
-	if p.podIdForbidden(newManifest.ID()) {
-		logger.WithField("pod", newManifest.ID()).Errorln("Cannot use this pod ID")
+	if newManifest.ID() != POD_ID && p.podIdForbidden(newManifest.RunAsUser()) {
+		logger.WithFields(logrus.Fields{
+			"pod":    newManifest.ID(),
+			"run_as": newManifest.RunAsUser(),
+		}).Errorln("Invalid run_as user")
 		p.tryRunHooks(hooks.AFTER_AUTH_FAIL, pod, newManifest, logger)
 		return false
 	}
@@ -229,7 +225,7 @@ func (p *Preparer) installAndLaunchPod(newManifest *pods.Manifest, pod Pod, logg
 	newSHA, _ := newManifest.SHA()
 
 	// if new or the manifest is different, launch
-	newOrDifferent := err == pods.NoCurrentManifest || currentSHA != newSHA
+	newOrDifferent := (err == pods.NoCurrentManifest) || (currentSHA != newSHA)
 	if newOrDifferent {
 		logger.WithFields(logrus.Fields{
 			"old_sha": currentSHA,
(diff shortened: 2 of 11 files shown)


pkg/preparer/orchestrate.go
```diff
@@ -150,22 +151,28 @@ func (p *Preparer) handlePods(podChan <-chan pods.Manifest, quit <-chan struct{}
...
-				// Still need to ensure that preparer launches correctly
-				// as root
-				if pod.Id == POD_ID {
-					pod.RunAs = "root"
-				}
-
-				ok := p.installAndLaunchPod(&manifestToLaunch, pod, manifestLogger)
-				if ok {
-					manifestToLaunch = pods.Manifest{}
+				if (manifestToLaunch.RunAsUser() == "root") && (pod.Id != POD_ID) {
```
mpuncel: I was kind of wondering if this logic should go inside of installAndLaunchPod, but I guess if you call that it should mean you have the intention of launching it. Also the way we've redefined the Pod interface in this file means I can't get to pod.RunAs or pod.Id in the other function

pkg/preparer/orchestrate_test.go
```diff
@@ -89,7 +95,7 @@ func (f *fakeHooks) RunHookType(hookType hooks.HookType, pod hooks.Pod, manifest
 		f.ranAfterLaunch = true
 		return f.afterLaunchErr
 	case hooks.AFTER_AUTH_FAIL:
-		f.ranAfterAuthFail = true
+		f.ranAfterAuthFail <- struct{}{}
```
mpuncel: I'm worried that this will be a gotcha for people writing new tests. If their test case calls the after auth fail hook then they can deadlock their test

mpuncel: The alternative is to keep this a bool and do something like

```    waiting := true
    go func() {
        fmt.Println("waiting for a sec")
        <-time.After(1 * time.Second)
        waiting = false
    }()

    success := false
    for waiting && !success {
        if hooks.ranAfterAuthFail {
            fmt.Println("after auth fail ran")
            success = true
        }
    }```

File: pkg/preparer/orchestrate_test.go [hunk 3]
```diff
@@ -360,3 +366,22 @@ func TestPreparerWillAcceptSignatureWhenKeyringIsNil(t *testing.T) {
...
+
+	p, hooks, _ := testPreparer(t, &FakeStore{})
+	podChan := make(chan pods.Manifest)
+	quitChan := make(chan struct{})
+	go p.handlePods(podChan, quitChan)
+
+	podChan <- illegalManifest
+
+	// wait a second to see if auth fail hook is run. Unfortunately can't assert that pod wasn't installed
+	// because the pod is created inside p.handlePods()
```
mpuncel: I can't get a handle on the pod to make sure installation didn't happen because it is created within `p.handlePods`. I'm unable to test that the preparer is allowed to run as root either without making the test take at least a second for the run after auth hook to NOT be called

pkg/preparer/orchestrate.go
anthonybishopric: Why not add `RunAs` to the interface if you need it?

tummychow: I would support a refactor to move the authz logic into installAndLaunchPod. This function is difficult to test (as we're finding out here...) so we should minimize the amount of actual work it does.

I'm also not convinced that the sleep is semantically equivalent to selecting on a timed channel. Was that change necessary for testability?

anthonybishopric: Also, this should be testing if you're in the prohibited users list, not just "root"

pkg/preparer/orchestrate.go
```diff
@@ -150,22 +151,28 @@ func (p *Preparer) handlePods(podChan <-chan pods.Manifest, quit <-chan struct{}
 			if !working {
 				p.tryRunHooks(hooks.AFTER_AUTH_FAIL, pods.NewPod(manifestToLaunch.ID(), pods.PodPath(p.podRoot, manifestToLaunch.ID())), &manifestToLaunch, manifestLogger)
 			}
-		case <-time.After(1 * time.Second):
```
anthonybishopric: I think we should always be pausing here. There are lots of cases where the work done inside the `working` block will terminate quickly in case of errors.

pkg/preparer/orchestrate.go
mpuncel: ah yeah you're right that it's not equivalent. Taking the `sleep()` out of the `else` would fix that.  And yeah in its old form my test would always take a second or more because this function always waits at least a second to do work. If I do the refactor I can put it back the way it was.

Although I think the intention of the code is clearer if we use a sleep?

pkg/preparer/orchestrate.go
```diff
@@ -212,6 +208,14 @@ func (p *Preparer) verifySignature(manifest pods.Manifest, logger logging.Logger
 func (p *Preparer) installAndLaunchPod(newManifest *pods.Manifest, pod Pod, logger logging.Logger) bool {
 	// do not remove the logger argument, it's not the same as p.Logger
 
+	// HACK ZONE. When we have better authz, rewrite.
+	if (newManifest.RunAsUser() == "root") && (newManifest.ID() != POD_ID) {
```
anthonybishopric: Still need to compare with the list of prohibited user IDs, instead of just root.

mpuncel: that's still there (below this)

anthonybishopric: The existing test is `p.podIdForbidden(newManifest.ID())`. But that same test needs to apply to `RunAs`.

mpuncel:
ready for another look

pkg/preparer/listener.go
```diff
@@ -81,23 +80,6 @@ func (l *HookListener) Sync(quit <-chan struct{}, errCh chan<- error) {
...
-			}
-			runAs, err := hookConfig.ReadString("run_as")
-			if err != nil {
-				sub.WithField("err", err).Errorln("Hook configuration is invalid")
-				break
-			}
-			if runAs == "" {
-				runAs = hookPod.Id
-			}
-			hookPod.RunAs = runAs
```
anthonybishopric: So `manifest.RunAs()` will work for hooks too? Let's verify that's the case, nice improvement.

tummychow:
looks pretty good overall, nothing jumped out at me but i'll probably make another pass in a couple of hours

pkg/preparer/orchestrate.go
```diff
@@ -212,6 +206,14 @@ func (p *Preparer) verifySignature(manifest pods.Manifest, logger logging.Logger
 func (p *Preparer) installAndLaunchPod(newManifest *pods.Manifest, pod Pod, logger logging.Logger) bool {
 	// do not remove the logger argument, it's not the same as p.Logger
 
+	// HACK ZONE. When we have better authz, rewrite.
+	if ((newManifest.RunAsUser() == "root") && (newManifest.ID() != POD_ID)) || p.podIdForbidden(newManifest.RunAsUser()) {
+		logger.WithField("pod", newManifest.ID()).
+			Errorln(fmt.Sprintf("Cannot use %s in run_as", newManifest.RunAsUser()))
```
anthonybishopric: A rule of thumb - if you're `sprintf`ing the value into the error message, it should probably be a field in the structured log as well.

pkg/preparer/orchestrate.go
```diff
@@ -150,22 +151,28 @@ func (p *Preparer) handlePods(podChan <-chan pods.Manifest, quit <-chan struct{}
...
+
+					ok := p.installAndLaunchPod(&manifestToLaunch, pod, manifestLogger)
+					if ok {
+						manifestToLaunch = pods.Manifest{}
+						working = false
+					}
 				}
+			} else {
+				// prevent a busy loop
+				time.Sleep(1 * time.Second)
```
anthonybishopric: restore the `time.After`

pkg/preparer/orchestrate_test.go
anthonybishopric: The test would deadlock if they had not anticipated the auth failure, right? Seems like something you'd see in the ensuing panic.

pkg/preparer/orchestrate.go
```diff
@@ -212,6 +206,14 @@ func (p *Preparer) verifySignature(manifest pods.Manifest, logger logging.Logger
 func (p *Preparer) installAndLaunchPod(newManifest *pods.Manifest, pod Pod, logger logging.Logger) bool {
 	// do not remove the logger argument, it's not the same as p.Logger
 
+	// HACK ZONE. When we have better authz, rewrite.
+	if ((newManifest.RunAsUser() == "root") && (newManifest.ID() != POD_ID)) || p.podIdForbidden(newManifest.RunAsUser()) {
```
anthonybishopric: I don't think this boolean logic is correct. the `||` means it will fail the deploy of the preparer, even if it's the preparer ID.

mpuncel: oh is the preparer id in the list of forbidden?

tummychow:
okay reviewed again. needs rebase. plus, does the manifest in the preparer integration test need to be changed? other than that, looks good

PR #154, opened 2015-04-13
Allow zero launchables or zero executables

tummychow:
Note that you can still get some undesired behavior when scheduling an empty manifest (eg the service check will still be created if the status_port is in the manifest), but that's a separate issue


anthonybishopric:
:+1: a simple change with big consequences

PR #155, opened 2015-04-14
Misc fixes and cleanup

bwester:
This commit contains a variety of small tweaks, cleanups, fixes, and
clarifications. The only motivation behind these changes is the thought, "This
doesn't look quite right," as I read through the source code.

### DIFF
--- a/pkg/artifact/artifact.go
+++ b/pkg/artifact/artifact.go
@@ -10,13 +10,13 @@ import (
 	"fmt"
 	"io/ioutil"
 	"os"
-	"path"
+	"path/filepath"
 	"regexp"
 	"strings"
 )
 
 var (
-	tarAppNameParse = regexp.MustCompile("([a-zA-Z\\d\\-\\_]+)\\_([a-f0-9]+)\\.tar(\\.gz)?")
+	tarAppNameParse = regexp.MustCompile(`^([\w\-]+)_([a-zA-Z0-9]+)\.tar(\.gz)?$`)
 )
 
 type Artifact struct {
@@ -28,17 +28,18 @@ type Application struct {
 }
 
 func NewArtifact(path string) (*Artifact, error) {
-	if !tarAppNameParse.MatchString(path) {
-		return nil, fmt.Errorf("The path %s is not a valid path: it does ", path)
+	_, file := filepath.Split(path)
+	if !tarAppNameParse.MatchString(file) {
+		return nil, fmt.Errorf("%s: invalid artifact name", path)
 	}
 	if _, err := os.Stat(path); os.IsNotExist(err) {
-		return nil, fmt.Errorf("The path %s does not exist", path)
+		return nil, fmt.Errorf("%s: %s", path, err)
 	}
 	return &Artifact{path}, nil
 }
 
 func (a *Artifact) App() *Application {
-	_, tar := path.Split(a.Path)
+	_, tar := filepath.Split(a.Path)
 	splitUp := tarAppNameParse.FindStringSubmatch(tar)
 	return &Application{
 		Name: splitUp[1],
@@ -50,6 +51,7 @@ func (a *Artifact) AppManifest() (*AppManifest, error) {
 	if err != nil {
 		return nil, err
 	}
+	defer reader.Close()
 	tarReader := tar.NewReader(reader)
 	for {
 		next, err := tarReader.Next()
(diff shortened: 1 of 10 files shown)


pkg/artifact/artifact.go
```diff
@@ -10,13 +10,13 @@ import (
...
 	"os"
-	"path"
+	"path/filepath"
 	"regexp"
 	"strings"
 )
 
 var (
-	tarAppNameParse = regexp.MustCompile("([a-zA-Z\\d\\-\\_]+)\\_([a-f0-9]+)\\.tar(\\.gz)?")
+	tarAppNameParse = regexp.MustCompile(`^([\w\-]+)_([a-fA-F0-9]+)\.tar(\.gz)?$`)
```
anthonybishopric: Interestingly, we don't require that the suffix be a hex value - a fact we take advantage of when using `p2-bin2pod` to generate an artifact from a binary (it comes out using the opposite character class, ie `foo_hijzlpohyt.tar.gz`)

mpuncel: the second group there is a git sha, so it will only have lowercase a-f

pkg/cgroups/cgroups.go
```diff
@@ -45,13 +45,19 @@ func Find() (Subsystems, error) {
 	scanner := bufio.NewScanner(mountInfo)
 	for scanner.Scan() {
 		lineSegs := strings.Fields(scanner.Text())
-		if lineSegs[len(lineSegs)-3] != "cgroup" {
+		if len(lineSegs) != 11 {
+			return Subsystems{}, fmt.Errorf("mountinfo: unrecognized format")
```
mpuncel: :+1:

pkg/artifact/artifact.go
anthonybishopric: > the second group there is a git sha, so it will only have lowercase a-f

True, but not really applicable for this regex.

pkg/artifact/artifact.go
```diff
@@ -50,6 +51,7 @@ func (a *Artifact) AppManifest() (*AppManifest, error) {
 	if err != nil {
 		return nil, err
 	}
+	defer reader.Close()
```
anthonybishopric: Good catch

pkg/artifact/artifact.go
bwester: If it isn't a SHA1 hash, `([a-zA-Z0-9]+)`?

pkg/cgroups/cgroups.go
```diff
@@ -45,13 +45,19 @@ func Find() (Subsystems, error) {
 	scanner := bufio.NewScanner(mountInfo)
 	for scanner.Scan() {
 		lineSegs := strings.Fields(scanner.Text())
-		if lineSegs[len(lineSegs)-3] != "cgroup" {
+		if len(lineSegs) != 11 {
```
anthonybishopric: this seems a little less clear to me. Maybe leave a comment why we're testing 11, among the other values below

tummychow: actually I'm pretty sure this change is incorrect. The reason for indexing from the end is that one of the fields is optional and can be missing altogether.

bwester: Ah, looks like my first reading of the man page missed how the optional fields work. Yes, this is broken.  And I think it's not just one optional field but an arbitrary list of them.

pkg/artifact/artifact.go
anthonybishopric: > If it isn't a SHA1 hash, ([a-zA-Z0-9]+)?
Yep

pkg/runit/servicebuilder.go
```diff
@@ -55,8 +55,8 @@ func (template *SBTemplate) Write(out io.Writer) error {
 	if err != nil {
 		return util.Errorf("Could not marshal servicebuilder template %s as YAML: %s", template.name, err)
 	}
-	out.Write(b)
-	return nil
+	_, err = out.Write(b)
+	return err
```
anthonybishopric: :star:

anthonybishopric:
Nice cleanup! Make sure to run `rake integration` to ensure that the end-to-end test passes.

bwester:
Updated artifact regex and procfs field parsing.  Integration tests pass.

anthonybishopric:
:+1:

PR #156, opened 2015-04-15
Dynamically load PGP keys when verifying signatures

bwester:
Changes P2 to pass around an instance of the new `SigVerifier` interface, an
abstraction for something that can verify PGP signatures. The `KeyringFile`
implementation just loads a PGP keyring from a file every time a signature
needs verifying.

The current implementation loads a keyring from a file once at startup. If the
keyring file changes after that, the preparer would need to be restarted. For
comparison, the new behavior will pick up changes instantly without requiring a
restart.


pkg/auth/signatures.go
```diff
@@ -0,0 +1,67 @@
...
+	"golang.org/x/crypto/openpgp"
+)
+
+// Something that can verify PGP signatures
+type SigVerifier interface {
+	CheckDetachedSignature(signed, signature io.Reader) (*openpgp.Entity, error)
+}
+
+// A SignatureVerifier that reads its keyring from a file. The file is
+// read every time it is needed.
```
anthonybishopric: reading the keyring everytime will work well for guaranteeing freshness, but we'll probably want a way to do reasonable caching / modification detection. Good for now though :+1:

bwester: The way to do that is fairly straightforward, but I'm not convinced yet that it's really required.  How frequently does a signature need to be verified?  How big is this keyring going to get?  I suspect that it isn't worth the work to make this more sophisticated.

anthonybishopric: Well, currently the signatures for *all* manifests in the intent store are verified [every 5 minutes or when any manifests are scheduled](https://github.com/square/p2/blob/master/pkg/preparer/orchestrate.go#L150). So you're guaranteed to read this file, which could easily have a thousand keys in it, at least 5-10 times every 5 minutes. No, this isn't a dealbreaker, but yes, it is inefficient :)

bwester: I though manifests were verified only when their pod was starting.  That does change things :)

anthonybishopric: Yeah it's not obvious from the channel abstraction, but our Consul watches are basically in a long poll against `/intent/<hostname>/?recurse`, which means any modifications result in every manifest coming back on the channel.

bwester:
Updated:
- Changed to channel interface
- Doesn't reload every time—check file's mtime to detect changes and reload as needed
- Fixed existing unit test failures
- Added new unit test that checks if key reloading works

pkg/pods/pod.go
```diff
@@ -378,7 +378,7 @@ func (pod *Pod) Verify(manifest *Manifest, keyring openpgp.KeyRing) error {
 
 		var launchableDigest map[string]string
 
-		if stanza.DigestSignatureLocation == "" || keyring == nil {
+		if stanza.DigestSignatureLocation == "" || verifier == nil {
```
mpuncel: I don't think it's possible for `verifier` to be nil because it isn't a pointer

pkg/pods/manifest.go
```diff
@@ -191,9 +192,12 @@ func (manifest *Manifest) PlatformConfigFileName() (string, error) {
 // Returns the entity that signed the manifest, if any. If there was no
 // signature, both returns are nil. If the signer is not in the given keyring,
 // an openpgp.ErrUnknownIssuer will be returned.
-func (manifest *Manifest) Signer(keyring openpgp.KeyRing) (*openpgp.Entity, error) {
-	if manifest.signature == nil {
+func (manifest *Manifest) Signer(verifier auth.SigVerifier) (*openpgp.Entity, error) {
+	if manifest.signature == nil || verifier == nil {
```
mpuncel: same as below

anthonybishopric: `auth.SigVerifier` is an interface, so it's actually much worse than Mike points out - https://groups.google.com/forum/#!topic/golang-nuts/QzVDKv7p0Vs

mpuncel: it's a chan<- SigRequest not an interface

pkg/preparer/setup.go
```diff
@@ -116,9 +94,9 @@ func New(preparerConfig *PreparerConfig, logger logging.Logger) (*Preparer, erro
 	}
 
 	var err error
-	var keyring openpgp.KeyRing = nil
+	var verifier auth.SigVerifier
 	if preparerConfig.KeyringPath != "" {
-		keyring, err = loadKeyring(preparerConfig.KeyringPath)
+		verifier, err = auth.FileKeyringVerifier(preparerConfig.KeyringPath)
```
mpuncel: I don't know if you guys will like this idea, but I sort of think that we should get rid of the empty KeyringPath check above, and change `auth.FileKeyringVerifier` to expect an empty string and give you a verifier that always returns true. That gets us away from having to check if auth.SigVerifier is a zero value, etc

anthonybishopric: I'm into it. :+1:

pkg/auth/signatures.go
```diff
@@ -0,0 +1,138 @@
...
+			}
+			entity, err := openpgp.CheckDetachedSignature(
+				keyring,
+				req.Signed,
+				req.Signature,
+			)
+			req.ReplyChan <- SigReply{entity, err}
+		}
+	}()
+	return SigVerifier(reqs), nil
```
mpuncel: I've never seen parens here instead of curly braces, does this work?

anthonybishopric: You only use curly braces when initializing a struct, iirc

anthonybishopric: Although the return type of this function is still `openpgp.Keyring`. Not sure how that is working

mpuncel: oh it's a type conversion

pkg/auth/signatures.go
```diff
@@ -0,0 +1,138 @@
...
+			newInfo, err := os.Stat(filename)
+			if err != nil {
+				req.ReplyChan <- SigReply{nil, err}
+				continue
+			}
+			if info.ModTime() != newInfo.ModTime() {
+				// File has been modified! Reload it.
+				newKeyring, err := loadKeyring(filename)
+				if err != nil {
+					req.ReplyChan <- SigReply{nil, err}
```
mpuncel: question: if a reload fails, is it better to use a new keyring or start rejecting signatures?

pkg/pods/pod.go
bwester: Actually, the default value for a channel is `nil`. [[ref]](https://golang.org/ref/spec#Channel_types)

But I'm not sure whether we want it to be null.  The checks for `nil` throughout the code seem to suggest that the value means: "don't do any signature checks." It's explicitly allowed in [setup.go](https://github.com/square/p2/blob/master/pkg/preparer/setup.go#L120), and disallowed in [p2-preparer/main.go](https://github.com/square/p2/blob/master/bin/p2-preparer/main.go#L26). I guess this is used just for testing?

pkg/auth/signatures.go
mpuncel: i'd probably say it's better to reject it (the way it's currently written).

anthonybishopric: I think there's a case for using old keyring. We don't want there to be a strict online dependency with the keyserver - especially if it's a countdown to losing all deploy access. There's a security/access tradeoff I think we'll eventually want to be configurable.

pkg/pods/pod.go
mpuncel: we want people to be able to deploy the preparer without needing to gpg sign things, so if there is not a `keyring` key in the preparer's config then it shouldn't do signature checks. Right now we do that by checking if the verifier is nil everywhere, but another idea would be to have a stand in verifier that is always happy

mpuncel: also TIL that channels and some other "primitives" can be nil

pkg/preparer/setup.go
bwester: That's a bit more complex, because the verifier doesn't just return true: it returns the PGP entity of the signer. Grated, the _who_ is usually ignored (no signer? fail!) but in [one case](https://github.com/square/p2/blob/master/pkg/preparer/orchestrate.go#L182) there are additional checks on the signer to implement a hard-coded authorization policy.

I think long-term, authorization and authentication need to be better addressed.  For now, can we leave this?

pkg/auth/signatures.go
bwester: If I understand the overall system, the file that will be monitored with this is only going to be periodically changed by FCM, right? If that fails, it won't change the existing files on disk.

If you're thinking of an online method of collecting these keys, that can be supported, but it's a different `SigVerifier` instance. I think my preferred solution would be a layered approach:
1. In-memory keyring holding at least the keys to verify the currently-running pods
2. Disk cache of recent keys found
3. Master keyring distributed via FCM (to ensure that there is always someone who can deploy)
4. Online retriever that can retrieve keys as it needs when it tries to verify a signature
That and decoupling the possession of a key with authentication and authorization.

bwester:
Updated:
- Separate the entire authorization policy from other preparer internals. So instead of treating `nil` specially throughout the code, there is always some policy, even if it's the one that accepts everything.
- Use the cached keyring on file access errors
- Don't reload the keyring on every access.  Instead, check the file's mtime and reload it only when needed.
- Got rid of the purely channel-based approach and went back to defining authorization as an interface. In retrospect, I think forcing channels on all implementations is a bad design. It led to code where the implementation was a single function, adapted to make it asynchronous, adapted to make it synchronous.

pkg/auth/policy.go
```diff
@@ -0,0 +1,268 @@
...
+// its actions. It is possible for implementations to rely on other
+// services for these behaviors, so these calls may be slow or
+// transiently fail.
+type Policy interface {
+	// Check if a pod is authorized to be installed and run on this
+	// node. This involves checking that the pod has a valid signature
+	// and that the signer is authorized to install/run the pod. If
+	// the action is authorized, `nil` will be returned.
+	AuthorizePod(
+		id string,
```
anthonybishopric: would it make more sense to pass a `pods.Manifest`?

pkg/auth/policy.go
```diff
@@ -0,0 +1,268 @@
...
+	// the action is authorized, `nil` will be returned.
+	CheckDigest(digest, signature io.Reader) error
+
+	// Release any resources held by the policy implementation.
+	Close()
+}
+
+// auth.Error wraps all errors generated by the authorization layer,
+// allowing errors to carry structured data.
+type Error struct {
```
anthonybishopric: we already have `util.StackError`, you may or may not wish to use that

pkg/auth/policy.go
bwester: Accepting a `pods.Manifest` directly introduces a circular dependency.  This is one way to mitigate—there are certainly others.

pkg/auth/policy.go
```diff
@@ -0,0 +1,268 @@
...
+func (p NullPolicy) Close() {
+}
+
+// Assert that NullPolicy is a Policy
+var _ Policy = NullPolicy{}
+
+// The FixedKeyring policy holds one keyring. A pod is authorized to be
+// deployed iff:
+// 1. The manifest is signed by a key on the keyring, and
+// 2. If the manifest has an authorization list, the signing key is on
```
anthonybishopric: `s/manifest/pod ID/`

pkg/auth/policy.go
bwester: The main reason for this error type is so that structured data can be added to the error (which will wind up in a Logrus entry).  As I see it, the choices are:
1. Don't use structured logging
2. Pass a logger along the call stack and log errors the instant they're generated
3. Pass up exceptions with context information, and log them centrally
I went with #3 because I've used this pattern in past projects, and it's worked decently well. As I've seen strategy #2 applied, low-level errors tend to generate many log messages, one for each layer of the stack they're handled at.

pkg/auth/policy.go
bwester: :+1:

pkg/auth/policy.go
anthonybishopric: In that case, I would suggest making an `auth.Manifest` interface. I'd like to be able to write an authorization policy access the structured fields in a manifest (for example, asserting who can launch apps as certain users)

pkg/auth/policy.go
anthonybishopric: :+1:

pkg/auth/policy.go
bwester: :+1: Sounds reasonable

tummychow: does it make more sense to break the cycle? eg move pod.Verify somewhere else, maybe integrate it into the new auth package. if pod needs auth to verify itself, and auth needs pod to determine what to verify, then that indicates (to me) that the division between the packages is poorly defined.

anthonybishopric: That change seems sane. It should be easy to alter the only caller to `pod.Verify()`

bwester: The way `pod.Verify()` is implemented, a temp directory is created, files are saved to it, then the files are re-read to perform the verification.  Is there any reason to keep this design?  If not, I'd prefer to re-write how it gathers its signature data to expose an interface similar to what I did with the `pod.Manifest`: expose a `SignatureData()` method that returns the data needed to verify the signature.

tummychow: temp files are just for the artifact signatures. sooner or later they have to be downloaded, but they don't have to be downloaded there. The reason I actually put that code there was because 1) I wasn't really convinced that the verification logic belonged in hoist.launchable and 2) I didn't want to pass the keyring into the launchable (which would have been necessary to verify the signature of the launchable's digest).

if the auth logic is factored out to some third party (as this pr is doing), then sure move the downloading somewhere else too

anthonybishopric:
Please verify with the integration test (and it looks like Travis is broken too) but otherwise :+1:

bwester:
One of the unit tests generates fresh RSA keys to test with, and it's causing Travis CI to timeout. (Nothing appears wrong with the logic.) Looks like I'll be embedding some keys manually!

bwester:
Updated:
- Authorize manifest & digest interfaces, not concrete types
- Embed test RSA keys into the unit tests
- Fix typo in comment
Unit, integration tests pass locally

bwester:
Updated:
- Rebase to master & squash diffs (no source changes)
- Change _pkg/preparer/setup.go_ to use the dynamically-loading policy

PR #157, opened 2015-04-21
Refactor URI fetcher to support streaming

bwester:
The "uri" module presented a nice interface for copying a file from an
arbitrary URI to a local file. But what if you just want to consume the data
stream and not store it? This commit refactors the `uri.Copier` type into the
`uri.Fetcher` interface, which exposes the URI data stream directly (you can
still make a local copy if you want).

By accessing the data stream, users can avoid allocating a temporary directory,
writing a file to it, operating on the file, then deleting it.

### DIFF
--- a/pkg/hoist/hoist_launchable.go
+++ b/pkg/hoist/hoist_launchable.go
@@ -19,23 +19,17 @@ import (
 	"github.com/square/p2/pkg/util"
 )
 
-type Fetcher func(string, string) error
-
 // A HoistLaunchable represents a particular install of a hoist artifact.
 type Launchable struct {
 	Location         string         // A URL where we can download the artifact from.
 	Id               string         // A unique identifier for this launchable, used when creating runit services
 	RunAs            string         // The user to assume when launching the executable
 	ConfigDir        string         // The value for chpst -e. See http://smarden.org/runit/chpst.8.html
-	FetchToFile      Fetcher        // Callback that downloads the file from the remote location.
+	Fetcher          uri.Fetcher    // Callback that downloads the file from the remote location.
 	RootDir          string         // The root directory of the launchable, containing N:N>=1 installs.
 	P2exec           string         // The path to p2-exec
 	CgroupConfig     cgroups.Config // Cgroup parameters to use with p2-exec
 	CgroupConfigName string         // The string in PLATFORM_CONFIG to pass to p2-exec
-}
-
-func DefaultFetcher() Fetcher {
-	return uri.URICopy
 }
 
 func (hl *Launchable) Halt(serviceBuilder *runit.ServiceBuilder, sv *runit.SV) error {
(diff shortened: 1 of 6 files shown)


tummychow:
other than that comment, looks pretty good

pkg/hoist/hoist_launchable.go
```diff
@@ -239,30 +233,12 @@ func (hl *Launchable) Install() error {
...
-	outDir, err := ioutil.TempDir("", hl.Version())
-	defer os.RemoveAll(outDir)
-	if err != nil {
-		return util.Errorf("Could not create temporary directory to install %s: %s", hl.Version(), err)
-	}
-
-	outPath := filepath.Join(outDir, hl.Version())
-
-	err = hl.FetchToFile(hl.Location, outPath)
+	data, err := hl.Fetcher.Open(hl.Location)
```
anthonybishopric: This is a neat simplification. Keep in mind that we want *some* strategy to allow caching across multiple pods if we share launchables. A shared jdk artifact is one example. We can't totally guarantee what will happen to an install directory after expansion unless we do some magic, so the tar was going to be the next logical step.

bwester: I definitely agree with you.  There should be some strategy for caching artifacts, not just for reuse by different pods but also for resilience in case we need to undo a rollout and go back to using an earlier version of a pod.

That said, I don't want to address that in this diff.  This diff is a refactor of existing behavior.  And I suspect that writing an artifact cache will be enough extra work that this simplification won't really get in the way for now.

pkg/uri/uri.go
```diff
@@ -3,75 +3,104 @@ package uri
...
+	Client *http.Client
+}
+
+func (f BasicFetcher) Open(srcUri string) (io.ReadCloser, error) {
+	uri, err := url.Parse(srcUri)
+	if err != nil {
+		return nil, util.Errorf("%s: bad URI", srcUri)
+	}
+	switch uri.Scheme {
+	case "file", "":
```
anthonybishopric: Does `url.Parse` correctly return the empty string for the scheme if it's just a file path?

anthonybishopric:
Left one more question in there, otherwise :+1:

pkg/uri/uri.go
bwester: http://play.golang.org/p/ZdLFeLBoHZ

Short answer: yes, for reasonable file names, but I there are a lot of edge cases. I'm planning on going back to stripping `file:(//)?` off of the beginning.

bwester:
Updated:
- Simplified parsing of file URIs
- Deferred closing in `uri.CopyLocal()`

anthonybishopric:
Looks good, although I humbly request that we keep the commit history pristine - it makes looking at small updates easier without squashing.

PR #158, opened 2015-04-21
Rewrite replication

tummychow:
- replication uses goroutines for concurrent rollout across nodes. Significant speed improvement (especially for no-ops) because there's no sleep. Each goroutine has its own watches on consul, but overall fewer requests are made.
- replication attempts to acquire a lock on all keys before rolling out. If two replications collide, one will bail out.
- pkg/health is decoupled from pkg/kp and no longer uses reality store functionality. Reality polling for replications has been moved to pkg/replication itself.
- Health checks are passed around in an array, so inspecting individual checks out of a group is easier.

replication and health were rewritten, so the diffs are probably easier to digest side-by-side.


bin/p2-replicate/main.go
```diff
@@ -77,19 +71,67 @@ func main() {
...
+		log.Fatalf("Could not retrieve user: %s", err)
+	}
+	lock, err := store.NewLock(fmt.Sprintf("%q from %q at %q", thisUser.Username, thisHost, time.Now()))
+	if err != nil {
+		log.Fatalf("Could not generate lock: %s", err)
+	}
+	defer lock.Destroy()
+	go func() {
+		for range time.Tick(10 * time.Second) {
+			err := lock.Renew()
```
anthonybishopric: What happens if `lock.Renew()` is called after `lock.Destroy()`? Wondering about the ordering of the defer and this goroutine.

tummychow: renew should return an error since the target session no longer exists, which would kill main. But the lock destruction is deferred on main, so it would only be encountered on the way out of repl.Enact.

There are definitely some questions about destroying the lock in the case of a nonstandard exit where main does not terminate, eg a ctrl-C (or actually, even an os.Exit). I'm going to poke around that some more today.

pkg/kp/kv.go
```diff
@@ -108,7 +108,6 @@ func (s *Store) ListPods(keyPrefix string) ([]ManifestResult, time.Duration, err
 // caller's responsibility to filter out unchanged manifests.
 func (s *Store) WatchPods(keyPrefix string, quitChan <-chan struct{}, errChan chan<- error, podChan chan<- ManifestResult) {
 	defer close(podChan)
-	defer close(errChan)
```
anthonybishopric: Why just this channel?

pkg/replication/README.md
```diff
@@ -1,44 +0,0 @@
-# Replication: facilities for scheduling a pod across multiple nodes
```
anthonybishopric: Why kill the README? It's nice to have per-package docs, or if you are replacing with a wiki page at least link to it.

pkg/kp/kv.go
tummychow: Makes it possible for multiple goroutines to share the same error channel without one closing it under the feet of the others.

After this, I was thinking more about how to handle channel arguments and I think that, in general, it makes more sense for the sender to create and close the channel, while passing it back to the receiver. Right now all of our APIs pass channels into the function, which leaves ownership in the hands of the caller. In that case, I think it makes more sense to avoid closing the channels you're given.

pkg/replication/README.md
tummychow: whoops, I must have blown the readme away when i was replacing the old package with the rewritten one. Will restore

bin/p2-replicate/main.go
tummychow: fixed session destruction when the keys can't be locked (had to manually delete the lock since deferring on main doesn't really work), going to look at ctrl-C as well

pkg/replication/order.go
```diff
@@ -0,0 +1,110 @@
...
+	Nodes() []string
+}
+
+type MaxOrder struct {
+	active int
+	nodes  []string
+}
+
+// create a new MaxOrder using the given node list. The maximum number of
+// concurrent updates permitted by this order is active. For example, if
```
anthonybishopric:
```
The maximum number ofconcurrent updates permitted by this order is active.
```
This was really confusing.  You mean "the parameter `active` designates the maximum number of concurrent updates"

tummychow: yeah that's much better wording, fixing

pkg/replication/order_test.go
```diff
@@ -0,0 +1,45 @@
...
+
+	go order.Execute(next, done, errs, quit)
+
+	count := 0
+OUTER:
+	for {
+		select {
+		case <-next:
+			count++
+		case <-time.After(500 * time.Millisecond):
```
anthonybishopric: Is there any way to rewrite this loop such that it doesn't guarantee an added 500ms to the test suite?

tummychow: I don't know of one. Select ordering is pseudorandom, so the solution you suggested earlier would still be random (as is this test, though, so it's not like this is much better). The only way to really know that the channel has "stopped" is to block on it until you feel confident that it's going to keep blocking in the foreseeable future.

fwiw, the real purpose of the test is to make sure that the `active` constraint doesn't get violated. This particular test verifies the constraint by waiting for the `next` channel to "block". I don't think there's any really deterministic way to test this, but if there's a less time-sensitive option than this one, I'm for it.

bin/p2-replicate/main.go
```diff
@@ -77,19 +72,85 @@ func main() {
...
-		if err != nil {
-			log.Fatalf("p2 is not running on host %s: %s", host, err)
-		}
+	healthResults, err := healthChecker.Service(manifest.ID())
+	if err != nil {
+		log.Fatalf("Could not get initial health results: %s", err)
+	}
+	order, err := replication.NewMaxOrder(*hosts, len(*hosts)-*minNodes, healthResults)
+	if err != nil {
+		log.Fatalf("Could not instantiate rollout order", err)
```
mpuncel: %s

pkg/kp/lock.go
```diff
@@ -0,0 +1,100 @@
...
+
+type AlreadyLockedError struct {
+	Key string
+}
+
+func (err AlreadyLockedError) Error() string {
+	return fmt.Sprintf("Key %q is already locked", err.Key)
+}
+
+func (s *Store) NewLock(name string) (Lock, error) {
```
mpuncel: to make sure i'm understanding correctly, this does not ACTUALLY lock anything it just gives you a struct that you can then lock? i.e. gives you an "unlocked lock" back

pkg/kp/constants.go
```diff
@@ -23,3 +24,7 @@ func RealityPath(args ...string) string {
 func HookPath(hookType hooks.HookType, args ...string) string {
 	return strings.Join(append([]string{HOOK_TREE, hookType.String()}, args...), "/")
 }
+
+func LockPath(args ...string) string {
```
mpuncel: Food for thought, I thought i'd find this in lock.go. Looking at the rest of the contents of this file it makes sense though.

pkg/kp/constants.go
```diff
@@ -10,6 +10,7 @@ const (
 	INTENT_TREE  string = "intent"
 	REALITY_TREE string = "reality"
 	HOOK_TREE    string = "hooks"
+	LOCK_TREE    string = "lock"
```
mpuncel: this is to get around that Consul api limitation right?

pkg/kp/lock.go
tummychow: yeah, it basically gives you a thing that you can use to claim locks on kv keys. maybe `Lock` was not the best name. I didn't really like the name `Session` either (which is the consul name for this object)

pkg/kp/constants.go
tummychow: yeah. If we could take locks on the intent tree without writing the keys by accident, this new tree would have been unnecessary.

pkg/replication/replication.go
```diff
@@ -1,195 +1,173 @@
...
-	if err != nil {
-		repl.Logger.WithField("err", err).Errorln("Couldn't communicate with health checker on first run")
-		return
+// Execute the replication.
+// note: error management could use some improvement, errors coming out of
+// updateOne need to be scoped to the node that they came from
+func (r Replicator) Enact(errCh chan<- error, quitCh <-chan struct{}) {
+	nodeQueue := make(chan string)
+	done := make(chan string)
+	innerQuit := make(chan struct{})
```
mpuncel: does anything write to innerQuit?

mpuncel: oh duh the defer close()

tummychow: yeah the trick here is to terminate an unknown number of goroutines without having to actually send on the channel (which might block if you send more values than there are receivers).

mpuncel:
Looks good to me

PR #159, opened 2015-04-25
Move servicebuilder functionality into p2

tummychow:
Managed to PR this before leaving the office for the last time! Summary:

- no dependency on external servicebuilder script
- the new servicebuilder does not explicitly stop and start services like the old one did. runsvdir does that already when new directories are added or removed.
  - I was considering writing the down file to make all services downed-by-default, which would give p2 more control over the moment that the service starts, but I don't think the distinction is particularly significant here, because the next thing we do (right after creating the service directory) is starting the service.
- servicebuilder functionality invoked as part of hoist launchable.Launch, removed from pod.Launch.
  - To preserve the behavior of servicebuilder wrt cleaning up old services, pod invokes the new Prune method to remove nonexistent services. launchables also invoke prune during their halt, which preserves the other behavior we wanted (deploying a pod with no launchables will remove the services for that pod). The pod-invoked Prune might actually be redundant because of that, now that I think about it.
- post-activate now kills deploys if it crashes. I know we used to make this a non-halting failure, but based on some of the services we've seen, I think bailing here makes more sense, especially since enable is also a killing script.
- post-activate is unexported from launchable, since pod.Launch no longer needs to call it. this should shrink the shared API surface for docker launchables.

### DIFF
--- a/pkg/pods/pod.go
+++ b/pkg/pods/pod.go
@@ -127,35 +127,8 @@ func (pod *Pod) Launch(manifest *Manifest) (bool, error) {
 		return false, err
 	}
 
-	var successes []bool
-	for _, launchable := range launchables {
-		err := launchable.MakeCurrent()
-		if err != nil {
-			// being unable to flip a symlink is a catastrophic error
-			return false, err
-		}
-
-		out, err := launchable.PostActivate()
-		if err != nil {
-			// if a launchable's post-activate fails, we probably can't
-			// launch it, but this does not break the entire pod
-			pod.logLaunchableError(launchable.Id, err, out)
-			successes = append(successes, false)
-		} else {
-			if out != "" {
-				pod.logInfo(out)
-			}
-			successes = append(successes, true)
-		}
-	}
-
-	err = pod.buildRunitServices(launchables)
-
 	success := true
-	for i, launchable := range launchables {
-		if !successes[i] {
-			continue
-		}
+	for _, launchable := range launchables {
 		err = launchable.Launch(pod.ServiceBuilder, pod.SV) // TODO: make these configurable
 		if err != nil {
 			// Log the failure but continue
@@ -170,37 +143,11 @@ func (pod *Pod) Launch(manifest *Manifest) (bool, error) {
 		pod.logInfo("Launched pod but one or more services failed to start")
 	}
 
-	return success, nil
-}
-
-// Write servicebuilder *.yaml file and run servicebuilder, which will register runit services for this
-// pod.
-func (pod *Pod) buildRunitServices(launchables []hoist.Launchable) error {
-	// if the service is new, building the runit services also starts them, making the sv start superfluous but harmless
-	sbTemplate := runit.NewSBTemplate(pod.Id)
-	for _, launchable := range launchables {
-		executables, err := launchable.Executables(pod.ServiceBuilder)
-		if err != nil {
-			return err
-		}
-		for _, executable := range executables {
-			sbTemplate.AddEntry(executable.Service.Name, executable.Exec)
-		}
-		if err != nil {
-			// Log the failure but continue
-			pod.logLaunchableError(launchable.Id, err, "Unable to launch launchable")
-		}
-	}
-	_, err := pod.ServiceBuilder.Write(sbTemplate)
+	err = pod.ServiceBuilder.Prune()
 	if err != nil {
-		return err
+		pod.logError(err, "Could not prune services")
 	}
-
-	_, err = pod.ServiceBuilder.Rebuild()
-	if err != nil {
-		return err
-	}
-	return nil
+	return success, err
 }
 
 func (pod *Pod) WriteCurrentManifest(manifest *Manifest) (string, error) {
@@ -284,24 +231,12 @@ func (pod *Pod) Uninstall() error {
 		return err
 	}
 
-	// halt launchables
+	// halt launchables and remove their runit services
 	for _, launchable := range launchables {
 		err = launchable.Halt(runit.DefaultBuilder, runit.DefaultSV) // TODO: make these configurable
 		if err != nil {
-			// log and continue
+			pod.logLaunchableError(launchable.Id, err, "could not halt launchable")
 		}
-	}
-
-	// remove runit services
-	sbTemplate := runit.NewSBTemplate(pod.Id)
-	err = pod.ServiceBuilder.Remove(sbTemplate)
-	if err != nil {
-		return err
-	}
-
-	_, err = pod.ServiceBuilder.Rebuild()
-	if err != nil {
-		return err
 	}
 
 	// remove pod home dir
--- a/pkg/runit/servicebuilder.go
+++ b/pkg/runit/servicebuilder.go
@@ -18,51 +18,85 @@
 package runit
 
 import (
-	"bytes"
 	"fmt"
-	"io"
+	"io/ioutil"
 	"os"
-	"os/exec"
-	"path"
+	"path/filepath"
 
 	"github.com/square/p2/pkg/util"
 
 	"gopkg.in/yaml.v2"
 )
 
-type SBTemplate struct {
-	name       string
-	runClauses map[string]map[string][]string
+type ServiceTemplate struct {
+	Run      []string `yaml:"run"`
+	Log      []string `yaml:"log,omitempty"`
+	Sleep    int      `yaml:"sleep,omitempty"`
+	LogSleep int      `yaml:"logsleep,omitempty"`
 }
 
-type SBTemplateEntry struct {
-	baseCmd []string
-}
-
-func NewSBTemplate(name string) *SBTemplate {
-	return &SBTemplate{
-		name:       name,
-		runClauses: make(map[string]map[string][]string),
+func (s ServiceTemplate) RunScript() ([]byte, error) {
+	if len(s.Run) == 0 {
+		return nil, util.Errorf("empty run script")
 	}
-}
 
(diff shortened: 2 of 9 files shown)


pkg/pods/pod.go
```diff
@@ -292,17 +239,7 @@ func (pod *Pod) Uninstall() error {
...
-	err = pod.ServiceBuilder.Remove(sbTemplate)
-	if err != nil {
-		return err
-	}
-
-	_, err = pod.ServiceBuilder.Rebuild()
-	if err != nil {
-		return err
-	}
+	// remove runit services here
```
anthonybishopric: Why not do anything here?

pkg/runit/servicebuilder.go
```diff
@@ -18,51 +18,85 @@
...
+func (s ServiceTemplate) LogScript() ([]byte, error) {
+	if len(s.Log) == 0 {
+		// use a default log script that makes a logdir, chowns it and execs
+		// svlogd into it
+		return []byte(`#!/usr/bin/ruby
+require 'fileutils'
+FileUtils.mkdir_p('./main')
+FileUtils.chown('nobody', 'nobody', './main')
+sleep 2
+exec('chpst', '-unobody', 'svlogd', '-tt', './main')
```
anthonybishopric: Why use `chpst` instead of `p2-exec`? We shouldn't re-introduce the dependency

anthonybishopric: also, do we need to configure the path to `svlogd` as we do with `sv`?

pkg/runit/servicebuilder.go
```diff
@@ -18,51 +18,85 @@
...
+	if s.LogSleep != 0 {
+		sleep = s.LogSleep
+	}
+
+	ret := fmt.Sprintf(`#!/usr/bin/ruby
+require 'yaml'
+sleep %v
+exec *YAML.load(DATA.read)
+__END__
+%s`, sleep, args)
```
anthonybishopric: copy pasta

pkg/runit/servicebuilder.go
```diff
@@ -75,52 +109,160 @@ var DefaultBuilder = &ServiceBuilder{
...
+	}
 
-	err := os.Remove(yamlPath)
+	links, err := ioutil.ReadDir(s.RunitRoot)
 	if err != nil {
 		return err
 	}
+	for _, link := range links {
+		if _, exists := configs[link.Name()]; !exists {
+			if err = os.Remove(filepath.Join(s.RunitRoot, link.Name())); err != nil {
```
anthonybishopric: Is this consistent with `servicebuilder`? I didn't realize it would remove *any* service under the runit root dir without it being provided in the configs.

pkg/util/write.go
```diff
@@ -0,0 +1,28 @@
+package util
+
+import (
+	"bytes"
+	"io/ioutil"
+	"os"
+)
+
+func WriteIfChanged(filename string, data []byte, perm os.FileMode) (bool, error) {
```
anthonybishopric: Kind of weird semantics. What if I want to keep the `perm` I had previously, but don't know it?

anthonybishopric:
Overall this is great :star: Just a few questions / clarifications.

In general I'm curious about how this performed in the integration test, but more importantly, if it works cooperatively with the ruby `servicebuilder`.

pkg/pods/pod.go
tummychow: good question... I removed the old code after eliminating those methods but it should probably delete the service and prune(). will fix

pkg/runit/servicebuilder.go
tummychow: used chpst because that's what our old log script was doing and I didn't see much reason to change it. we could probably use p2-exec instead, I'll take a look. I would rather not introduce configurable paths for svlogd just because I don't think there's any practical reason to do so and passing those paths down the structs is a pain.

pkg/runit/servicebuilder.go
tummychow: not quite, the run script reopens stderr where as the log script doesn't, which is why I've never copied them over. I'm not sure if reopening stderr would have any impact on svlogd, i'll check

pkg/runit/servicebuilder.go
tummychow: yes, this is consistent behavior. After servicebuilder generates new services, it goes over the list of all service symlinks in /var/service and deletes the ones that were not found in configuration, then it repeats the process with the staged folders under /var/service-stage.

pkg/util/write.go
tummychow: set perm=0. there might actually be a bug where open syscall fails to create a new file because the permissions are zero, but fortunately we haven't run into that yet because cgroup parameter files always exist. I'll check

tummychow:
integration test passes (after adding the URI fix; I just rebased over master to incorporate that), and I'm fairly certain that this behavior is compatible with that of the existing servicebuilder. Note that we create a yaml file for each launchable rather than one for the entire pod; that's because service creation happens at the launchable level. Having multiple yaml files is functionally equivalent to having one file with all the services in it.

I mentioned this in the opening comment, but the only real change is that this servicebuilder implementation does not explicitly invoke sv anywhere, where as servicebuilder invoked sv restart upon creating or modifying a service, and it invoked sv stop before deleting one. In practice none of these actions are necessary so I took them out:

- runsvdir automatically reaps runsv processes for directories that no longer exist, which will cause the child service process to die. no need for explicit sv stop
- runsvdir automatically instantiates runsv processes for new directories, and those processes will start the target service immediately because there is no down file. no need for explicit sv start (unless the down file exists. In that case you do have to start it the first time. But we aren't writing that file.)
- there's no need to sv restart as part of changing the service, because the hoistlaunchable code is going to do that anyway

There is only one servicebuilder feature I know of that this does not support. In servicebuilder yaml files, you can specify down to say that the down file actually should be written. p2 never uses that file (and as I mentioned on slack on friday, I think we should stay that way) so support for that configuration was not implemented. This should not have impact on hoist-managed servicebuilder configuration, since this implementation generates services one by one rather than parsing the entire directory. (ie if a hoist-managed file has the down configuration in it, the p2 servicebuilder implementation will not blow that file away by accident.)

pkg/pods/pod.go
tummychow: whoops, actually not necessary. When the launchable is halted, it also prunes its runit service. I accidentally forgot to remove the comment, I'll nuke that.

pkg/util/write.go
tummychow: okay, so if you set perm=0 for a file that doesn't exist, the open syscall succeeds, and it creates a file with all the permission bits turned off. Kinda makes sense to be honest.

Probably the more logical behavior in this scenario is that if perm=0, the file should be opened with perms 0777 (and then umask is applied). If the file already exists then the 0777 would have no effect. I'll fix it.

tummychow:
Fixed comments. Integration test still passes.

anthonybishopric:
:+1:

bwester:
This merge causes rake integration tests to fail.  Output:
```
{"Counter":0,"PID":3435,"level":"info","msg":"Successfully installed","pod":"consul","time":"2015-05-13T22:21:50-04:00"}
{"Counter":1,"PID":3435,"level":"info","msg":"Successfully launched","pod":"consul","time":"2015-05-13T22:21:50-04:00"}
{"Counter":2,"PID":3435,"err":"servicebuilder.go:261: service with name consul__consul__launch was defined twice (from consul__consul.yaml)","level":"error","msg":"Could not prune services","pod":"consul","time":"2015-05-13T22:21:50-04:00"}
2015/05/13 22:21:50 Could not install consul: bootstrap.go:95: Can't launch Consul, aborting: servicebuilder.go:261: service with name consul__consul__launch was defined twice (from consul__consul.yaml)
```

tummychow:
@bwester strange, the test passes cleanly without issues on my end. Did you try destroying the VM first (`(cd integration/single-node-slug-deploy && vagrant destroy -f)`)?

bwester:
@tummychow: recreating the VM does the trick. I'm surprised that there would be any files "left over" from a previous version of p2-preparer if it's shut down correctly. Could this just be a problem with the integration test not cleaning up properly?

tummychow:
The issue is that the integration test runs in a vagrant VM, and the VM is not destroyed unless you explicitly do so yourself (which, I agree, is an example of not cleaning up properly). Repeated runs of the integration test don't actually recreate the VM from scratch, so if you installed p2 in the VM once, reusing that VM will retain the state of the old install.

The fix would be to `vagrant destroy -f` right before running the test. (Sometimes I do want the VM to stick around afterwards to debug a failure, but when it's time to run again, I think I would want the old one blown away.) I can PR that tomorrow.

bwester:
1. It takes a while to completely recreate a vagrant VM. Your proposal will be a huge slowdown.
2. I would expect that if I start the preparer, get it to load something, get it to stop the something, then stop the preparer, it should leave my system in the same state.  It isn't, and the current trouble highlights this.  I think _this_ is what needs fixing.

tummychow:
The integration test doesn't actually stop or uninstall any pods, it just turns the VM off if the test passes (note that this is `vagrant halt`, not `vagrant destroy`), which is why when you turn the VM back on to run the test again, the VM boots up with the same state as in the previous run.

We could implement a "reverse p2-launch" to do the cleanup inside the integration test, which would result in the behavior you're thinking of. Now that I think of it, there are definitely other useful things we could do in tests with such a utility (eg running multiple integration tests in travis without retaining state between them). I'll be online tomorrow, so we can discuss this in more detail on slack.

PR #160, opened 2015-04-29
allow passing a http.RoundTripper to NewHeaderClient

mpuncel:


pkg/pods/pod.go
```diff
@@ -186,10 +186,6 @@ func (pod *Pod) buildRunitServices(launchables []hoist.Launchable) error {
 		for _, executable := range executables {
 			sbTemplate.AddEntry(executable.Service.Name, executable.Exec)
 		}
```
bwester: Removing this error message seems unrelated to the TLS configuration change. Can you put it in its own commit?

mpuncel: sure, although it's not really a change since it's unreachable code

pkg/util/net/client.go
```diff
@@ -22,14 +23,22 @@ func (ht headerTransport) RoundTrip(r *http.Request) (*http.Response, error) {
 // NewHeaderClient returns an http.Client. It has the same behavior as
 // DefaultClient, except that the key-value pairs in extras are added as headers
 // on every request.
-func NewHeaderClient(extras map[string]string) *http.Client {
+func NewHeaderClient(extras map[string]string, tlsConfig *tls.Config) *http.Client {
```
bwester: Exposing only the TLS config seems rather arbitrary. Instead of accepting at TLS config and building the inner transport from that, what about accepting an `http.RoundTripper` and letting the caller decide how to initialize it?  Or perhaps a `*http.RoundTripper`, so that `nil` can be provided to get a sensible default?

mpuncel: sounds good

mpuncel: http.RoundTripper is an interface so `nil` is valid for it

bin/p2-inspect/main.go
```diff
@@ -51,7 +51,7 @@ func main() {
 	opts := kp.Options{
 		Address: *consulUrl,
 		Token:   *consulToken,
-		Client:  net.NewHeaderClient(*headers),
+		Client:  net.NewHeaderClient(*headers, nil),
```
anthonybishopric: I'd feel more comfortable explicitly passing `http.DefaultTransport`, but I won't press that point.

pkg/pods/pod.go
anthonybishopric: I would like to do this in a separate commit only because I think we need to replace this message with something else. Ideally we provide information to the service healthcheck about the uptime of services and the PIDs they're running as.

mpuncel: by separate commit do you mean separate PR? Also this if statement is totally superfluous and will never execute if you read the code above it

bin/p2-inspect/main.go
mpuncel: that sounds good

PR #161, opened 2015-05-01
Return an error when BasicFetcher gets an HTTP error

bwester:
HTTP allows meaningful data in the body of HTTP responses for many status
codes. As far as `uri.BasicFetcher.Open()` is concerned, anything except a 200
status means the data can't be accessed, so it should return an error.


anthonybishopric:
:+1:

mpuncel:
lgtm

PR #162, opened 2015-05-05
include inner error in error message when health status check fails

mpuncel:


bwester:
Nice! :+1:

Could you post an example error message generated by the affected statement?

mpuncel:
example error message: ```Unable to get initial health results: health.go:32: /health/service failed for "slug": Unexpected response code: 400 (Invalid Datacenter/Zone
)```

tummychow:
are we sure this doesn't leak ACLs into preparer logs? It was intentional that errors coming directly from the API were always concealed and their stringifications were never output from pkg/health or pkg/kp. The health endpoints don't take ACLs, but the client package might still be attaching them to the URL ([ref](https://github.com/hashicorp/consul/blob/master/api/api.go#L323)). If consul becomes unreachable, writing the error content out may result in ACLs appearing in the preparer's stdout.

I know it's a pain to not actually see the output of the error when it's meaningful, but I didn't have a good solution to that problem other than to hide the entire error content.

PR #163, opened 2015-05-08
Travis

tummychow:
resuming previous work

### DIFF
--- a/bin/p2-exec/main.go
+++ b/bin/p2-exec/main.go
@@ -174,7 +174,12 @@ func cgEnter(platconf, launchableName, cgroupName string) error {
 		return util.Errorf("Could not find cgroupfs mount point: %s", err)
 	}
 	err = cg.Write(cgConfig)
-	if err != nil {
+	if _, ok := err.(cgroups.UnsupportedError); ok {
+		// if a subsystem is not supported, just log
+		// and carry on
+		log.Printf("Unsupported subsystem (%s), continuing\n", err)
+		return nil
+	} else if err != nil {
 		return util.Errorf("Could not set cgroup parameters: %s", err)
 	}
 	return cg.AddPID(cgConfig.Name, os.Getpid())
--- a/integration/travis.sh
+++ b/integration/travis.sh
@@ -1,0 +1,15 @@
+#!/usr/bin/env bash
+
+set -ex
+
+sudo groupadd nobody
+sudo useradd hello
+
+sudo /sbin/stop runsvdir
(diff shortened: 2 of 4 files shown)


tummychow:
finally green! this is ready for review

bin/p2-exec/main.go
```diff
@@ -174,7 +174,12 @@ func cgEnter(platconf, launchableName, cgroupName string) error {
 		return util.Errorf("Could not find cgroupfs mount point: %s", err)
 	}
 	err = cg.Write(cgConfig)
-	if err != nil {
+	if _, ok := err.(cgroups.UnsupportedError); ok {
+		// if a subsystem is not supported, just log
+		// and carry on
+		log.Printf("Unsupported subsystem (%s), continuing", err)
```
anthonybishopric: `continuing\n"`

integration/travis.sh
```diff
@@ -0,0 +1,15 @@
+#!/usr/bin/env bash
+
+set -ex
+
+sudo groupadd nobody
+sudo useradd hello
```
anthonybishopric: This useradd should be taken care of in a hook

bin/p2-exec/main.go
tummychow: fixed

integration/travis.sh
```diff
@@ -0,0 +1,15 @@
...
+
+sudo /sbin/stop runsvdir
+sudo sed -i -e 's;/usr/sbin/runsvdir-start;/usr/bin/runsvdir /var/service;g' /etc/init/runsvdir.conf
+sudo /sbin/start runsvdir
+sudo mkdir -p /etc/servicebuilder.d /var/service-stage /var/service
```
anthonybishopric: Do we still need this section?

integration/travis.sh
anthonybishopric: "should" -> was already part of the test to add the user in the hook

tummychow: It is, but with the wrong path - the hook created in the integration test calls `/sbin/adduser`; on ubuntu the adduser binary is under `/usr/sbin`. As a result, the hook does nothing and the user doesn't exist unless created earlier. This is one of those moments where clearing PATH in hooks really bites us because these are two different distributions.

integration/travis.sh
tummychow: yes. The default debian/ubuntu init script for runsvdir starts it under /etc/service instead of /var/service.

anthonybishopric:
:+1:

tummychow:
rebased onto new servicebuilder implementation, still green, merging

PR #164, opened 2015-05-08
Parse URIs for fetching

tummychow:
Fixes a bug in the handling of `file://` URIs; the leading slash on an absolute path was being stripped off (notably our integration test uses these). Question marks in the URL path will have to be query-escaped.

### DIFF
--- a/pkg/uri/uri.go
+++ b/pkg/uri/uri.go
@@ -3,15 +3,12 @@ package uri
 import (
 	"io"
 	"net/http"
+	"net/url"
 	"os"
-	"regexp"
-	"strings"
+	"path/filepath"
 
 	"github.com/square/p2/pkg/util"
 )
-
-// Internet Standard: STD 66
-var leadingScheme = regexp.MustCompile(`^([a-zA-Z][a-zA-Z0-9+\-.]*):`)
 
 // A UriFetcher presents simple methods for fetching URIs from
 // different schemes.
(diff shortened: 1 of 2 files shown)


bwester:
Could you amend your commit message to describe the functional changes to the parser, beyond just "fixes a bug?"

I'm not sure what you mean by "Question marks in the URL path will have to be query-escaped." Could you give an example?

tummychow:
Sure, I'll amend the commit in a sec. Basically after this change, `file:///path/to/my?file` will result in a path of `/path/to/my`, and the `?file` will be the query on the URL (which is ignored, probably we should return an error there for sanity reasons so I'll fix that too). If you want the question mark in the path, you have to %-escape it, eg `file:///path/to/my%3Ffile`.

I think the reason we weren't actually parsing the string before was specifically to ignore those query metacharacters in references to local files. I would rather just parse the thing to avoid bugs like this one and provide a more consistent interpretation of the string. The practical impact of the change will be small because we don't use local file paths except in tests.

bwester:
> I think the reason we weren't actually parsing the string before was specifically to ignore those query metacharacters in references to local files.

Yes, that was the reason. It's valid to name a file `my?file` (or even `my%3Ffile`). However, testing on a random selection of normal user tools (OS X "open", Firefox, and curl) shows that the URI authority, query, and fragment are routinely ignored in the `file:` scheme anyway. If that's the common expectation, sure, let's follow suite.

PR #166, opened 2015-05-11
Add user-based authorization policy

bwester:
Adds a new deploy authorization policy to the preparer based on an external
user-based policy file.  This PR consists of a two commits that are pure
refactors and one commit to implement the new policy.

### DIFF
--- a/pkg/auth/policy.go
+++ b/pkg/auth/policy.go
@@ -3,9 +3,12 @@ package auth
 import (
 	"bytes"
 	"fmt"
+	"io/ioutil"
 	"os"
 
 	"golang.org/x/crypto/openpgp"
+	"golang.org/x/crypto/openpgp/errors"
+	"gopkg.in/yaml.v2"
 
 	"github.com/square/p2/pkg/logging"
 	"github.com/square/p2/pkg/util"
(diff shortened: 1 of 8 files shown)


bwester:
I'm somewhat unsatisfied with requiring the preparer configuration to depend on the email domain. It's a consequence of having email addresses in PGP keys and email addresses in the input file. But I don't think this is a critical problem, and I'm willing to go ahead with the proposed solution.

As an alternative, the policy format could represent users as:
- email addresses
- key IDs
Any thoughts?

pkg/auth/policy.go
```diff
@@ -197,76 +200,257 @@ func LoadKeyring(path string) (openpgp.EntityList, error) {
...
+// returning the data.
+type fileWatcher struct {
+	reqChan chan<- (chan<- interface{})
 }
 
-func NewFileKeyringPolicy(
-	keyringPath string,
-	authorizedDeployers map[string][]string,
-) (Policy, error) {
+func newFileWatcher(
```
mpuncel: this is pretty cool :+1:

anthonybishopric:
Just checking email addresses seems like a nice simplification. I don't think it matters a ton in practice since most organizations using P2 will, more than likely, share an email domain. That said, i imagine it will be a teeny bit more flexible and enable weirder cases when necessary.

pkg/auth/policy.go
anthonybishopric: Agreed, can we move it to a utility package?

bwester: Sure

pkg/preparer/setup.go
```diff
@@ -26,16 +26,30 @@ type LogDestination struct {
...
-	AuthorizedDeployers  []string         `yaml:"authorized_deployers,omitempty"`
-	ForbiddenPodIds      []string         `yaml:"forbidden_pod_ids,omitempty"`
-	ExtraLogDestinations []LogDestination `yaml:"extra_log_destinations,omitempty"`
+	NodeName             string                 `yaml:"node_name"`
+	ConsulAddress        string                 `yaml:"consul_address"`
+	ConsulTokenPath      string                 `yaml:"consul_token_path,omitempty"`
+	HooksDirectory       string                 `yaml:"hooks_directory"`
+	CAPath               string                 `yaml:"ca_path,omitempty"`
+	PodRoot              string                 `yaml:"pod_root,omitempty"`
+	Auth                 map[string]interface{} `yaml:"auth,omitempty"`
```
anthonybishopric: Any reason why this shouldn't just be `map[string]string`?

bwester: Changing it from `interface{}` to `string` wouldn't impact the current use cases.

anthonybishopric:
Looks good overall :+1:

petertseng:
Code-wise looking good. I would also be happy with using emails in the policy file rather than just usernames.

bwester:
Changes made. This should be good to go.

bwester:
Latest update was to merge several commits and rebase onto master. No code was changed.

PR #167, opened 2015-05-12
allow the replicator to forcibly lock hosts (to override an in progre…

mpuncel:
…ss rollout)

### DIFF
--- a/pkg/replication/replication.go
+++ b/pkg/replication/replication.go
@@ -63,17 +63,16 @@ func (r Replicator) lock(lock kp.Lock, lockPath string, overrideLock bool) error
 			// which prevents other parties from acquiring the lock for a
 			// limited time
 			return util.Errorf("Lock for %q is blocked due to delay by previous holder", lockPath)
-		} else {
-			if overrideLock {
-				err = r.Store.DestroyLockHolder(id)
-				if err != nil {
-					return util.Errorf("Unable to destroy the current lock holder (%s) for %q: %s", holder, lockPath, err)
-				}
-
-				// try acquiring the lock again, but this time don't destroy holders so we don't try forever
-				return r.lock(lock, lockPath, false)
+		} else if overrideLock {
+			err = r.Store.DestroyLockHolder(id)
+			if err != nil {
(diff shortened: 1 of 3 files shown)


pkg/replication/replication.go
```diff
@@ -33,32 +33,54 @@ func (r Replicator) CheckPreparers() error {
...
+	err := lock.Lock(lockPath)
+
+	if _, ok := err.(kp.AlreadyLockedError); ok {
+		holder, id, err := r.Store.LockHolder(lockPath)
+		if err != nil {
+			return util.Errorf("Lock already held for %q, could not determine holder due to error: %s", lockPath, err)
+		} else if holder == "" {
+			// we failed to acquire this lock, but there is no outstanding
+			// holder
+			// this indicates that the previous holder had a LockDelay,
```
anthonybishopric: Is this really the only sign of a lock delay?

anthonybishopric: Also seems like it could be frustrating if you are passing `--override-lock` but you are being thwarted by a lock that is already gone.

mpuncel: we set lock delay to 1 * time.Nanosecond so I doubt this will be a problem in practice

mpuncel: see https://github.com/square/p2/blob/master/pkg/kp/lock.go#L32

pkg/replication/replication.go
```diff
@@ -33,32 +33,54 @@ func (r Replicator) CheckPreparers() error {
...
+			// which prevents other parties from acquiring the lock for a
+			// limited time
+			return util.Errorf("Lock for %q is blocked due to delay by previous holder", lockPath)
+		} else {
+			if overrideLock {
```
petertseng: I don't mind so much, but this means you could remove a level of indentation by making this an `else if overrideLock {`

Line 77 would become the `else` clause.

I dunno if you feel that makes the logic a bit unclear, but could consider it.

mpuncel: that seems logical

anthonybishopric:
:+1:

PR #170, opened 2015-05-14
Show PGP key ID when signer is not found

bwester:
When a manifest is signed with a key that isn't in the keyring, P2 will now
display an error message that includes the key ID:

    unknown signer: 0123456780ABCDEF

This should make it easier to understand why a manifest was rejected.


anthonybishopric:
:+1:

PR #171, opened 2015-05-14
servicebuilder reimplementation, round 2

tummychow:
The first two commits are the same as in the original branch. For the third commit, the original branch moved servicebuilder into the hoistlaunchable, which caused one file to be created for each launchable. This causes conflicts with existing servicebuilder files.

This new branch instead uses the new servicebuilder in the existing pod.buildRunitServices. This preserves the one-file-per-pod behavior that we were using before, which means conflicts will not be created between old p2 servicebuilder files and new ones.

Thorough testing is in progress, but the code is ready for review. My travis branch will be rebased back on top of this after merge.

### DIFF
--- a/pkg/cgroups/cgroups.go
+++ b/pkg/cgroups/cgroups.go
@@ -2,13 +2,14 @@ package cgroups
 
 import (
 	"bufio"
-	"bytes"
 	"fmt"
 	"io/ioutil"
 	"os"
 	"path/filepath"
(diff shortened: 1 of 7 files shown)


tummychow:
integration passes, universe behaves soundly, confirmed that only one file is created per pod. should be good to go

anthonybishopric:
:+1:

PR #173, opened 2015-05-16
Add package utility

anthonybishopric:
Standard rake task to create builds to use in tagged versions of P2

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -1,3 +1,5 @@
+require 'json'
+
 def e(cmd)
   puts cmd
   system(cmd) || abort("Error running `#{cmd}`")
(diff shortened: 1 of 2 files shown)


Rakefile
```diff
@@ -1,3 +1,5 @@
+require 'json'
```
petertseng: Is there actually any jsonning being done here?

petertseng:
Other than the json thing I'm okay with this in general. I did `rake build` and got a `builds/darwin/` dir with reasonable contents (`p2-0.0.3-40-gecd5ef8-darwin`, `p2-0.0.3-40-gecd5ef8-darwin.tar.gz` and `p2.tar.gz`)

petertseng:
The tar has the `p2-*` binaries under a `./bin` dir, seems reasonable

Rakefile
anthonybishopric: No, there isn't - will remove

PR #175, opened 2015-05-22
Add status port for p2-preparer

petertseng:
This implementation counts consecutive errors from either hooks or pods. Status goes critical if the number of consecutive errors >= 10 (the constant tooManyErrors).

I have been testing on a universe by stopping consul in the universe and seeing the errors count up. I then restart consul to see if it goes healthy again. I'm not 100% clear on whether I have the "OK I'm healthy again" part right. Going to do some more testing and report back on this PR.

### DIFF
--- a/bin/p2-preparer/main.go
+++ b/bin/p2-preparer/main.go
@@ -48,7 +48,7 @@ func main() {
 		func(w http.ResponseWriter, r *http.Request) {
 			fmt.Fprintf(w, "p2-preparer OK")
 		})
-	go http.ListenAndServe(":8080", nil)
+	go http.ListenAndServe(fmt.Sprintf(":%d", preparerConfig.StatusPort), nil)
 
 	waitForTermination(logger, quitMainUpdate, quitHookUpdate)
 
--- a/bin/p2-preparer/main.go
+++ b/bin/p2-preparer/main.go
@@ -31,12 +31,13 @@ func main() {
 	defer prep.Close()
 
 	logger.WithFields(logrus.Fields{
-		"starting":  true,
-		"node_name": preparerConfig.NodeName,
-		"consul":    preparerConfig.ConsulAddress,
-		"hooks_dir": preparerConfig.HooksDirectory,
-		"keyring":   preparerConfig.Auth["keyring"],
-		"version":   version.VERSION,
+		"starting":    true,
+		"node_name":   preparerConfig.NodeName,
+		"consul":      preparerConfig.ConsulAddress,
+		"hooks_dir":   preparerConfig.HooksDirectory,
+		"status_port": preparerConfig.StatusPort,
+		"keyring":     preparerConfig.Auth["keyring"],
+		"version":     version.VERSION,
 	}).Infoln("Preparer started successfully")
 
 	quitMainUpdate := make(chan struct{})
--- a/bin/p2-preparer/main.go
+++ b/bin/p2-preparer/main.go
@@ -1,6 +1,8 @@
 package main
 
 import (
+	"fmt"
+	"net/http"
 	"os"
 	"os/signal"
 	"syscall"
@@ -41,6 +43,12 @@ func main() {
 	quitHookUpdate := make(chan struct{})
 	go prep.WatchForPodManifestsForNode(quitMainUpdate)
 	go prep.WatchForHooks(quitHookUpdate)
+
+	http.HandleFunc("/_status",
+		func(w http.ResponseWriter, r *http.Request) {
+			fmt.Fprintf(w, "p2-preparer OK")
+		})
+	go http.ListenAndServe(":8080", nil)
 
 	waitForTermination(logger, quitMainUpdate, quitHookUpdate)
 
--- a/bin/p2-preparer/main.go
+++ b/bin/p2-preparer/main.go
@@ -44,11 +44,13 @@ func main() {
 	go prep.WatchForPodManifestsForNode(quitMainUpdate)
 	go prep.WatchForHooks(quitHookUpdate)
 
-	http.HandleFunc("/_status",
-		func(w http.ResponseWriter, r *http.Request) {
-			fmt.Fprintf(w, "p2-preparer OK")
-		})
-	go http.ListenAndServe(fmt.Sprintf(":%d", preparerConfig.StatusPort), nil)
+	if preparerConfig.StatusPort != 0 {
+		http.HandleFunc("/_status",
+			func(w http.ResponseWriter, r *http.Request) {
+				fmt.Fprintf(w, "p2-preparer OK")
+			})
+		go http.ListenAndServe(fmt.Sprintf(":%d", preparerConfig.StatusPort), nil)
+	}
 
 	waitForTermination(logger, quitMainUpdate, quitHookUpdate)
 
--- a/pkg/preparer/setup.go
+++ b/pkg/preparer/setup.go
@@ -32,6 +32,7 @@ type PreparerConfig struct {
 	HooksDirectory       string                 `yaml:"hooks_directory"`
 	CAPath               string                 `yaml:"ca_path,omitempty"`
 	PodRoot              string                 `yaml:"pod_root,omitempty"`
+	StatusPort           int                    `yaml:"status_port"`
 	Auth                 map[string]interface{} `yaml:"auth,omitempty"`
 	ForbiddenPodIds      []string               `yaml:"forbidden_pod_ids,omitempty"`
 	ExtraLogDestinations []LogDestination       `yaml:"extra_log_destinations,omitempty"`


pkg/preparer/setup.go
```diff
@@ -32,6 +32,7 @@ type PreparerConfig struct {
 	HooksDirectory       string                 `yaml:"hooks_directory"`
 	CAPath               string                 `yaml:"ca_path,omitempty"`
 	PodRoot              string                 `yaml:"pod_root,omitempty"`
+	StatusPort           int                    `yaml:"status_port"`
```
anthonybishopric: In a future change, we should have the preparer write the status port from the root values of the pod manifest to the `PLATFORM_CONFIG_PATH`. As written, this requires the `status_port` field to be repeated as top-level conf and as a value in the config stanza. In general, we need to figure out how to address better port values.

petertseng:
OK on my universe, running `watch curl ...` to see what's going on.
I stop consul, both error counts start increasing, we go CRITICAL when we reach 10
I restart consul.
At this point, the errors stop increasing, but they don't go to 0 either. Will need to figure out something for that.

pkg/preparer/listener.go
```diff
@@ -124,6 +124,8 @@ func (l *HookListener) Sync(quit <-chan struct{}, errCh chan<- error) {
 				sub.WithField("err", err).Errorln("Could not write hook link")
 			} else {
 				sub.NoFields().Infoln("Updated hook")
+				// Send back a nil error to signal success!
+				errCh <- nil
```
anthonybishopric: We shouldn't use nil as a sentinel value for success. One wrong access and this whole process panics.

There are a few alternatives I can think of - one, we actually have the preparer type provide an `eventStream` channel. The status handler code could `select` on the event stream success and the error channel. This has the benefit of being useful in other ways (for example, to create a local history).

A less preferable approach creates a `MaybeError` channel.

bin/p2-preparer/main.go
```diff
@@ -29,24 +32,85 @@ func main() {
...
+			Status:                status,
+			ConsecutiveHookErrors: consecutiveHookErrors,
+			ConsecutivePodErrors:  consecutiveMainErrors,
+			LastHookError:         fmt.Sprintf("%+v", lastHookError),
+			LastPodError:          fmt.Sprintf("%+v", lastMainError),
+		})
+		if err != nil {
+			fmt.Fprintf(w, "{status:\"Unknown (couldn't marshal: %s)\"}\n", err)
+		} else {
+			fmt.Fprintf(w, "%s\n", response)
```
anthonybishopric: Where is the http 500?

pkg/preparer/orchestrate.go
```diff
@@ -94,6 +98,7 @@ func (p *Preparer) WatchForPodManifestsForNode(quitAndAck chan struct{}) {
 				go p.handlePods(podChanMap[podId], quitChanMap[podId])
 			}
 			podChanMap[podId] <- result.Manifest
+			errs <- nil
```
anthonybishopric: Have you tested this? I am pretty sure the `waitForTermination` method will hang forever in main because it sends a value on the `err` channel, but will never be received if no `StatusPort` was configured.

pkg/preparer/listener.go
petertseng: I think I'm for this. There's a related thing I want to talk about. It's related to making this happen, but I would also need to make the change in either case.

Looks like I need to make a change to `WatchPods`. `WatchPods` right now sends errors on its error channel and new manifests on its manifest channel. I think I need to add a third channel on which it sends something if it wants to say "Hi, I read from consul without errors but I have no new manifests"

petertseng: What kind of information are you thinking of including on such `events` on the `eventStream` channel?

petertseng: I now have explicit success channels that just pass `struct{}` around instead of putting `nil`s in the error channels.

bin/p2-preparer/main.go
petertseng: OK added them

pkg/preparer/orchestrate.go
petertseng: So you are saying I should test the case where `StatusPort` is not configured. OK, will do. Testing it in universe.

petertseng:
Changes made to address some comments, but still some things to think about.

* I must still test with no `StatusPort` configured.
* Figure out why integration test is no good.
* My tests on Universe with `StatusPort` configured had the following results: I disable consul, error counts go up. I re-enable consul. Now, error counts go back to 0 after about 10 minutes. That's when the timeout in WatchPods expires, and WatchPods reports "OK, but no new manifests". It would go back faster if *both* a new pod and a new hook got scheduled, but that is not always the case. I'm not sure we like to wait 10 minutes for things to go healthy again. Is there a better way to do it? Should we reduce the timeout? Is that an OK thing to do?

petertseng:
Integration test is good. It was broken precisely because it had no status port, and I was not handling that (nothing was reading out of the channels, so the senders were blocking)

bin/p2-watch/main.go
```diff
@@ -56,9 +56,10 @@ func main() {
 	log.Printf("Watching manifests at %s\n", path)
 
 	quit := make(chan struct{})
+	noManifests := make(chan struct{})
 	errChan := make(chan error)
 	podCh := make(chan kp.ManifestResult)
-	go store.WatchPods(path, quit, errChan, podCh)
+	go store.WatchPods(path, quit, noManifests, errChan, podCh)
```
petertseng: Oh wait since I'm not reading this channel I'm pretty sure I just broke p2-watch.

pkg/replication/replication.go
```diff
@@ -140,10 +140,11 @@ func (r Replicator) updateOne(node string, done chan<- string, errCh chan<- erro
 	}
 
 	realityResults := make(chan kp.ManifestResult)
+	realityNoManifests := make(chan struct{})
 	realityErr := make(chan error)
 	realityQuit := make(chan struct{})
 	defer close(realityQuit)
-	go r.Store.WatchPods(kp.RealityPath(node, r.Manifest.ID()), realityQuit, realityErr, realityResults)
+	go r.Store.WatchPods(kp.RealityPath(node, r.Manifest.ID()), realityQuit, realityNoManifests, realityErr, realityResults)
```
petertseng: Need to read from this channel!!!

bwester:
I think that correctly reading from all these notification channels is going to be an ongoing pain point. (You wrote the changes, and your first implementation didn't get it right!) Since we're pretending that the core parts of P2 are a general library that might be reused by other tools, let's keep them hardened like a library:

1. Don't block writing to the metric channels. I think it's fine if a message gets lost.  Giving the channels a small buffer would help ensure that in all practical cases, all errors are caught.
2. Since the user (e.g., the things in _/bin_ and those externally-written tools) is responsible for creating these channels and passing them into the library, *document* No.1.

bwester:
I'm not sure that just counting the number of errors is a good way to determine whether a node is unhealthy. The kinds of errors that the node is experiencing are very important.  I'm going to broadly generalize and say that errors could be:

1. _Logic errors_ in the programming or configuration. The existence of _one_ of these should be enough to cause the node to be unhealthy (or just crash... what monitors crashes of the p2-preparer?)
2. _Transient errors_ caused by a peer being temporarily faulty. These should be expected in some small number; it's only a problem if the faults persist.
3. _Input errors_ caused by users giving it bad data. Bad input is most likely not affected by the correct functioning of the preparer.

In particular, if I push an invalid manifest to a preparer's intent store, will that eventually set the node into "critical" status?

bin/p2-preparer/main.go
```diff
@@ -29,24 +32,101 @@ func main() {
...
+		case <-successes:
+			*consecutive = 0
+		}
+	}
+}
+
+// if the number of consecutive errors >= tooManyErrors, the status is critical.
+const tooManyErrors = 10
+
+func statusHandler(mainSuccesses, hookSuccesses <-chan struct{}, mainErrors, hookErrors <-chan error) http.HandlerFunc {
```
bwester: What are the semantics of the reply that this generates?  Please document.

bin/p2-preparer/main.go
```diff
@@ -29,24 +32,101 @@ func main() {
...
+	consecutiveMainErrors := 0
```
bwester: There are data races accessing all four of these variables. Add appropriate synchronization.

bin/p2-preparer/main.go
```diff
@@ -29,24 +32,101 @@ func main() {
...
-	go prep.WatchForPodManifestsForNode(quitMainUpdate)
-	go prep.WatchForHooks(quitHookUpdate)
+	go prep.WatchForPodManifestsForNode(successMainUpdate, errMainUpdate, quitMainUpdate)
+	go prep.WatchForHooks(successHookUpdate, errHookUpdate, quitHookUpdate)
+
+	if preparerConfig.StatusPort != 0 {
+		http.HandleFunc("/_status", statusHandler(successMainUpdate, successHookUpdate, errMainUpdate, errHookUpdate))
+		go http.ListenAndServe(fmt.Sprintf(":%d", preparerConfig.StatusPort), nil)
+	} else {
+		// Discard everything from these channel!
```
bwester: If you make the senders non-blocking, you don't need this.

anthonybishopric:
I'm inclined to agree with @bwester that this may not be the right way to gauge healthiness of the preparer, especially given the added complexity in managing all these concurrent routines. In some ways, I think I'd just be happy to have the `/_status` endpoint always return a healthy status.

Some rationalization for that: the main use of the status port is to prevent a rollout of the preparer itself to prevent future deploys. If a new preparer can't be started _at all_ then this status port will have achieved its purpose.

Additionally, we could launch a dummy pod right at the beginning of the preparer's execution (basically a `pod.Launch()` on a pod we know should work). As long as that first pod works, we always return 200. This prevents the deployment of a preparer that can't deploy anything at all.

Heuristics about how many errors signify a serious issue won't be helpful during rollout - not enough errors would accumulate to actually provide feedback to `p2-replicate`.

bwester:
+1 for `/_status` always returning success, for now. That will be an improvement to the current state, and we can implement something more sophisticated later, when we know what we want.

petertseng:
Yup, HTTP version done. I could make it HTTPS but I need some certs and we currently don't have those, it seems?
Because of this consideration, we'll need to remember to add `status_http` to our manifest.
Or we could take some time to make some certs...

bwester:
:ship:

PR #178, opened 2015-05-26
Balancer Proof-of-Concept 

anthonybishopric:
This is a simple implementation of a least-connections load balancer that uses Consul to determine healthy backends to route to. It has two server modes - either a dedicated server for a particular service name on a particular port, or using SNI-based virtual hosts.

When using virtual hosting, the host named in the SNI will be used to determine the service name.

Alternate load balancing strategies and backend monitors are possible (ie, use round-robin with a reality-tree based lookup instead of Consul services)

### DIFF
--- a/bin/p2-balancer/main.go
+++ b/bin/p2-balancer/main.go
@@ -20,15 +20,18 @@ func main() {
 	logger := logging.DefaultLogger
 
 	if err != nil {
-		logger.NoFields().Fatalf("Could not init serpro: %s", err)
+		logger.WithField("err", err).Fatalf("Could not load configuration for p2-balancer")
 	}
 
 	service, err := cfg.ReadString("service_to_proxy")
 	if err != nil {
-		logger.WithField("err", err).Fatalln("Could not get the service to proxy")
+		logger.WithField("err", err).Fatalln("Could not read the service_to_proxy the field from config")
 	}
 
 	vhost, err := cfg.ReadString("vhost")
+	if err != nil {
+		logger.WithField("err", err).Fatalln("Could not read the vhost field from config")
+	}
 
 	if service == "" && vhost == "" {
 		logger.NoFields().Fatalln("Did not specify a service or vhost key in config")
@@ -90,6 +93,7 @@ func doVhost(port int, signalCh chan os.Signal, monitor balancer.Monitor) error 
 	})
 	logger.WithField("port", port).Infoln("Serving TLS traffic for all registered services")
 	vhostServer := balancer.NewVhostServer(monitor, &logger)
+	defer vhostServer.Cleanup()
 	errCh := make(chan error)
 	go func() {
 		err := vhostServer.Serve(port)
--- a/bin/p2-balancer/main.go
+++ b/bin/p2-balancer/main.go
@@ -1,0 +1,105 @@
+package main
+
+import (
+	"os"
+	"os/signal"
+	"syscall"
+
+	"github.com/Sirupsen/logrus"
+	"github.com/hashicorp/consul/api"
+	"github.com/square/p2/pkg/balancer"
+	"github.com/square/p2/pkg/config"
+	"github.com/square/p2/pkg/logging"
+)
+
+// serpro (service proxy) is a simple TCP proxy for services registered
+// in the Consul registry.
+
+func main() {
+	cfg, err := config.LoadFromEnvironment()
+	logger := logging.DefaultLogger
+
+	if err != nil {
+		logger.NoFields().Fatalf("Could not init serpro: %s", err)
+	}
+
+	service, err := cfg.ReadString("service_to_proxy")
+	if err != nil {
+		logger.WithField("err", err).Fatalln("Could not get the service to proxy")
+	}
+
+	vhost, err := cfg.ReadString("vhost")
+
+	if service == "" && vhost == "" {
+		logger.NoFields().Fatalln("Did not specify a service or vhost key in config")
+	}
+
+	port, err := cfg.ReadInt("port")
+	if err != nil {
+		logger.WithField("err", err).Fatalln("Could not get port number")
+	}
+	if port == 0 {
+		port = 443
+	}
+
+	monitor, err := balancer.NewConsulMonitor(api.DefaultConfig(), &logger)
+
+	if err != nil {
+		logger.NoFields().Fatalln(err)
+	}
+
+	signalCh := make(chan os.Signal, 2)
+	signal.Notify(signalCh, syscall.SIGTERM, os.Interrupt)
+
+	if service != "" {
+		err = doServer(service, port, signalCh, monitor)
+	} else {
+		err = doVhost(port, signalCh, monitor)
+	}
+	if err != nil {
+		logger.WithField("err", err).Fatalln("Error running")
+	}
+}
+
+func doServer(service string, port int, signalCh chan os.Signal, monitor balancer.Monitor) error {
+	logger := logging.NewLogger(logrus.Fields{
+		"service": service,
+	})
+	logger.WithField("port", port).Infoln("Serving TLS traffic for service")
+
+	server := balancer.NewServer(monitor, &logger)
+
+	errCh := make(chan error)
+	go func() {
+		leastConnections, _ := balancer.NewLeastConnectionsStrategy([]string{})
+		err := server.Serve(service, port, leastConnections)
+		errCh <- err
+	}()
+	select {
+	// add cleanup here
+	case err := <-errCh:
+		return err
+	case <-signalCh:
+		return nil
+	}
+}
+
+func doVhost(port int, signalCh chan os.Signal, monitor balancer.Monitor) error {
+	logger := logging.NewLogger(logrus.Fields{
+		"vhost": true,
+	})
+	logger.WithField("port", port).Infoln("Serving TLS traffic for all registered services")
+	vhostServer := balancer.NewVhostServer(monitor, &logger)
+	errCh := make(chan error)
+	go func() {
+		err := vhostServer.Serve(port)
+		errCh <- err
+	}()
+	select {
+	// add cleanup here
+	case err := <-errCh:
+		return err
+	case <-signalCh:
+		return nil
+	}
+}
--- a/bin/p2-balancer/main.go
+++ b/bin/p2-balancer/main.go
@@ -5,8 +5,8 @@ import (
 	"os/signal"
 	"syscall"
 
-	"github.com/Sirupsen/logrus"
-	"github.com/hashicorp/consul/api"
+	"github.com/square/p2/Godeps/_workspace/src/github.com/Sirupsen/logrus"
+	"github.com/square/p2/Godeps/_workspace/src/github.com/hashicorp/consul/api"
 	"github.com/square/p2/pkg/balancer"
 	"github.com/square/p2/pkg/config"
 	"github.com/square/p2/pkg/logging"
@@ -15,34 +15,28 @@ import (
 // serpro (service proxy) is a simple TCP proxy for services registered
 // in the Consul registry.
 
+type balancerConfig struct {
+	serviceToProxy string `yaml:"service_to_proxy"`
+	vhost          string `yaml:"vhost"`
+	port           int    `yaml:"port"`
+}
+
 func main() {
-	cfg, err := config.LoadFromEnvironment()
+	cfg := &balancerConfig{}
+	err := config.LoadFromEnvInto(cfg)
+
 	logger := logging.DefaultLogger
 
 	if err != nil {
 		logger.WithField("err", err).Fatalf("Could not load configuration for p2-balancer")
 	}
 
-	service, err := cfg.ReadString("service_to_proxy")
-	if err != nil {
-		logger.WithField("err", err).Fatalln("Could not read the service_to_proxy the field from config")
-	}
-
-	vhost, err := cfg.ReadString("vhost")
-	if err != nil {
-		logger.WithField("err", err).Fatalln("Could not read the vhost field from config")
-	}
-
-	if service == "" && vhost == "" {
+	if cfg.serviceToProxy == "" && cfg.vhost == "" {
 		logger.NoFields().Fatalln("Did not specify a service or vhost key in config")
 	}
 
-	port, err := cfg.ReadInt("port")
-	if err != nil {
-		logger.WithField("err", err).Fatalln("Could not get port number")
-	}
-	if port == 0 {
-		port = 443
+	if cfg.port == 0 {
+		cfg.port = 443
 	}
 
 	monitor, err := balancer.NewConsulMonitor(api.DefaultConfig(), &logger)
@@ -54,10 +48,10 @@ func main() {
 	signalCh := make(chan os.Signal, 2)
 	signal.Notify(signalCh, syscall.SIGTERM, os.Interrupt)
 
-	if service != "" {
-		err = doServer(service, port, signalCh, monitor)
+	if cfg.serviceToProxy != "" {
+		err = doServer(cfg.serviceToProxy, cfg.port, signalCh, monitor)
 	} else {
-		err = doVhost(port, signalCh, monitor)
+		err = doVhost(cfg.port, signalCh, monitor)
 	}
 	if err != nil {
 		logger.WithField("err", err).Fatalln("Error running")
--- a/integration/single-node-slug-deploy/check.go
+++ b/integration/single-node-slug-deploy/check.go
@@ -371,6 +371,7 @@ func verifyHelloRunning() error {
 		if err != nil {
 			return err
 		}
+		defer resp.Body.Close()
 		if resp.StatusCode != http.StatusOK {
 			body, _ := ioutil.ReadAll(resp.Body)
 			return util.Errorf("Did not OK response from hello: %s %s", resp.Status, string(body))
--- a/integration/single-node-slug-deploy/check.go
+++ b/integration/single-node-slug-deploy/check.go
@@ -104,8 +104,8 @@ func signManifest(manifestPath string, workdir string) (string, error) {
 }
 
 func generatePreparerPod(workdir string) (string, error) {
-	// build the artifact from HEAD
-	err := exec.Command("go", "build", "github.com/square/p2/bin/p2-preparer").Run()
+	// build the artifact with -race to check for race conditions
+	err := exec.Command("go", "build", "-race", "github.com/square/p2/bin/p2-preparer").Run()
 	if err != nil {
 		return "", util.Errorf("Couldn't build preparer: %s", err)
 	}
--- a/integration/single-node-slug-deploy/check.go
+++ b/integration/single-node-slug-deploy/check.go
@@ -311,6 +311,8 @@ func postHelloManifest(dir string) error {
 		LaunchableType: "hoist",
 		Location:       hello,
 	}
+	manifest.StatusPort = 43770
+	manifest.StatusHTTP = true
 	manifest.LaunchableStanzas = map[string]pods.LaunchableStanza{
 		"hello": stanza,
 	}
@@ -363,6 +365,16 @@ func verifyHelloRunning() error {
 	case <-time.After(20 * time.Second):
 		return fmt.Errorf("Couldn't start hello after 15 seconds:\n\n %s", targetLogs())
 	case <-helloPidAppeared:
+		log.Println("Hello PID appeared, letting start up and then checking for responsiveness")
+		<-time.After(10 * time.Second)
+		resp, err := http.Get("http://localhost:43770/_status")
+		if err != nil {
+			return err
+		}
+		if resp.StatusCode != http.StatusOK {
+			body, _ := ioutil.ReadAll(resp.Body)
+			return util.Errorf("Did not OK response from hello: %s %s", resp.Status, string(body))
+		}
 		return nil
 	}
 }
--- a/integration/suite.rb
+++ b/integration/suite.rb
@@ -72,11 +72,7 @@ Dir.glob(File.join(path, '*/')).each do |test_dir|
       puts "#{test_name} #{'SUCCEEDED'.green}"
     ensure
       if options[:vm_maintenance] && succeeded
-        puts "Halting VM".yellow
-        unless system("vagrant halt")
-          puts "Tried to halt #{test_name} but it failed".red
-        end
-        puts "Halted VM"
+        
       end
     end
   end if options[:regex].match(test_name)
--- a/integration/suite.rb
+++ b/integration/suite.rb
@@ -72,7 +72,11 @@ Dir.glob(File.join(path, '*/')).each do |test_dir|
       puts "#{test_name} #{'SUCCEEDED'.green}"
     ensure
       if options[:vm_maintenance] && succeeded
-        
+        puts "Halting VM".yellow
+        unless system("vagrant halt")
+          puts "Tried to halt #{test_name} but it failed".red
+        end
+        puts "Halted VM"
       end
     end
   end if options[:regex].match(test_name)
--- a/pkg/balancer/consul_monitor.go
+++ b/pkg/balancer/consul_monitor.go
@@ -5,8 +5,8 @@ import (
 	"strings"
 	"time"
 
-	"github.com/Sirupsen/logrus"
-	"github.com/hashicorp/consul/api"
+	"github.com/square/p2/Godeps/_workspace/src/github.com/Sirupsen/logrus"
+	"github.com/square/p2/Godeps/_workspace/src/github.com/hashicorp/consul/api"
 	"github.com/square/p2/pkg/health"
 	"github.com/square/p2/pkg/logging"
 )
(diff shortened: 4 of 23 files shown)


bwester:
> The final implementation of `Routable`...

Does this mean you have further changes to push before this PR is ready to be reviewed?

integration/single-node-slug-deploy/check.go
```diff
@@ -288,6 +291,16 @@ func verifyHelloRunning() error {
 		preparerT.Run()
 		return fmt.Errorf("Couldn't start hello after 15 seconds: \n\n hello tail: \n%s\n\n preparer tail: \n%s", helloTail.String(), preparerTail.String())
 	case <-helloPidAppeared:
+		log.Println("Hello PID appeared, letting start up and then checking for responsiveness")
+		<-time.After(10 * time.Second)
```
bwester: [`time.Sleep()`](http://golang.org/pkg/time/#Sleep)

And why 10 seconds?

anthonybishopric: Totally heuristic

anthonybishopric:
@bwester there's a later commit in this PR that revises the API.

integration/single-node-slug-deploy/check.go
```diff
@@ -288,6 +291,16 @@ func verifyHelloRunning() error {
...
+		resp, err := http.Get("http://localhost:43770/_status")
+		if err != nil {
+			return err
+		}
+		if resp.StatusCode != http.StatusOK {
```
bwester: Need to close `resp.Body`.

integration/suite.rb
```diff
@@ -72,11 +72,7 @@ def pink
       puts "#{test_name} #{'SUCCEEDED'.green}"
     ensure
       if options[:vm_maintenance] && succeeded
-        puts "Halting VM".yellow
-        unless system("vagrant halt")
-          puts "Tried to halt #{test_name} but it failed".red
-        end
-        puts "Halted VM"
+
```
bwester: The blank line seems conspicuous. Is something missing?  Is there another way to clean up the VM?

anthonybishopric: Halting the VM doesn't clean up at all (you need to destroy it to actually clean it up). My initial goal with halting the VM after the test was to save memory as you moved onto the next test.

I'm happy to revert this change, it was convenient for testing the balancer locally to have a VM with a functioning preparer on it. It turns out that halting the VM with a standalone Consul instance causes Consul to lose all of the registered services in the catalog (this does not happen with Consul instances in a quorum) which makes the balancer fail tremendously.

pkg/balancer/server.go
```diff
@@ -0,0 +1,82 @@
...
+	"fmt"
+	"io"
+	"net"
+	"time"
+
+	"github.com/square/p2/pkg/logging"
+	"github.com/square/p2/pkg/util"
+)
+
+type Handle func(outbound net.Conn) error
```
bwester: I feel like `Handler` as a name makes the code a bit easier to read.

integration/suite.rb
bwester: I admit there are different levels of "cleaning up," but leaving a VM running in the background seems like a bad thing to do. It sounds like you were using the integration test's VM as a quick way to launch a Consul instance. That's great for development, but it shouldn't be the preferred way to do integration tests.  Maybe we should set up a separate Vagrant file for live testing while writing new features?

bin/p2-balancer/main.go
```diff
@@ -0,0 +1,105 @@
...
+)
+
+// serpro (service proxy) is a simple TCP proxy for services registered
+// in the Consul registry.
+
+func main() {
+	cfg, err := config.LoadFromEnvironment()
+	logger := logging.DefaultLogger
+
+	if err != nil {
```
bwester: I'd prefer more precise and actionable error messages throughout here:

* could not load configuration from $CONFIG_PATH: …
* missing required configuration: "service_to_proxy"

…

integration/suite.rb
anthonybishopric: +1, will revert and make a dev Vagrantfile

pkg/balancer/least_connections.go
```diff
@@ -0,0 +1,128 @@
...
+		if err != nil {
+			return nil, err
+		}
+	}
+	return l, nil
+}
+
+func (l *LeastConnections) Route(fn Handle) error {
+	address, err := l.acquireAddress()
+	defer l.releaseAddress(address)
```
bwester: I think you want this _after_ the check for `err`, not before.

anthonybishopric: Good catch

bin/p2-balancer/main.go
```diff
@@ -0,0 +1,105 @@
...
+	"github.com/square/p2/pkg/balancer"
+	"github.com/square/p2/pkg/config"
+	"github.com/square/p2/pkg/logging"
+)
+
+// serpro (service proxy) is a simple TCP proxy for services registered
+// in the Consul registry.
+
+func main() {
+	cfg, err := config.LoadFromEnvironment()
```
mpuncel: Unrelated to this PR, but I feel like the API for config should be LoadFromEnvironment(root string, some_struct interface{})

It will yaml marshal into the struct pointer you pass in. Then you don't need all these ReadString and ReadMap etc

mpuncel: I guess this is all in the top level? in any case you should be able to pass a struct with the fields you want

pkg/balancer/consul_monitor.go
```diff
@@ -0,0 +1,80 @@
...
+				address := fmt.Sprintf("%s:%d", serviceEntry.Node.Address, serviceEntry.Service.Port)
+				for _, check := range serviceEntry.Checks {
+					state := health.ToHealthState(check.Status)
+					if state == health.Critical || state == health.Unknown {
+						healthy = false
+						c.logger.WithFields(logrus.Fields{
+							"service": check.ServiceName,
+							"status":  string(state),
+							"address": address,
+						}).Warningln("Address is reporting unhealthily")
```
mpuncel: Maybe this would make more sense in the wild but this warning seems confusing

mpuncel: Looks like Address must be a balancer primitive lower down in this PR, my comment may be irrelevant

pkg/balancer/least_connections.go
```diff
@@ -0,0 +1,128 @@
...
+	if err != nil {
+		return util.Errorf("Could acquire %s: %s", address, err)
```
mpuncel: Could *not* acquire

pkg/balancer/least_connections.go
```diff
@@ -0,0 +1,128 @@
...
+
+	"github.com/square/p2/pkg/util"
+)
+
+type lcBackend struct {
+	conns    int
+	disabled bool
+}
+
+type LeastConnections struct {
```
mpuncel: I know this is just an experiment, but any reason you didn't just round robin? If a host is somehow in the pool and connections fail immediately, this lb strategy would send all new requests to it (because it has the least connections)

pkg/balancer/least_connections.go
```diff
@@ -0,0 +1,128 @@
...
+
+func (l *LeastConnections) RemoveAddress(address string) error {
+	l.mapMux.Lock()
+	defer l.mapMux.Unlock()
+	backend, ok := l.backends[address]
+	if !ok {
+		return nil
+	}
+	// disabling allows us to gracefully release connections without assigning
+	// new connections to the backend.
```
mpuncel: i'd prefer this comment be in the definition of the lcBackend struct

pkg/balancer/least_connections.go
```diff
@@ -0,0 +1,128 @@
...
+	if !ok {
+		return util.Errorf("Address %s not in set, cannot release", address)
+	}
+	backend.conns--
+	return nil
+}
+
+// least connections is routable as long as there is at least one non-disabled
+// backend
+func (l *LeastConnections) Routable(duration time.Duration) error {
```
mpuncel: Should probably be called WaitRoutable. I expected this to return a bool and not block based on the name

pkg/balancer/server.go
mpuncel: agreed

pkg/balancer/server.go
```diff
@@ -0,0 +1,82 @@
...
+	logger  *logging.Logger
+	monitor Monitor
+}
+
+func DuplexConnection(inbound net.Conn) Handle {
+	return func(outbound net.Conn) error {
+		done := make(chan struct{})
+		go func() {
+			io.Copy(outbound, inbound)
+			outbound.Close()
```
mpuncel: how do you know that these Close() calls won't interrupt the other Copy()?

mpuncel:
```        // Close closes the connection.
        // Any blocked Read or Write operations will be unblocked and return errors.
        Close() error```

File: bin/p2-balancer/main.go (hunk 7 above)
Comment (bwester) [inline bin/p2-balancer/main.go]: :+1:

File: pkg/balancer/server.go [hunk 14]
```diff
@@ -0,0 +1,82 @@
...
+	AddAddress(addr string) error
+	RemoveAddress(addr string) error
+	Routable(wait time.Duration) error
+}
+
+type Monitor interface {
+	MonitorHosts(service string, strategy Strategy, quitCh <-chan struct{})
+}
+
+type Server struct {
```
mpuncel: should probably not export this to enforce use of the NewServer()?

pkg/balancer/vhost_server.go
```diff
@@ -0,0 +1,96 @@
...
+	"time"
+
+	vhost "github.com/inconshreveable/go-vhost"
+	"github.com/square/p2/pkg/logging"
+	"github.com/square/p2/pkg/util"
+)
+
+// VhostServer uses TLS virtual hosting using SNI to direct requests to
+// an appropriate backend.
+type VhostServer struct {
```
mpuncel: don't export?

pkg/balancer/vhost_server.go
```diff
@@ -0,0 +1,96 @@
...
+}
+
+func NewVhostServer(monitor Monitor, logger *logging.Logger) *VhostServer {
+	return &VhostServer{make(map[string]Strategy), monitor, logger, make([]chan struct{}, 0), sync.Mutex{}}
+}
+
+func (m *VhostServer) Serve(port int) error {
+	l, err := net.Listen("tcp", fmt.Sprintf(":%d", port))
+	if err != nil {
+		return util.Errorf("Could begin listening: %s", err)
```
mpuncel: could not*

pkg/balancer/vhost_server.go
```diff
@@ -0,0 +1,96 @@
...
+	}
+	if err != nil {
```
mpuncel: you're checking error twice in a row

pkg/balancer/vhost_server.go
```diff
@@ -0,0 +1,96 @@
...
+		return util.Errorf("Could not init muxer on port %d: %s", port, err)
+	}
+	for {
+		conn, err := l.Accept()
+		if err != nil {
+			m.logger.WithField("err", err).Warnln("Connection failed")
+		}
+		go m.handleConnection(conn)
```
mpuncel: you're still doing this even if Accept() had an error

pkg/balancer/vhost_server.go
```diff
@@ -0,0 +1,96 @@
...
+		routableErrCh := make(chan error)
+		go func() {
+			err := strategy.Routable(15 * time.Second)
+			if err != nil {
+				routableErrCh <- err
+			} else {
+				routableCh <- struct{}{}
+			}
+		}()
+		go m.monitor.MonitorHosts(service, strategy, monitorQuitCh)
```
mpuncel: this is the only place where service will be a hostname instead of just a consul service name right? Probably better to do the strings.Split() herre instead of in the MonitorHosts method

pkg/balancer/vhost_server.go
```diff
@@ -0,0 +1,96 @@
...
+		select {
+		case err = <-routableErrCh:
+			monitorQuitCh <- struct{}{}
+			return nil, util.Errorf("Could not find routable backend to %s: %s", service, err)
+		case <-routableCh:
+			m.toQuit = append(m.toQuit, monitorQuitCh)
```
mpuncel: seems like m.toQuit isn't used anywhere

mpuncel:
pretty cool stuff

pkg/balancer/least_connections.go
anthonybishopric: Least connections is one of the benefits of having a centralized load balancer (client-side load balancers cannot know who is less loaded) so I wanted to highlight it. Round-robin would certainly be easier. Depending on the workloads, servers can end up better off with least-connections balancing over round-robin.

You bring up an interesting problem. The host should be removed from the pool by the monitor relatively quickly in such cases. We could potentially aggressively disable hosts where we notice that `net.Dial` is failing, but we can't detect serious issues at the HTTP level since we aren't terminating TLS.

pkg/balancer/server.go
anthonybishopric: Good catch. `net.TCPConn` has `CloseRead` and `CloseWrite` - I'll use those

pkg/balancer/vhost_server.go
anthonybishopric: it's used in `bin/balancer`

pkg/balancer/least_connections.go
mpuncel: yeah there will always be potential for a bad host that you can't detect. Health check might work but a typo in a rails project might cause requests to fail for example

pkg/balancer/vhost_server.go
anthonybishopric: I actually had it here first, but moved it to the monitor. This was anticipating smarter service-discovery clients with more sophisticated cluster specification schemes being embedded in the hostname, and I didn't want to force any one monitor to be restricted to one form of service naming.

pkg/balancer/vhost_server.go
anthonybishopric: will remove

pkg/balancer/vhost_server.go
anthonybishopric: good catch

pkg/balancer/vhost_server.go
mpuncel: looks like you use the constructor though? i assume you mean bin/p2-balancer/main.go?

anthonybishopric: good point, I will unexport this and the normal server as well.

bin/p2-balancer/main.go
anthonybishopric: A lot of the clarifying details about the errors will be included in the "err" field, but I will revise some of these to be clearer.

pkg/balancer/consul_monitor.go
```diff
@@ -0,0 +1,80 @@
...
+		client,
+		logger,
+	}, nil
+}
+
+func (c *ConsulMonitor) MonitorHosts(service string, strategy Strategy, quitCh <-chan struct{}) {
+	// strip everything after the first . in case the service passed is a vhost
+	service = strings.Split(service, ".")[0]
+
+	backoff := time.Duration(5)
```
bwester: `backoff` is really a poll interval; you only start an exponential backoff when an error is returned. What do you hope to gain from an exponential backoff?

Since you refer to "5" and "30" (below) twice each, make them const values.

pkg/balancer/least_connections.go
```diff
@@ -0,0 +1,129 @@
...
+
+	"github.com/square/p2/pkg/util"
+)
+
+type lcBackend struct {
+	conns int
+
+	// disabling allows us to gracefully release connections without assigning
+	// new connections to the backend.
+	disabled bool
```
bwester: Two suggestions:

1. Flip this variable to be `enabled bool`
2. Change `LeastConnections.backends` below to be a pointer-less `map[string]lcBackend`

Then the struct's zero value corresponds to a disabled host with no connections, and you can rely on that zero value to simplify the logic when reading from `backends`.

bwester: I think disabling a connection record without removing its connection count is the right way to handle removing an address, but I don't understand what you mean by enabling/disabling allowing for a "graceful release" of connections. Could you clarify?

pkg/balancer/least_connections.go
```diff
@@ -0,0 +1,129 @@
...
+}
+
+func (l *LeastConnections) RemoveAddress(address string) error {
+	l.mapMux.Lock()
+	defer l.mapMux.Unlock()
+	backend, ok := l.backends[address]
+	if !ok {
+		return nil
+	}
+	backend.disabled = true
```
bwester: Here and in `releaseAddress()` below, once the record becomes disabled with no connections, it should eventually be removed from `backends` to prevent a memory leak.

pkg/balancer/least_connections.go
```diff
@@ -0,0 +1,129 @@
...
+	backend.conns--
+	return nil
+}
+
+// least connections is routable as long as there is at least one non-disabled
+// backend
+func (l *LeastConnections) WaitRoutable(duration time.Duration) error {
+	after := time.After(duration)
+	for {
+		l.mapMux.Lock()
```
bwester: How about using a [`sync.Cond`](http://golang.org/pkg/sync/#Cond) instead of busy-waiting?

bin/p2-balancer/main.go
```diff
@@ -0,0 +1,109 @@
...
+		leastConnections, _ := balancer.NewLeastConnectionsStrategy([]string{})
+		err := server.Serve(service, port, leastConnections)
+		errCh <- err
+	}()
+	select {
+	// add cleanup here
+	case err := <-errCh:
+		return err
+	case <-signalCh:
+		return nil
```
bwester: If this returns instantly on a SIGTERM, the main goroutine will exit, severing all open connections. Could you instead route the exit request through the `Server` to give it a chance to stop accepting new connections and wait for its existing ones to end?

pkg/balancer/least_connections.go
```diff
@@ -0,0 +1,129 @@
...
+		for _, backend := range l.backends {
+			if !backend.disabled {
+				l.mapMux.Unlock()
```
bwester: `defer l.mapMux.Unlock()`

pkg/balancer/server.go
```diff
@@ -0,0 +1,82 @@
...
+	MonitorHosts(service string, strategy Strategy, quitCh <-chan struct{})
+}
+
+type server struct {
+	logger  *logging.Logger
+	monitor Monitor
+}
+
+func DuplexConnection(inbound net.Conn) Handler {
+	return func(outbound net.Conn) error {
```
bwester: O, how I miss partial application, working with golang... :broken_heart:

pkg/balancer/least_connections.go
anthonybishopric: You would still need a solution that prevents a <wait time> -long mutex from being held, since no requests can be routed while this lock is held. I think that involves some amount of busy waiting. Go's maps are not threadsafe and could panic if we attempted to loop during concurrent modifications.

pkg/balancer/server.go
```diff
@@ -0,0 +1,82 @@
...
+		done := make(chan struct{})
+		defer inbound.Close()
+		defer outbound.Close()
```
bwester: You're already closing `inbound` in `server.handleConnection()` and `outbound` in `LeastConnections.Route()`

pkg/balancer/server.go
```diff
@@ -0,0 +1,82 @@
...
+
+		go func() {
+			io.Copy(outbound, inbound)
+			close(done)
+		}()
+		io.Copy(inbound, outbound)
```
bwester: I'm not sure if this is a problem, but will these two goroutines actually terminate if either connection is closed? Early? How much did you test this?

pkg/balancer/least_connections.go
bwester: > You would still need a solution that prevents a -long mutex from being held

Like a [condition variable](http://en.wikipedia.org/wiki/Monitor_%28synchronization%29)?

anthonybishopric: Ah yes, good call. Sorry I misunderstood your suggestion

pkg/balancer/least_connections.go
```diff
@@ -0,0 +1,129 @@
...
+	// new connections to the backend.
+	disabled bool
+}
+
+type LeastConnections struct {
+	backends map[string]*lcBackend
+	mapMux   sync.Mutex
+}
+
+func NewLeastConnectionsStrategy(initialAddresses []string) (*LeastConnections, error) {
```
bwester: I think you can get rid of `initialAddresses`. It's always given a zero value in non-test code, and even for test code, you could get the same behavior by creating a test-only wrapper function that calls `AddAddress()` in a loop.  There's nothing special about this constructor.

pkg/balancer/vhost_server.go
```diff
@@ -0,0 +1,100 @@
...
+	m.strategyMux.Lock()
+	defer m.strategyMux.Unlock()
+	strategy, ok := m.muxed[service]
+	if !ok {
+		var err error
+		strategy, err = NewLeastConnectionsStrategy([]string{})
+		if err != nil {
+			return nil, err
+		}
+		// Wait for 15 seconds for routability while starting the monitoring process
```
bwester: _All_ new connections will block whenever one connection accesses a new service.  A better strategy: drop the mutex, then create the strategy, then reacquire the mutex and install the strategy into `m.muxed`.

bwester: Are you talking about the case where many incoming connections want to access the same new service? I agree that you want only one connection to initiate a strategy; other connections should wait for the first one to finish. Serializing requests through a goroutine would work. If you wanted to continue using locks, there are solutions there too.

bin/p2-balancer/main.go
```diff
@@ -0,0 +1,109 @@
...
+	cfg, err := config.LoadFromEnvironment()
+	logger := logging.DefaultLogger
+
+	if err != nil {
+		logger.WithField("err", err).Fatalf("Could not load configuration for p2-balancer")
+	}
+
+	service, err := cfg.ReadString("service_to_proxy")
+	if err != nil {
+		logger.WithField("err", err).Fatalln("Could not read the service_to_proxy the field from config")
```
petertseng: `the service_to_proxy the field`

pkg/balancer/least_connections.go
```diff
@@ -0,0 +1,129 @@
...
+	if err != nil {
+		return util.Errorf("Could not acquire %s: %s", address, err)
+	}
+	defer l.releaseAddress(address)
+	outbound, err := net.Dial("tcp", address)
+	if err != nil {
+		return util.Errorf("Could not dial %s: %s", address, err)
+	}
+	defer outbound.Close()
+	err = fn(outbound.(*net.TCPConn))
```
petertseng: I was a bit surprised. Is http://golang.org/pkg/net/#DialTCP unsuitable for some reason or other?

pkg/balancer/server.go
```diff
@@ -0,0 +1,82 @@
...
+	"time"
+
+	"github.com/square/p2/pkg/logging"
+	"github.com/square/p2/pkg/util"
+)
+
+type Handler func(outbound net.Conn) error
+
+type Strategy interface {
+	// Route is expected to synchronously call the given Handle with an outbound connection.
```
petertseng: Now that you changed `Handle` to `Handler`, perhaps this comment as well.

pkg/balancer/vhost_server.go
```diff
@@ -0,0 +1,100 @@
...
+func (m *vhostServer) handleConnection(conn net.Conn) {
+	defer conn.Close()
+	tlsConn, err := vhost.TLS(conn)
+	if err != nil {
+		m.logger.WithField("err", err).Errorln("Failed to recognize a TLS connection")
+		return
+	}
+	defer tlsConn.Close()
+	service := tlsConn.Host()
+	strategy, err := m.determineStrategy(service)
```
petertseng: this `err` never checked (immediately overwritten on next line)

integration/single-node-slug-deploy/check.go
```diff
@@ -288,6 +291,17 @@ func verifyHelloRunning() error {
 		preparerT.Run()
 		return fmt.Errorf("Couldn't start hello after 15 seconds: \n\n hello tail: \n%s\n\n preparer tail: \n%s", helloTail.String(), preparerTail.String())
 	case <-helloPidAppeared:
+		log.Println("Hello PID appeared, letting start up and then checking for responsiveness")
+		<-time.After(10 * time.Second)
+		resp, err := http.Get("http://localhost:43770/_status")
```
petertseng: If you rebase on master you could use https://github.com/square/p2/blob/master/integration/single-node-slug-deploy/check.go#L137 as `checkStatus(43770, "hello")` that I added in PR #180

anthonybishopric:
killing this PR for now

PR #179, opened 2015-05-26
Reuse error handling logic in policy test

bwester:
Some slight fixes to 653cf1e to preserve the common usage of `testHarness`.


petertseng:
OK looks good

PR #181, opened 2015-05-28
Authorize hook deployment as if they were the preparer

bwester:
Hooks are not the same as apps: they are the method of extending the preparer,
not something that users will use for pushing apps. This commit separates
`auth.Policy.AuthorizePod()` into `AuthorizeApp()` for use on apps and
`AuthorizeHook()` for use on hooks. In most policies, the two are treated the
same. In `UserPolicy`, hooks are authorized using the name of the preparer, not
the name of the hook's pod.


anthonybishopric:
:+1:

mpuncel:
ship it!

bwester:
Updated: fix grammar in the commit title

PR #183, opened 2015-06-03
Force `p2-schedule` to terminate if a manifest is unschedulable

anthonybishopric:
Also exposes an `UnsafeError` field on the KVError type. Consul's
errors are not safe to print since they frequently include ACL tokens
in the text, so this can be selectively inspected during triage.


petertseng:
LGTM (though I mention my preference would have been to put the UnsafeError in separate commit)

PR #184, opened 2015-06-09
small logging improvements

anthonybishopric:
Allows callers of p2-inspect to read the unsafe error instead of
the sanitized one (assumes that all callers of p2-inspect are human)

Prints name of hook executed instead of an unqualified name.

### DIFF
--- a/bin/p2-inspect/main.go
+++ b/bin/p2-inspect/main.go
@@ -59,11 +59,11 @@ func main() {
 
 	intents, _, err := store.ListPods(kp.INTENT_TREE)
 	if err != nil {
-		log.Fatalf("Could not list intent kvpairs: %s", err)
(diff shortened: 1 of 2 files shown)


bin/p2-inspect/main.go
```diff
@@ -59,11 +59,11 @@ func main() {
 
 	intents, _, err := store.ListPods(kp.INTENT_TREE)
 	if err != nil {
-		log.Fatalf("Could not list intent kvpairs: %s", err)
+		log.Fatalf("Could not list intent kvpairs: %s", err.(kp.KVError).UnsafeError)
```
petertseng: I'm in general OK with this, but what will happen in code paths where `err` is not a `kp.KVError`? I believe there is such a code path. See at https://github.com/square/p2/blob/master/pkg/kp/kv.go#L94 and https://github.com/square/p2/blob/master/pkg/pods/manifest.go#L81 - are you OK with that?

PR #185, opened 2015-06-09
make kp.Store into an interface, with consulStore implementing that i…

mpuncel:
…nterface

### DIFF
--- a/pkg/kp/kv.go
+++ b/pkg/kp/kv.go
@@ -17,12 +17,24 @@ type ManifestResult struct {
 	Path     string
 }
 
-type Store struct {
+type Store interface {
+	SetPod(key string, manifest pods.Manifest) (time.Duration, error)
+	Pod(key string) (*pods.Manifest, time.Duration, error)
+	WatchPods(keyPrefix string, quitChan <-chan struct{}, errChan chan<- error, podChan chan<- ManifestResult)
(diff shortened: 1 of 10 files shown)


anthonybishopric:
:+1:

pkg/kp/kv.go
```diff
@@ -17,12 +17,19 @@ type ManifestResult struct {
...
+}
+
+type consulStore struct {
 	client *api.Client
 }
 
-func NewStore(opts Options) *Store {
-	return &Store{client: NewConsulClient(opts)}
+func NewStore(opts Options) Store {
+	return &consulStore{client: NewConsulClient(opts)}
```
bwester: I don't think you need to return a pointer here.

bwester:
From the CI build failures, there are some more call sites to convert, but otherwise this is looking good.

bwester:
Github has problems displaying your commit message.  Is it a problem with your commit itself, or does Github just not like long lines?

mpuncel:
the commit message just wrapped into the description. The full commit message is "make kp.Store into an interface, with consulStore implementing that interface"

petertseng:
github is cutting off at 70 chars (counting the trailing "…" as one character)

bwester:
:ship:

PR #187, opened 2015-06-09
Authorize a pod based on its manifest's user, not its name

bwester:
Changes the UserPolicy authorization policy to look for authorization based on
the user the pod will run as, not the name of the pod. This makes the tool more
similar to "sudo", in that users are authorized to run commands (deploy
software) as other users.

This commit also renames the configuration parameter "forbidden_pod_ids" to
"forbidden_users".  Even before this commit, it was blocking users in the
"run_as" field, but its name did not convey this.

### DIFF
--- a/pkg/preparer/orchestrate.go
+++ b/pkg/preparer/orchestrate.go
@@ -35,15 +35,15 @@ type Store interface {
 }
 
 type Preparer struct {
-	node            string
-	store           Store
-	hooks           Hooks
-	hookListener    HookListener
-	Logger          logging.Logger
-	podRoot         string
-	forbiddenPodIds map[string]struct{}
-	caPath          string
-	authPolicy      auth.Policy
+	node           string
+	store          Store
+	hooks          Hooks
+	hookListener   HookListener
+	Logger         logging.Logger
+	podRoot        string
+	forbiddenUsers map[string]bool
+	caPath         string
+	authPolicy     auth.Policy
 }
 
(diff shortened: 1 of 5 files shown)


pkg/preparer/orchestrate.go
```diff
@@ -174,21 +169,23 @@ func (p *Preparer) verifySignature(manifest pods.Manifest, logger logging.Logger
...
 	}
-	return true
-}
-
-func (p *Preparer) installAndLaunchPod(newManifest *pods.Manifest, pod Pod, logger logging.Logger) bool {
-	// do not remove the logger argument, it's not the same as p.Logger
 
-	if newManifest.ID() != POD_ID && p.podIdForbidden(newManifest.RunAsUser()) {
+	// Even if the authorization policy allows it, apply another
+	// filter to keep apps from being deployed as system users.
```
anthonybishopric: For future work - we should probably consolidate system users into the main authorization system since the question asked by the authz system specifically handles users and not apps. That has the nice side effect of making "systems" deployable with P2.

anthonybishopric:
Probably worth running the integration test to make sure that it still functions as expected. Otherwise, :+1:

bwester:
I haven't been able to reproduce the integration test failure on my end, and a re-run of the Travis CI build passed its test.  Considering that the integration test is non-deterministic because it relies on time, I'm going to call it a transient failure.

PR #188, opened 2015-06-16
move our tar extraction function to a reusable package

mpuncel:


pkg/gzip/gzip.go
```diff
@@ -0,0 +1,159 @@
+package gzip
```
mpuncel: i made this a new package because it creates an import cycle if it's in `util` (`util` imports `user` and vice-versa)

bwester:
Looks like you just moved the function over without many changes. Works for me!

:ship:

PR #189, opened 2015-06-19
show the name of the hook in the log message signifying a hook failure

mpuncel:


anthonybishopric:
:+1:

pkg/hooks/hooks.go
```diff
@@ -86,7 +86,7 @@ func runDirectory(dirpath string, environment []string, logger logging.Logger) e
 				"err":    err,
 				"path":   fullpath,
 				"output": hookOut.String(),
-			}).Warnln("Could not execute hook")
+			}).Warnf("Could not execute hook %s", path.Base(fullpath))
```
bwester: nit: I'd prefer `f.Name()` to `path.Base(path.Join(dirpath, f.Name()))`. They should be the same value anyway.

PR #190, opened 2015-06-23
preserve the original order of bytes read from a manifest file, so th…

mpuncel:
…at we can compare equality of a manifest as bytes to the original file

### DIFF
--- a/pkg/pods/manifest.go
+++ b/pkg/pods/manifest.go
@@ -74,6 +74,11 @@ func ManifestFromString(str string) (*Manifest, error) {
 func ManifestFromBytes(bytes []byte) (*Manifest, error) {
 	manifest := &Manifest{}
 
+	// Preserve the raw manifest so that manifest.Bytes() returns bytes in
+	// the same order that they were passed to this function
+	manifest.raw = make([]byte, len(bytes))
+	copy(manifest.raw, bytes)
+
 	signed, _ := clearsign.Decode(bytes)
 	if signed != nil {
 		signature, err := ioutil.ReadAll(signed.ArmoredSignature.Body)
@@ -81,9 +86,6 @@ func ManifestFromBytes(bytes []byte) (*Manifest, error) {
 			return nil, fmt.Errorf("Could not read signature from pod manifest: %s", err)
 		}
 		manifest.signature = signature
-
-		manifest.raw = make([]byte, len(bytes))
-		copy(manifest.raw, bytes)
 
 		// the original plaintext is in signed.Plaintext, but the signature
 		// corresponds to signed.Bytes, so that's what we need to save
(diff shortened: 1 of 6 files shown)


pkg/pods/manifest.go
```diff
@@ -93,6 +90,10 @@ func ManifestFromBytes(bytes []byte) (*Manifest, error) {
 		bytes = signed.Plaintext
 	}
 
+	// Preserve the raw manifest so that manifest.Bytes() returns bytes in
+	// the same order that they were passed to this function
+	manifest.raw = make([]byte, len(bytes))
+	copy(manifest.raw, bytes)
```
anthonybishopric: This means the bytes will always be stripped of the signature. (notice how the `manifest.raw` is assigned the signed bytes inside the if statement before the plaintext is extracted)

mpuncel: moved this to before the if statement

anthonybishopric:
:+1:  once tests are fixed

pkg/pods/manifest.go
```diff
@@ -123,6 +128,13 @@ func (manifest *Manifest) Bytes() ([]byte, error) {
 	return yaml.Marshal(manifest)
 }
 
+// Returns bytes correlating to the (potentially mutated) struct, unlike
+// Bytes() which guarantees that the bytes returned will be the same bytes the
+// manifest struct was built from
+func (manifest *Manifest) Marshal() ([]byte, error) {
```
mpuncel: I realize that callers can just call `yaml.Marshal()` themselves, but I think this will make the codebase clearer

pkg/pods/manifest.go
```diff
@@ -111,6 +113,9 @@ func (manifest *Manifest) Write(out io.Writer) error {
 	return nil
 }
 
+// Bytes() always returns the bytes from the original manifest file that was
```
bwester: This seems like a very significant API change!  Before this PR, `Bytes()` on an unsigned manifest returned the current state of the manifest. After this PR, `Bytes()` on an unsigned manifest _might_ return an older version of the struct.

To make it clear that it has such different behavior, can you rename this to `OriginalBytes()` and audit all the references to this method to clarify whether they should be updated to the new name or switched to `Marshal()`?

mpuncel: manifest.Bytes() is only used in tests, I modified the ones that depended on being able to programmatically change a manifest

mpuncel: made your suggested rename :+1:

pkg/pods/manifest.go
anthonybishopric: If it becomes anything more complex than a `yaml.Marshal`, which I think is possible, then this is a good method to have.

anthonybishopric:
:+1:

PR #193, opened 2015-06-25
Adds GetTransport method

robertabbott:
This method enables the creation of a p2 headerTransport
object. This lets you set the Transport field of an
existing http.Client object


anthonybishopric:
:+1:

PR #194, opened 2015-06-29
Creates inspect pkg

robertabbott:
Some of the structs in p2-inspect may be useful elsewhere (ie p2-show-health),
so I have moved them into their own package.

Also adds JSONManifest to the pods package. JSONManifest is a JSON encodable
version of the Manifest struct in the pods package.

### DIFF
--- a/bin/p2-inspect/main.go
+++ b/bin/p2-inspect/main.go
@@ -2,14 +2,12 @@ package main
 
 import (
 	"encoding/json"
-	"fmt"
 	"log"
 	"net/http"
 	"os"
-	"sort"
-	"strings"
 
 	"github.com/square/p2/pkg/health"
+	"github.com/square/p2/pkg/inspect"
 	"github.com/square/p2/pkg/kp"
 	"github.com/square/p2/pkg/util/net"
 	"github.com/square/p2/pkg/version"
@@ -24,26 +22,6 @@ var (
 	headers        = kingpin.Flag("header", "An HTTP header to add to requests, in KEY=VALUE form. Can be specified multiple times.").StringMap()
 	https          = kingpin.Flag("https", "Use HTTPS").Bool()
 )
-
-const (
-	INTENT_SOURCE = iota
-	REALITY_SOURCE
-)
-
-type NodePodStatus struct {
-	NodeName           string             `json:"node,omitempty"`
-	PodId              string             `json:"pod,omitempty"`
-	IntentManifestSHA  string             `json:"intent_manifest_sha"`
-	RealityManifestSHA string             `json:"reality_manifest_sha"`
-	IntentLocations    []string           `json:"intent_locations"`
-	RealityLocations   []string           `json:"reality_locations"`
-	Health             health.HealthState `json:"health,omitempty"`
-}
-
-type NodeHealth struct {
-	Status string `json:"status"`
-	Output string `json:"output"`
-}
 
 func main() {
 	kingpin.Version(version.VERSION)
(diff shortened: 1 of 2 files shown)


robertabbott:
I'm trying to get rid of the JSONManifest struct since it may not actually be necessary

pkg/pods/manifest.go
```diff
@@ -41,6 +41,14 @@ type Manifest struct {
 	signature []byte
 }
 
+type JSONManifest struct {
```
petertseng: Hmm. I was about to ask "Can we embed JSONManifest into Manifest so we don't forget and accidentally get them out of sync with each other?" but then I saw the json annotations and realized it may not be possible. Unless it's possible to have both a json annotation and a yaml annotation on a struct member...? I'd at least investigate the possibility. If not, at least to comment on both structs saying they need to be kept in sync.

bwester: You don't need to add a new type to specify the JSON encoding for these fields. Just reuse `Manifest`. Go struct tags can have multiple key-value pairs if you separate them with spaces [[ref]](http://golang.org/pkg/reflect/#StructTag).

```go
Id string `yaml:"id" json:"id"`
```

pkg/pods/manifest.go
```diff
@@ -41,6 +41,14 @@ type Manifest struct {
...
+	// JSON encodable version of Manifest
+	Id                string                      `json:"id"`
+	Path              string                      `json:"path"`
```
petertseng: oh, where did this one come from? sorry if I'm a bit behind the times

pkg/pods/manifest.go
robertabbott: The problem I'm having which is the reason I created the struct in the first place is that Encode(some pods.Manifest) fails. I thought it was because Encode can't handle the map[interface{}]interface{}, though that may be incorrect.

pkg/pods/manifest.go
robertabbott: I created this struct for p2-show-health because it was failing to encode the pods.Manifest struct

bin/p2-inspect/main.go
```diff
@@ -112,7 +98,7 @@ func main() {
 	enc.Encode(statusMap)
 }
 
-func addKVPToMap(result kp.ManifestResult, source int, filterNode, filterPod string, statuses map[string]map[string]NodePodStatus) error {
+func addKVPToMap(result kp.ManifestResult, source int, filterNode, filterPod string, statuses map[string]map[string]inspect.NodePodStatus) error {
```
bwester: Since you've moved the main data structure to its own module, would it make sense to move the functions that operate on the data structure to the module too?

robertabbott:
Got rid of JSONManifest struct and moved AddKVPToMap from p2-inspect to the inspect package.

bwester:
I think this PR is in pretty good shape. There is one more thing I'd like to see though. There are two logical, unrelated changes happening here: (1) Refactoring inspection and aggregation logic out of _p2-inspect_, and (2) changing the config interface.  Could you put these in separate commits?

The reason for having separate commits is that it helps manage the changes independently. Imagine that the config change introduced a problem and we needed to remove it.  If it is in a separate commit, it's easy to ask git to backout just those changes. If both changes are in the same commit, it would involve more work:  someone would have to re-learn and reason about the diff, separate the logical components, and manually create a patch.

robertabbott:
Split into two commits. The other one is here: https://github.com/square/p2/pull/195

robertabbott:
looks like test flakiness...

bwester:
Personally, as long as the changes are related or dependent, I think it's fine to include multiple commits in one pull request. You didn't necessarily need to create #195, but it isn't wrong that you did either. This keeps related discussion together in one place.

pkg/inspect/inspect.go
```diff
@@ -0,0 +1,77 @@
...
+	NodeName           string             `json:"node,omitempty"`
+	PodId              string             `json:"pod,omitempty"`
+	IntentManifestSHA  string             `json:"intent_manifest_sha"`
+	RealityManifestSHA string             `json:"reality_manifest_sha"`
+	IntentLocations    []string           `json:"intent_locations"`
+	RealityLocations   []string           `json:"reality_locations"`
+	Health             health.HealthState `json:"health,omitempty"`
+}
+
+type NodeHealth struct {
```
bwester: Sorry, I should have caught this earlier.  It looks like this structure is never used. Maybe now, during a refactor, is a good time to get rid of it. Can you amend the commit?

bwester: Sorry, I should have caught this earlier. It looks like this structure is never used. Maybe now, during a refactor, is a good time to get rid of it. Can you amend the commit?

anthonybishopric:
@robertabbott yep, pull requests are supposed to include multiple dependent commits. If this can be merged w/o the other one, then you're good to merge this IMO :+1:

PR #195, opened 2015-06-30
Changes pods.Manifest.Config to map[string]interface{}

robertabbott:
Before, Config was a map[interface{}]interface{}. This
was causing problems because the structure could not
be encoded as JSON.


anthonybishopric:
Is it worth putting explicit `json` tags so the values get mapped to the same field names as the YAML ones do?

bwester:
Ship it! :ship:

I am expecting the struct fields to get JSON tags in a separate diff.

robertabbott:
After working more with this problem this change is not the best solution, and is not likely to be helpful to others.

PR #196, opened 2015-07-02
Remove "forbidden_users" preparer option

bwester:
The "forbidden_users" option was an extra authorization check performed after
the normal authorization policy allowed an app to be installed. It allowed
certain "run_as" usernames to be blacklisted. Since the "user" policy is
already verifying the "run_as" user, the extra check is redundant.

This commit removes the "forbidden_users" option from the config struct and all
internal uses of it. It can continue to be specified in the preparer's config
without error; it will simply have no effect.

### DIFF
--- a/pkg/preparer/orchestrate.go
+++ b/pkg/preparer/orchestrate.go
@@ -35,15 +35,14 @@ type Store interface {
 }
 
 type Preparer struct {
-	node           string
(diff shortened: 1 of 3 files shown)


anthonybishopric:
:+1: glad we did this but also glad we don't need to support it anymore.

PR #197, opened 2015-07-09
Abbott/health checking

robertabbott:
I'm going to write unit tests in parallel to this PR being reviewed. I wanted to get it out so review could begin and if their are any structural changes I would like to make those before writing more unit tests.

### DIFF
--- a/bin/p2-preparer/main.go
+++ b/bin/p2-preparer/main.go
@@ -15,7 +15,6 @@ import (
 )
 
 func main() {
-	var token string
 	logger := logging.NewLogger(logrus.Fields{})
 	configPath := os.Getenv("CONFIG_PATH")
 	if configPath == "" {
@@ -47,11 +46,6 @@ func main() {
 	quitHookUpdate := make(chan struct{})
 	quitWatchHealth := make(chan struct{})
 
-	token, err = preparer.LoadConsulToken(preparerConfig.ConsulTokenPath)
-	if err != nil {
-		logger.WithField("inner_err", err).Fatalln("Could not load consul token")
-	}
-
 	go prep.WatchForPodManifestsForNode(quitMainUpdate)
 	go prep.WatchForHooks(quitHookUpdate)
 
@@ -65,11 +59,7 @@ func main() {
 
 	// Launch health checking watch. This watch tracks health of
 	// all pods on this host and writes the information to consul
-	go watch.WatchHealth(preparerConfig.NodeName,
-		preparerConfig.ConsulAddress,
-		token,
-		&logger,
-		quitWatchHealth)
+	go watch.WatchHealth(preparerConfig, &logger, quitWatchHealth)
 
 	waitForTermination(logger, quitMainUpdate, quitHookUpdate, quitWatchHealth)
 
--- a/bin/p2-preparer/main.go
+++ b/bin/p2-preparer/main.go
@@ -58,6 +58,7 @@ func main() {
 	go watch.WatchHealth(preparerConfig.NodeName,
 		preparerConfig.ConsulAddress,
 		token,
+		&logger,
 		quitWatchHealth)
 
 	if preparerConfig.StatusPort != 0 {
--- a/bin/p2-preparer/main.go
+++ b/bin/p2-preparer/main.go
@@ -15,6 +15,7 @@ import (
 )
 
 func main() {
+	var token string
 	logger := logging.NewLogger(logrus.Fields{})
 	configPath := os.Getenv("CONFIG_PATH")
 	if configPath == "" {
@@ -45,9 +46,13 @@ func main() {
 	quitMainUpdate := make(chan struct{})
 	quitHookUpdate := make(chan struct{})
 	quitWatchHealth := make(chan struct{})
-	token, err := preparer.LoadConsulToken(preparerConfig.ConsulTokenPath)
-	if err != nil {
-		logger.WithField("inner_err", err).Fatalln("Could not load consul token")
+	if preparerConfig.ConsulTokenPath != "" {
+		token, err = preparer.LoadConsulToken(preparerConfig.ConsulTokenPath)
+		if err != nil {
+			logger.WithField("inner_err", err).Fatalln("Could not load consul token")
+		}
+	} else {
+		token = ""
 	}
 
 	go prep.WatchForPodManifestsForNode(quitMainUpdate)
@@ -55,7 +60,7 @@ func main() {
 
 	// Launch health checking watch. This watch tracks health of
 	// all pods on this host and writes the information to consul
-	go watch.WatchHealth(preparerConfig.NodeName,
+	go watch.WatchHealth("localhost:8500",
 		preparerConfig.ConsulAddress,
 		token,
 		&logger,
--- a/bin/p2-preparer/main.go
+++ b/bin/p2-preparer/main.go
@@ -44,7 +44,7 @@ func main() {
 
 	quitMainUpdate := make(chan struct{})
 	quitHookUpdate := make(chan struct{})
-	quitWatchHealth := make(chan struct{})
+	quitMonitorPodHealth := make(chan struct{})
 
 	go prep.WatchForPodManifestsForNode(quitMainUpdate)
 	go prep.WatchForHooks(quitHookUpdate)
@@ -59,20 +59,20 @@ func main() {
 
 	// Launch health checking watch. This watch tracks health of
 	// all pods on this host and writes the information to consul
-	go watch.WatchHealth(preparerConfig, &logger, quitWatchHealth)
+	go watch.MonitorPodHealth(preparerConfig, &logger, quitMonitorPodHealth)
 
-	waitForTermination(logger, quitMainUpdate, quitHookUpdate, quitWatchHealth)
+	waitForTermination(logger, quitMainUpdate, quitHookUpdate, quitMonitorPodHealth)
 
 	logger.NoFields().Infoln("Terminating")
 }
 
-func waitForTermination(logger logging.Logger, quitMainUpdate, quitHookUpdate chan struct{}, quitWatchHealth chan struct{}) {
+func waitForTermination(logger logging.Logger, quitMainUpdate, quitHookUpdate chan struct{}, quitMonitorPodHealth chan struct{}) {
 	signalCh := make(chan os.Signal, 2)
 	signal.Notify(signalCh, syscall.SIGTERM, os.Interrupt)
 	received := <-signalCh
 	logger.WithField("signal", received.String()).Infoln("Stopping work")
 	quitHookUpdate <- struct{}{}
 	quitMainUpdate <- struct{}{}
-	quitWatchHealth <- struct{}{}
+	quitMonitorPodHealth <- struct{}{}
 	<-quitMainUpdate // acknowledgement
 }
--- a/bin/p2-preparer/main.go
+++ b/bin/p2-preparer/main.go
@@ -11,6 +11,7 @@ import (
 	"github.com/square/p2/pkg/logging"
 	"github.com/square/p2/pkg/preparer"
 	"github.com/square/p2/pkg/version"
+	"github.com/square/p2/pkg/watch"
 )
 
 func main() {
@@ -43,8 +44,13 @@ func main() {
 
 	quitMainUpdate := make(chan struct{})
 	quitHookUpdate := make(chan struct{})
+	quitWatch := make(chan struct{})
 	go prep.WatchForPodManifestsForNode(quitMainUpdate)
 	go prep.WatchForHooks(quitHookUpdate)
+	go watch.WatchHealth(preparerConfig.NodeName,
+		preparerConfig.ConsulAddress,
+		preparerConfig.Auth["keyring"],
+		quitWatch)
 
 	if preparerConfig.StatusPort != 0 {
 		http.HandleFunc("/_status",
@@ -66,5 +72,6 @@ func waitForTermination(logger logging.Logger, quitMainUpdate, quitHookUpdate ch
 	logger.WithField("signal", received.String()).Infoln("Stopping work")
 	quitHookUpdate <- struct{}{}
 	quitMainUpdate <- struct{}{}
+	quitWatch <- struct{}{}
 	<-quitMainUpdate // acknowledgement
 }
--- a/bin/p2-preparer/main.go
+++ b/bin/p2-preparer/main.go
@@ -46,25 +46,14 @@ func main() {
 	quitMainUpdate := make(chan struct{})
 	quitHookUpdate := make(chan struct{})
 	quitWatchHealth := make(chan struct{})
-	if preparerConfig.ConsulTokenPath != "" {
-		token, err = preparer.LoadConsulToken(preparerConfig.ConsulTokenPath)
-		if err != nil {
-			logger.WithField("inner_err", err).Fatalln("Could not load consul token")
-		}
-	} else {
-		token = ""
+
+	token, err = preparer.LoadConsulToken(preparerConfig.ConsulTokenPath)
+	if err != nil {
+		logger.WithField("inner_err", err).Fatalln("Could not load consul token")
 	}
 
 	go prep.WatchForPodManifestsForNode(quitMainUpdate)
 	go prep.WatchForHooks(quitHookUpdate)
-
-	// Launch health checking watch. This watch tracks health of
-	// all pods on this host and writes the information to consul
-	go watch.WatchHealth("localhost:8500",
-		preparerConfig.ConsulAddress,
-		token,
-		&logger,
-		quitWatchHealth)
 
 	if preparerConfig.StatusPort != 0 {
 		http.HandleFunc("/_status",
@@ -73,6 +62,14 @@ func main() {
 			})
 		go http.ListenAndServe(fmt.Sprintf(":%d", preparerConfig.StatusPort), nil)
 	}
+
+	// Launch health checking watch. This watch tracks health of
+	// all pods on this host and writes the information to consul
+	go watch.WatchHealth(preparerConfig.NodeName,
+		preparerConfig.ConsulAddress,
+		token,
+		&logger,
+		quitWatchHealth)
 
 	waitForTermination(logger, quitMainUpdate, quitHookUpdate, quitWatchHealth)
 
--- a/bin/p2-preparer/main.go
+++ b/bin/p2-preparer/main.go
@@ -44,13 +44,21 @@ func main() {
 
 	quitMainUpdate := make(chan struct{})
 	quitHookUpdate := make(chan struct{})
-	quitWatch := make(chan struct{})
+	quitWatchHealth := make(chan struct{})
+	token, err := preparer.LoadConsulToken(preparerConfig.ConsulTokenPath)
+	if err != nil {
+		logger.WithField("inner_err", err).Fatalln("Could not load consul token")
+	}
+
 	go prep.WatchForPodManifestsForNode(quitMainUpdate)
 	go prep.WatchForHooks(quitHookUpdate)
+
+	// Launch health checking watch. This watch tracks health of
+	// all pods on this host and writes the information to consul
 	go watch.WatchHealth(preparerConfig.NodeName,
 		preparerConfig.ConsulAddress,
-		preparerConfig.Auth["keyring"],
-		quitWatch)
+		token,
+		quitWatchHealth)
 
 	if preparerConfig.StatusPort != 0 {
 		http.HandleFunc("/_status",
@@ -60,18 +68,18 @@ func main() {
 		go http.ListenAndServe(fmt.Sprintf(":%d", preparerConfig.StatusPort), nil)
 	}
 
-	waitForTermination(logger, quitMainUpdate, quitHookUpdate)
+	waitForTermination(logger, quitMainUpdate, quitHookUpdate, quitWatchHealth)
 
 	logger.NoFields().Infoln("Terminating")
 }
 
-func waitForTermination(logger logging.Logger, quitMainUpdate, quitHookUpdate chan struct{}) {
+func waitForTermination(logger logging.Logger, quitMainUpdate, quitHookUpdate chan struct{}, quitWatchHealth chan struct{}) {
 	signalCh := make(chan os.Signal, 2)
 	signal.Notify(signalCh, syscall.SIGTERM, os.Interrupt)
 	received := <-signalCh
 	logger.WithField("signal", received.String()).Infoln("Stopping work")
 	quitHookUpdate <- struct{}{}
 	quitMainUpdate <- struct{}{}
-	quitWatch <- struct{}{}
+	quitWatchHealth <- struct{}{}
 	<-quitMainUpdate // acknowledgement
 }
--- a/integration/single-node-slug-deploy/check.go
+++ b/integration/single-node-slug-deploy/check.go
@@ -374,17 +374,6 @@ func verifyHealthChecks(config *preparer.PreparerConfig, services []string) erro
 	}
 	store := kp.NewConsulStore(opts)
 
-	// consul kvs sanity check (probably not necessary...)
-	_, _, err := store.PutHealth("testsv", "testNode", "seamus")
-	if err != nil {
-		return err
-	}
-	t, err := store.GetHealth("testsv", "testNode")
-	if err != nil {
-		fmt.Println(t)
-		return err
-	}
-
 	time.Sleep(2)
 	// check consul for health information for each app
 	for _, sv := range services {
--- a/integration/single-node-slug-deploy/check.go
+++ b/integration/single-node-slug-deploy/check.go
@@ -3,7 +3,6 @@ package main
 import (
 	"bytes"
 	"encoding/json"
-	"errors"
 	"fmt"
 	"io/ioutil"
 	"log"
@@ -30,6 +29,9 @@ func main() {
 	// 3. Execute bootstrap with premade consul pod and preparer pod
 	// 4. Deploy hello pod manifest by pushing to intent store
 	// 5. Verify that hello is running (listen to syslog? verify Runit PIDs? Both?)
+
+	// list of services running on integration test host
+	services := []string{"consul", "p2-preparer", "hello"}
 	tempdir, err := ioutil.TempDir("", "single-node-check")
 	log.Printf("Putting test manifests in %s\n", tempdir)
 	if err != nil {
@@ -85,8 +87,7 @@ func main() {
 	if err != nil {
 		log.Fatalf("Couldn't get hello running: %s", err)
 	}
-
-	err = verifyHealthChecks(config)
+	err = verifyHealthChecks(config, services)
 	if err != nil {
 		log.Fatalf("Could not get health check info from consul: %s", err)
 	}
@@ -365,7 +366,7 @@ func verifyHelloRunning() error {
 	}
 }
 
-func verifyHealthChecks(config *preparer.PreparerConfig) error {
+func verifyHealthChecks(config *preparer.PreparerConfig, services []string) error {
 	// do consulGet: key = hello
 	opts := kp.Options{
 		Address: config.ConsulAddress,
@@ -373,33 +374,28 @@ func verifyHealthChecks(config *preparer.PreparerConfig) error {
 	}
 	store := kp.NewConsulStore(opts)
 
-	// consul kvs sanity check
-	_, err := store.Put("testkey", "seamus")
+	// consul kvs sanity check (probably not necessary...)
+	_, err := store.PutHealth("testsv", "testNode", "seamus")
 	if err != nil {
 		return err
 	}
-	time.Sleep(1 * time.Second)
-	t, err := store.Get("testkey")
+	t, err := store.GetHealth("testsv", "testNode")
 	if err != nil {
 		fmt.Println(t)
 		return err
 	}
 
-	man, _, err := store.ListPods("reality/localhost.localdomain")
-	if err != nil {
-		return err
-	}
-	fmt.Println(man)
-	//////////////////////////
-
-	res, err := store.Get("consul")
-	if err != nil {
-		return err
-	} else if res == nil {
-		err = errors.New("no results for hello")
-	} else {
-		fmt.Println("Got value!!!")
-		fmt.Println(string(res.Value))
+	time.Sleep(2)
+	// check consul for health information for each app
+	for _, sv := range services {
+		res, err := store.GetHealth(sv, "localhost.localdomain")
+		if err != nil {
+			return err
+		} else if res == "" {
+			fmt.Errorf("No results for %s", sv)
+		} else {
+			fmt.Println(res)
+		}
 	}
 
 	// if it reaches here it means health checks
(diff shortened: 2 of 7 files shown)


bin/p2-preparer/main.go
```diff
@@ -43,6 +45,16 @@ func main() {
 
 	quitMainUpdate := make(chan struct{})
 	quitHookUpdate := make(chan struct{})
+	quitWatchHealth := make(chan struct{})
+	if preparerConfig.ConsulTokenPath != "" {
```
anthonybishopric: The logic of loading the `ConsulTokenPath` should really happen in `setup.go` - this file is purposely quite lean to encourage modularity.

integration/single-node-slug-deploy/check.go
```diff
@@ -348,3 +366,42 @@ func verifyHelloRunning() error {
...
+
+	time.Sleep(2)
+	// check consul for health information for each app
+	for _, sv := range services {
+		path := fmt.Sprintf("%s/%s", sv, "localhost.localdomain")
+		res, err := store.Get(path)
+		if err != nil {
+			return err
+		} else if res == nil {
+			err = errors.New("no results for hello")
```
anthonybishopric: I'm guessing this should be `fmt.Errorf("No results for %s", sv)`

pkg/kp/kv.go
```diff
@@ -50,6 +52,32 @@ func (err KVError) Error() string {
...
+	if writeMeta != nil {
+		retDur = writeMeta.RequestTime
+	}
+	if err != nil {
+		return retDur, err
+	}
+	return retDur, nil
+}
+
+func (c consulStore) Get(key string) (*api.KVPair, error) {
```
anthonybishopric: We shouldn't expose primitives of the Consul API SDK in the `kp` package. This is intended to be a P2-specific key value wrapper that is abstracted from the underlying storage mechanism.

anthonybishopric: Specifically, we shouldn't return an `*api.KVPair`

pkg/watch/health.go
```diff
@@ -0,0 +1,240 @@
...
+	// key =  service/node
+	// if status == passing: value = status
+	// else: values = health.Status/health.Output
+	key := fmt.Sprintf("%s/%s", res.Service, res.Node)
+	if res.Status == health.Passing {
+		value = "passing"
+	} else {
+		value = fmt.Sprintf("%s/%s", res.Status, res.Output)
+	}
+	_, err := store.Put(key, value)
```
anthonybishopric: Maybe an alternative would be to move the health structs in `health` to `kp` and make them part of the interface to `Store`. Specifically something like:

```
func writeToConsul(res kp.healthResult, store kp.Store) error {
   store.SetHealth(res)
}
```
This lets the underlying schema in Consul be managed by the `kp` package directly

pkg/watch/health.go
```diff
@@ -0,0 +1,240 @@
...
+	if err != nil {
+		return err
+	}
+	return nil
```
anthonybishopric: The above three lines are equivalent to `return err`

pkg/watch/health.go
```diff
@@ -0,0 +1,240 @@
...
+}
+
+//
+// Methods for tracking pods that should be monitored
+//
+
+func updateHealthMonitors(store kp.Store, watchedPods []*PodWatch, node string, logger *logging.Logger) []*PodWatch {
+	path := fmt.Sprintf("reality/%s", node)
```
anthonybishopric: `path := kp.RealityPath(node)`

pkg/watch/health.go
```diff
@@ -0,0 +1,240 @@
...
+		})
+	} else {
+		store = kp.NewConsulStore(kp.Options{
+			Address: consul,
+			HTTPS:   false,
+			Client:  net.NewHeaderClient(nil, http.DefaultTransport),
+		})
+	}
+
+	go startTimer(tochan, POLL_KV_FOR_PODS)
```
anthonybishopric: Have you seen `time.After`?

pkg/watch/health.go
```diff
@@ -0,0 +1,240 @@
...
+	for {
+		select {
+		case _ = <-tochan:
```
anthonybishopric: `case <- time.After(POLL_KV_FOR_PODS):`

pkg/watch/health.go
```diff
@@ -0,0 +1,240 @@
...
+	"github.com/square/p2/pkg/kp"
+	"github.com/square/p2/pkg/logging"
+	"github.com/square/p2/pkg/util/net"
+)
+
+// number of milliseconds between reality store checks
+const POLL_KV_FOR_PODS = 4000
+
+// number of milliseconds between health checks
+const HEALTHCHECK_INTERVAL = 200
```
anthonybishopric: We will probably want to both increase this interval (a healthcheck every 200 millis is pretty aggressive) and enable some form of backoff.

pkg/watch/health.go
```diff
@@ -0,0 +1,240 @@
...
+		ID:      p.manifest.Manifest.Id,
+		Node:    node,
+		Service: p.manifest.Manifest.Id,
+		Status:  healthstate,
+		Output:  res,
+	}
+
+	err = writeToConsul(health, store)
+	if err != nil {
+		p.logger.WithField("CONSUL WRITE ERROR", err).Fatalln("failed to write health data to consul")
```
anthonybishopric: The "WithField" is for a key in a JSON document most likely. You probably just want to use the string "err" instead of "CONSUL WRITE ERROR" here.

pkg/watch/health.go
```diff
@@ -0,0 +1,240 @@
...
+	healthstate, res, err := check(healthCheck)
+	health := health.Result{
+		ID:      p.manifest.Manifest.Id,
+		Node:    node,
+		Service: p.manifest.Manifest.Id,
+		Status:  healthstate,
+		Output:  res,
+	}
+
+	err = writeToConsul(health, store)
```
anthonybishopric: Two things here:

First, I don't see any notion of time / TTL on the health information we are sending to Consul and there should be some indication.

Second, and this may indicate we want an explicit expiry on the health state (not just a timestamp indicating when it was generated), but we may want the ability to abstain from sending a redundant health update to Consul if we can avoid the roundtrip. Lets say we are doing health checks every second, and we explicitly set a TTL of 10 seconds on all health reports we make. If on the 2nd check the health has not changed we can avoid rewriting the value to Consul and have a pretty good guarantee that clients won't find the value to be stale.

pkg/watch/health.go
```diff
@@ -0,0 +1,240 @@
...
+			p.checkHealth(statusCheck, node, store)
+			go startTimer(tochan, HEALTHCHECK_INTERVAL)
+		case _ = <-shutdownCh:
+			return
+		}
+	}
+}
+
+func (p *PodWatch) checkHealth(healthCheck, node string, store kp.Store) {
+	healthstate, res, err := check(healthCheck)
```
anthonybishopric: We should reimplement `check(healthcheck)` in pure Go so we don't need to worry about shelling out. That is important since we need to be able to do richer things with the check in the future, including verifying that the `/_status` endpoint does not return a header indicating degradation.

pkg/kp/kv.go
```diff
@@ -50,6 +52,32 @@ func (err KVError) Error() string {
 	return fmt.Sprintf("%s failed for path %s", err.Op, err.Key)
 }
 
+func (c consulStore) Put(key, value string) (time.Duration, error) {
+	value = appendTimeStamp(value)
```
anthonybishopric: Whatever value we write into the KV store it should be in a standard data representation that we have libraries for marshaling and unmarshaling. My strong suggestion is to use JSON for this purpose.

anthonybishopric:
Good start! Lets work through some of the design issues and figure out the unit testing situation.

robertabbott:
I think all the constants should at some point be part of a config file and the hardcoded values can be defaults.

integration/single-node-slug-deploy/check.go
```diff
@@ -348,3 +365,29 @@ func verifyHelloRunning() error {
...
+
+func verifyHealthChecks(config *preparer.PreparerConfig, services []string) error {
+	// do consulGet: key = hello
+	opts := kp.Options{
+		Address: config.ConsulAddress,
+		HTTPS:   false,
+	}
+	store := kp.NewConsulStore(opts)
+
+	time.Sleep(5)
```
anthonybishopric: This reads as `sleep for 5 nanoseconds`. Is that what you wanted?

integration/single-node-slug-deploy/check.go
```diff
@@ -348,3 +365,29 @@ func verifyHelloRunning() error {
...
+	// check consul for health information for each app
+	for _, sv := range services {
+		res, err := store.GetHealth(sv, "localhost.localdomain")
```
anthonybishopric: Maybe use `os.Hostname()` instead of hardcoding this?

integration/single-node-slug-deploy/check.go
```diff
@@ -348,3 +365,29 @@ func verifyHelloRunning() error {
...
+		if err != nil {
+			return err
+		} else if res == "" {
+			fmt.Errorf("No results for %s", sv)
```
anthonybishopric: This doesn't print anything - it just returns an `error` for use. You need to actually `Print` it.

pkg/kp/kv.go
```diff
@@ -50,6 +53,42 @@ func (err KVError) Error() string {
 	return fmt.Sprintf("%s failed for path %s", err.Op, err.Key)
 }
 
+func (c consulStore) PutHealth(res map[string]string) (time.Time, time.Duration, error) {
+	key := fmt.Sprintf("%s%s/%s", res["service"], "Health", res["node"])
```
anthonybishopric: Lets consolidate the path making function since this is repeated below and will be useful for others.

`func HealthPath(service, node string) string`

anthonybishopric: make the string `"Health"` lowercase to be consistent with other subtrees.

pkg/kp/kv.go
```diff
@@ -50,6 +53,42 @@ func (err KVError) Error() string {
...
+		Key:   key,
+		Value: data,
+	}
+
+	writeMeta, err := c.client.KV().Put(keyPair, nil)
+	var retDur time.Duration
+	if writeMeta != nil {
+		retDur = writeMeta.RequestTime
+	}
+	if err != nil {
```
anthonybishopric: This needs to be wrapped with `KVError`

pkg/kp/service.go
```diff
@@ -38,3 +40,16 @@ func (c consulStore) RegisterService(manifest pods.Manifest, caPath string) erro
 
 	return c.client.Agent().ServiceRegister(podService)
 }
+
+// Go version of http status check
+func HttpStatusCheck(node string, port int) *http.Response {
+	url := fmt.Sprintf("http://%s:%s/_status", node, strconv.Itoa(port))
```
anthonybishopric: You don't need the `strconv` if you use `%d` in the format string

pkg/watch/health.go
```diff
@@ -0,0 +1,257 @@
...
+
+import (
+	"io/ioutil"
+	"net/http"
+	"time"
+
+	"github.com/square/p2/pkg/health"
+	"github.com/square/p2/pkg/kp"
+	"github.com/square/p2/pkg/logging"
+	"github.com/square/p2/pkg/preparer"
```
anthonybishopric: it's a little strange to import the preparer here, but I realize that's an effect of moving the config stuff out of `bin/preparer`. Fine for now but I think we can do this better with a little reordering.

pkg/watch/health.go
```diff
@@ -0,0 +1,257 @@
...
+	} else {
+		store = kp.NewConsulStore(kp.Options{
+			Address: consul,
+			HTTPS:   false,
+			Client:  net.NewHeaderClient(nil, http.DefaultTransport),
+		})
+	}
+	for {
+		select {
+		case _ = <-time.After(POLL_KV_FOR_PODS):
```
anthonybishopric: Nit: don't need the `_ =`

pkg/watch/health.go
```diff
@@ -0,0 +1,257 @@
...
+
+	consul := config.ConsulAddress
+	node := config.NodeName
+	pods := []*PodWatch{}
+	authtoken, err := preparer.LoadConsulToken(config.ConsulTokenPath)
+	if err != nil {
+		logger.WithField("inner_err", err).Warningln("Could not load consul token")
+	}
+
+	if authtoken != "" {
```
anthonybishopric: There is no difference between the if/else stanzas - in Go, the zero value for a string is the same as a blank string.

pkg/watch/health.go
```diff
@@ -0,0 +1,257 @@
...
+// in their preparer config...
+
+// number of milliseconds between reality store checks
+const POLL_KV_FOR_PODS = 2000
+
+// number of milliseconds between health checks
+const HEALTHCHECK_INTERVAL = 1000
+
+// TTL on healthchecks in seconds
+const TTL = 60
```
anthonybishopric: I'd feel more comfortable if we multiplied these by their units during initialization instead of trusting all the clients to multiply correctly:

```
const POLL_KV_FOR_PODS = 2 * time.Second
const HEALTHCHECK_INTERVAL = 1 * time.Second
const TTL = 60 * time.Second
```

pkg/watch/health.go
```diff
@@ -0,0 +1,257 @@
...
+}
+
+// Monitor Health is a go routine that runs as long as the
+// service it is monitoring. Every HEALTHCHECK_INTERVAL it
+// performs a health check and writes that information to
+// consul
+func (p *PodWatch) MonitorHealth(node string, store kp.Store, shutdownCh chan bool) {
+	for {
+		select {
+		case _ = <-time.After(HEALTHCHECK_INTERVAL):
```
anthonybishopric: ... case in point, this currently is polling every 2000 nanoseconds.

pkg/watch/health.go
```diff
@@ -0,0 +1,257 @@
...
+	resp := kp.HttpStatusCheck(node, port)
+	health, err := resultFromCheck(resp)
+	if err != nil {
+		p.logger.WithField("err", err).Warningln("Failed to read status check response")
+	}
+	health.ID = p.manifest.Manifest.Id
+	health.Node = node
+	health.Service = p.manifest.Manifest.Id
+
+	if p.updateNeeded(health, TTL) == true {
```
anthonybishopric: Nit: unnecessary `== true`

pkg/watch/health.go
```diff
@@ -0,0 +1,257 @@
...
+			return
+		}
+	}
+}
+
+func (p *PodWatch) checkHealth(node string, port int, store kp.Store) {
+	resp := kp.HttpStatusCheck(node, port)
+	health, err := resultFromCheck(resp)
+	if err != nil {
+		p.logger.WithField("err", err).Warningln("Failed to read status check response")
```
anthonybishopric: Need to return here?

pkg/watch/health.go
```diff
@@ -0,0 +1,257 @@
...
+// pod's goroutine if the pod is removed from the reality
+// tree, and a bool that indicates whether or not the pod
+// has a running MonitorHealth go routine
+type PodWatch struct {
+	manifest kp.ManifestResult
+
+	// For tracking/controlling the go routine that performs health checks
+	// on the pod associated with this PodWatch
+	shutdownCh chan bool
+	hasMonitor bool // indicates whether this pod is being monitored
```
anthonybishopric: I don't fully understand why this boolean is necessary. Doesn't the existence of this `PodWatch` constitute being monitored? If it's just a function of figuring out what new pods need watches, it seems like it's an artifact of the implementation of `updatePods`, not an inherent property of this struct. What do you think?

pkg/watch/health.go
```diff
@@ -0,0 +1,257 @@
...
+
+func resultFromCheck(resp *http.Response) (health.Result, error) {
+	var err error
+	res := health.Result{}
+	if resp == nil {
+		return res, nil
+	} else if resp.StatusCode == 200 {
+		res.Status = health.Passing
+	} else if resp.StatusCode > 200 && resp.StatusCode < 300 {
+		res.Status = health.Warning
```
anthonybishopric: Any 2xx should be Passing probably.

pkg/kp/kv.go
```diff
@@ -50,6 +53,42 @@ func (err KVError) Error() string {
 	return fmt.Sprintf("%s failed for path %s", err.Op, err.Key)
 }
 
+func (c consulStore) PutHealth(res map[string]string) (time.Time, time.Duration, error) {
+	key := healthPath(res["service"], res["node"])
```
anthonybishopric: Magic keys are a really bad idea - it makes it impossible to statically check callers correctness of this program. This should at the very least be a struct with named fields that are marshaled in a specific way.

pkg/watch/health.go
```diff
@@ -0,0 +1,257 @@
...
+}
+
+func (p *PodWatch) updateNeeded(res health.Result, ttl int) bool {
+	// if status has changed indicate that consul needs to be updated
+	if p.lastStatus != res.Status {
+		return true
+	}
+	// if more than TTL / 4 seconds have elapsed since previous check
+	// indicate that consul needs to be updated
+	if time.Since(p.lastCheck) > time.Duration(ttl/4)*time.Second {
```
anthonybishopric: Nit: `return time.Since(p.lastCheck) > time.Duration(ttl/4)*time.Second`

pkg/watch/health.go
```diff
@@ -0,0 +1,257 @@
...
+// WatchHealth tracks the reality store and manages
+// a health checking go routine for each service in the
+// reality store
+
+// PodWatch houses a pod's manifest, a channel to kill the
+// pod's goroutine if the pod is removed from the reality
+// tree, and a bool that indicates whether or not the pod
+// has a running MonitorHealth go routine
+type PodWatch struct {
+	manifest kp.ManifestResult
```
anthonybishopric: I noticed you use `manifest.Manifest` a lot of places in this code. Probably a sign that this should just be a `*pods.Manifest` instead of a `kp.ManifestResult`

pkg/watch/health.go
```diff
@@ -0,0 +1,257 @@
...
+		for _, man := range reality {
+			if man.Manifest.Id == pod.manifest.Manifest.Id {
+				inReality = true
+				break
+			}
+		}
+
+		// if this podwatch is not in the reality store kill its go routine
+		// else add this podwatch to newCurrent
+		if inReality == false {
```
anthonybishopric: Nit: Awkward test `if !thing {a} else {b}`. Invert so that it's `if thing {b} else {a}`

pkg/watch/health.go
```diff
@@ -0,0 +1,257 @@
...
+				break
+			}
+		}
+
+		// if a manifest is in reality but not current a podwatch is created
+		// with that manifest and added to newCurrent
+		if missing == true {
+			newCurrent = append(newCurrent, &PodWatch{
+				manifest:   man,
+				shutdownCh: make(chan bool, 1),
```
anthonybishopric: Why use a buffered channel?

pkg/watch/health.go
```diff
@@ -0,0 +1,257 @@
...
+	logger *logging.Logger
+}
+
+// WatchHealth is meant to be a long running go routine.
+// WatchHealth reads from a consul store to determine which
+// services should be running on the host. WatchHealth
+// runs a CheckHealth routine to monitor the health of each
+// service and kills routines for services that should no
+// longer be running.
+func WatchHealth(config *preparer.PreparerConfig, logger *logging.Logger, shutdownCh chan struct{}) {
```
mpuncel: I think this name is a little confusing since you're not using a watch operation, and you're not watching consul for data like the other Watch*. All that said, I don't have a better name :). Maybe SyncAndMonitorPods?

pkg/watch/health.go
anthonybishopric: Still redundant

pkg/kp/kv.go
```diff
@@ -50,6 +62,42 @@ func (err KVError) Error() string {
 	return fmt.Sprintf("%s failed for path %s", err.Op, err.Key)
 }
 
+func (c consulStore) PutHealth(res WatchResult) (time.Time, time.Duration, error) {
+	key := healthPath(res.Service, res.Node)
+
+	t, value := addTimeStamp(res)
```
anthonybishopric: Is it okay for clients to pass a `Time` on the watch result? It looks like this will always override it. Probably worth saying that somewhere or adding a guard clause.

integration/single-node-slug-deploy/check.go
```diff
@@ -348,3 +365,34 @@ func verifyHelloRunning() error {
...
+			return err
+		}
+		res, err := store.GetHealth(sv, name)
+		if err != nil {
+			return err
+		} else if res == "" {
+			err = fmt.Errorf("No results for %s", sv)
+			return err
+		} else {
+			fmt.Println(res)
```
anthonybishopric: Next PR: change this to actually verify that services are _healthy_, not just _present_. Needs client integration.

pkg/kp/kv.go
```diff
@@ -50,6 +62,42 @@ func (err KVError) Error() string {
...
+	if writeMeta != nil {
+		retDur = writeMeta.RequestTime
+	}
+	if err != nil {
+		return t, retDur, KVError{Op: "get", Key: key, UnsafeError: err}
+	}
+	return t, retDur, nil
+}
+
+func (c consulStore) GetHealth(service, node string) (string, error) {
```
anthonybishopric: Shouldn't this return a `WatchResult`?

integration/single-node-slug-deploy/check.go
```diff
@@ -348,3 +365,35 @@ func verifyHelloRunning() error {
...
+	time.Sleep(5 * time.Second)
+	// check consul for health information for each app
+	for _, sv := range services {
+		name, err := os.Hostname()
+		if err != nil {
+			return err
+		}
+		res, err := store.GetHealth(sv, name)
+		if err != nil {
+			fmt.Errorf("sldkajflkfjasldkfjsalkfjd")
```
anthonybishopric: ?

robertabbott: sorry

pkg/watch/health.go
```diff
@@ -0,0 +1,239 @@
...
+				break
+			}
+		}
+
+		// if a manifest is in reality but not current a podwatch is created
+		// with that manifest and added to newCurrent
+		if missing == true {
+			newPod := PodWatch{
+				manifest:   man.Manifest,
+				shutdownCh: make(chan bool, 1),
```
anthonybishopric: I think my previous incarnation of this question got erased: why use a buffered channel here?

anthonybishopric: You can achieve non-blocking quits by executing `close(shutdownCh)` instead of using a buffer, fwiw

anthonybishopric:
I'm okay to roll with this as a first edition. :+1:

robertabbott:
:smiley:

integration/single-node-slug-deploy/check.go
```diff
@@ -348,3 +365,34 @@ func verifyHelloRunning() error {
 		return nil
 	}
 }
+
+func verifyHealthChecks(config *preparer.PreparerConfig, services []string) error {
+	// do consulGet: key = hello
```
petertseng: I'm not sure key stated in this comment is accurate

integration/single-node-slug-deploy/check.go
```diff
@@ -348,3 +365,34 @@ func verifyHelloRunning() error {
...
+	opts := kp.Options{
+		Address: config.ConsulAddress,
+		HTTPS:   false,
+	}
+	store := kp.NewConsulStore(opts)
+
+	time.Sleep(5 * time.Second)
+	// check consul for health information for each app
+	for _, sv := range services {
+		name, err := os.Hostname()
```
petertseng: I would say this can be moved out of the loop (hostname won't change will it?)

integration/single-node-slug-deploy/check.go
```diff
@@ -348,3 +365,34 @@ func verifyHelloRunning() error {
...
+		if err != nil {
+			return err
+		}
+		res, err := store.GetHealth(sv, name)
+		if err != nil {
+			return err
+		} else if (res == kp.WatchResult{}) {
```
petertseng: if I could style nit I'd say it seems the parentheses are unnecessary

pkg/kp/kv.go
```diff
@@ -171,3 +222,16 @@ func (c consulStore) Ping() error {
...
+	currentTime := time.Now()
+	if value.Time == "" {
+		value.Time = currentTime.String()
+	}
+
+	return currentTime, value
+}
+
+func healthPath(service, node string) string {
+	return fmt.Sprintf("%s%s/%s", service, "health", node)
```
petertseng: should there be a slash between the first two `%s`? If not, then this implies service should end with a slash?

robertabbott: its servicehealth/nodename. It could just be service/node or service/health/node

pkg/kp/service.go
```diff
@@ -38,3 +39,16 @@ func (c consulStore) RegisterService(manifest pods.Manifest, caPath string) erro
 
 	return c.client.Agent().ServiceRegister(podService)
 }
+
+// Go version of http status check
+func HttpStatusCheck(node string, port int) *http.Response {
+	url := fmt.Sprintf("http://%s:%d/_status", node, port)
+	resp, _ := http.Get(url)
```
petertseng: do you think it might be prudent to check for `err` here and adjust accordingly if there is an error?

pkg/kp/service.go
```diff
@@ -38,3 +39,16 @@ func (c consulStore) RegisterService(manifest pods.Manifest, caPath string) erro
...
+	return resp
+}
+
+func HttpsStatusCheck(node string, port int) *http.Response {
+	url := fmt.Sprintf("https://%s:%d/_status", node, port)
+	resp, _ := http.Get(url)
```
petertseng: same question about checking for error

robertabbott: if Get has a non-2xx response it returns an err. If figured we would rather handle such an error by looking at the response.

pkg/watch/health.go
```diff
@@ -0,0 +1,239 @@
...
+	"github.com/square/p2/pkg/logging"
+	"github.com/square/p2/pkg/pods"
+	"github.com/square/p2/pkg/preparer"
+	"github.com/square/p2/pkg/util/net"
+)
+
+// These constants should probably all be something the p2 user can set
+// in their preparer config...
+
+// number of milliseconds between reality store checks
```
petertseng: I think these comments are a bit inconsistent with the units. `3 * time.Second` appears to actually be a nanosecond count. It might just be best to say it's a Duration
http://golang.org/pkg/time/#Duration

pkg/watch/health.go
```diff
@@ -0,0 +1,239 @@
...
+
+func resultFromCheck(resp *http.Response) (health.Result, error) {
+	var err error
+	res := health.Result{}
+	if resp == nil {
+		res.Status = health.Critical
+		return res, nil
+	}
+	if resp.StatusCode == 200 {
+		res.Status = health.Passing
```
petertseng: hmm but could it be possible that a 200 status will have output as well? do we not care in this case?

robertabbott: I think we don't care in that case. Though I may be mistaken

petertseng: I guess we don't really care, it's OK and that's all we need t know

pkg/kp/kv.go
petertseng: hmm ok. I was asking because I was wondering if we would set an ACL on the entire tree (like we have `intent/` and `reality/` and it seems harder to do unless they're all organized under `/health/<service>/<node>` or something, but if y'all are ok with it then can ship

pkg/kp/service.go
petertseng: ah ok then. As long as resp isn't nil or have a nil body or something I guess we're OK. I don't know what guarantees the api provides so I had to ask

petertseng:
sorry on being late to the show in reviewing. good to go from me. I mainly only took a quick glance but looks like Anthony has taken a more detailed look. If there was one last thing I'd ask to think about it's the `healthPath` thing

pkg/kp/service.go
anthonybishopric: @petertseng is right that this needs to be addressed. It's very likely you will get a ConnectionRefused if the service is not available, making `resp` nil.

pkg/watch/health.go
anthonybishopric: I think we do care. It's an easy feature to provide and someone will ask for it.

pkg/kp/kv.go
anthonybishopric: Yeah, sorry just noticing this too. I think it's probably a better idea to have a separate part of the tree instead of concatenating the words together. `/health/{service}/{node}` is probably a good bet because it matches the pattern of having namespaces for types of data, ie `/intent`, `/reality`. As written this would make it very hard to execute a sensible watch against the health of all services.

pkg/watch/health.go
robertabbott: I'll handle the err then.

pkg/kp/service.go
robertabbott: I dealt with this by making the result critical if the resp is nil. In the case of resp == nil I would think we would still want the critical result. Would the best thing to do be log the error? and still write the health result.

anthonybishopric: Ideally we'd include the text in `err` (currently `_`) in a critical health result. Logging as well couldn't hurt.

robertabbott: Ok. I will log it and return the error as Result.Output

pkg/watch/health.go
```diff
@@ -0,0 +1,239 @@
...
+		res.Status = health.Critical
+		res.Output = err.Error()
+		return res, nil
+	}
+
+	res.Output, err = getBody(resp)
+	if resp.StatusCode == 200 {
+		res.Status = health.Passing
+	} else if resp.StatusCode > 200 && resp.StatusCode < 300 {
+		res.Status = health.Passing
```
petertseng: looks like you can merge this with the previous now since they're both doing the same thing

pkg/watch/health_test.go
```diff
@@ -0,0 +1,87 @@
...
+		Status: health.Critical,
+	}
+	Assert(t).AreEqual(true, p.updateNeeded(res, 1000), "should need update since Result.Status changed")
+
+	res.Status = health.Passing
+	Assert(t).AreEqual(true, p.updateNeeded(res, 0), "TTL is 0 so should always need update")
+	Assert(t).AreEqual(false, p.updateNeeded(res, 1000), "TTL is >> than time since ti was created and status is unchanged")
+}
+
+func TestResultFromCheck(t *testing.T) {
```
petertseng: looks like you could add a case for `resultFromCheck(resp, non-nil-error)` and make sure it's critical

robertabbott: added tests for this in most recent update

petertseng: :+1:

petertseng:
:+1:

PR #198, opened 2015-07-20
Abbott/healthcheck

robertabbott:
hopefully this isn't a bad way to check both the kv store and the health catalog


bin/p2-inspect/main.go
```diff
@@ -81,7 +81,7 @@ func main() {
 			}
 
 			old := statusMap[podId][node]
-			_, old.Health = health.FindWorst(results)
+			_, old.Health = health.GetMultisourceResult(results)
```
anthonybishopric: Why does the public API on pkg/health need to change?

pkg/health/health.go
```diff
@@ -28,24 +30,33 @@ type Result struct {
...
 		return nil, util.Errorf("/health/service failed for %q: %s", serviceID, err)
 	}
 
-	ret := make(map[string][]Result)
-	for _, entry := range entries {
-		ret[entry.Node.Node] = make([]Result, 0, len(entry.Checks))
+	ret := make(map[string][][]Result)
+	for _, entry := range catalogEntries {
+		// get entry in KV store for this service/node
+		kvEntry, err := h.consulStore.GetHealth(serviceID, entry.Node.Node)
```
anthonybishopric: This results in N queries even though we know we can query for the health of the service using one. Worth refactoring.

Also, this implementation implies that all services will have an entry in the catalog but this is definitely not the case.

bin/p2-inspect/main.go
robertabbott: I doesn't need to. I just thought the name FindWorst didn't represent super accurately what the method would do.

pkg/health/health.go
```diff
@@ -61,9 +72,14 @@ func (h ConsulHealthChecker) WatchNodeService(nodename string, serviceID string,
 			})
 			if err != nil {
 				errCh <- err
+			}
+			kvCheck, err := h.consulStore.GetHealth(nodename, serviceID)
```
anthonybishopric: Is there an error if the health is not present for that node? It needs to be the case that a service:node is healthy when either health checks are present, not necessarily both.

bin/p2-inspect/main.go
anthonybishopric: We should try to minimize public API changes if possible. The fact that health lives in two places isn't important to clients. `FindWorst` is an important aspect of `pkg/replication` since we always want to ensure we're working on hosts that are the "least healthy" at any given time.

pkg/health/health_test.go
```diff
@@ -0,0 +1,52 @@
+package health
+
+import (
+	"testing"
+
+	. "github.com/anthonybishopric/gotcha"
+)
```
anthonybishopric: Need tests for:

* Catalog is healthy but KV is not present
* Catalog is healthy but KV is not
* KV is healthy but Catalog is not
* KV is healthy but Catalog is not present at all

All the above cases should result in passing health checks from `pkg/health`

bin/p2-inspect/main.go
robertabbott: That makes sense. I'll revert to using the name FindWorst

pkg/health/health.go
```diff
@@ -28,24 +30,33 @@ type Result struct {
...
+		}
+
+		res := make([]Result, 0, len(entry.Checks))
 		for _, check := range entry.Checks {
-			ret[entry.Node.Node] = append(ret[entry.Node.Node], consulCheckToResult(*check))
+			res = append(res, consulCheckToResult(*check))
 		}
+		kvRes := []Result{consulWatchToResult(kvEntry)}
+		// place worst result from catalog and result from KV store in map
+		ret[entry.Node.Node] = [][]Result{kvRes, res}
```
anthonybishopric: How are clients expected to use this nested array? Could you explain why this change is necessary?

robertabbott: Rather than return a single list of results it now returns one list per source of health info. I think this would be confusing for the client as is. The other options were to append kv store results to the catalog results and the api would be the same as it was before. However if this is done there is no way to distinguish between the two sources of health.

I think the best way would be to return FindWorst(all results) and return a single result. This would result in changes to the API, but unless there is a strong reason for wanting to return all the results this option might be the most clear. All calls to Service and WatchNodeService call FindWorst on the results anyway. The downside of this means we're deciding for the client what 'worst' is.

pkg/health/health.go
```diff
@@ -28,24 +30,33 @@ type Result struct {
 	Output  string
 }
 
-func (h ConsulHealthChecker) Service(serviceID string) (map[string][]Result, error) {
-	entries, _, err := h.client.Health().Service(serviceID, "", false, nil)
+func (h ConsulHealthChecker) Service(serviceID string) (map[string][][]Result, error) {
```
petertseng: could we have a comment here to describe the characteristics of the `map[string][][]Result`? Kind of our fault for not documenting it before--I think even in its previous form `map[string][]Result` it could have deserved a comment about the structure (what will the keys be? given a key, what does the corresponding value represent?). But since you are changing it to have another array, I think it's a good time to doc.

anthonybishopric: We discussed this a bit in person, but we decided to consolidate this back to a `map[string]Result`. Yes, documenting this is valuable and encouraged.

pkg/health/health.go
robertabbott: It will be healthy if either check is present. There should not be an error if the health is not present for that node. GetHealth doesn't error on a nil response and consul's Get doesn't either.

robertabbott: Also i added an integration test to ensure that this is the case.

pkg/kp/kv.go
```diff
@@ -101,6 +104,28 @@ func (c consulStore) GetHealth(service, node string) (WatchResult, error) {
...
+	key := healthPath(service, "")
+	res, _, err := c.client.KV().List(key, nil)
+	if err != nil {
+		return healthRes, KVError{Op: "get", Key: key, UnsafeError: err}
+	} else if res == nil {
+		return healthRes, nil
+	}
+	for _, kvp := range res {
+		watch := &WatchResult{}
+		err = json.Unmarshal(kvp.Value, watch)
```
petertseng: you're setting this here but not checking it until after the `for` loop - does this mean that all errors except the very last one will be ignored? Is that OK or should the error check be moved into the `for` loop?

robertabbott: error check should be in for loop. Fixing now

mpuncel: maybe worth looking into https://github.com/hashicorp/go-multierror ?

pkg/health/health_test.go
```diff
@@ -0,0 +1,129 @@
...
+		ID:     "testwarn",
+		Status: Warning,
+	}
+	c := Result{
+		ID:     "testpass",
+		Status: Passing,
+	}
+
+	id, _ := FindWorst([]Result{a, b})
+	Assert(t).AreEqual(id, a.ID, "FindWorst should have returned testcrit")
```
petertseng: I think we can do with a bit of descriptiveness like "FindWorst between Critical and Warning should have returned testcrit"

pkg/health/health_test.go
```diff
@@ -0,0 +1,129 @@
...
+func TestPickServiceResult(t *testing.T) {
+	t1 := mockServiceEntry("1", "passing")
+	t2 := mockServiceEntry("2", "passing")
+	t3 := mockServiceEntry("3", "passing")
+
+	// Catalog is healthy but KV is not present
+	testMap := getResult([]*api.ServiceEntry{t1, t2, t3})
+	catalog := []*api.ServiceEntry{t1, t2, t3}
+	res, err := selectResult(catalog, nil)
+	Assert(t).AreEqual(err, nil, "selectResult returned err")
```
petertseng: Hmm. I think we might benefit from something to distinguish this "selectResult returned err" from the other three instances, so that if a test fails on us we have an easy way to see which one failed.

What would you say to splitting this up into four functions like `TestPickServiceResultCatalogHealthyKvAbsent`, `TestPickServiceResultCatalogHealthyKvUnhealthy`? I think @bwester  and @anthonybishopric  had a discussion about our trend toward bigger test functions, but I don't remember the result of the conversation.

I would either 1) split it up into multiple test functions, one for each of the four cases or 2) at least distinguish the error messages for the four cases

pkg/health/health_order.go
```diff
@@ -18,13 +18,13 @@ func (s SortOrder) Swap(i, j int) {
 
 func (s SortOrder) Less(i, j int) bool {
 	iHealth := Unknown
-	if res, ok := s.Health[s.Nodes[i]]; ok && len(res) > 0 {
-		_, iHealth = FindWorst(res)
+	if res, ok := s.Health[s.Nodes[i]]; ok && &res != nil {
```
petertseng: hmm I need to think about this one for a bit. Can taking the address of a variable result in nil? I might need an explanation on this one.

petertseng:
Yup I think I'm mostly happy with this, just a question about more easily telling which test failed out of four, and whether `&var` can ever be `nil`

pkg/health/health_order.go
robertabbott: I don't think it can. I'm not sure why I put this in.

pkg/health/health_test.go
robertabbott: I took option 2, but if anyone has a strong preference for smaller methods I can split it up.

petertseng: sure

petertseng:
:+1:  from me

pkg/health/health_test.go
bwester: The whole point of these tests is (a) to make sure that your code behaves the way you think it should, (b) to make sure it stays that way as you rewrite it, and (c) when _a_ or _b_ break, to make it easy for you to understand what went wrong. Your change makes _c_ easier, so I'm happy.

(I don't remember this earlier conversation at all, btw.)

integration/single-node-slug-deploy/check.go
```diff
@@ -391,7 +396,39 @@ func verifyHealthChecks(config *preparer.PreparerConfig, services []string) erro
 		}
 	}
 
+	res, err := store.GetServiceHealth(services[1])
+	getres, _ := store.GetHealth(services[1], name)
+	if err != nil {
+		return err
+	}
+	val := res[fmt.Sprintf("health/p2-preparer/%s", name)]
```
anthonybishopric: Use the health path function here instead?

integration/single-node-slug-deploy/check.go
```diff
@@ -391,7 +396,39 @@ func verifyHealthChecks(config *preparer.PreparerConfig, services []string) erro
...
+	if getres.Id != val.Id || getres.Service != val.Service || getres.Status != val.Status {
+		return fmt.Errorf("GetServiceHealth failed %+v", res)
+	}
+
 	// if it reaches here it means health checks
 	// are being written to the KV store properly
 	return nil
 }
+
+func emptyHealthCheck(config *preparer.PreparerConfig) error {
```
anthonybishopric: This seems a little out of place in the integration test, since the integration test is dedicated to launching the hello world service. I'd have suggested writing unit tests for this, possibly even emulating the upstream Consul service with the `httptest` package if you wanted to get sophisticated.

pkg/health/health.go
```diff
@@ -71,14 +60,58 @@ func (h ConsulHealthChecker) WatchNodeService(nodename string, serviceID string,
 					if outResult.Service != serviceID && outResult.Service != "" {
 						continue
 					}
-					out = append(out, outResult)
+					catalogResults = append(catalogResults, outResult)
 				}
-				resultCh <- out
+				kv := consulWatchToResult(kvCheck)
```
anthonybishopric: `s/kv/kvCheckResult`

pkg/health/health.go
```diff
@@ -71,14 +60,58 @@ func (h ConsulHealthChecker) WatchNodeService(nodename string, serviceID string,
...
+				catalog := findWorstResult(catalogResults)
```
anthonybishopric: `s/catalog/catalogResult`

anthonybishopric:
Had a few nits, but looked much better on the second go-around. :star:

integration/single-node-slug-deploy/check.go
robertabbott: Consul has a package called testutil that would be very good for tests that require consul API calls. Using the consul package is probably the easiest way to write these tests.

robertabbott:
Godep made that a struggle but this should now be good to go

petertseng:
:+1: !!!

PR #200, opened 2015-07-23
fix integration test for health check

robertabbott:
Had to fix logging in the watch package. Was trying
to reference nil pointers because PodWatch's logger
field was nil.

The manifest for hello which is populated in check.go
did not have the StatusPort or StatusHTTP fields set.

Integration test for watch pkg now asserts that all
services being health checked are passing. I don't
know why I didn't do that before...

### DIFF
--- a/integration/single-node-slug-deploy/check.go
+++ b/integration/single-node-slug-deploy/check.go
@@ -15,6 +15,7 @@ import (
 	"runtime"
 	"time"
 
+	"github.com/square/p2/pkg/health"
 	"github.com/square/p2/pkg/kp"
 	"github.com/square/p2/pkg/pods"
 	"github.com/square/p2/pkg/preparer"
@@ -31,7 +32,7 @@ func main() {
 	// 5. Verify that hello is running (listen to syslog? verify Runit PIDs? Both?)
 
 	// list of services running on integration test host
-	services := []string{"consul", "p2-preparer", "hello"}
+	services := []string{"p2-preparer", "hello"}
 	tempdir, err := ioutil.TempDir("", "single-node-check")
 	log.Printf("Putting test manifests in %s\n", tempdir)
 	if err != nil {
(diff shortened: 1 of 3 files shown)


integration/single-node-slug-deploy/check.go
```diff
@@ -31,7 +32,7 @@ func main() {
 	// 5. Verify that hello is running (listen to syslog? verify Runit PIDs? Both?)
 
 	// list of services running on integration test host
-	services := []string{"consul", "p2-preparer", "hello"}
```
petertseng: so since you have some uses of `services[1]` below, does that mean you need to change them to `services[0]`?

robertabbott: It should really just check all of them, which is what I'm having it do now.

petertseng: ah OK. that makes more sense

pkg/watch/health.go
```diff
@@ -57,8 +58,9 @@ type PodWatch struct {
 // runs a CheckHealth routine to monitor the health of each
 // service and kills routines for services that should no
 // longer be running.
-func MonitorPodHealth(config *preparer.PreparerConfig, logger *logging.Logger, shutdownCh chan struct{}) {
+func MonitorPodHealth(config *preparer.PreparerConfig, shutdownCh chan struct{}) {
```
petertseng: could you explain real quick why no longer passing in a logger?

petertseng: by that I mean... I think one possible alternative could have been to change `*logging.Logger` to `logging.Logger` and still pass the logger in from `p2-preparer/main.go` - is there reason we should favor one approach over the other

robertabbott: I originally removed it to help figure out where the nil pointer dereference error was coming from. It had nothing to do with the error, but I left the change in because it simplifies the MonitorPodHealth API. Though now that I think about it thats not really a good thing cause it makes the logging opaque and inflexible for the client.

I will revert this.

petertseng:
:+1:

PR #201, opened 2015-07-24
removes catalog health checking

robertabbott:


pkg/watch/health.go
```diff
@@ -226,7 +226,7 @@ func updatePods(current []PodWatch, reality []kp.ManifestResult, logger *logging
 
 		// if a manifest is in reality but not current a podwatch is created
 		// with that manifest and added to newCurrent
-		if missing == true {
+		if missing == true && man.Manifest.StatusPort != 0 {
```
petertseng: `== true` is redundant isn't it? I maybe should have said something before, but while we're changing this line might as well, right?

pkg/health/health_test.go
```diff
@@ -41,7 +41,7 @@ func TestPickServiceResult(t *testing.T) {
 	testMap := getResult([]*api.ServiceEntry{t1, t2, t3})
 	catalog := []*api.ServiceEntry{t1, t2, t3}
 	res, err := selectResult(catalog, nil)
-	Assert(t).AreEqual(err, nil, "catalog is health, kv not present and selectResult returned err")
+	Assert(t).AreEqual(err, nil, "catalog is healthy, kv not present and selectResult returned err")
```
petertseng: in an ideal world these would be in a separate commit reading "fix typos in health test messages" or something

pkg/watch/health_test.go
```diff
@@ -60,8 +60,11 @@ func TestUpdateNeeded(t *testing.T) {
 }
 
 func TestResultFromCheck(t *testing.T) {
+	http.HandleFunc("/_status", statusHandler)
+	go http.ListenAndServe(":8080", nil)
 	client := http.DefaultClient
-	resp, _ := client.Get("http://ifconfig.co/all.json")
```
petertseng: interesting...

petertseng:
This is looking good 👍

Small nits that could be fixed.

petertseng:
COOL

PR #202, opened 2015-07-27
preparer: Support HTTPS

petertseng:
The support is there in `kp.Options`, we just were not exposing the option in preparer config.


petertseng:
`../Development/go/src/github.com/square/p2/pkg/preparer/setup_test.go:19: gotcha.Assert(t).True undefined (type *gotcha.Asserter has no field or method True)`

I am not very good at this. just a second.

robertabbott:
LGTM

PR #204, opened 2015-07-29
Add TLS configuration options for Consul connection

bwester:
Adds the `cert_file`, `key_file`, and `ca_file` configuration options for the
p2-preparer.  These options are used to configure the TLS transport options for
the client used to communicate with the Consul agent.


mpuncel:
:+1:

PR #205, opened 2015-07-30
Fix servicebuilder discrepencies

bwester:
Updates the servicebuild clone to fix a couple of discrepencies compared with
the canonical version:
* Didn't create the "app/log/main" directory
* Used timeout value in "run" and "log/run" scripts instead of "2"

Also makes the following improvements:
* Make it harder to accidentally remove mandatory trailing whitespace when
  using editors that remove trailing whitespaces
* Simplify some of the logic in the unit tests

### DIFF
--- a/pkg/runit/test_helper.go
+++ b/pkg/runit/test_helper.go
@@ -1,6 +1,7 @@
 package runit
 
 import (
+	"io/ioutil"
 	"os"
 	"path/filepath"
 	"runtime"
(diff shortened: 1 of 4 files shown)


pkg/runit/test_helper.go
```diff
@@ -20,24 +22,56 @@ func FakeChpst() string {
...
-	os.MkdirAll(stagingRoot, 0755)
-	runitRoot := filepath.Join(testDir, "/var/service")
-	os.MkdirAll(runitRoot, 0755)
-
-	return &ServiceBuilder{
-		ConfigRoot:  configRoot,
-		StagingRoot: stagingRoot,
-		RunitRoot:   runitRoot,
-		Bin:         fakeSBBinPath,
+	fmt.Printf("created test root: %s\n", root)
```
petertseng: hmm, you OK with keeping this?

petertseng:
👍 yup seems good, one small thing

pkg/runit/test_helper.go
bwester: I'm OK with it in (it's only visible if the test fails), but I'm also OK removing it.

bwester:
Amended:
* Don't print the test root when creating test ServiceBuilder

PR #206, opened 2015-07-30
Expose the wait time parameter to kp.Options so that clients can conf…

mpuncel:
…igure it

### DIFF
--- a/pkg/kp/config.go
+++ b/pkg/kp/config.go
@@ -2,6 +2,7 @@ package kp
 
 import (
 	"net/http"
+	"time"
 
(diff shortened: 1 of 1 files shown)


bwester:
Code looks fine... so far. Aren't you forgetting to include some place where `Options.WaitTime` can be set?

mpuncel:
@bwester you're right that this doesn't give much flexibility, but I guess I feel unable to anticipate which places will want to override WaitTime on a per-request basis.

It would also mean adding an argument to a lot of places that will be either `nil` or an empty `kp.Options` struct

PR #207, opened 2015-07-30
Add "--show-consul-errors-unsafe" flag to show Consul errors

bwester:
We don't want to show Consul errors all the time—they can leak ACL tokens—but
when debugging a problem involving Consul, that kind of information is very
important to understanding what's going on. This commit adds a command line
flag that will enable detailed log messages, to be used only when needed.

### DIFF
--- a/bin/p2-preparer/main.go
+++ b/bin/p2-preparer/main.go
@@ -8,6 +8,8 @@ import (
 	"syscall"
 
 	"github.com/Sirupsen/logrus"
+	"gopkg.in/alecthomas/kingpin.v1"
+
 	"github.com/square/p2/pkg/logging"
 	"github.com/square/p2/pkg/preparer"
 	"github.com/square/p2/pkg/version"
@@ -15,6 +17,9 @@ import (
 )
 
 func main() {
+	// Other packages define flags, and they need parsing here.
+	kingpin.Parse()
+
 	logger := logging.NewLogger(logrus.Fields{})
 	configPath := os.Getenv("CONFIG_PATH")
 	if configPath == "" {
--- a/pkg/kp/kv.go
+++ b/pkg/kp/kv.go
@@ -9,9 +9,16 @@ import (
 	"time"
 
 	"github.com/hashicorp/consul/api"
+	"gopkg.in/alecthomas/kingpin.v1"
+
 	"github.com/square/p2/pkg/pods"
 	"github.com/square/p2/pkg/util"
 )
+
+var showConsulErrors = kingpin.Flag(
+	"show-consul-errors-unsafe",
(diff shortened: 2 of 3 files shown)


pkg/kp/kv.go
```diff
@@ -81,7 +92,7 @@ func (c consulStore) PutHealth(res WatchResult) (time.Time, time.Duration, error
 		retDur = writeMeta.RequestTime
 	}
 	if err != nil {
-		return t, retDur, KVError{Op: "get", Key: key, UnsafeError: err}
+		return t, retDur, KVError{Op: "put", Key: key, UnsafeError: err}
```
petertseng: oops

pkg/kp/lock.go
```diff
@@ -78,7 +78,7 @@ func (l Lock) Lock(key string) error {
 	}, nil)
 
 	if err != nil {
-		return util.Errorf("Could not acquire lock on %s", key)
+		return KVError{Op: "acquire lock", Key: key, UnsafeError: err}
```
petertseng: Ah, I've had to do this before, good that it will be on master now

mpuncel:
lgtm

pkg/kp/kv.go
```diff
@@ -9,10 +9,17 @@ import (
 	"time"
 
 	"github.com/hashicorp/consul/api"
+	"gopkg.in/alecthomas/kingpin.v1"
+
 	"github.com/square/p2/pkg/pods"
 	"github.com/square/p2/pkg/util"
 )
 
+var showConsulErrors = kingpin.Flag(
```
petertseng: ah interesting, haven't thought of putting a kingpin flag in a pkg before (as opposed to something in bin), but it seems to work for what we want here, eh

bin/p2-preparer/main.go
```diff
@@ -15,6 +17,7 @@ import (
 )
 
 func main() {
+	kingpin.Parse()
```
petertseng: do you anticipate that someone reading this code might be confused at "there are no kingpin flags in this file! what is going on! why is it being parsed?!" ? Maybe something to explain if so.

petertseng:
OK from me, maybe a thought about whether a comment could help avoid confusion

pkg/kp/kv.go
bwester: This is a pretty common use case for [gflags](https://gflags.github.io/gflags/) which got significantly simplified to become the standard [`flag`](http://golang.org/pkg/flag/) package.  I would have used `flag` if other p2 binaries hadn't already started using kingpin, and the two packages don't seem to get along well together.

bwester:
Amended:
* Added note about kingpin flags coming from other packages

petertseng:
👍

PR #209, opened 2015-08-03
Better error message regarding absent preparers

anthonybishopric:
Currently any error when checking for the reality store entry
for a preparer pod will indicate that no preparer is present. This
commit improves the error message so that communication errors are
more apparent.


petertseng:
sure, why not. 👍

PR #212, opened 2015-08-05
Fix bug in f3658d76

bwester:
Commit f3658d76 reordered parameter arguments in a call to Sprintf(). This
commit fixes the bug.


petertseng:
looks much better. 👍

anthonybishopric:
:+1:

PR #213, opened 2015-08-05
Two bug fixes + godep update

anthonybishopric:
This PR includes three commits, two of which represent bug fixes.

* When handling responses from Consul, check that `err != nil` prior to using `meta`. 
* Don't write (and actively remove) empty servicebuilder files.

The third commit simply updates the project's dependencies.


pkg/health/health.go
```diff
@@ -40,10 +47,9 @@ func (h ConsulHealthChecker) WatchNodeService(nodename string, serviceID string,
 		case <-quitCh:
 			return
 		case <-time.After(1 * time.Second):
-			checks, meta, err := h.client.Health().Node(nodename, &api.QueryOptions{
-				WaitIndex: curIndex,
-				WaitTime:  h.WaitTime,
-			})
+			var err error
+			var checks []*api.HealthCheck
+			checks, curIndex, err = h.fetchNodeHealth(nodename, curIndex)
```
bwester: `curIndex` doesn't appear to be used anywhere... why produce it?

anthonybishopric:
@petertseng correct, we were just holding it wrong.

pkg/health/health.go
anthonybishopric: it's used on this line and passed to the health call, just as before.

bwester: Ah, of course! I was just misreading the code.

bwester:
Looks fine to me
:ship:

PR #214, opened 2015-08-05
KV health checks now use https when the statusHttp field is set

robertabbott:
The preparer creates http client from config. This is passed to
the go routines that monitor pods that require https.


robertabbott:
still working on the integration test for https

pkg/watch/health.go
```diff
@@ -36,6 +36,7 @@ const TTL = 60 * time.Second
 // has a running MonitorHealth go routine
 type PodWatch struct {
 	manifest pods.Manifest
+	client   *http.Client
```
petertseng: hmm, it doesn't seem to me like having this declaration is terribly necessary. Do you have something in mind I'm not thinking of?

pkg/preparer/setup.go
```diff
@@ -151,22 +151,38 @@ func getTLSClient(certFile, keyFile, caFile string) (*http.Client, error) {
...
 		Address: c.ConsulAddress,
 		HTTPS:   c.ConsulHttps,
 		Token:   token,
 		Client:  client,
-	}), nil
+	}, nil
+}
+
+func (c *PreparerConfig) GetClient() (*http.Client, error) {
+	client, err := getTLSClient(c.CertFile, c.KeyFile, c.CAFile)
```
petertseng: Will this be OK in the case of `c.ConsulHttps == false`?

pkg/watch/health.go
robertabbott: I want to use the same http client for health checks that is used by everything else in the preparer. I thought about adding methods to the Store interface for actually performing health checks but that didn't really make sense to me.

petertseng: Oops. I misread the code. deleting this comment.

pkg/preparer/setup.go
robertabbott: probably not. Will fix

pkg/watch/health.go
petertseng: actually I misread the code a bit. this seems fine now that I look at it again

petertseng: (I thought this was in a function, not in a struct, so was confused. Upon further review I could not have been correct because if it were in a function it needs to be `var client *http.Client`)

I think what you have written here is preferable. I agree with you that adding methods to `Store` for checking seems a bit weird.

integration/single-node-slug-deploy/check.go
```diff
@@ -296,7 +296,7 @@ func postHelloManifest(dir string) error {
 	manifest := &pods.Manifest{}
 	manifest.Id = "hello"
 	manifest.StatusPort = 43770
-	manifest.StatusHTTP = true
+	manifest.StatusHTTP = false
```
anthonybishopric: ... or don't set it at all

pkg/preparer/setup.go
```diff
@@ -151,22 +151,38 @@ func getTLSClient(certFile, keyFile, caFile string) (*http.Client, error) {
...
+	if err != nil {
+		return &http.Client{}, err
```
bwester: In idiomatic Go, the `&http.Client{}` should never be examined. In cases like this, just use `nil`. (The exception being if you return a struct by value, where you _must_ construct an empty value to return.)

After that transform, you can simplify this function further.

pkg/kp/service.go
```diff
@@ -60,7 +60,7 @@ func HttpStatusCheck(node string, port int) (*http.Response, error) {
 	return http.Get(url)
 }
 
-func HttpsStatusCheck(node string, port int) (*http.Response, error) {
+func HttpsStatusCheck(client *http.Client, node string, port int) (*http.Response, error) {
```
bwester: Why not do this to `HttpStatusCheck` too?

pkg/watch/health.go
```diff
@@ -186,7 +200,7 @@ func (p *PodWatch) updateNeeded(res health.Result, ttl time.Duration) bool {
 
 // compares services being monitored with services that
 // need to be monitored.
-func updatePods(current []PodWatch, reality []kp.ManifestResult, logger *logging.Logger, store kp.Store, node string) []PodWatch {
+func updatePods(store kp.Store, client *http.Client, current []PodWatch, reality []kp.ManifestResult, node string, logger *logging.Logger) []PodWatch {
```
bwester: nit: Now that there are so many arguments here, can you switch formatting styles? This will help keep future diffs small and precise, and there is less room for error putting new arguments in the wrong slot (like I just did in my other diff).

```
func updatePods(
    store kp.Store,
    ...,
    logger *logging.Logger,
) []PodWatch {
```

pkg/preparer/setup.go
anthonybishopric: +1

pkg/watch/health.go
```diff
@@ -85,19 +93,25 @@ func MonitorPodHealth(config *preparer.PreparerConfig, logger *logging.Logger, s
...
 	}
 }
 
-func (p *PodWatch) checkHealth(node string, port int, store kp.Store) {
-	resp, err := kp.HttpStatusCheck(node, port)
+func (p *PodWatch) checkHealth(store kp.Store, node string, port int) {
+	var resp *http.Response
+	var err error
+	if p.manifest.StatusHTTP == true {
+		resp, err = kp.HttpsStatusCheck(p.client, node, port)
```
bwester: This has gotten too unwieldy. The code is passing around a node name, a port, an HTTP client, and a manifest just so it can have the data available right here to perform the status check it needs.  This state comes from a mix of fields in a struct definition and parameters threaded through half a dozen functions.

I want you consider an alternate way of structuring this check: define a new `StatusCheck` type that encapsulates all the state it needs to perform the check and pass around an instance of that.  The following might work

    type StatusCheck struct {
        ID string
        Node string // maybe?
        URI string
        Client *http.Client
    }

    func (sc StatusCheck) Check() health.Result { ... }

You wouldn't need quite so many parameters in all your functions, and it decouples the _gathering_ of a check from the _handling_ of one.

robertabbott:
finally

integration/single-node-slug-deploy/check.go
```diff
@@ -23,6 +23,10 @@ import (
 )
 
 const preparerStatusPort = 32170
+const preparerLog = "/var/service/p2-preparer__p2-preparer__launch/log/main/current"
```
mpuncel: nit: use `const (
preparerStatusPort = 32170
... )` syntax

integration/single-node-slug-deploy/check.go
```diff
@@ -63,6 +67,10 @@ func main() {
 	fmt.Println("Executing bootstrap")
 	err = executeBootstrap(signedPreparerManifest, signedConsulManifest)
 	if err != nil {
+		cmd := exec.Command("cat", preparerLog)
```
mpuncel: why use cat instead of `ioutil.ReadFile`?

integration/single-node-slug-deploy/check.go
```diff
@@ -384,6 +394,12 @@ func verifyHealthChecks(config *preparer.PreparerConfig, services []string) erro
 		} else if (res == kp.WatchResult{}) {
 			return fmt.Errorf("No results for %s", sv)
 		} else if res.Status != string(health.Passing) {
+			cmd := exec.Command("curl", "localhost:8500/v1/kv/health/hello?recurse")
```
mpuncel: why `curl` instead of an http get?

robertabbott: no reason I guess. These were mostly for debugging travis. Is it worth changing them to http get and ReadFile? or should I just remove them?

integration/common-provision.sh
```diff
@@ -15,6 +15,22 @@ export GOPATH=/usr/local/share/go
...
+localityName=SF
+commonName=$HOSTNAME
+organizationalUnitName=Vel
+emailAddress=doesntmatter@something.edu
+"
+
+CERTPATH=/var/tmp/certs
+mkdir -p $CERTPATH
+openssl req -x509 -newkey rsa:2048 -keyout $CERTPATH/key.pem -out $CERTPATH/cert.pem -nodes -days 300 -subj "$(echo -n "$subj" | tr "\n" "/")"
+cp $CERTPATH/cert.pem $CERTPATH/cert2.pem
```
bwester: No need to have a second file.

integration/single-node-slug-deploy/check.go
```diff
@@ -129,11 +131,13 @@ func generatePreparerPod(workdir string) (string, error) {
 			"type":    "keyring",
 			"keyring": util.From(runtime.Caller(0)).ExpandPath("pubring.gpg"),
 		},
+		"ca_file":     certpath + "/cert2.pem",
```
bwester: Just use _cert.pem_.

integration/single-node-slug-deploy/check.go
```diff
@@ -63,6 +64,7 @@ func main() {
 	fmt.Println("Executing bootstrap")
 	err = executeBootstrap(signedPreparerManifest, signedConsulManifest)
 	if err != nil {
+
```
bwester: nit: extraneous line

integration/travis.sh
```diff
@@ -12,4 +12,19 @@ sudo mkdir -p /etc/servicebuilder.d /var/service-stage /var/service
...
+ST=CA
+O=SQ
+localityName=SF
+commonName=$HOSTNAME
+organizationalUnitName=Vel
+emailAddress=doesntmatter@something.edu
+"
+CERTPATH=/var/tmp/certs
+mkdir -p $CERTPATH
+openssl req -x509 -newkey rsa:2048 -keyout $CERTPATH/key.pem -out $CERTPATH/cert.pem -nodes -days 300 -subj "$(echo -n "$subj" | tr "\n" "/")"
```
bwester: I'm surprised this works in Travis!  Last time I tried submitting code that created asymmetric keys in a test, it caused the test to time out.

In developing this test, have you seen _any_ of your test cases fail due to runtime problems?

pkg/kp/service.go
```diff
@@ -42,6 +42,7 @@ func (c consulStore) RegisterService(manifest pods.Manifest, caPath string) erro
 				manifest.StatusPort,
 			)
 		} else {
+			fmt.Println("saklfdhsdakfjashfkjasdfhs ", caPath, "    ", manifest.ID())
```
bwester: Stray debugging output?

pkg/watch/health.go
```diff
@@ -64,17 +74,23 @@ func MonitorPodHealth(config *preparer.PreparerConfig, logger *logging.Logger, s
 			WithField("inner_err", err).
 			Fatalf("error creating health monitor KV store")
 	}
+	client, err := config.GetClient()
+	if err != nil {
+		logger.
+			WithField("inner_err", err).
+			Warningln("failed to get http client for this preparer")
```
bwester: I noticed you changed from a `Fatalln()` to a `Warningln()`. Under what conditions can `GetClient()` fail here?

pkg/watch/health.go
```diff
@@ -85,44 +101,48 @@ func MonitorPodHealth(config *preparer.PreparerConfig, logger *logging.Logger, s
...
 		case <-shutdownCh:
 			return
 		}
 	}
 }
 
-func (p *PodWatch) checkHealth(node string, port int, store kp.Store) {
-	resp, err := kp.HttpStatusCheck(node, port)
+func (p *PodWatch) checkHealth(store kp.Store, sc StatusCheck) {
+	resp, err := sc.Check()
```
bwester: FYI: Even with a multiple-value return, Go lets you [compose function calls](http://play.golang.org/p/0xOe-dHgGU) like these together!

pkg/watch/health.go
```diff
@@ -85,44 +101,48 @@ func MonitorPodHealth(config *preparer.PreparerConfig, logger *logging.Logger, s
...
 	health, err := resultFromCheck(resp, err)
 	if err != nil {
 		return
 	}
-	health.ID = p.manifest.Id
-	health.Node = node
-	health.Service = p.manifest.Id
+	health.ID = sc.ID
```
bwester: ret := foo.Check()
    ret.X = foo.X
    ret.Y = foo.Y
    return ret

This is not a good pattern to see. If the check is going to return a health result, the check itself should fill in as much of the result as it can, not rely on the caller to copy data between structs. In general, the more extra steps are required to use an API correctly, the greater the chance of misusing it. I'd like to see the check be as self contained as possible.

    return foo.Check()

pkg/watch/health.go
```diff
@@ -85,44 +101,48 @@ func MonitorPodHealth(config *preparer.PreparerConfig, logger *logging.Logger, s
...
 			return
 		}
 	}
 }
 
-func (p *PodWatch) checkHealth(node string, port int, store kp.Store) {
-	resp, err := kp.HttpStatusCheck(node, port)
+func (p *PodWatch) checkHealth(store kp.Store, sc StatusCheck) {
+	resp, err := sc.Check()
 	health, err := resultFromCheck(resp, err)
```
bwester: Not that you want to be composing these functions, at least not here. To present the caller with a simple, hard-to-misuse interface, this should probably be folded into `Check()`. It does make sense to keep it as a separate function, so you can unit test it without faking/mocking HTTP calls.

pkg/watch/health.go
```diff
@@ -85,44 +101,48 @@ func MonitorPodHealth(config *preparer.PreparerConfig, logger *logging.Logger, s
...
-	health.Node = node
-	health.Service = p.manifest.Id
+	health.ID = sc.ID
+	health.Node = sc.Node
+	health.Service = sc.ID
 
-	if p.updateNeeded(health, TTL) {
-		p.lastCheck, err = writeToConsul(health, store)
-		p.lastStatus = health.Status
+	if sc.updateNeeded(health, TTL) {
```
bwester: I think there are two actions going on here:

* Get an app's status
* Write a status to Consul when needed

The procedures and data needed to do each action are completely separate; they don't gain anything by being bundled into the same structure. Can you break the latter task into its own unit?

integration/travis.sh
robertabbott: I have yet to see it failed due to runtime problems

integration/single-node-slug-deploy/check.go
```diff
@@ -129,11 +130,13 @@ func generatePreparerPod(workdir string) (string, error) {
 			"type":    "keyring",
 			"keyring": util.From(runtime.Caller(0)).ExpandPath("pubring.gpg"),
 		},
+		"ca_file":     certpath + "/cert.pem",
```
anthonybishopric: nit: `filepath.Join(certpath, "cert.pem")`, etc

robertabbott: fixed

integration/travis.sh
anthonybishopric: Yeah, I think we will see issues doing this eventually. Travis's virtual machines just don't generate enough entropy to do things like this. Can we not just generate these and check them in?

robertabbott: I'm not sure. The reason I did it this way is because the certificates only work on hosts who's hostname == the commonName field of the key. Since the travis machines don't all have the same hostname I generated them on the fly. If there was a way to make keys/certs that work everywhere then we could just check them in.

anthonybishopric: Ok, lets leave it as is for now - but I think we may want to figure out a better approach eventually.

robertabbott: I could also make the key smaller to decrease the chances we run into this kind of issue.

anthonybishopric: Could be okay, although I'm not sure it would improve things terribly. I've seen 100 bits of entropy take about a minute to produce on an XL instance in AWS.

pkg/preparer/setup.go
```diff
@@ -151,22 +151,41 @@ func getTLSClient(certFile, keyFile, caFile string) (*http.Client, error) {
...
 		HTTPS:   c.ConsulHttps,
 		Token:   token,
 		Client:  client,
-	}), nil
+	}, err
+}
+
+func (c *PreparerConfig) GetClient() (*http.Client, error) {
+	client, err := getTLSClient(c.CertFile, c.KeyFile, c.CAFile)
+	if err != nil {
```
bwester: The conditionals here are not needed. You can write this function in one line.

pkg/watch/health.go
```diff
@@ -64,69 +78,181 @@ func MonitorPodHealth(config *preparer.PreparerConfig, logger *logging.Logger, s
...
+		// if a manifest is in reality but not current a podwatch is created
+		// with that manifest and added to newCurrent
+		if missing && man.Manifest.StatusPort != 0 {
+			newPod := PodWatch{
+				manifest: man.Manifest,
+				store:    store,
+				statusChecker: StatusChecker{
+					ID:     man.Manifest.Id,
+					Node:   node,
+					URI:    fmt.Sprintf("%s:%d", node, man.Manifest.StatusPort),
```
bwester: I strongly suggest that you include the full URI here, e.g.: "http://foo.square:1212/_status". This will help remove even more redundant state and code.

pkg/preparer/setup.go
```diff
@@ -151,22 +151,41 @@ func getTLSClient(certFile, keyFile, caFile string) (*http.Client, error) {
...
+	opts, err := c.getOpts()
 	if err != nil {
 		return nil, err
 	}
+	return kp.NewConsulStore(opts), nil
+}
 
-	client, err := getTLSClient(c.CertFile, c.KeyFile, c.CAFile)
+func (c *PreparerConfig) getOpts() (kp.Options, error) {
+	client := &http.Client{}
```
petertseng: any preference between `http.DefaultClient` versus `&http.Client{}` ?

(they're currently the same thing so I'm not fussed, but I ask anyway in case someone really cares)

robertabbott: we use defaultClient elsewhere so I'll go with that.

pkg/watch/health.go
robertabbott: done

pkg/watch/health.go
```diff
@@ -64,69 +78,179 @@ func MonitorPodHealth(config *preparer.PreparerConfig, logger *logging.Logger, s
...
+		}
+
+		// if a manifest is in reality but not current a podwatch is created
+		// with that manifest and added to newCurrent
+		if missing && man.Manifest.StatusPort != 0 {
+			sc := StatusChecker{
+				ID:     man.Manifest.Id,
+				Node:   node,
+				Client: client,
+				HTTPS:  man.Manifest.StatusHTTP,
```
bwester: You have this backwards. Don't assign _StatusHTTP_ to a variable called _HTTP**S**_.

For that matter, you don't use `StatusChecker.HTTPS` anywhere except the line below.  Get rid of the variable from the structure.

integration/travis.sh
```diff
@@ -12,4 +12,19 @@ sudo mkdir -p /etc/servicebuilder.d /var/service-stage /var/service
...
+C=US
+ST=CA
+O=SQ
+localityName=SF
+commonName=$HOSTNAME
+organizationalUnitName=Vel
+emailAddress=doesntmatter@something.edu
+"
+CERTPATH=/data/pods/hello/hello/installs/hoisted-hello_def456/bin/
+find / -name launch
```
bwester: What's going on here?

bwester:
Once the http/https bug is fixed and Travis is happy, I'm fine signing off on this.

robertabbott:
Made Benji's https/http change. I had to revert to the hello server with hard-coded cert path to stop travis from complaining. I'll fix that in a separate PR.

anthonybishopric:
:ship:

pkg/watch/health.go
```diff
@@ -64,69 +77,178 @@ func MonitorPodHealth(config *preparer.PreparerConfig, logger *logging.Logger, s
...
+		// if a manifest is in reality but not current a podwatch is created
+		// with that manifest and added to newCurrent
+		if missing && man.Manifest.StatusPort != 0 {
+			sc := StatusChecker{
+				ID:     man.Manifest.Id,
+				Node:   node,
+				Client: client,
+			}
+			if man.Manifest.StatusHTTP {
+				sc.URI = fmt.Sprintf("https://%s:%d/_status", node, man.Manifest.StatusPort)
```
bwester: NO!!!!

PR #215, opened 2015-08-06
I had http and https backwards

robertabbott:
sorry

### DIFF
--- a/integration/single-node-slug-deploy/check.go
+++ b/integration/single-node-slug-deploy/check.go
@@ -138,6 +138,7 @@ func generatePreparerPod(workdir string) (string, error) {
 	}
 	manifest.RunAs = "root"
 	manifest.StatusPort = preparerStatusPort
+	manifest.StatusHTTP = true
 
 	manifestBytes, err := manifest.Marshal()
 	if err != nil {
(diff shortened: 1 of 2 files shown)


anthonybishopric:
:+1: I think we can improve these two status related fields significantly - the naming isn't entirely obvious. `StatusHTTP` being false could be assumed to mean there are no checks at all.

integration/single-node-slug-deploy/check.go
```diff
@@ -300,7 +301,6 @@ func postHelloManifest(dir string) error {
 	manifest := &pods.Manifest{}
 	manifest.Id = "hello"
 	manifest.StatusPort = 43770
-	manifest.StatusHTTP = true
```
anthonybishopric: So how was this passing before? Hello should have been listening on HTTPs before?

pkg/watch/health.go
```diff
@@ -166,9 +166,9 @@ func updatePods(store kp.Store,
 				Client: client,
 			}
 			if man.Manifest.StatusHTTP {
-				sc.URI = fmt.Sprintf("https://%s:%d/_status", node, man.Manifest.StatusPort)
-			} else {
 				sc.URI = fmt.Sprintf("http://%s:%d/_status", node, man.Manifest.StatusPort)
+			} else {
+				sc.URI = fmt.Sprintf("https://%s:%d/_status", node, man.Manifest.StatusPort)
```
bwester: :+1:

integration/single-node-slug-deploy/check.go
robertabbott: The tests and the watch pkg were both backwards. Check was listening on https and when statusHTTP was true the watch pkg was using https

bwester:
:ship:

PR #216, opened 2015-08-06
Add extra logging information to log messages that have an 'err' fiel…

mpuncel:
…d. This enables another process to consume log messages and publish them to an exception reporting service such as Airbrake or Bugsnag

### DIFF
--- a/bin/p2-preparer/main.go
+++ b/bin/p2-preparer/main.go
@@ -27,12 +27,12 @@ func main() {
 	}
 	preparerConfig, err := preparer.LoadConfig(configPath)
 	if err != nil {
-		logger.WithField("inner_err", err).Fatalln("could not load preparer config")
+		logger.WithError(err).Fatalln("could not load preparer config")
 	}
 
 	prep, err := preparer.New(preparerConfig, logger)
 	if err != nil {
-		logger.WithField("inner_err", err).Fatalln("Could not initialize preparer")
+		logger.WithError(err).Fatalln("Could not initialize preparer")
 	}
 	defer prep.Close()
 
--- a/pkg/kp/kv.go
+++ b/pkg/kp/kv.go
@@ -6,6 +6,8 @@ import (
 	"bytes"
 	"encoding/json"
 	"fmt"
+	"path/filepath"
+	"runtime"
 	"time"
 
 	"github.com/hashicorp/consul/api"
@@ -67,14 +69,50 @@ type KVError struct {
 	Op          string
 	Key         string
 	UnsafeError error
+	filename    string
+	function    string
+	lineNumber  int
 }
+
+var _ util.CallsiteError = KVError{}
+
+const KVErrorFormat = "%s failed for path %s%s"
 
 func (err KVError) Error() string {
 	cerr := ""
 	if *showConsulErrors {
 		cerr = fmt.Sprintf(": %s", err.UnsafeError)
 	}
-	return fmt.Sprintf("%s failed for path %s%s", err.Op, err.Key, cerr)
+	return fmt.Sprintf(KVErrorFormat, err.Op, err.Key, cerr)
+}
+
+func (err KVError) Filename() string {
+	return err.filename
+}
+
+func (err KVError) Function() string {
+	return err.function
+}
+
+func (err KVError) LineNumber() int {
+	return err.lineNumber
+}
+
+func NewKVError(op string, key string, unsafeError error) KVError {
+	var function string
+	// Skip one stack frame to get the file & line number of caller.
+	pc, file, line, ok := runtime.Caller(1)
+	if ok {
+		function = runtime.FuncForPC(pc).Name()
+	}
+	return KVError{
+		Op:          op,
+		Key:         key,
+		UnsafeError: unsafeError,
+		filename:    filepath.Base(file),
+		function:    function,
+		lineNumber:  line,
+	}
 }
 
 func (c consulStore) PutHealth(res WatchResult) (time.Time, time.Duration, error) {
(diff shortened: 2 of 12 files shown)


mpuncel:
sample error message: 
```
2015-08-06_22:08:58.57061 {"Counter":0,"PID":2253,"err":"setup.go:223: error configuring keyring auth: stat /usr/local/share/go/src/github.com/square/p2/integration/single-node-slug-deploy/pubring.gpg: no such file or directory","filename":"setup.go","function":"github.com/square/p2/pkg/preparer.New","level":"fatal","line_number":223,"msg":"Could not initialize preparer","time":"2015-08-06T18:08:58-04:00"}
```

pkg/logging/logging.go
```diff
@@ -61,6 +61,23 @@ func (l *Logger) NoFields() *logrus.Entry {
 }
 
 func (l *Logger) WithFields(fields logrus.Fields) *logrus.Entry {
+	// If an err field is provided, and it conforms to the util.StackError
+	// interface, add some fields to the log message that can be used for
+	// exception reporting by a buddy launchable (e.g. to bugsnag or
+	// airbrake)
+	var stackErr util.StackError
+	if err, ok := fields["err"]; ok {
```
mpuncel: maybe it would be better to test all fields for conformity to the `util.StackError` interface? that way `err` isn't a magic key

pkg/logging/logging.go
```diff
@@ -61,6 +61,23 @@ func (l *Logger) NoFields() *logrus.Entry {
...
+		stackErr, _ = err.(util.StackError)
+	} else {
+		if err, ok := l.baseFields["err"]; ok {
+			stackErr, _ = err.(util.StackError)
```
anthonybishopric: I think we should figure out a safer way to detect whether an error contains a stack trace, using interface upgrades. This approach is really brittle and highly coupled to the util package.

anthonybishopric: My mistake, was confused with `stackError`, `StackError` is indeed an interface.

bin/p2-preparer/main.go
```diff
@@ -27,12 +27,12 @@ func main() {
...
 	preparerConfig, err := preparer.LoadConfig(configPath)
 	if err != nil {
-		logger.WithField("inner_err", err).Fatalln("could not load preparer config")
+		logger.WithField("err", err).Fatalln("could not load preparer config")
 	}
 
 	prep, err := preparer.New(preparerConfig, logger)
 	if err != nil {
-		logger.WithField("inner_err", err).Fatalln("Could not initialize preparer")
+		logger.WithField("err", err).Fatalln("Could not initialize preparer")
```
anthonybishopric: We should make a first-class error method to initialize a field set (and get rid of the magic `"err"`s everywhere).

```
logger.WithError(err).Fatalln("Could not initialize preparer")
```

pkg/pods/pod.go
```diff
@@ -518,15 +518,14 @@ func (pod *Pod) getLaunchable(launchableStanza LaunchableStanza, runAsUser strin
...
-		"err": err.Error(),
-	}).Error(message)
+	p.logger.WithError(err).
+		Error(message)
 }
 
 func (p *Pod) logLaunchableError(launchableId string, err error, message string) {
-	p.logger.WithFields(logrus.Fields{
+	p.logger.WithError(err).
+		WithFields(logrus.Fields{
```
anthonybishopric: `.WithError(err).WithField("launchable", launchableId).Error(message)`

anthonybishopric:
:+1:

pkg/preparer/orchestrate.go
```diff
@@ -110,8 +109,8 @@ func (p *Preparer) WatchForPodManifestsForNode(quitAndAck chan struct{}) {
 func (p *Preparer) tryRunHooks(hookType hooks.HookType, pod hooks.Pod, manifest *pods.Manifest, logger logging.Logger) {
 	err := p.hooks.RunHookType(hookType, pod, manifest)
 	if err != nil {
-		logger.WithFields(logrus.Fields{
-			"err":   err,
+		logger.WithError(err).
+			WithFields(logrus.Fields{
```
petertseng: this can be a `WithField` now since there's only one, right?

pkg/preparer/orchestrate.go
```diff
@@ -197,9 +196,8 @@ func (p *Preparer) installAndLaunchPod(newManifest *pods.Manifest, pod Pod, logg
 	// if the old manifest is corrupted somehow, re-launch since we don't know if this is an update.
 	problemReadingCurrentManifest := (err != nil && err != pods.NoCurrentManifest)
 	if problemReadingCurrentManifest {
-		logger.WithFields(logrus.Fields{
-			"sha":       newSHA,
-			"inner_err": err,
+		logger.WithError(err).WithFields(logrus.Fields{
```
petertseng: can make it a `WithField`?

petertseng:
yup seems good

pkg/kp/kv.go
```diff
@@ -67,14 +69,50 @@ type KVError struct {
...
 	Key         string
 	UnsafeError error
+	filename    string
+	function    string
+	lineNumber  int
 }
 
+var _ util.CallsiteError = KVError{}
+
+const KVErrorFormat = "%s failed for path %s%s"
```
bwester: Any particular reason for moving away from it's use?

pkg/kp/kv.go
```diff
@@ -96,7 +134,7 @@ func (c consulStore) PutHealth(res WatchResult) (time.Time, time.Duration, error
 		retDur = writeMeta.RequestTime
 	}
 	if err != nil {
-		return t, retDur, KVError{Op: "put", Key: key, UnsafeError: err}
+		return t, retDur, NewKVError("put", key, err)
```
bwester: Any particular reason for only converting this one `KVError` to the new format?  Any problems with me doing the same for the rest?

pkg/kp/kv.go
mpuncel: I'm not sure I understand the question

pkg/kp/kv.go
mpuncel: nope, that was a goof on my part.

PR #217, opened 2015-08-06
find*Result in health package handles empty slice inputs

robertabbott:
In some cases findWorstResult and findBestResult were being
passed empty result slices. This was cause panics. Now when
they are passed an empty slice they return a HealthEmptyError.
HealthEmptyError wraps an error and allows callers to handle
it as they see fit.

Updates health_test to reflect these changes and includes a
test for empty slices passed to the aformentioned methods.

### DIFF
--- a/pkg/health/health.go
+++ b/pkg/health/health.go
@@ -78,13 +78,14 @@ func (h ConsulHealthChecker) WatchNodeService(nodename string, serviceID string,
 				kvCheckResult := consulWatchToResult(kvCheck)
 				catalogCheckResult, HEErr := findWorstResult(catalogResults)
 				if HEErr != nil {
-					errCh <- HEErr
-				}
-				best, HEErr := findBestResult([]Result{kvCheckResult, catalogCheckResult})
-				if HEErr != nil {
-					errCh <- HEErr
+					resultCh <- kvCheckResult
 				} else {
-					resultCh <- best
+					best, HEErr := findBestResult([]Result{kvCheckResult, catalogCheckResult})
+					if HEErr != nil {
+						errCh <- HEErr
+					} else {
+						resultCh <- best
+					}
 				}
 			}
 		}
--- a/pkg/health/health.go
+++ b/pkg/health/health.go
@@ -37,6 +37,14 @@ type Result struct {
 	Output  string
 }
 
+type HealthEmpty struct {
+	Res string
+}
+
+func (h *HealthEmpty) Error() string {
+	return h.Res
+}
+
 func (h ConsulHealthChecker) WatchNodeService(nodename string, serviceID string, resultCh chan<- Result, errCh chan<- error, quitCh <-chan struct{}) {
 	defer close(resultCh)
 
@@ -68,8 +76,16 @@ func (h ConsulHealthChecker) WatchNodeService(nodename string, serviceID string,
 					catalogResults = append(catalogResults, outResult)
 				}
 				kvCheckResult := consulWatchToResult(kvCheck)
-				catalogCheckResult := findWorstResult(catalogResults)
-				resultCh <- findBestResult([]Result{kvCheckResult, catalogCheckResult})
+				catalogCheckResult, HEErr := findWorstResult(catalogResults)
+				if HEErr != nil {
+					errCh <- HEErr
+				}
+				best, HEErr := findBestResult([]Result{kvCheckResult, catalogCheckResult})
+				if HEErr != nil {
+					errCh <- HEErr
+				} else {
+					resultCh <- best
+				}
 			}
 		}
 	}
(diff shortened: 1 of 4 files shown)


pkg/health/health.go
```diff
@@ -170,10 +214,15 @@ func findWorstResult(results []Result) Result {
...
 		}
 	}
-	return retVal
+	return retVal, nil
 }
 
-func findBestResult(results []Result) Result {
+func findBestResult(results []Result) (Result, *HealthEmptyError) {
+	if len(results) == 0 {
+		return Result{Status: Critical}, &HealthEmptyError{
```
robertabbott: I figured passing a critical result is >= a completely empty result in terms of usefulness.

pkg/health/health.go
```diff
@@ -107,26 +119,37 @@ func (h ConsulHealthChecker) Service(serviceID string) (map[string]Result, error
 }
 
 func selectResult(catalogEntries []*api.ServiceEntry, kvEntries map[string]kp.WatchResult) (map[string]Result, error) {
+	var HEErr *HealthEmptyError
```
anthonybishopric: Rename to `HealthEmpty`.

pkg/health/health.go
```diff
@@ -107,26 +119,37 @@ func (h ConsulHealthChecker) Service(serviceID string) (map[string]Result, error
...
+
+	// will return this error if there are no results
+	err := fmt.Errorf("no results found for catalog or kv")
 	for _, entry := range catalogEntries {
 		res := make([]Result, 0, len(entry.Checks))
 		for _, check := range entry.Checks {
 			res = append(res, consulCheckToResult(*check))
 		}
-		ret[entry.Node.Node] = findWorstResult(res)
+		ret[entry.Node.Node], HEErr = findWorstResult(res)
```
robertabbott: might be better to only set the value in the map if HEErr == nil

robertabbott: as opposed to a Result{Status: Critical}  if HEErr != nil

pkg/health/health.go
```diff
@@ -107,26 +119,37 @@ func (h ConsulHealthChecker) Service(serviceID string) (map[string]Result, error
 }
 
 func selectResult(catalogEntries []*api.ServiceEntry, kvEntries map[string]kp.WatchResult) (map[string]Result, error) {
+	var HEErr *HealthEmptyError
 	ret := make(map[string]Result)
+
+	// will return this error if there are no results
+	err := fmt.Errorf("no results found for catalog or kv")
```
anthonybishopric: Won't this be apparent if the `ret` is empty?

robertabbott: yes. Then maybe the HealthEmpty's in this method can just be ignored

robertabbott: not ignored but when they are non-nil nothing will be put in the map.

pkg/health/health.go
```diff
@@ -151,17 +171,38 @@ func consulWatchToResult(w kp.WatchResult) Result {
...
+		return "", Critical, HEErr
+	}
+	return best.ID, best.Status, HEErr
 }
 
-func findWorstResult(results []Result) Result {
+func findWorstResult(results []Result) (Result, *HealthEmpty) {
+	if len(results) == 0 {
+		return Result{Status: Critical}, &HealthEmpty{
+			Err: fmt.Errorf("no results were passed to findWorstResult"),
```
anthonybishopric: I don't think you need this nested error

anthonybishopric: Also, technically, you should make `HealthEmpty` conform to the `error` interface.

pkg/health/health.go
```diff
@@ -68,8 +76,16 @@ func (h ConsulHealthChecker) WatchNodeService(nodename string, serviceID string,
 					catalogResults = append(catalogResults, outResult)
 				}
 				kvCheckResult := consulWatchToResult(kvCheck)
-				catalogCheckResult := findWorstResult(catalogResults)
-				resultCh <- findBestResult([]Result{kvCheckResult, catalogCheckResult})
+				catalogCheckResult, HEErr := findWorstResult(catalogResults)
+				if HEErr != nil {
+					errCh <- HEErr
+				}
+				best, HEErr := findBestResult([]Result{kvCheckResult, catalogCheckResult})
```
anthonybishopric: `HEErr` is always nil in this case

pkg/health/health.go
```diff
@@ -68,8 +76,16 @@ func (h ConsulHealthChecker) WatchNodeService(nodename string, serviceID string,
 					catalogResults = append(catalogResults, outResult)
 				}
 				kvCheckResult := consulWatchToResult(kvCheck)
-				catalogCheckResult := findWorstResult(catalogResults)
-				resultCh <- findBestResult([]Result{kvCheckResult, catalogCheckResult})
+				catalogCheckResult, HEErr := findWorstResult(catalogResults)
+				if HEErr != nil {
+					errCh <- HEErr
```
anthonybishopric: In a world where we exclusively do KV checks, this will be an unending stream of errors. The outside world should only care about the case where _both_ error checks are blank.

pkg/health/health.go
```diff
@@ -68,8 +76,17 @@ func (h ConsulHealthChecker) WatchNodeService(nodename string, serviceID string,
 					catalogResults = append(catalogResults, outResult)
 				}
 				kvCheckResult := consulWatchToResult(kvCheck)
-				catalogCheckResult := findWorstResult(catalogResults)
-				resultCh <- findBestResult([]Result{kvCheckResult, catalogCheckResult})
+				catalogCheckResult, HEErr := findWorstResult(catalogResults)
+				if HEErr != nil {
+					resultCh <- kvCheckResult
```
anthonybishopric: You need to check if `kvCheckResult` is actually present since it could be empty, right?

robertabbott: GetHealth (on line 64) will return an error if there is no kv result in which case it won't reach this point. tldr kvCheckResult cant be empty here.

pkg/health/health.go
```diff
@@ -53,23 +61,54 @@ func (h ConsulHealthChecker) WatchNodeService(nodename string, serviceID string,
...
 			kvCheck, err := h.consulStore.GetHealth(nodename, serviceID)
-			if err != nil {
-				errCh <- err
+			pickHealthResult(catalogResults, kvCheck, err, resultCh, errCh)
+		}
+	}
+}
+
+// if there are no health results available will place error in errCh then return
+func pickHealthResult(catalogResults []Result, kvCheck kp.WatchResult, err error, resultCh chan<- Result, errCh chan<- error) {
```
anthonybishopric: Please rename `err` to `kvCheckErr` so it's clear what it's supposed to be (`err` is fairly nondescript)

pkg/health/health_test.go
```diff
@@ -9,6 +9,45 @@ import (
 	"github.com/square/p2/pkg/kp"
 )
 
+func TestPickHealthResult(t *testing.T) {
+	catalogResults := []Result{Result{Status: Passing}}
+	kvCheck := kp.WatchResult{Status: string(Critical)}
+	errCh := make(chan error, 1)
+	resultCh := make(chan Result, 1)
```
anthonybishopric: Why buffer the channels?

pkg/health/health.go
```diff
@@ -53,23 +61,54 @@ func (h ConsulHealthChecker) WatchNodeService(nodename string, serviceID string,
...
+	if err != nil {
+		// if there are neither kv nor catalog results
+		if len(catalogResults) == 0 {
+			errCh <- err
+			return
+		} else {
+			// there are no kv results but there are catalog results
+			catalogCheckResult, HEErr := findWorstResult(catalogResults)
```
anthonybishopric: `HEErr` is redundant here, yes?

robertabbott: yes and theres another redundant one on line 108

robertabbott: removed them in most recent update

anthonybishopric:
:ship:

PR #218, opened 2015-08-13
Put writing to /v1/kv/health behind a config flag so it can be easily…

mpuncel:
… disabled

### DIFF
--- a/bin/p2-preparer/main.go
+++ b/bin/p2-preparer/main.go
@@ -49,7 +49,7 @@ func main() {
 
 	quitMainUpdate := make(chan struct{})
 	quitHookUpdate := make(chan struct{})
-	quitMonitorPodHealth := make(chan struct{})
+	quitChans := []chan struct{}{quitHookUpdate}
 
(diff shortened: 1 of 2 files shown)


bin/p2-preparer/main.go
```diff
@@ -64,20 +64,25 @@ func main() {
...
 	// all pods on this host and writes the information to consul
-	go watch.MonitorPodHealth(preparerConfig, &logger, quitMonitorPodHealth)
+	if preparerConfig.WriteKVHealth {
+		quitMonitorPodHealth := make(chan struct{})
+		go watch.MonitorPodHealth(preparerConfig, &logger, quitMonitorPodHealth)
+		quitChans = append(quitChans, quitMonitorPodHealth)
+	}
 
-	waitForTermination(logger, quitMainUpdate, quitHookUpdate, quitMonitorPodHealth)
+	waitForTermination(logger, quitMainUpdate, quitChans...)
```
anthonybishopric: why variadic and not a slice?

anthonybishopric:
:+1:

PR #220, opened 2015-08-17
Turns off kv health integration test

robertabbott:
Since kv health has been turned off due to the
excessive io use, its integration test needs
to be turned off as well

### DIFF
--- a/integration/single-node-slug-deploy/check.go
+++ b/integration/single-node-slug-deploy/check.go
@@ -34,7 +34,7 @@ func main() {
 	// 5. Verify that hello is running (listen to syslog? verify Runit PIDs? Both?)
 
 	// list of services running on integration test host
-	services := []string{"p2-preparer", "hello"}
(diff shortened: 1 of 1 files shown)


integration/single-node-slug-deploy/check.go
```diff
@@ -86,10 +86,6 @@ func main() {
 	if err != nil {
 		log.Fatalf("Couldn't get hello running: %s", err)
 	}
-	err = verifyHealthChecks(config, services)
-	if err != nil {
-		log.Fatalf("Could not get health check info from consul: %s", err)
-	}
```
mpuncel: you commented out `services` but deleted this. Do you anticipate it coming back?

mpuncel:
lgtm, just a little strange that you deleted parts and commented out parts, should probably be consistent (and if you choose commenting, leave a note about when to uncomment)

PR #221, opened 2015-08-17
Fixes preparer config in integration tests

robertabbott:


bwester:
Could you amend your commit to give some context?  What problem is this commit solving?  How does the commit solve it?

robertabbott:
updated commit message

bwester:
:ship:

PR #222, opened 2015-08-17
Fail tests that get skipped by consul's test package and add the -tes…

mpuncel:
…t.v flag to 'go test' so skips are reported

### DIFF
--- a/pkg/kp/kv_test.go
+++ b/pkg/kp/kv_test.go
@@ -53,6 +53,14 @@ func TestGetHealthWithEntry(t *testing.T) {
 }
 
 func makeStore(t *testing.T) (Store, *testutil.TestServer) {
+	// Detect if the test gets skipped by testutil.NewTestServerConfig.
+	// This happens if 'consul' isn't in $PATH
+	defer func() {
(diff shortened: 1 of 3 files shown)


pkg/kp/kv_test.go
```diff
@@ -53,6 +53,14 @@ func TestGetHealthWithEntry(t *testing.T) {
 }
 
 func makeStore(t *testing.T) (Store, *testutil.TestServer) {
+	// Detect if the test gets skipped by testutil.NewTestServerConfig.
+	// This happens if 'consul' isn't in $PATH
+	defer func() {
+		if t.Skipped() {
+			t.Fatal("test skipped by testutil package")
```
bwester: After looking at the output generated by a skipped-then-failed run, something like "failing skipped test" would make more sense in context.

mpuncel: works for me

bwester:
Looks like the Travis CI script doesn't install `consul`, so unit tests once skipped are now failing. Fix that, and this will be good to go.

PR #223, opened 2015-08-18
Rework pkg/replication to be easier for clients to work with.

mpuncel:
1) kp.NewLock() now will auto renew the lock in consul until explicitly
stopped
2) The functionality of Replicator was split into two structs,
Replicator and Replication. Replicator has an InitializeReplication()
function that returns a single-use instance of Replication, which has
the necessary data to perform a replication of a pod to a set of nodes.

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -23,9 +23,14 @@ task :build do
   e "go build -v ./..."
 end
 
-desc 'Test all projects'
+desc 'Test all projects (short only)'
 task :test => [:build] do
-  e "go test -timeout 10s ./..."
+  e "go test -short -timeout 10s ./..."
+end
+
+desc 'Test all projects'
+task :test_all => [:build] do
+  e "go test -timeout 120s ./..."
 end
 
 desc 'Update all dependencies'
--- a/pkg/replication/common_test_setup.go
+++ b/pkg/replication/common_test_setup.go
@@ -1,0 +1,193 @@
+package replication
+
+import (
+	"fmt"
+	"io/ioutil"
+	"sync/atomic"
+	"testing"
+
+	"github.com/square/p2/pkg/health"
+	"github.com/square/p2/pkg/kp"
+	"github.com/square/p2/pkg/logging"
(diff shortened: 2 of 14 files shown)


Rakefile
```diff
@@ -23,9 +23,14 @@ task :build do
...
-desc 'Test all projects'
+desc 'Test all projects (short only)'
 task :test => [:build] do
-  e "go test -timeout 10s ./..."
+  e "go test -short -timeout 10s ./..."
+end
+
+desc 'Test all projects'
+task :test_all => [:build] do
+  e "go test -short -timeout 120s ./..."
```
bwester: Which tests take longer than 10s?

bwester: And, shouldn't you leave out the `-short` flag here?

mpuncel: the tests that use `github.com/hashicorp/consul/testutil` are very slow, ~5s each. 

I was thinking that local developers would still use `rake test` and have fast tests and travis would run the full suite with long tests included, happy to take any suggestions

mpuncel: and yes I should leave out the `-short` flag

bwester:
I'm pretty much overwhelmed by such a large single commit that is a mix of bug fixes, refactors, and new functionality. I can't give it as thorough a review as usual, but from what I can follow, it seems fine. Get one more person to take a glance at this, and I'll approve it for merging.

pkg/replication/common_test_setup.go
```diff
@@ -0,0 +1,193 @@
...
+		Address: server.HTTPAddr,
+	})
+	return store, server
+}
+
+// Adds preparer manifest to reality tree to fool replication library into
+// thinking it is installed on the test nodes
+func setupPreparers(server *testutil.TestServer) {
+	for _, node := range testNodes {
+		key := fmt.Sprintf("reality/%s/p2-preparer", node)
```
anthonybishopric: Use the path helpers here

mpuncel:
breaking this up into multiple PRs

PR #224, opened 2015-08-20
Preliminary commits before session-based health checks

bwester:
This PR contains some preliminary commits that need to be in place before I introduce the main PR for session-based health checks. Those changes aren't quite ready for a PR, but I think the changes here are of high quality and not likely to change, so I'd like y'all to start looking these over.

I recommend reviewing one commit at a time, rather than looking at the entire "files changed" view for the PR.

### DIFF
--- a/pkg/kp/kv.go
+++ b/pkg/kp/kv.go
@@ -76,14 +76,12 @@ type KVError struct {
 
 var _ util.CallsiteError = KVError{}
 
-const KVErrorFormat = "%s failed for path %s%s"
-
 func (err KVError) Error() string {
 	cerr := ""
 	if *showConsulErrors {
 		cerr = fmt.Sprintf(": %s", err.UnsafeError)
 	}
-	return fmt.Sprintf(KVErrorFormat, err.Op, err.Key, cerr)
+	return fmt.Sprintf("%s failed for path %s%s", err.Op, err.Key, cerr)
 }
 
 func (err KVError) Filename() string {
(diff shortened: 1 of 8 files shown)


mpuncel:
lgtm, i think yours might conflict with my PR on the consul test fixture stuff but not a big deal

pkg/kp/kv.go
```diff
@@ -52,6 +52,26 @@ type WatchResult struct {
 	Status  string
 	Output  string
 	Time    time.Time
+	Expires time.Time `json:"expires,omitempty"`
```
robertabbott: Should the other fields have json tags as well? Why just this one?

bwester: I added this field comment so I could specify the "omitempty" option, indicating that this field might not be present. The default JSON marshaller handles the other fields well enough; they don't need special handling.

But I do need a change here. As mentioned in another PR, the JSON field name should be capitalized, `Expires`, to match the other fields.  Interestingly, the YAML marshaller adjusts the case of its fields, and I was expecting that behavior from JSON too.

pkg/kp/kv.go
```diff
@@ -52,6 +52,26 @@ type WatchResult struct {
...
+		r.Output == s.Output
+}
+
+// IsStale returns true when the result is stale according to the local clock.
+func (r WatchResult) IsStale() bool {
+	expires := r.Expires
+	if expires.IsZero() {
+		expires = r.Time.Add(TTL)
+	}
+	return time.Now().After(r.Expires)
```
bwester: Oops, should be `expires`.

bwester:
Rebased to address review feedback

pkg/util/stream/publisher.go
```diff
@@ -0,0 +1,154 @@
...
+	case s.values <- val:
+		close(s.initialized)
+	case <-s.canceled:
+	}
+}
+
+// close the subscription's data stream if the subscription wasn't previously canceled
+func (s *StringSubscription) close() {
+	select {
+	default:
```
robertabbott: was putting the default case first done for style/readability reasons?

bwester: The main job of the method is to close its output socket, so I wrote that first. It's also makes the method similar to `waitForReady()` and `sendFirst()` in that checking for cancellation is always written last.

petertseng:
I think I can get behind this PR, so 👍

minor nit for using `Fatal` rather than `Fatalln`, which I trust you'll resolve as you see fit

bwester:
Amended to use `Fatalln()` per project convention instead of `Fatal()`.

PR #225, opened 2015-08-21
Session-based health checks

bwester:
Adds support for session-based health checks. Instead of having each health check write a fresh result to Consul, this PR allow Consul to write one long-lived health check that is updated only when it changes. The health check result is locked to a session, and if the preparer dies, its health checks will time out.

Like the preliminary PR, I recommend reviewing each individual commit rather than the entire PR at once.


robertabbott:
leader election takes 2-3 seconds apparently and thats why rake test times out

bwester:
@robertabbott: The unit tests that use `testutil` do take a while to start up, but I don't think that's the main reason why tests are timing out. There are many passing tests that use testutil; not all fail.  The Travis CI failure here is deeper than just a slow startup time, and I haven't taken the time to dive into it yet.

bwester:
@robertabbott So, I think the only "deep" problem is that I didn't understand how "go test" applied timeouts. (It's a timeout on the total runtime of all test functions in one package.) It also turns out that it interacts poorly with the Consul tester:  when the timeout happens, the test just kinda hangs.

pkg/kp/healthmanager.go
```diff
@@ -0,0 +1,340 @@
...
+	logger logging.Logger,
+) HealthManager {
+	return &consulSimpleHealthManager{
+		store:  c,
+		logger: logger,
+	}
+}
+
+// NewUpdater implements the HealthManager interface.
+func (m *consulSimpleHealthManager) NewUpdater(_service string) HealthUpdater {
```
anthonybishopric: Unused parameter?

pkg/kp/healthmanager.go
```diff
@@ -0,0 +1,340 @@
...
+	store      consulStore
+	logger     logging.Logger
+	lastCheck  time.Time
+	lastStatus string
+}
+
+// newSimpleHealthManager creates a simple health manager that writes a service's health
+// status to Consul every 15 seconds or whenever its status changes.
+func (c consulStore) newSimpleHealthManager(
+	_node string,
```
anthonybishopric: Unused?

pkg/kp/healthmanager.go
bwester: Yes. The underscore is meant to suggest that it's intentional that the parameter isn't used. The other implementation of `HealthUpdater` uses those parameters to validate that updates are applied to the right nodes.  While those checks apply to `consulSimpleHealthManager` too, I didn't bother putting them in.  If you think it's important, I can do so.

pkg/kp/healthmanager.go
bwester: Intentional (as with previous comment)

pkg/kp/healthmanager.go
anthonybishopric: I'd have expected a truly unused param to be named `_`, which is why I was confused. If you're leaving them in for documentation-ish purposes, maybe an explanatory comment is in order.

pkg/kp/healthmanager.go
```diff
@@ -0,0 +1,340 @@
...
+	}
+}
+
+// Close implements the HealthManager interface. consulSimpleHealthManager has no
+// resources to release.
+func (m *consulSimpleHealthManager) Close() {}
+
+// PutHealth implements the HealthUpdater interface.
+func (u *consulSimpleHealthUpdater) PutHealth(health WatchResult) error {
+	if health.Status != u.lastStatus || time.Since(u.lastCheck) > TTL/4 {
```
anthonybishopric: I believe the `> TTL/4` is the previous implementation's attempt to reduce the write traffic to Consul. Maybe just delete this so there isn't any confusion about what the `consulSimpleHealthUpdater`'s contract is (since it is advertised as always updating the health every time)

pkg/kp/healthmanager.go
```diff
@@ -0,0 +1,340 @@
...
+	logger logging.Logger,
+) HealthManager {
+	// Create a stream of sessions
+	sessionChan := make(chan string)
+	done := make(chan struct{})
+	// Current time of "Jan 2, 15:04:05" turns into "0102-150405"
+	timeStr := time.Now().Format("0102-150405")
+	go ConsulSessionManager(
+		api.SessionEntry{
+			Name:      fmt.Sprintf("health:%s:%s", node, timeStr),
```
anthonybishopric: Should we include the PID?

pkg/kp/healthmanager.go
bwester: Sure, if you think it helps improve the readability.

pkg/kp/healthmanager.go
bwester: The appropriate advertisement should be: "it implements whatever behavior was there before, but this one uses a new interface."

And actually, I think a TTL of 60 seconds is far too long. 10-15 seconds seems more reasonable for detecting that a preparer has died and all the services on the machine could be in questionable state. (A 15s TTL would, incidentally, make the revised check equal to `oldTTL / 4`.)

pkg/kp/healthmanager.go
```diff
@@ -0,0 +1,340 @@
...
+				"service": service,
+				"node":    m.node,
+			}),
+		)
+	}()
+	return u
+}
+
+func (u *consulHealthUpdater) PutHealth(health WatchResult) error {
+	if health.Id != u.service || health.Node != u.node || health.Service != u.service {
```
anthonybishopric: TBD: implications of `health.Id != health.Service`

pkg/kp/healthmanager.go
bwester: Sure, it might be useful.

pkg/kp/healthmanager.go
bwester: The existing code didn't make any distinction, so my changes didn't either.  In general, I'd say that if a distinction were required, `NewUpdater(service string)` should become `NewUpdater(id, service string)`.  Want that in this stack of commits?

mpuncel:
looks good to me

pkg/kp/healthmanager.go
anthonybishopric: That's alright, this PR is pretty solid honestly. I am thinking of a future in which we have possibly multiple status ports and/or overlay the preparer's health on the service's health in the same manner that eviction from the Consul serf cluster marks an app unhealthy.

anthonybishopric:
:+1:

pkg/kp/healthmanager.go
bwester: After considering this a bit more, I'd rather not touch the old implementation's timeouts in this PR. This kind of behavior is necessary for a watchdog service: you will always want to refresh the timeout well in advance of its expiration. It increases, not decreases, the write traffic by 4×.

The contract being advertised is:
> // newSimpleHealthManager creates a simple health manager that writes a service's health
// status to Consul every 15 seconds or whenever its status changes.

pkg/kp/healthmanager.go
```diff
@@ -0,0 +1,372 @@
...
+						if !ok {
+							return fmt.Errorf("write denied")
+						}
+						return nil
+					})
+				} else {
+					go sendHealthUpdate(writeLogger, w, localHealth, func() error {
+						_, err := client.KV().Put(kv, nil)
+						if err != nil {
+							return NewKVError("acquire", kv.Key, err)
```
bwester: :dizzy_face:

bwester:
Squashed all feedback into their respective commits & rebased onto master to handle merge conflicts

PR #227, opened 2015-08-24
make session locks self renewal so clients don't have to implement th…

mpuncel:
…is behavior

### DIFF
--- a/pkg/kp/lock.go
+++ b/pkg/kp/lock.go
@@ -12,17 +12,33 @@ type Lock struct {
 	client  *api.Client
 	session string
 	name    string
+
+	// Coordination channels
+	//
+	// signals that continual renewal of lock should stop
+	quitCh chan struct{}
+
+	// communicates any error occuring during renewal
+	renewalErrCh chan error
+
+	// signals when a renewal on the consul lock should be performed
+	renewalCh <-chan time.Time
 }
 
 type AlreadyLockedError struct {
 	Key string
 }
 
+const (
+	lockTTL         = "15s"
+	renewalInterval = 10 * time.Second
+)
+
 func (err AlreadyLockedError) Error() string {
 	return fmt.Sprintf("Key %q is already locked", err.Key)
 }
 
-func (c consulStore) NewLock(name string) (Lock, error) {
+func (c consulStore) NewLock(name string, renewalCh <-chan time.Time) (Lock, chan error, error) {
 	session, _, err := c.client.Session().CreateNoChecks(&api.SessionEntry{
 		Name: name,
 		// if the lock delay is zero, it becomes the default value, which is 15s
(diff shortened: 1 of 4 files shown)


pkg/kp/lock_test.go
```diff
@@ -0,0 +1,39 @@
...
+import (
+	"testing"
+	"time"
+)
+
+const (
+	lockMessage = "Locked by lock_test.go"
+)
+
+func TestLock(t *testing.T) {
```
anthonybishopric: I want this test to be clearer about what it's verifying. From reading the test I can tell that it's establishing a lock and showing that a client that does not use the lock cannot access a key. However, I don't see tests that show that owning the lock does allow reads and writes.

anthonybishopric: While I don't want to have fully blown-out test names a la Rspec land, I do think it's valuable to know exactly what is being tested in the form of comments so we can establish how comprehensive coverage is at a glance.

mpuncel: I don't believe that locking does prevent reads or writes, we just use a convention that any process performing a replication must first acquire locks named a particular way.

pkg/kp/kv_test.go
```diff
@@ -1,8 +1,6 @@
 package kp
 
-import (
-	"testing"
-)
+import "testing"
```
bwester: nit: Did an automatic tool make this change? I'd rather you keep the old formatting if possible.

pkg/kp/lock.go
```diff
@@ -97,6 +145,11 @@ func (l Lock) Renew() error {
 
 // destroy a lock, releasing and deleting all the keys it holds
 func (l Lock) Destroy() error {
+	l.quitCh <- struct{}{}
```
bwester: `close(l.quitCh)` would be more idiomatic Go.

pkg/kp/lock.go
```diff
@@ -86,6 +117,23 @@ func (l Lock) Lock(key string) error {
...
+		select {
+		case <-l.renewalCh:
+			err := l.Renew()
+			if err != nil {
+				l.renewalErrCh <- err
+				l.Destroy()
+				return
+			}
+		case <-l.quitCh:
+			close(l.renewalErrCh)
```
bwester: Could you boost this to the top of the method as `defer close(l.renewalErrCh)`?  That give the `Lock` object a nice invariant: the error channel will always be closed once its producer quits producing.

pkg/kp/kv_test.go
mpuncel: not sure how that happened, i like the old format too.

pkg/kp/lock.go
mpuncel: addressed

anthonybishopric:
:+1:

bwester:
:ship: Just make sure you squash your commits.

PR #228, opened 2015-08-24
Fix buggy unit test

bwester:
The unit tests for the "stream" package failed with output:

    --- FAIL: TestUnsubscribeIndependence-2 (0.00s)
            publisher_test.go:251: sub2 got unexpected sequence: []

The unit test created a StringValuePublisher, subscribed to it. A value was
sent to the publisher, and afterwards the subscription was canceled. It was
expected that "[]" would not be received because the send of the value would
block until the subscriber had received at least the initial value.

After taking a second look at the synchronization, that behavior wasn't
accurate. Even if it was, the test was flaky because it depended on the
implementation details of the publisher rather than its high-level behavioral
contract. The only truly buggy behavior is if the subscription receives a value
that was published after the subscription was canceled.


mpuncel:
:+1:

PR #229, opened 2015-08-24
Rework pkg/replication to be easier for clients to work with.

mpuncel:
1) kp.NewLock() now will auto renew the lock in consul until explicitly
stopped
2) The functionality of Replicator was split into two structs,
Replicator and Replication. Replicator has an InitializeReplication()
function that returns a single-use instance of Replication, which has
the necessary data to perform a replication of a pod to a set of nodes.

### DIFF
--- a/pkg/replication/replication.go
+++ b/pkg/replication/replication.go
@@ -1,61 +1,99 @@
 package replication
 
 import (
+	"sort"
 	"time"
 
-	"github.com/square/p2/Godeps/_workspace/src/github.com/Sirupsen/logrus"
 	"github.com/square/p2/pkg/health"
 	"github.com/square/p2/pkg/health/checker"
 	"github.com/square/p2/pkg/kp"
 	"github.com/square/p2/pkg/logging"
 	"github.com/square/p2/pkg/pods"
-	"github.com/square/p2/pkg/preparer"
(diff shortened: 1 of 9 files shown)


pkg/replication/replication.go
```diff
@@ -81,69 +169,70 @@ func (r Replicator) lock(lock kp.Lock, lockPath string, overrideLock bool) error
 	return err
 }
 
-// Execute the replication.
-// note: error management could use some improvement, errors coming out of
-// updateOne need to be scoped to the node that they came from
-func (r Replicator) Enact(errCh chan<- error, quitCh <-chan struct{}) {
-	nodeQueue := make(chan string)
-	done := make(chan string)
-	innerQuit := make(chan struct{})
```
mpuncel: man this diff came out badly, `Enact()` is virtually unchanged from before

petertseng:
looks like some binary files made it in (at least bin/p2-preparer/p2-preparer , haven't looked at others), could remove those first

petertseng:
I took a quick glance through and didn't see anything else I'd like to change, probably want at least one more to review though. Do remove that binary though

mpuncel:
rebased and also tested this via `p2-replicate`, will merge if CI passes

PR #231, opened 2015-08-26
Mpuncel/immutable-pod-manifests

mpuncel:
The first commit has the meat of the change, the second commit is just the fallout. Will squash before merging because code is broken in between

### DIFF
--- a/pkg/pods/manifest.go
+++ b/pkg/pods/manifest.go
@@ -7,7 +7,6 @@ package pods
 import (
 	"crypto/sha256"
 	"encoding/hex"
-	"fmt"
 	"io"
 	"io/ioutil"
 	"os"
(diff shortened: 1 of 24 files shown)


pkg/pods/manifest.go
```diff
@@ -28,31 +27,132 @@ type LaunchableStanza struct {
...
+	GetConfig() map[interface{}]interface{}
+	SHA() (string, error)
+	OriginalBytes() ([]byte, error)
+	SetID(string) error
+	SetConfig(config map[interface{}]interface{})
+	AddLaunchableStanza(launchableId string, stanza LaunchableStanza)
+	GetStatusPort() int
+	GetStatusHTTP() bool
+	SetRunAsUser(user string) error
+	SetStatusPort(port int)
```
anthonybishopric: should we make `pods.Manifest` immutable and only `pods.UnsignedManifest` have setters?

pkg/pods/manifest.go
```diff
@@ -28,31 +27,132 @@ type LaunchableStanza struct {
...
 }
 
-func ManifestFromPath(path string) (*Manifest, error) {
+func (manifest *unsignedManifest) SetRunAsUser(user string) error {
+	manifest.RunAs = user
+	return nil
+}
+
+func (manifest *signedManifest) SetRunAsUser(user string) error {
+	return util.Errorf("Cannot change RunAsUser for a signed manifest")
```
anthonybishopric: If we make it so that `signedManifest` doesn't even have a `SetRunAsUser` method, we can protect against this error at compile-time, not run-time.

bwester:
Refactoring a struct into a single interface with a lot of getters and setters on its own doesn't make the object any less flaky.  This version is still easy to abuse. I can call `setStatusPort()` on a signed object, and the field will appear changed until I write it.  And I still think the multiple serialization methods, which depend on now-hidden state, are confusing.

I'm in favor of making the basic structure immutable. I'd split it up a little differently than abishopric though. Have a `Manifest` interface with getters, a `MutableManifest` with Manifest and the setters, and a `SignedManifest` with Manifest and the signature-related methods. `Manifest` also needs a new method `Copy() MutableManifest` so that changes can be made.

Making `Manifest` immutable makes it very explicit which methods might alter the original bytes and solves my first complaint.  Having a separate immutable `SignedManifest` can also make it more obvious when a serialization might return a signed version.

anthonybishopric:
:+1: to @bwester's comments

mpuncel:
Okay i've basically redone this PR.

`pods.Manifest` is now an interface with getters only. It also has a method called `GetBuilder()` which returns a `pods.ManifestBuilder` which has setters, as well as a `GetManifest()` function.

mpuncel:
the meat of the changes are still in the 1st commit, the rest is just fallout. Will squash before merging because the build is broken in between.

pkg/pods/manifest.go
```diff
@@ -29,33 +28,140 @@ type LaunchableStanza struct {
 	CgroupConfig            cgroups.Config `yaml:"cgroup,omitempty"`
 }
 
-type Manifest struct {
+// Manifests are mutable when unsigned
```
anthonybishopric: Is this comment correct? It seems like `Manifest`s aren't mutable at all.

pkg/pods/manifest.go
```diff
@@ -29,33 +28,140 @@ type LaunchableStanza struct {
...
 	plaintext []byte
 	signature []byte
 }
 
-func (manifest *Manifest) ID() string {
+func (m *manifest) GetBuilder() ManifestBuilder {
+	builder := manifestBuilder{
+		&manifest{},
+	}
+	*builder.manifest = *m
```
anthonybishopric: Delete the signature data at this step?

pkg/pods/manifest.go
mpuncel: sorry holdout from the original attempt

pkg/pods/manifest.go
```diff
@@ -134,7 +240,10 @@ func (manifest *Manifest) Write(out io.Writer) error {
 // OriginalBytes() always returns the bytes from the original manifest file
 // that was passed to ManifestFromBytes(). Any mutations to the struct will not
 // appear in the returned bytes. If mutations are desired, use Marshal()
-func (manifest *Manifest) OriginalBytes() ([]byte, error) {
+func (manifest *manifest) OriginalBytes() ([]byte, error) {
```
anthonybishopric: Do we need a `Bytes()` vs `OriginalBytes()` anymore?  Seems like we can remove `OriginalBytes` and make `Bytes` return what `OriginalBytes` returned.

pkg/pods/manifest.go
```diff
@@ -29,33 +28,140 @@ type LaunchableStanza struct {
...
 }
 
-func (manifest *Manifest) ID() string {
+func (m *manifest) GetBuilder() ManifestBuilder {
+	builder := manifestBuilder{
+		&manifest{},
+	}
+	*builder.manifest = *m
+	builder.manifest.plaintext = []byte{}
+	builder.manifest.signature = []byte{}
```
anthonybishopric: `manifest.raw`?

pkg/pods/manifest.go
```diff
@@ -131,10 +237,13 @@ func (manifest *Manifest) Write(out io.Writer) error {
 	return nil
 }
 
-// OriginalBytes() always returns the bytes from the original manifest file
-// that was passed to ManifestFromBytes(). Any mutations to the struct will not
-// appear in the returned bytes. If mutations are desired, use Marshal()
-func (manifest *Manifest) OriginalBytes() ([]byte, error) {
+// Returns bytes correlating to the (potentially mutated) struct, unlike
+// Bytes() which guarantees that the bytes returned will be the same bytes the
+// manifest struct was built from
```
anthonybishopric: If we delete `manifest.raw` when getting a builder as well as the signature data, then you don't even need a different method `Marshal` and `Bytes`. Effectively can just have the behavior of `Marshal` be sufficient.

anthonybishopric:
:ship: pending tests

pkg/pods/manifest.go
```diff
@@ -29,33 +28,141 @@ type LaunchableStanza struct {
...
+	return m.manifest
+}
+
+type manifestBuilder struct {
+	*manifest
+}
+
+// Manifest contains the functionality in common between SignedManifest and
+// UnsignedManifest. It is immutable (read-only) and does not expose signature
+// information
```
bwester: Take another look at this documentation.

pkg/pods/manifest.go
```diff
@@ -29,33 +28,141 @@ type LaunchableStanza struct {
...
+func (manifest *manifest) GetLaunchableStanzas() map[string]LaunchableStanza {
+	return manifest.LaunchableStanzas
+}
+
+func (manifest *manifest) AddLaunchableStanza(launchableId string, stanza LaunchableStanza) {
+	if manifest.LaunchableStanzas == nil {
+		manifest.LaunchableStanzas = make(map[string]LaunchableStanza)
+	}
+
+	manifest.LaunchableStanzas[launchableId] = stanza
```
bwester: The builder only gets a shallow copy of the original data structure, so this map reference will alias with the original.  You should probably make a copy of the map in `GetBuilder()`.

Or, change the builder interface to have `SetLaunchables(...)` instead of `Add...`.

bwester:
You might also want to amend the commit title to something more appropriate, once you've squashed all your changes.

mpuncel:
@bwester not sure how to change a branch name without losing PR history, the commit message is correct though

PR #232, opened 2015-08-27
Fix replication bug

bwester:
The GetHealth method is defined as:

    GetHealth(service, node string)

but was being called with (node, service).


anthonybishopric:
:+1:

PR #233, opened 2015-08-31
Only update KV health for status changes

bwester:
Some services include unique output in their status checks, e.g., a current
timestamp, and this was causing the services' KV health checks to be updated on
every check. With this commit, a new health check will be written only when
"Status" changes in a check result. "Output" is still present, but it cannot be
relied upon to be the most recent value.

### DIFF
--- a/pkg/kp/healthmanager.go
+++ b/pkg/kp/healthmanager.go
@@ -334,7 +334,7 @@ func processHealthUpdater(
 // Helper to processHealthUpdater()
 func healthEquiv(x *WatchResult, y *WatchResult) bool {
 	return x == nil && y == nil ||
-		x != nil && y != nil && x.ValueEquiv(*y)
+		x != nil && y != nil && x.Status == y.Status
 }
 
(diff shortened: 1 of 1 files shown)


anthonybishopric:
:+1:

bwester:
The "Output" field might be too useful for debugging to just ignore.

bwester:
The other solution to this problem is PR #237.  Let's pick one and go with it.

bwester:
To be fair, even if we don't ever update when "Output" alone changes, rate limiting could conceivably be useful if there are many services flapping.  But it's not such a pressing concern.

anthonybishopric:
I really like the idea of supporting a "flapping" state in the health checks, now that I've seen it. It's not unusual for an application to be stuck in such a state and could be very valuable in an emergency debugging context. So I'd be in favor of keeping #237's rate limiting + removing output from the status check.

anthonybishopric:
:ship:

PR #235, opened 2015-09-01
Write KV health status for all services

bwester:
In this commit, even services that don't declare any health check will still
have a health check written to the KV with "passing" status. This happens to be
the same behavior that the catalog-based health checks have.

### DIFF
--- a/pkg/watch/health.go
+++ b/pkg/watch/health.go
@@ -156,13 +156,15 @@ func updatePods(
 
 		// if a manifest is in reality but not current a podwatch is created
 		// with that manifest and added to newCurrent
-		if missing && man.Manifest.StatusPort != 0 {
+		if missing {
 			sc := StatusChecker{
 				ID:     man.Manifest.Id,
 				Node:   node,
 				Client: client,
 			}
-			if man.Manifest.StatusHTTP {
+			if man.Manifest.StatusPort == 0 {
+				sc.URI = ""
+			} else if man.Manifest.StatusHTTP {
 				sc.URI = fmt.Sprintf("http://%s:%d/_status", node, man.Manifest.StatusPort)
 			} else {
 				sc.URI = fmt.Sprintf("https://%s:%d/_status", node, man.Manifest.StatusPort)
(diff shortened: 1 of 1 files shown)


pkg/watch/health.go
```diff
@@ -212,7 +214,25 @@ func (p *PodWatch) checkHealth() {
...
+		return sc.resultFromCheck(sc.StatusCheck())
+	} else {
+		// "unknown" is probably more accurate, but automated tools can't handle an app that is
+		// always non-"passing". For instance, p2-replicate by default waits for a node to
+		// become "passing" before it considers the deployment a success.
+		//
+		// TODO: P2 has the capacity to check whether the app's process is running. This would
+		// make a great default status check! However, that information isn't easily accessible
+		// over here in the watch package. It would take a lot of refactoring to make this
+		// happen.
```
anthonybishopric: That would definitely be a good default check.

anthonybishopric: Except that frequently people deploy apps intended to not run at all, even if they have running processes - basically using P2 as a tar delivery mechanism. Perhaps the right solution is to provide a generic runit monitor sidekick that can be added to a pod manifest

pkg/watch/health.go
```diff
@@ -156,13 +156,15 @@ func updatePods(
...
-		if missing && man.Manifest.StatusPort != 0 {
+		if missing {
 			sc := StatusChecker{
 				ID:     man.Manifest.Id,
 				Node:   node,
 				Client: client,
 			}
-			if man.Manifest.StatusHTTP {
+			if man.Manifest.StatusPort == 0 {
+				sc.URI = ""
```
anthonybishopric: Probably A Bad Idea ®: make the default URI an endpoint on the _preparer_ that effectively emits the result of `sv stat` for a given pod ID.

anthonybishopric:
:ship:

pkg/watch/health.go
bwester: I don't even follow what you're suggesting.  But since it's a bad idea, we probably don't want to do it anyway.

anthonybishopric: :+1:

PR #236, opened 2015-09-01
Labels

tummychow:
Big diff due to kubernetes packages. This adds basic support for attaching labels to pods or nodes in consul. The API is not exercised anywhere right now. The labels for an object are stored as a single JSON document under `/labels/node/<nodename>` and `/labels/pod/<podname>`. We'll still have to decide what exactly a "pod name" is in this context, but slash-separated strings will probably work.


tummychow:
ugh, CI failure because a linux-only dependency wasn't `godep save`d. will fix shortly

tummychow:
- merged linux-only dependencies into first commit to avoid horrible ugly giant diffs
- changed `POD` and `NODE` to be constants
- used enums for label operators instead of strings

will implement auto retry logic shortly plus tests

tummychow:
- implemented optional auto-retry after CAS failure
- merged all implementation commits into one

anthonybishopric:
I think I'm satisfied :+1:

tummychow:
- removed redundant nil check in convertkvptolabeled

tummychow:
- added type argument to GetMatches

petertseng:
Yup :+1:

PR #237, opened 2015-09-01
Limit output

bwester:
A couple patches to reduce the Consul write rate for troublesome services.

### DIFF
--- a/pkg/kp/healthmanager.go
+++ b/pkg/kp/healthmanager.go
@@ -92,7 +92,7 @@ var (
 	HealthRetryTimeSec = param.Int("health_retry_time_sec", 5)
 
 	// SessionTTLSec sets the TTL time for each session created by consulHealthManager. This
-	// parmeter controls how long it takes for clients to notice that health checks have
+	// parameter controls how long it takes for clients to notice that health checks have
 	// stopped.
 	SessionTTLSec = param.Int("health_session_ttl_sec", 15)
 )
--- a/pkg/kp/healthmanager.go
+++ b/pkg/kp/healthmanager.go
@@ -9,7 +9,9 @@ import (
 	"github.com/square/p2/Godeps/_workspace/src/github.com/Sirupsen/logrus"
 	"github.com/square/p2/Godeps/_workspace/src/github.com/hashicorp/consul/api"
 
+	"github.com/square/p2/pkg/health"
 	"github.com/square/p2/pkg/logging"
+	"github.com/square/p2/pkg/util/limit"
 	"github.com/square/p2/pkg/util/param"
 	"github.com/square/p2/pkg/util/stream"
 )
@@ -95,6 +97,20 @@ var (
 	// parameter controls how long it takes for clients to notice that health checks have
 	// stopped.
 	SessionTTLSec = param.Int("health_session_ttl_sec", 15)
+
+	// HealthWritesPerMinute sets the average number of writes per minute per service that
+	// will be sent to Consul to update health.
+	HealthWritesPerMinute = param.Int("health_writes_per_minute", 4)
+
+	// HealthMaxBucketSize sets the maximum token bucket size per service used to
+	// rate-limit Consul writes.
+	HealthMaxBucketSize = param.Int64("health_max_bucket_size", 16)
+
+	// HealthResumeLimit sets the lower bound on the number of tokens at which updates will
+	// be resumed. If a service runs out of tokens, its last update will be to set the
+	// health status to "unknown" with an error message, and further updates will be
+	// throttled until enough tokens have been accumulated.
+	HealthResumeLimit = param.Int64("health_resume_limit", 4)
 )
 
 // consulHealthManager maintains a Consul session for all the local node's health checks,
@@ -205,8 +221,9 @@ func (u *consulHealthUpdater) Close() {
 }
 
 type writeResult struct {
-	Health *WatchResult // The health that was just written
-	OK     bool         // Whether the write succeeded
+	Health   *WatchResult // The health that was just written
+	OK       bool         // Whether the write succeeded
+	Throttle bool         // Whether to turn on the write throttle on success
 }
 
 // processHealthUpdater() runs in a goroutine to keep Consul in sync with the local health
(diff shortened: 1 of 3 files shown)


bwester:
I think I want to boos those rate limiting constants into configuration options, so expect a subsequent revision. But the rest of this is ready to be reviewed.

pkg/kp/healthmanager.go
```diff
@@ -94,6 +95,24 @@ const (
 	// parmeter controls how long it takes for clients to notice that health checks have
 	// stopped.
 	SessionTTL = 15 * time.Second
+
+	// MaxWritesPerMinute sets the average number of writes per minute per service
+	// that will be sent to Consul to update health.
+	MaxWritesPerMinute = 1
```
anthonybishopric: For the purposes of replication, this sets a lower bound on how long a deploy can take, since you are required to wait at least 60 seconds from the time a node goes unhealthy (from the old version being stopped) to the new version being brought up. I'd like to raise this number as much as we can tolerate - is 4 okay?

pkg/kp/healthmanager.go
```diff
@@ -269,7 +300,14 @@ func processHealthUpdater(
 			write = nil
 			if result.OK {
 				remoteHealth = result.Health
+				if result.Throttle && throttle == nil {
```
anthonybishopric: This function has gotten quite large - what do you think about moving this logic to a `LeakyBucket{}`?

pkg/kp/healthmanager.go
```diff
@@ -332,9 +385,29 @@ func processHealthUpdater(
...
+	}
+	return x.ValueEquiv(*y)
+}
+
+func toThrottled(wr *WatchResult) *WatchResult {
+	return &WatchResult{
+		Node:    wr.Node,
+		Id:      wr.Id,
+		Service: wr.Service,
+		Status:  "throttled",
```
anthonybishopric: Please make sure this is enumerated with the other health states. That said, I really like the idea of a state just intended to represent service flapping.

pkg/kp/healthmanager.go
bwester: Maybe. The uncertainty here is why I want to make it configurable.  We can see what happens.

But I don't mind making the built-in default 4.

pkg/kp/healthmanager.go
bwester: I don't quite understand your reference to the leaky bucket. I don't think that rate control algorithm applies to this controller.  But I agree that the logic here has gotten a bit unwieldy, and there could be a better way to express the state transitions and actions.

pkg/kp/healthmanager.go
bwester: np

pkg/kp/healthmanager.go
anthonybishopric: My mistake, I do see there is a token bucket implementation - didn't fully grok that tokens were being added to the bucket, not removed, over time.

pkg/kp/kv_test.go
```diff
@@ -49,3 +51,57 @@ func TestGetHealthWithEntry(t *testing.T) {
 		t.Fatalf("watchRes and watch Service did not match. GetHealth failed: %#v", watchRes)
 	}
 }
+
+func jsonMustMarshal(value interface{}) string {
+	encoded, err := json.Marshal(value)
+	if err != nil {
+		panic(err)
```
anthonybishopric: Perhaps `t.Fatal` would be more appropriate in a test context?

pkg/kp/kv.go
```diff
@@ -79,14 +81,29 @@ type WatchResult struct {
 	Expires time.Time `json:"Expires,omitempty"`
 }
 
+func normalizeOutput(in string) interface{} {
+	var obj map[string]interface{}
+	if json.Unmarshal([]byte(in), &obj) != nil {
+		return in
+	}
+	for key := range obj {
+		if strings.ToLower(key) == "time" {
```
anthonybishopric: Do we really need this if we also have the rate limiting algorithm?

bwester: For a subset of hosts in our infrastructure, this normalization will allow the service to have an average write rate of 0.  With just the rate limiter, they will write once every 15 seconds.

With the way our infrastructure has been behaving, I anticipate that it's worthwhile to reduce the write workload whenever possible.  But is it _necessary_?  We'll see what happens when I roll out this change ;)

anthonybishopric: I'm guessing we can kill any output normalization if we retain #233

pkg/kp/config.go
```diff
@@ -7,6 +7,45 @@ import (
...
+
+// DefaultHealthOptions contains the default health options
+var DefaultHealthOptions = HealthOptions{
+	HealthRetrySeconds: 5,
+	SessionTTLSeconds:  15,
+	WritesPerMinute:    4,
+	BucketSize:         60,
+	OutputLimit:        30,
+	ResumeLimit:        8,
+}
```
anthonybishopric: :star:

pkg/health/checker/checker.go
```diff
@@ -0,0 +1,202 @@
+package checker
```
anthonybishopric: Anything in this file that should be reviewed?

bwester: It's a cherry-pick of PR #238. Review comments should go there. (Preview: I only moved around a bunch of functions.)

pkg/kp/kv.go
bwester: Agreed

anthonybishopric:
what's the status of this PR?

bwester:
I think this PR is still relevant, and I've updated it to account for various changes made over the last several weeks.

pkg/kp/healthmanager.go
```diff
@@ -95,6 +97,20 @@ var (
 	// parmeter controls how long it takes for clients to notice that health checks have
```
mpuncel: may as well fix this typo

bwester: Sure, I'll fix the typo.

anthonybishopric:
:+1: pending test flakes.

bwester:
I keep forgetting that this PR is still open...

PR #238, opened 2015-09-02
Refactor pkg/health

bwester:
Splits the ConsulHealthChecker and related functions into its own package
"pkg/health/checker" out of "pkg/health".  This resolves a circular import
problem, wherein the "pkg/kp" package couldn't import "pkg/health" to access
the canonical health values.

This is a pure refactor. Code is copied around and packages & methods are renamed. That's it.


anthonybishopric:
:+1:

PR #242, opened 2015-09-04
Kill the preparer when its status server dies

bwester:
If the preparer is unable to correctly report its health, it isn't functioning
properly. This commit causes the whole preparer to die when its HTTP status
server dies, so that the preparer has a chance of fixing itself when it starts
up again.

### DIFF
--- a/bin/p2-preparer/main.go
+++ b/bin/p2-preparer/main.go
@@ -59,7 +59,10 @@ func main() {
 			func(w http.ResponseWriter, r *http.Request) {
 				fmt.Fprintf(w, "p2-preparer OK")
 			})
-		go http.ListenAndServe(fmt.Sprintf(":%d", preparerConfig.StatusPort), nil)
+		go func() {
+			err := http.ListenAndServe(fmt.Sprintf(":%d", preparerConfig.StatusPort), nil)
+			logger.WithError(err).Fatalln("status server exited")
+		}()
 	}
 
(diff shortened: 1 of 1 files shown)


bin/p2-preparer/main.go
```diff
@@ -59,7 +59,10 @@ func main() {
 			func(w http.ResponseWriter, r *http.Request) {
 				fmt.Fprintf(w, "p2-preparer OK")
 			})
-		go http.ListenAndServe(fmt.Sprintf(":%d", preparerConfig.StatusPort), nil)
+		go func() {
+			err := http.ListenAndServe(fmt.Sprintf(":%d", preparerConfig.StatusPort), nil)
+			logger.WithError(err).Fatalln("status server exited")
```
anthonybishopric: When does this happen?

anthonybishopric: The http server crashing seems unrelated from its own status check and ability to send health information

bwester: Practically speaking, once the server gets going, it should never exit.  I'm more concerned with it never getting started, if, for instance, the port it's binding to is already in use.

In either case, this preparer will forevermore be considered "critical".

Interestingly, the server exiting is very unrelated from its ability to check apps' statuses, and the preparer will gladly continue to post that its own health is "critical."

bwester:
Ping?  This still PR seems valuable.

anthonybishopric:
:ship:

PR #243, opened 2015-09-04
Remove OS X firewall message

bwester:
Changes the status handler in the "pkg/watch" unit test to only bind to the
"localhost" address. This change prevents the Mac OS X firewall from prompting
me "Do you want the application 'watch.test' to accept incoming network
connections?" every time I ran unit tests.


anthonybishopric:
:+1:

PR #244, opened 2015-09-04
Increase headroom for cgroup memory limits

bwester:
An extra 10% appears to be too restrictive in practice. It would be nice to
make this configurable, but that can come later.


anthonybishopric:
:+1: for now

PR #245, opened 2015-09-09
Abishopric/process monitor

anthonybishopric:
WIP - this is a possible sidekick process for services that simply need to know if runit believes their process has been up for enough time.


tummychow:
looks fine

PR #246, opened 2015-09-09
Absorb kubernetes label dependencies

tummychow:
As we discussed. I pulled in as few files as possible to make things work without API or functional changes. The last commit drops the godeps that I added in my previous PR.


anthonybishopric:
Please link to the PR against k8s, otherwise :+1:

tummychow:
it'll take multiple PRs to k8s before we can use it again, but the first one is here: https://github.com/kubernetes/kubernetes/pull/13754

PR #247, opened 2015-09-11
Add replication controller interface, implementation, and tests

petertseng:
Includes an rcstore (replication controller store) for storing replication controllers. Currently rcstore only has a fake implementation (in memory, for tests) - a Consul-backed implementation will come in a later PR.

Replication Controller functionality is complete and includes tests.


pkg/rc/replication_controller.go
```diff
@@ -0,0 +1,40 @@
...
+
+type ReplicationController interface {
+	// It is expected that the below functions will be used by the orchestrator,
+	// in response to user-requested deploy events.
+	// They affect the desired behavior of the replication controller.
+
+	// IncreaseReplicas(n) increases the number of desired replicas by n.
+	// This returns the new desired number if successful, or the old number and an error.
+	// It is expected that the replication controller will attempt to schedule its pod onto a node returned by EligibleNodes().
+	IncreaseReplicas(n int) error
```
robertabbott: should these be *creaseReplicas(n int) (int, error)?

petertseng: ah that would be consistent with my documentation comment...

pkg/rc/replication_controller.go
```diff
@@ -0,0 +1,40 @@
...
+	DecreaseReplicas(n int) (int, error)
+
+	// Disable() causes the replication controller to stop reacting to changes in the number of desired replicas.
+	Disable() error
+
+	// It is expected that the below functions are used by the replication controller
+	// to determine whether observed state matches desired state.
+
+	// EligibleNodes() returns the nodes this replication controller could possibly run on.
+	EligibleNodes() ([]string, error)
```
anthonybishopric: This doesn't seem like a thing you'd want to ask the RC for directly. I'd expect an RC to have a `LabelSelector` which can be used to select nodes.

pkg/rc/replication_controller.go
```diff
@@ -0,0 +1,40 @@
...
+
+	// CurrentNodes() returns the nodes this replication controller is currently running on.
+	CurrentNodes() ([]string, error)
```
anthonybishopric: Same sort of thing here, I'd expect this to be a label selector for pods.

pkg/rc/replication_controller.go
petertseng: yes, the RC I have in mind internally has a LabelSelector and uses it when this function is called

pkg/rc/replication_controller.go
```diff
@@ -0,0 +1,40 @@
...
+
+	// DesiredReplicas() returns the number of desired replicas for this replication controller.
+	DesiredReplicas() (int, error)
+
+	// It is expected that the below functions are used by the replication controller
+	// to bring observed state in line with desired state.
+
+	// Schedule(node) schedules the pod managed by this replication controller onto the specified node
+	Schedule(node string) error
```
anthonybishopric: These two methods seem incompatible with label selectors.

petertseng: that's true but I'd expect replication controllers to only call this with a node that is in the `Eligible - Current` set. It may be the case that this will be non-exported. I'll see as I continue to write more implementation code.

bin/p2-rc/main.go
```diff
@@ -0,0 +1,110 @@
...
+	"github.com/square/p2/pkg/pods"
+	"github.com/square/p2/pkg/rc"
+	"github.com/square/p2/pkg/util/net"
+	"github.com/square/p2/pkg/version"
+)
+
+var (
+	replicate = kingpin.New("p2-rc", `p2-rc uses the rc package to schedule deployment of a pod across multiple nodes. See the rc package's README and godoc for more information.
+
+	Example invocation: p2-rc --min-nodes 2 helloworld.yaml aws{1,2,3}.example.com
```
petertseng: this is totally not correct. fix this j0nx

bin/p2-rc/main.go
```diff
@@ -0,0 +1,110 @@
...
+		nodeSelector,
+		kp.NewFromClient(consulClient),
+		nodeApplicator,
+		podApplicator,
+	)
+	if err != nil {
+		log.Fatalf("Couldn't create RC: %s", err)
+	}
+
+	_, err = replicationController.IncreaseReplicas(1)
```
petertseng: any way to make this user-requestable?

pkg/rc/consul_backed.go
```diff
@@ -0,0 +1,133 @@
...
+
+type kvPair struct {
+	key   string
+	value []byte
+}
+
+// consulRc implements the ReplicationController interface.
+var _ ReplicationController = &consulRc{}
+
+func NewConsulBacked(consulClient *api.Client, innerRc ReplicationController) (*consulRc, error) {
```
petertseng: come on, document this

petertseng: resume from consul saved state as well please

pkg/rc/memory_backed.go
```diff
@@ -0,0 +1,231 @@
...
+	nodeSelector    labels.Selector
+	manifest        *pods.Manifest
+	replicasDesired int
+	disabled        bool
+}
+
+// memoryRc implements the ReplicationController interface.
+var _ ReplicationController = &memoryRc{}
+
+func NewMemoryBacked(
```
petertseng: highly likely that it will be beneficial to document the implications of this being in memory only (not persisted, etc.)

petertseng:
Consider adding in some sort of additional filtering beyond label selectors. For example, we may want to spread load around. should have some sort of `Scheduler(Pod, LabelSelector) -> nodes` interface that is used by `EligibleNodes` to determine what nodes are eligible.

For now the `Scheduler` can just pass-through whatever the LabelSelector returns, but this will allow for future expansions.

petertseng:
Thoughts on labels: We'll want to add `pod_id` at least (for a future world where it is not obvious from the key structure)

In the case where there are mutliple RCs deploying the same manifest, we want a way to tell them apart. So the RC should be able to take a set of selectors that 1) will be applied to pods it deploys and 2) are used to select pods so that the RC knows what node it's running on.

pkg/kp/kv.go
```diff
@@ -111,6 +111,16 @@ func NewConsulStore(opts Options) Store {
 	}
 }
 
+func NewFromClient(client *api.Client) Store {
```
anthonybishopric: Just use `NewConsulStore` and pass an empty `Options{}`

pkg/rc/consul_backed.go
```diff
@@ -0,0 +1,133 @@
...
+	rc := &consulRc{
+		innerRc: innerRc,
+		kv:      consulClient.KV(),
+		kpStore: kp.NewFromClient(consulClient),
+	}
+
+	// Store everything related to the RC into consul, bail if errored.
+	_, err := rc.kpStore.SetPod(rc.consulPrefix()+"/pod_manifest", *innerRc.Manifest())
```
anthonybishopric: `"/pod_manifest"` should probably be relegated to somewhere in `kp`. Same with the paths listed below.

anthonybishopric: Or just enumerated in this pkg since this owns RCs. One thing I like about the `pods` package is that there are no references to Consul inside it - `kp` maps Consul values onto `pod` objects

pkg/rc/consul_backed.go
```diff
@@ -0,0 +1,133 @@
...
+func (rc *consulRc) mutateDesired(n int) error {
+	return rc.consulPutOne("replicas_desired", []byte(strconv.Itoa(n)))
+}
+
+func (rc *consulRc) DesiredReplicas() int {
+	return rc.innerRc.DesiredReplicas()
+}
+
+func (rc *consulRc) Disable() error {
+	err := rc.consulPutOne("disabled", []byte("true"))
```
anthonybishopric: A number of these methods suffer from split brain risks I think. Right now, if a `consulPutOne` succeeded but the `innerRc` failed to `Disable()` you'd end up not really understanding the true state of the cluster.

anthonybishopric: I'd suggest possibly doing the Consul operations optimistically but undoing them if the `innerRc` fails. What do you think

pkg/rc/consul_backed.go
```diff
@@ -0,0 +1,133 @@
...
+		return err
+	}
+	return rc.innerRc.Disable()
+}
+
+func (rc *consulRc) Meet() error {
+	return rc.innerRc.Meet()
+}
+
+func (rc *consulRc) MeetEvery(every <-chan interface{}) chan<- error {
```
anthonybishopric: I think I'm a bit confused about the purpose of the `consulRc`. It looks like you would use it to send instructions to the RC to update things, ie the replica count. But then it immediately goes on to do the update itself (via `innerRc`), implying that this is the server code. I would have imagined the `consulRc` basically doing what I see `memoryRc` doing but managed by some parent goroutine that is holding a watch on the rc's consul endpoint. How did you envision clients sending updates to Consul for RC changes?

pkg/rc/memory_backed.go
```diff
@@ -0,0 +1,231 @@
...
+func (rc *memoryRc) DesiredReplicas() int {
+	return rc.replicasDesired
+}
+
+func (rc *memoryRc) Disable() error {
+	rc.disabled = true
+	return nil
+}
+
+func (rc *memoryRc) Meet() error {
```
anthonybishopric: This `Meet()` is analogous to the watch loop in the preparer - it seems desirable to want to recreate that same pattern of watching + responding.

pkg/rc/memory_backed.go
```diff
@@ -0,0 +1,231 @@
...
+		nodeSelector:    nodeSelector,
+		replicasDesired: 0,
+		disabled:        false,
+	}
+
+	return rc, nil
+}
+
+func (rc *memoryRc) ID() string {
+	return rc.manifestSha
```
petertseng: consider a uuid instead

pkg/kp/kv.go
petertseng: I'm super unsure about that, since I do want to pass a consul client that has a certain httpClient. Will revisit after I'm done with refactor though.

petertseng:
Shortened this PR to just be the necessary interfaces.

pkg/kp/rc.go
```diff
@@ -0,0 +1,14 @@
+package kp
+
+import (
+	"github.com/square/p2/pkg/labels"
+	"github.com/square/p2/pkg/pods"
+)
+
+type ReplicationControllerStore interface {
+	Create(manifest pods.Manifest, nodeSelector, podSelector labels.Selector) (string, error)
```
petertseng: someone is going to mix up the order of the arguments `nodeSelector` and `podSelector` and it is going to suck. Any way to make better?

anthonybishopric: Shouldn't there also be a `podLabels` set? You need to know how to label the pod as well as select it.

Probably worth documenting the use of each in the interface as well.

pkg/kp/rc.go
```diff
@@ -0,0 +1,22 @@
...
+)
+
+type ReplicationControllerStore interface {
+	// Create creates a replication controller with the specified manifest and selectors.
+	// The node selector is used to determine what nodes the replication controller may schedule on.
+	// The pod selector is:
+	// 1) applied to every pod the replication controller schedules,
+	// 2) used (in addition to replication controller ID) to determine what pods are already scheduled.
+	// Returns the ID of the replication controller just created, and an error if any.
+	Create(manifest pods.Manifest, nodeSelector, podSelector labels.Selector) (string, error)
```
petertseng: someone is going to mix up the order of `nodeSelector` and `podSelector` and it is going to suck. Any way to make better?

pkg/kp/rc.go
petertseng: oh I think I did mean labels, not label selector. OK that will differentiate them well enough.

petertseng:
`podLabels` now, not `podSelector`

pkg/rc/replication_controller.go
```diff
@@ -0,0 +1,21 @@
...
+	"github.com/square/p2/pkg/pods"
+)
+
+type ReplicationController interface {
+	ID() string
+	Manifest() *pods.Manifest
+	NodeSelector() labels.Selector
+	Disabled() bool
+	DesiredReplicas() int
+
```
petertseng: what else does the orchestrator need?

* `PodLabels()`? I say no.
* `CurrentNodes()`? I say yes.
* `EligibleNodes()`? I say no.

pkg/kp/rc.go
petertseng: > You need to know how to label the pod as well as select it.

I actually made it so that it's the label set (so the replication controller knows how to label the pod), and I expect the replication controller to make the selector based on that label set.

bwester:
Without seeing documentation, an implementation, or a use case, it's difficult for me to understand why this interface is the way it is.

pkg/kp/rc.go
```diff
@@ -0,0 +1,22 @@
...
+)
+
+type ReplicationControllerStore interface {
+	// Create creates a replication controller with the specified manifest and selectors.
+	// The node selector is used to determine what nodes the replication controller may schedule on.
+	// The pod label set is:
+	// 1) applied to every pod the replication controller schedules,
+	// 2) used (in addition to replication controller ID) to determine what pods are already scheduled.
+	// Returns the ID of the replication controller just created, and an error if any.
+	Create(manifest pods.Manifest, nodeSelector labels.Selector, podLabels labels.Set) (string, error)
```
bwester: > the ID of the replication controller just created.

What is a "replication controller" in this context?

pkg/kp/rc.go
```diff
@@ -0,0 +1,22 @@
...
+	ReplicationController(id string) rc.ReplicationController
```
bwester: Are these `id string` parameters related to the string that's returned by `Create()`?  Why not have `Create()` return an object on which these methods can be called?

pkg/rc/replication_controller.go
```diff
@@ -0,0 +1,22 @@
...
+type ReplicationController interface {
+	ID() string
+	Manifest() *pods.Manifest
+	NodeSelector() labels.Selector
+	Disabled() bool
+	DesiredReplicas() int
+	CurrentNodes() []string
+
+	// WatchDesires() spawns a goroutine that continually watches desired state
```
bwester: What is the desired state?

pkg/kp/rc.go
```diff
@@ -0,0 +1,22 @@
...
+	Delete(id string) error
+	Increase(id string, n int) (int, error)
```
bwester: What are we increasing?  Could we instead have a `Set()` method that doesn't require blindly adjusting some hidden value?

pkg/kp/rc.go
bwester: `ReplicationControllerStore` is an interface, so there could conceivably be more than one implementation of this interface.  Are these ids scoped to the implementation that generated them?  Do they have a generic meaning in some context?

petertseng: That's a good idea, I made it `ReplicationControllerHandle`

pkg/rc/replication_controller.go
petertseng: yup, clarified that it's desired replicas

pkg/kp/rc.go
petertseng: yeah that's better, jsut made it `SetDesiredReplicas`

petertseng:
The intended use of this is:

The Orchestrator service receives a request:

* deploy a manifest
* to N hosts
* the hosts being subject to some label selector,
* and apply these pod labels to the pods you deploy

The Orchestrator calls `Create(manifest, labelSelector, podLabels)` on a consul-backed implementation of `ReplicationControllerStore`. This stores the request into Consul, as a replication controller with some default initial state.

Separately, a service which I'll call the replication controller manager (not necessarily what it will be called -- really it could just be a separate goroutine in the Orchestrator service) has been watching Consul for new replication controllers. When it sees that a new replication controller has been created, it creates a new `ReplicationController` from the information in Consul and calls `WatchDesires` on it. This spawns a goroutine that watches Consul for any changes to the number of desired replicas for this replication controller.

The Orchestrator calls `SetDesiredReplicas(1)` on its `ReplicationControllerHandle`, setting the desired replica count to 1 in Consul.

The `WatchDesires` goroutine sees the change, and schedules the manifest onto an eligible node, labeling it with the desired labels.

This process repeats until the manifest is scheduled onto N nodes.

Another case to handle is if the replication controller manager crashes, when it comes back up it should query Consul for all currently active replication controllers. It creates `ReplicationController`s for each and calls `WatchDesires` for each.

petertseng:
A thought about the above use case: When the orchestrator received the request of "deploy this manifest to N hosts`, N is not getting persisted anywhere. If the orchestrator crashes, as written, the above process does not resume any rollouts - it just resumes any replication controllers. Should we allow for that?

pkg/kp/rc.go
```diff
@@ -0,0 +1,30 @@
...
+type ReplicationControllerStore interface {
+	// Create creates a replication controller with the specified manifest and selectors.
+	// The node selector is used to determine what nodes the replication controller may schedule on.
+	// The pod label set is:
+	// 1) applied to every pod the replication controller schedules,
+	// 2) used (in addition to replication controller ID) to determine what pods are already scheduled.
+	Create(manifest pods.Manifest, nodeSelector labels.Selector, podLabels labels.Set) (ReplicationControllerHandle, error)
+
+	// List lists all of the replication controllers.
+	List() ([]ReplicationControllerHandle, error)
```
petertseng: actually I might only care about *active* replication controllers

petertseng: and should there be a method that queries for a *specific* replication controller, taking in some ID?

petertseng: most definitely needs `Watch` to support the uses I described.

bwester: What makes a controller "active"?

---

I assume that the data in Consul will allow:
1. Querying for the status of a deployment as it's happening.
2. Resuming a deployment if one node happens to crash.

For (1), it seems reasonable to want status from a specific deployment, but I wouldn't bother with it until we have direct use case (e.g., until said UI is built).  For (1) and (2), you need discovery, and `List()` is probably sufficient for a first pass.

But also, to support (2) in the presence of many orchestrator instances, do we need some kind of locking on each request?

bwester:
It seems odd to call the data in Consul a "ReplicationController". I usually think of a "controller" as something active that manipulates state, not the state itself.  Could you call it a `ReplicationRequest`?

pkg/rc/replication_controller.go
```diff
@@ -0,0 +1,231 @@
...
+
+func (rc *replicationController) WatchDesires(quitAndAck chan struct{}) <-chan error {
+	/*
+		!!! TODO: This API now feels a bit strange to me.
+		The only reason a replicationController cares about its rcStore is so that it can call Watch on it.
+		HOWEVER, wouldn't it be smarter to have ReplicationControllerData provide a Watch() function???
+		That way, replicationControllers don't actually care how they're stored!
+		Then the Watch() function can be removed from ReplicationControllerStore.
+	*/
+	desiresChannel, errChannel := rc.rcStore.Watch(rc.ID(), quitAndAck)
```
petertseng: Throwing comment here so I can easily see what this used to look like. I wonder if it'll catch the entire comment.

petertseng:
There we go, ready to review again. This time an implementation and test have made it in.

consulRcStore not included here (not done yet nor necessary for test), I will probably just put it in a separate PR

pkg/kp/rcstore/rc.go
```diff
@@ -0,0 +1,95 @@
...
+	// List lists all of the replication controllers.
+	List() ([]ReplicationRequest, error)
+}
+
+type desireWatcher interface {
+	// watch(quitAndAck) watches for any changes to the ReplicationControllerDesires associated with the watcher.
+	// watch() returns a channel on which new ReplicationControllerDesires are sent,
+	// and a channel on which errors are reported.
+	// Send a value on quitAndAck to stop watching.
+	// watch() signals it has stopped by sending back a value on quitAndAck.
```
petertseng: It may not be necessary to send back a value on the quit channel to confirm that it has quit. Closing its output channels could be sufficient. I have no specific reason for sending back an ack on the quit channel other than that preparers do it too, but they probably had specific reasons for needing it whereas I... don't think I do.

pkg/rc/replication_controller.go
```diff
@@ -0,0 +1,239 @@
...
+	PodLabels() labels.Set
+	Disabled() bool
+	DesiredReplicas() int
+	CurrentNodes() ([]string, error)
+
+	// WatchDesires() spawns a goroutine that continually watches the desired replica count
+	// of this replication controller.
+	// Upon seeing any changes, it will converge observed state to desired state.
+	// The goroutine continues running until a value is received on the quitAndAck channel.
+	// The goroutine signals it has stopped by sending back a value on quitAndAck.
```
petertseng: same comment as before - closing the error channel would probably be sufficient to ack the quit request.

pkg/kp/rcstore/rc.go
```diff
@@ -0,0 +1,95 @@
...
+	// It is expected that these fields will never be changed after an RC is created.
+	id           string
+	manifest     *pods.Manifest
+	nodeSelector labels.Selector
+	podLabels    labels.Set
+
+	watcher desireWatcher
+
+	// It is expected that these fields can be changed.
+	ReplicationControllerDesires
```
anthonybishopric: Why distinguish this from the other fields on this struct?

anthonybishopric: I would argue that mutability vs immutability isn't a great reason to separate them.

pkg/kp/rcstore/rc.go
```diff
@@ -0,0 +1,95 @@
...
+	// The pod label set is:
+	// 1) applied to every pod the replication controller schedules,
+	// 2) used (in addition to replication controller ID) to determine what pods are already scheduled.
+	Create(manifest pods.Manifest, nodeSelector labels.Selector, podLabels labels.Set) (ReplicationRequest, error)
+
+	// List lists all of the replication controllers.
+	List() ([]ReplicationRequest, error)
+}
+
+type desireWatcher interface {
```
anthonybishopric: Is `desire` intentionally distinct from `intent`? Any reason to not just make it `intent`?

anthonybishopric: This is a naming question, not a storage / API question

pkg/kp/rcstore/rc.go
petertseng: That's true. I did this so that the internal `watch` function on `desireWatcher` would only need to return those fields, since those are the only ones expected to change. Now that I think about it though, I could just make the `desireWatcher` modify in place instead of returning the fields. In that case `ReplicationControllerDesires` could just go away.

I may even structure the console keys in a way such that these keys live under e.g. `replication_controllers/some_id_here/desires/` and I just watch that instead of `replication_controllers/some_id_here`.

I've also realized `ReplicationControllerDesires` isn't even referenced anywhere outside the `rcstore` package (everything else is using exported methods on `ReplicationControllerData`), which means I could just not export it.

anthonybishopric: I honestly don't even think we know enough about the design yet to want to split out fields that could change from those that can't. I'm not 100% convinced that, say, `nodeSelector` is immutable. I don't think it makes sense to couple that requirement to its layout in Consul, since that implies the need for data migrations if they do change.

I'd be more comfortable making fewer assumptions at this stage.

pkg/kp/rcstore/rc.go
petertseng: It could very well be `intent`.  we'd have to be careful not to confuse replication controller intent with pod intent though. Is the following sentence confusing to you?

"Replication controllers try to bring their reality in line with their intent by scheduling manifests into the intent store"

pkg/kp/rcstore/rc.go
```diff
@@ -0,0 +1,95 @@
...
+	watch(quitAndAck chan struct{}) (<-chan ReplicationControllerDesires, <-chan error)
+}
+
+// Methods called on instances of ReplicationRequest are able to affect underlying data persisted in consul.
+type ReplicationRequest interface {
+	Data() ReplicationControllerData
+	Delete() error
+	SetDesiredReplicas(n int) error
+	Disable() error
+}
```
anthonybishopric: This request is also itself capable of performing operations like `Delete`? Seems like two things in one.

petertseng: Perhaps `ReplicationControllerHandle` is a better name? It used to be that but then @bwester  thought https://github.com/square/p2/pull/247#issuecomment-142730556

pkg/rc/replication_controller.go
```diff
@@ -0,0 +1,239 @@
...
+	WatchDesires(quitAndAck chan struct{}) <-chan error
+}
+
+// I just copied this from kp.Store because it was a huge interface and I really only care about this one method!
```
anthonybishopric: s/one/these/

pkg/kp/rcstore/rc.go
petertseng: Sure, I can assume that any fields can change except for ID. I think for some internal things to work, I still need to have a separate struct called `rcFields` (so that the watcher can still send updates), but I'll no longer export it.

pkg/kp/rcstore/rc.go
petertseng: this question is now moot, I'm renaming it to `rcWatcher` since we're not explicitly separating fields that can change from those that can't.

pkg/kp/rcstore/rc.go
petertseng: done.

pkg/rc/replication_controller.go
petertseng: got it

pkg/rc/replication_controller.go
anthonybishopric: :+1:

pkg/rc/replication_controller.go
```diff
@@ -0,0 +1,239 @@
...
+	kpStore       kpStore
+	scheduler     Scheduler
+	podApplicator labels.Applicator
+}
+
+// replicationController implements the ReplicationController interface.
+var _ ReplicationController = &replicationController{}
+
+func New(
+	rcData rcstore.ReplicationControllerData,
```
anthonybishopric: It seems backwards to have `pkg/rc` require something from `pkg/kp/rcstore`, since the rc package is what should define the fields an rc needs. The `pods` package is the opposite of this.

pkg/rc/replication_controller.go
```diff
@@ -0,0 +1,258 @@
...
+	rc.logger.NoFields().Infof("Eligible to be on nodes %s", eligible)
+
+	current, err := rc.CurrentNodes()
+	if err != nil {
+		return err
+	}
+	rc.logger.NoFields().Infof("Currently on nodes %s", current)
+
+	if rc.DesiredReplicas() > len(current) {
+		possible := difference(eligible, current)
```
anthonybishopric: I'd leave some kind of scheduler to do here. This code currently assumes that the label selector - existing nodes currently used by the pod is a sufficient decision making process for whether an app should run. Note that Docker/runc will be free of the constraint of not running more than once on a node. I'd like to acknowledge that in this code.

petertseng: ok, did it. let me know if I should include more in the todo

pkg/rc/replication_controller.go
petertseng: I can probably make it happen. Need some thought on how to avoid a circular dependency: `pkg/rc` needs to know about `rcstore` so it can perform watches, and `rcstore` needs to be provide enough information about replication controllers in order to be able to recreate them from the store. That implies that the *data* of an RC needs to live in a separate package from this `pkg/rc` (which is mainly in the business of allowing RCs to meet their desires)

anthonybishopric:
Good first start, happy to merge this and then resolve some of the practical considerations over the next week.

petertseng:
Refactor complete, let me know if you like this or the previous better (which I'm keeping around at https://github.com/petertseng/p2/commits/rc-old )

pkg/rc/fields/fields.go
```diff
@@ -0,0 +1,18 @@
+package fields
+
+import (
+	"github.com/square/p2/pkg/labels"
+	"github.com/square/p2/pkg/pods"
+)
+
+type RC struct {
+	// It is assumed that the ID field never changes for a given replication controller object.
```
petertseng: Why is this important? What code relies on this assumption? Clarify.

pkg/rc/replication_controller.go
```diff
@@ -0,0 +1,252 @@
...
+type replicationController struct {
+	fields.RC
+
+	manifest        pods.Manifest
+	nodeSelector    labels.Selector
+	podLabels       labels.Set
+	replicasDesired int
+	disabled        bool
+
+	// All other fields could possibly change.
```
petertseng: stray leftover comment. Can remove.

pkg/rc/replication_controller.go
```diff
@@ -0,0 +1,252 @@
...
+// Replication controllers have no need of any methods other than these.
+type kpStore interface {
+	SetPod(key string, manifest pods.Manifest) (time.Duration, error)
+	DeletePod(key string) (time.Duration, error)
+}
+
+type replicationController struct {
+	fields.RC
+
+	manifest        pods.Manifest
```
petertseng: remove leftover fields.

pkg/kp/rcstore/store.go
```diff
@@ -0,0 +1,32 @@
...
+type Store interface {
+	// Create creates a replication controller with the specified manifest and selectors.
+	// The node selector is used to determine what nodes the replication controller may schedule on.
+	// The pod label set is:
+	// 1) applied to every pod the replication controller schedules,
+	// 2) used (in addition to replication controller ID) to determine what pods are already scheduled.
+	Create(manifest pods.Manifest, nodeSelector labels.Selector, podLabels labels.Set) (fields.RC, error)
+
+	List() ([]fields.RC, error)
+	SetDesiredReplicas(fields.RC, int) error
```
petertseng: these three methods (`SetDesiredReplicas`, `Disable`, `Delete`) could possibly just take an ID (uniquely identifying a replication controller) if desired. I'm not terribly fussed either way.

pkg/kp/rcstore/fake_store.go
```diff
@@ -0,0 +1,134 @@
...
+		return updatesOut, errors
+	}
+
+	errors := make(chan error)
+	updatesIn := make(chan struct{})
+	entry.watchers = append(entry.watchers, updatesIn)
+
+	go func() {
+		for _ = range updatesIn {
+			rc.ReplicasDesired = entry.ReplicasDesired
```
petertseng: prone to breakage if we ever add any other fields that can change. make it general?

petertseng: fixed

pkg/kp/rcstore/store.go
```diff
@@ -0,0 +1,32 @@
...
+)
+
+// Store represents an interface for persisting replication controllers to Consul,
+// as well as restoring replication controllers from Consul.
+type Store interface {
+	// Create creates a replication controller with the specified manifest and selectors.
+	// The node selector is used to determine what nodes the replication controller may schedule on.
+	// The pod label set is:
+	// 1) applied to every pod the replication controller schedules,
+	// 2) used (in addition to replication controller ID) to determine what pods are already scheduled.
```
petertseng: wait, this is an inaccurate comment. Currently I'm only using the replication controller ID to determine what pods are scheduled. Given that RC IDs are going to be UUIDs, that should uniquely identify pods that have been scheduled by this very same replication controller.

Is this the behavior we want, though? If I create a second replication controller that is identical to the first (same manifest, same node selector, same pod labels), *should* the latter see the pods scheduled by the former?

If the answer is yes, I'll change `currentNodes` to match.
If the answer is no, I need to remove this line as it's an inaccurate comment.

pkg/rc/replication_controller.go
```diff
@@ -0,0 +1,245 @@
...
+
+	return nil
+}
+
+func (rc *replicationController) eligibleNodes() ([]string, error) {
+	return rc.scheduler.EligibleNodes(rc.Manifest, rc.NodeSelector)
+}
+
+func (rc *replicationController) CurrentNodes() ([]string, error) {
+	selector := labels.Everything().Add(rcIdLabel, labels.EqualsOperator, []string{rc.Id})
```
petertseng: note this line (related to previous comment)

pkg/rc/replication_controller.go
```diff
@@ -0,0 +1,245 @@
...
+	for k, v := range rc.PodLabels {
+		selector = selector.Add(k, labels.EqualsOperator, []string{v})
```
petertseng: please consider using `rc.PodLabels.AsSelector()` instead.

pkg/kp/rcstore/store.go
anthonybishopric: Good question. Given that users can submit additional labels and selectors for replication controllers, it seems easy to implement totally distinct controllers on top of overlapping controllers. Harder to go the other way if the RC requires its own ID always be present in the selector. @bwester @mpuncel what do you think?

petertseng: I'll tentatively say that the answer is "yes, two replication controllers with the same pod labels will see each others' pods", code and comment updated to match.

pkg/kp/rcstore/store.go
petertseng: using an ID now, typedefed it in case anything changes, but I don't think anything will

pkg/rc/replication_controller.go
```diff
@@ -0,0 +1,243 @@
...
+	}
+}
+
+func (rc *replicationController) WatchDesires(quit <-chan struct{}) <-chan error {
+	desiresChanged, errInChannel := rc.rcStore.Watch(&rc.RC, quit)
+
+	errOutChannel := make(chan error)
+	channelsClosed := make(chan struct{})
+
+	// When seeing any changes, try to meet then.
```
mpuncel: s/then/them

petertseng: thanks for catch, amended to fix

pkg/rc/replication_controller.go
```diff
@@ -0,0 +1,243 @@
...
+	// If meeting produces any error, send it on the output error channel.
+	go func() {
+		for _ = range desiresChanged {
```
mpuncel: nit: can just do `for range desiresChanged`

pkg/rc/replication_controller.go
```diff
@@ -0,0 +1,243 @@
...
+func (rc *replicationController) meetDesires() error {
+	rc.logger.NoFields().Infof("Meeting with desired replicas %d, disabled %v", rc.ReplicasDesired, rc.Disabled)
+
+	// If we're disabled, we do nothing, nor is it an error
+	// (it's a normal possibility to be disabled)
+	if rc.Disabled {
+		return nil
+	}
+
+	eligible, err := rc.eligibleNodes()
```
mpuncel: nit: you don't need to know eligible nodes if you already meet the desires. Unless you want that log line for debugging

pkg/rc/replication_controller.go
```diff
@@ -0,0 +1,243 @@
...
+
+func difference(subtractFrom, subtractThese []string) []string {
+	diff := []string{}
+	m := map[string]bool{}
+	for _, v := range subtractThese {
+		m[v] = true
+	}
+
+	for _, v := range subtractFrom {
+		if _, ok := m[v]; !ok {
```
petertseng: nit: `if !m[v]` works here because the default is the zero value and the zero value is false.

pkg/rc/replication_controller.go
```diff
@@ -0,0 +1,243 @@
...
+	})
+	if err != nil {
+		return err
+	}
+
+	_, err = rc.kpStore.DeletePod(intentPath)
+	return err
+}
+
+func difference(subtractFrom, subtractThese []string) []string {
```
mpuncel: maybe should use `StringSet.Difference()` from pkg/labels/set.go? Will mean you need to use set types instead of string slices which maybe we don't want if the node ordering matters?

pkg/rc/replication_controller.go
```diff
@@ -0,0 +1,243 @@
...
+	for i, node := range pods {
+		parts := strings.SplitN(node.ID, "/", 2)
+		// If it so happens that node.ID contains no "/",
+		// then parts[0] is the entire string, so we're OK
+		result[i] = parts[0]
+	}
+	return result, nil
+}
+
+func (rc *replicationController) forEachLabel(node string, f func(id, k, v string) error) error {
```
mpuncel: looks like this function can partially succeed, maybe worth a comment

pkg/rc/replication_controller.go
```diff
@@ -0,0 +1,243 @@
...
+	return rc.forEachLabel(node, func(id, k, v string) error {
+		return rc.podApplicator.SetLabel(labels.POD, id, k, v)
+	})
+}
+
+func (rc *replicationController) unschedule(node string) error {
+	intentPath := kp.IntentPath(node, rc.Manifest.ID())
+	rc.logger.NoFields().Infof("Uncheduling from %s", intentPath)
+
+	err := rc.forEachLabel(node, func(id, k, _ string) error {
```
mpuncel: what are the ramifications if this function fails for some labels and succeeds for others?

pkg/rc/replication_controller.go
petertseng: oh cool, thanks

pkg/rc/replication_controller.go
petertseng: yeah makes sense. moved it into the only branch where it's necessary then. and removed the separate log line anyway since there's another log line that covers it

pkg/rc/replication_controller.go
petertseng: See, this is where I wish consul could do transactions spanning multiple KV operations. Comment is good idea, will have

pkg/rc/replication_controller.go
```diff
@@ -0,0 +1,243 @@
...
+
+	if rc.ReplicasDesired > len(current) {
+		// TODO: With Docker or runc we would not be constrained to running only once per node.
+		// So it may be the case that we need to make the Scheduler interface smarter and use it here.
+		possible := difference(eligible, current)
+		toSchedule := rc.ReplicasDesired - len(current)
+
+		rc.logger.NoFields().Infof("Need to schedule %d nodes out of %s", toSchedule, possible)
+
+		if toSchedule > len(possible) {
```
mpuncel: shouldn't you at least schedule the ones you can?

pkg/rc/replication_controller.go
petertseng: Adding comment. This means that a pod gets orphaned (the RC will no longer realize that the pod belongs to it, but the pod will in fact still be running). Seems unfortunate, but less unfortunate than unscheduling, having the unschedule succeed, and then having the first label remove fail (in this case the pod is actually not running but the RC might think it is). Going to leave a TODO in on figuring this out.

pkg/rc/replication_controller.go
petertseng: sounds reasonable

pkg/rc/replication_controller.go
petertseng: Node ordering doesn't really matter too much, so I went ahead and used `StringSet`s. `difference` is therefore deleted.

pkg/rc/replication_controller.go
petertseng: done, though I want a test

petertseng: test added

petertseng:
OK, I added consul RC stores (the last two commits of this PR), the other commits are unchanged. Perhaps it makes sense to review them here as well. If someone's blocked on the interfaces I'll just split consul RC stores into a separate PR and merge the rest, but otherwise I'll leave them together

PR #249, opened 2015-09-14
Implement pod DELETE

tummychow:
please review code for sanity, not ready for merge. I haven't tested it or implemented intent safety checks.

### DIFF
--- a/pkg/preparer/orchestrate.go
+++ b/pkg/preparer/orchestrate.go
@@ -298,6 +298,8 @@ func (p *Preparer) stopAndUninstallPod(pair ManifestPair, pod Pod, logger loggin
 		logger.NoFields().Warnln("One or more launchables did not halt successfully")
 	}
 
+	p.tryRunHooks(hooks.BEFORE_UNINSTALL, pod, pair.Reality, logger)
+
 	err = pod.Uninstall()
 	if err != nil {
 		logger.WithError(err).Errorln("Uninstall failed")
--- a/pkg/preparer/orchestrate.go
+++ b/pkg/preparer/orchestrate.go
@@ -19,6 +19,7 @@ type Pod interface {
 	hooks.Pod
 	Launch(*pods.Manifest) (bool, error)
 	Install(*pods.Manifest) error
+	Uninstall() error
 	Verify(*pods.Manifest, auth.Policy) error
 	Halt(*pods.Manifest) (bool, error)
 }
@@ -28,8 +29,9 @@ type Hooks interface {
 }
 
 type Store interface {
-	Pod(string) (*pods.Manifest, time.Duration, error)
+	ListPods(keyPrefix string) ([]kp.ManifestResult, time.Duration, error)
 	SetPod(string, pods.Manifest) (time.Duration, error)
+	DeletePod(key string) (time.Duration, error)
 	RegisterService(pods.Manifest, string) error
 	WatchPods(string, <-chan struct{}, chan<- error, chan<- []kp.ManifestResult)
 }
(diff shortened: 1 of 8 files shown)


tummychow:
okay this has safety checking for the preparer's pod ID in the intent set. ready for review

pkg/preparer/orchestrate.go
```diff
@@ -84,22 +89,34 @@ func (p *Preparer) WatchForPodManifestsForNode(quitAndAck chan struct{}) {
...
-					go p.handlePods(podChanMap[podId], quitChanMap[podId])
+		case intentResults := <-podChan:
+			realityResults, _, err := p.store.ListPods(kp.RealityPath(p.node))
+			if err != nil {
+				errChan <- err
+			} else {
+				// if the preparer's own ID is missing from the intent set, we
+				// assume it was damaged and discard it
+				if !checkResultsForID(intentResults, POD_ID) {
+					p.Logger.NoFields().Errorln("Intent results set did not contain p2-preparer pod ID")
```
anthonybishopric: Lets add a bit more color to this message - "... pod ID - it is possible that Consul's data is corrupt". This might help someone looking at preparer logs who isn't aware of this class of error.

pkg/preparer/orchestrate.go
```diff
@@ -130,26 +147,37 @@ func (p *Preparer) handlePods(podChan <-chan pods.Manifest, quit <-chan struct{}
...
+					p.tryRunHooks(hooks.AFTER_AUTH_FAIL, pods.NewPod(nextLaunch.ID, pods.PodPath(p.podRoot, nextLaunch.ID)), nextLaunch.Intent, manifestLogger)
+				}
 			}
 		case <-time.After(1 * time.Second):
 			if working {
-				pod := pods.NewPod(manifestToLaunch.ID(), pods.PodPath(p.podRoot, manifestToLaunch.ID()))
+				pod := pods.NewPod(nextLaunch.ID, pods.PodPath(p.podRoot, nextLaunch.ID))
 
-				ok := p.installAndLaunchPod(&manifestToLaunch, pod, manifestLogger)
+				ok := p.installAndLaunchPod(nextLaunch, pod, manifestLogger)
```
anthonybishopric: Lets move removal to a separate function from install, for clarity purposes

pkg/preparer/result_set_test.go
```diff
@@ -0,0 +1,34 @@
...
+	intent := []kp.ManifestResult{
+		{Manifest: pods.Manifest{Id: "foo"}},
+		{Manifest: pods.Manifest{Id: "bar"}},
+	}
+	reality := []kp.ManifestResult{
+		{Manifest: pods.Manifest{Id: "baz"}},
+		{Manifest: pods.Manifest{Id: "bar"}},
+	}
+
+	ValidatePairList(t, ZipResultSets(intent, reality))
```
anthonybishopric: I'd like this test to be a bit more precise: have the expected outcomes explicitly stated, ie "there's a pair with intent foo, but no reality foo", etc

pkg/preparer/orchestrate.go
tummychow: fair enough, but in that case I think the sha comparison function should also be factored out

pkg/preparer/orchestrate.go
tummychow: will fix

pkg/preparer/result_set_test.go
tummychow: yep I can add a bit more verification here

tummychow:
- rebased
- split up sha comparison, install/launch and halt/uninstall into separate functions

will embellish tests momentarily

Let's also talk about hooks on deletion

tummychow:
- tests for result set improved
- actually deletes the pod from reality now as well

tummychow:
added hooks for before_uninstall. i'll give it a quick spin to convince myself it works

petertseng:
well it lgtm, anyone else?

pkg/preparer/orchestrate.go
```diff
@@ -176,92 +205,113 @@ func (p *Preparer) authorize(manifest pods.Manifest, logger logging.Logger) bool
...
+	if err != nil {
+		logger.WithError(err).Errorln("Pod halt failed")
+	} else if !success {
+		logger.NoFields().Warnln("One or more launchables did not halt successfully")
+	}
 
-			p.tryRunHooks(hooks.AFTER_LAUNCH, pod, newManifest, logger)
-		}
-		return err == nil && ok
+	p.tryRunHooks(hooks.BEFORE_UNINSTALL, pod, pair.Reality, logger)
```
anthonybishopric: :+1:

anthonybishopric:
:shipit:

PR #250, opened 2015-09-16
Support forceable restarts after timeout

anthonybishopric:
All restarts are now forcible by default with a 60 second timeout, except for the preparer pod, which will never use forcible restart to ensure that bootstrapping continues to function correctly.

Each launchable section now supports an optional `restart_timeout` flag which can express a `time.Duration` that will be passed (in seconds) to the `sv force-restart` command. If no restart duration is provided, we continue to use `sv restart` and do not use KILL.

### DIFF
--- a/pkg/preparer/orchestrate.go
+++ b/pkg/preparer/orchestrate.go
@@ -147,6 +147,11 @@ func (p *Preparer) handlePods(podChan <-chan pods.Manifest, quit <-chan struct{}
 			if working {
 				pod := pods.NewPod(manifestToLaunch.ID(), pods.PodPath(p.podRoot, manifestToLaunch.ID()))
 
+				// TODO better solution: force the preparer to have a 0s default timeout, prevent KILLs
+				if pod.Id == POD_ID {
+					pod.DefaultTimeout = time.Duration(0)
+				}
+
 				ok := p.installAndLaunchPod(&manifestToLaunch, pod, manifestLogger)
 				if ok {
 					manifestToLaunch = pods.Manifest{}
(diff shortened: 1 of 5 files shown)


pkg/preparer/orchestrate.go
```diff
@@ -147,6 +147,11 @@ func (p *Preparer) handlePods(podChan <-chan pods.Manifest, quit <-chan struct{}
 			if working {
 				pod := pods.NewPod(manifestToLaunch.ID(), pods.PodPath(p.podRoot, manifestToLaunch.ID()))
 
+				// TODO better solution: force the preparer to have a 0s default timeout, prevent KILLs
+				if pod.Id == POD_ID {
```
tummychow: not especially important, but random thought here: maybe we should constrain the 0 timeout to the preparer's launchable only in the future

tummychow:
pretty straightforward, lgtm

pkg/runit/runit_service.go
```diff
@@ -80,10 +85,6 @@ func (sv *SV) Stop(service *Service) (string, error) {
 	return convertToErr(sv.execOnService(service, "stop"))
```
tummychow: you might want to add the timeout behavior to stop as well? not sure if we actually use this anywhere though

pkg/preparer/orchestrate.go
anthonybishopric: If a preparer pod has other launchables, they can override the rule in their own `restart_timeout` stanzas.

pkg/runit/runit_service.go
anthonybishopric: gonna hold off since we don't use it - probably should just delete it.

PR #251, opened 2015-09-16
Do not stat absolute symlinks during verification

tummychow:
Tested in universe and verified that it works. It would be nice if we could fix the bug that dragged in the absolute symlinks, but ffi makefiles still don't make sense to me, and that gem creeps into almost every ruby app in the company, bringing an absolutely-symlinked header file with it. So I think this is an acceptable patch-up.


bwester:
This looks like a horrible thing to do, but it also looks like it's a practical necessity that we do it.

Reluctantly: :ship:

anthonybishopric:
:sob:  :shipit:

PR #252, opened 2015-09-17
De-duplicate ManifestFromURI code

petertseng:
You may think it's not so bad that it's duplicated in two places but I'm
about to add a third, so I think it helps to do this.

### DIFF
--- a/bin/p2-launch/main.go
+++ b/bin/p2-launch/main.go
@@ -1,13 +1,10 @@
 package main
 
 import (
-	"io/ioutil"
 	"log"
-	"os"
 
 	"github.com/square/p2/Godeps/_workspace/src/gopkg.in/alecthomas/kingpin.v1"
 	"github.com/square/p2/pkg/pods"
-	"github.com/square/p2/pkg/uri"
 	"github.com/square/p2/pkg/version"
 )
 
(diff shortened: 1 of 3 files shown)


anthonybishopric:
:+1:

bin/p2-launch/main.go
```diff
@@ -19,19 +16,10 @@ var (
...
 
-	err = uri.URICopy(*manifestURI, localMan.Name())
-	if err != nil {
-		log.Fatalf("Could not fetch manifest: %s", err)
-	}
-	manifest, err := pods.ManifestFromPath(localMan.Name())
+	manifest, err := pods.ManifestFromURI(*manifestURI)
 	if err != nil {
-		log.Fatalf("Invalid manifest: %s", err)
+		log.Fatalf("%s", err)
```
bwester: For next time: `log.Fatalln(err)`

pkg/pods/manifest.go
```diff
@@ -102,6 +103,21 @@ func ManifestFromBytes(bytes []byte) (*Manifest, error) {
 	return manifest, nil
 }
 
+func ManifestFromURI(manifestUri string) (*Manifest, error) {
+	// Fetch manifest (could be URI) into temp file
+	localMan, err := ioutil.TempFile("", "tempmanifest")
```
bwester: I never noticed this before.

I wonder how many `tempmanifest`-s my laptop has by now...

petertseng: hmm, well I think they'd only be left lying around if the `defer os.Remove` failed to run, right?

PR #253, opened 2015-09-17
Add HTTP Applicator and tests

petertseng:
This allows using an HTTP server that accepts GET requests and returns
JSON lists of strings representing the nodes that match the query.

### DIFF
--- a/pkg/labels/http_applicator.go
+++ b/pkg/labels/http_applicator.go
@@ -1,0 +1,85 @@
+package labels
+
+import (
+	"encoding/json"
+	"net/http"
+	"net/url"
+
+	"github.com/square/p2/pkg/logging"
+	"github.com/square/p2/pkg/util"
+)
+
+type httpApplicator struct {
+	client *http.Client
+	// The endpoint that will be queried for matches.
+	// GetMatches will add to this endpoint a query parameter with key "selector" and value selector.String()
+	matchesEndpoint *url.URL
+	logger          logging.Logger
+}
+
+var _ Applicator = &httpApplicator{}
+
+func NewHttpApplicator(client *http.Client, matchesEndpoint *url.URL) (*httpApplicator, error) {
+	if matchesEndpoint == nil {
+		return nil, util.Errorf("matches endpoint cannot be nil")
+	}
+
+	c := client
+	if c == nil {
+		c = http.DefaultClient
+	}
+
+	return &httpApplicator{
+		logger:          logging.DefaultLogger,
+		matchesEndpoint: matchesEndpoint,
+		client:          c,
+	}, nil
+}
+
+func (h *httpApplicator) SetLabel(labelType Type, id, name, value string) error {
+	return util.Errorf("SetLabel not implemented for HttpApplicator (type %s, id %s, name %s, value %s)", labelType, id, name, value)
+}
+
+func (h *httpApplicator) RemoveLabel(labelType Type, id, name string) error {
+	return util.Errorf("RemoveLabel not implemented for HttpApplicator (type %s, id %s, name %s)", labelType, id, name)
+}
+
+func (h *httpApplicator) GetLabels(labelType Type, id string) (Labeled, error) {
+	return Labeled{}, util.Errorf("GetLabels not implemented for HttpApplicator (type %s, id %s)", labelType, id)
+}
+
+func (h *httpApplicator) GetMatches(selector Selector, labelType Type) ([]Labeled, error) {
+	params := url.Values{}
+	params.Add("selector", selector.String())
+	params.Add("type", labelType.String())
+
+	// Make value copy of URL; don't want to mutate the URL in the struct.
+	urlToGet := *h.matchesEndpoint
+	urlToGet.RawQuery = params.Encode()
+
+	resp, err := h.client.Get(urlToGet.String())
+	if err != nil {
+		return []Labeled{}, err
+	}
+
+	defer resp.Body.Close()
+
+	matches := []string{}
+	decoder := json.NewDecoder(resp.Body)
+	err = decoder.Decode(&matches)
+	if err != nil {
+		return []Labeled{}, err
+	}
+
+	labeled := make([]Labeled, len(matches))
+
+	for i, s := range matches {
+		labeled[i] = Labeled{
+			ID:        s,
+			LabelType: labelType,
+			Labels:    Set{},
+		}
+	}
+
+	return labeled, nil
+}
--- a/pkg/labels/http_applicator_test.go
+++ b/pkg/labels/http_applicator_test.go
@@ -1,0 +1,55 @@
+package labels
+
+import (
+	"fmt"
+	"net/http"
+	"net/http/httptest"
+	"net/url"
+	"testing"
+
+	. "github.com/square/p2/Godeps/_workspace/src/github.com/anthonybishopric/gotcha"
+)
+
+const endpointSuffix = "/testendpoint"
+
+func getMatches(t *testing.T, httpResponse string) ([]Labeled, error) {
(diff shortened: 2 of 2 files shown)


pkg/labels/http_applicator.go
```diff
@@ -0,0 +1,80 @@
...
+func (h *httpApplicator) RemoveLabel(labelType Type, id, name string) error {
+	return util.Errorf("RemoveLabel not implemented for HttpApplicator (type %s, id %s, name %s)", labelType, id, name)
+}
+
+func (h *httpApplicator) GetLabels(labelType Type, id string) (Labeled, error) {
+	return Labeled{}, util.Errorf("GetLabels not implemented for HttpApplicator (type %s, id %s)", labelType, id)
+}
+
+func (h *httpApplicator) GetMatches(selector Selector, labelType Type) ([]Labeled, error) {
+	resp, err := h.client.Get(h.matchesEndpoint + selector.String())
```
tummychow: simple concat works in this case and it's the most generic implementation, but I think setting the query argument on the request URL would be safer. that being said i'm not sure if label selectors have to be escaped - it's never smacked me on the command line before - so it's probably not a huge deal.

pkg/labels/http_applicator.go
```diff
@@ -0,0 +1,80 @@
...
+	if err != nil {
+		return []Labeled{}, err
+	}
+
+	defer resp.Body.Close()
+	body, err := ioutil.ReadAll(resp.Body)
```
tummychow: stylistic alternative to reading all: `json.Decoder`? up to you, this is mostly a nit

pkg/labels/http_applicator.go
```diff
@@ -0,0 +1,80 @@
...
+
+	err = json.Unmarshal(body, &matches)
+	if err != nil {
+		return []Labeled{}, err
+	}
+
+	labeled := make([]Labeled, len(matches))
+
+	for i, v := range matches {
+		s, ok := v.(string)
```
tummychow: why not just unmarshal into `[]string{}`? seems unnecessary to type-assert each element, unless my backend implementation is doing something dumb in which case let me know

pkg/labels/http_applicator.go
```diff
@@ -0,0 +1,80 @@
...
+		if !ok {
+			return []Labeled{}, util.Errorf("when parsing %s: %+v not a string", body, v)
+		}
+		labeled[i] = Labeled{
+			ID:        s,
+			LabelType: labelType,
+			Labels:    Set{},
```
tummychow: I can adjust the backend implementation to return the labels on these objects if you prefer

pkg/labels/http_applicator.go
petertseng: I don't think that json.Unmarshal allows that to happen. It only unmarshals into `[]interface{}`. I'll test it out and let you know if it actually works.

pkg/labels/http_applicator.go
petertseng: The only user of `HttpApplicator.GetMatches` (the new RC) currently doesn't care, so only do that if you think someone will care in the future I suppose

pkg/labels/http_applicator.go
petertseng: pushed a commit that uses urls instead, I am a bit worried that it's now unsafe to run concurrently, take a look and see whether you like that better though

tummychow: yeah you might be mutating the same URL on every request. maybe save the string and parse it into a URL at the start of the getmatches function?

pkg/labels/http_applicator.go
petertseng: IT WORKS! Using strings directly.

pkg/labels/http_applicator.go
```diff
@@ -0,0 +1,84 @@
...
+func (h *httpApplicator) GetLabels(labelType Type, id string) (Labeled, error) {
+	return Labeled{}, util.Errorf("GetLabels not implemented for HttpApplicator (type %s, id %s)", labelType, id)
+}
+
+func (h *httpApplicator) GetMatches(selector Selector, labelType Type) ([]Labeled, error) {
+	params := url.Values{}
+	params.Add("selector", selector.String())
+
+	// This makes GetMatches unsafe to run concurrently!
+	h.matchesEndpoint.RawQuery = params.Encode()
```
petertseng: yeah it'd be great if I can get a `url.Clone` or semething

alternatively I could `url.Parse(url.String())` I guess...

pkg/labels/http_applicator.go
```diff
@@ -0,0 +1,84 @@
...
+	logger          logging.Logger
+}
+
+var _ Applicator = &httpApplicator{}
+
+func NewHttpApplicator(client *http.Client, matchesEndpoint *url.URL) *httpApplicator {
+	c := client
+	if c == nil {
+		c = http.DefaultClient
+	}
```
anthonybishopric: verify that `matchesEndpoint` is not nil?

pkg/labels/http_applicator.go
anthonybishopric: I'd suggest performing the query with a value copy of the URL and its query params. It's icky to reassign the field this way.

anthonybishopric:
:+1: with comments

pkg/labels/http_applicator.go
petertseng: ok, so doing `urlToGet := *h.matchesEndpoint` should just do a value copy, looks like I'll do that.

pkg/labels/http_applicator.go
petertseng: Interesting. I'll do it because then I won't import ioutil

petertseng:
Squashed it. Merge coming tomorrow.

pkg/labels/http_applicator.go
```diff
@@ -0,0 +1,84 @@
...
+
+func (h *httpApplicator) RemoveLabel(labelType Type, id, name string) error {
+	return util.Errorf("RemoveLabel not implemented for HttpApplicator (type %s, id %s, name %s)", labelType, id, name)
+}
+
+func (h *httpApplicator) GetLabels(labelType Type, id string) (Labeled, error) {
+	return Labeled{}, util.Errorf("GetLabels not implemented for HttpApplicator (type %s, id %s)", labelType, id)
+}
+
+func (h *httpApplicator) GetMatches(selector Selector, labelType Type) ([]Labeled, error) {
```
bwester: Is it expected/intentional that the query doesn't depend on `labelType` in any way?

bwester:
:ship:

pkg/labels/http_applicator.go
anthonybishopric: Pretty sure ignoring it is a bug

tummychow: well right now the backend only supports type=NODE, therefore the implementation only has an endpoint for type=NODE. adding support for type=POD to the backend would be fairly straightforward in which case same could be implemented here. for now I don't think it's a blocker?

tummychow:
lgtm

pkg/labels/http_applicator.go
petertseng: Given the current URL structure of our backend, I guess that would require some changes to this code (because when querying for node matches we include "node" in the URL rather than, say, "type=node" in query parameters). So I envision something like a map[Type]string indicating the path to query for a given Type. We can easily make that happen when the time comes.

I have just added some minor help to backends that support both NODE and POD by including "type=node" or "type=pod" in the query parameters since that was super easy, but the existing backend doesn't support it (it doesn't hurt it either, of course, otherwise I wouldn't add it).

The only current user we have of httpApplicator doesn't care (it only needs to look for nodes not pods), so I'm going to ship it as-is.

anthonybishopric: :+1:

PR #255, opened 2015-09-17
Also use force timeouts on pod.Halt()

anthonybishopric:
Since we currently do an `sv stop` followed by an `sv restart`, it makes sense that we should pass the time out to stop.


tummychow:
code looks good. what's up with integration test failure? other than that +1

anthonybishopric:
Adds logging, bug disappears #heisenbuglife

PR #256, opened 2015-09-17
Enable global hooks

tummychow:
next step would be disabling typed hooks, but we'll have to do that after we migrate all our hooks (including those from other teams) to the new system.

### DIFF
--- a/bin/p2-schedule/main.go
+++ b/bin/p2-schedule/main.go
@@ -14,9 +14,11 @@ import (
 )
 
 var (
-	manifests     = kingpin.Arg("manifests", "one or more manifest files to schedule in the intent store").Strings()
-	nodeName      = kingpin.Flag("node", "The node to do the scheduling on. Uses the hostname by default.").String()
-	hookTypeName  = kingpin.Flag("hook-type", "Schedule as a hook, not an intended pod, as the given hook type. Can be one of the hooks listed in hooks.go").String()
+	manifests    = kingpin.Arg("manifests", "one or more manifest files to schedule in the intent store").Strings()
+	nodeName     = kingpin.Flag("node", "The node to do the scheduling on. Uses the hostname by default.").String()
+	hookTypeName = kingpin.Flag("hook-type", "Schedule as a hook, not an intended pod, as the given hook type. Can be one of the hooks listed in hooks.go, or the word \"global\"").String()
+	// unhide this flag when typed hooks are removed
+	hookGlobal    = kingpin.Flag("hook", "Schedule as a global hook.").Hidden().Bool()
 	consulAddress = kingpin.Flag("consul", "The address of the consul node to use. Defaults to 0.0.0.0:8500").String()
 	consulToken   = kingpin.Flag("token", "The ACL to use for accessing consul.").String()
 	headers       = kingpin.Flag("header", "An HTTP header to add to requests, in KEY=VALUE form. Can be specified multiple times.").StringMap()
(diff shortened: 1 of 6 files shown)


pkg/hooks/hooks.go
```diff
@@ -144,5 +147,9 @@ func (h *HookDir) runHooks(dirpath string, hType HookType, pod Pod, podManifest
 
 func (h *HookDir) RunHookType(hookType HookType, pod Pod, manifest *pods.Manifest) error {
 	dirpath := path.Join(h.dirpath, hookType.String())
-	return h.runHooks(dirpath, hookType, pod, manifest)
+	if err := h.runHooks(dirpath, hookType, pod, manifest); err != nil {
```
anthonybishopric: Requested cleanup: distinguish `h.dirpath` and `dirpath` more clearly - not obvious that one is global

anthonybishopric:
:+1: pending small renaming request, also what's up with the test

pkg/hooks/hooks.go
tummychow: sure I'll change the variable name to make it clear that one is scoped to a certain type

tummychow:
trimmed log output for hooks, verified that it's not as profuse

pkg/hooks/hooks.go
```diff
@@ -143,6 +141,17 @@ func (h *HookDir) runHooks(dirpath string, hType HookType, pod Pod, podManifest
...
 
 func (h *HookDir) RunHookType(hookType HookType, pod Pod, manifest *pods.Manifest) error {
-	dirpath := path.Join(h.dirpath, hookType.String())
-	return h.runHooks(dirpath, hookType, pod, manifest)
+	logger := h.logger.SubLogger(logrus.Fields{
+		"pod":      manifest.ID(),
+		"pod_path": pod.Path(),
+		"event":    hookType.String(),
+	})
+	logger.NoFields().Infoln("Running hooks")
```
anthonybishopric: Can we do an `Infof` here with the `hookType` inlined? The dashboard currently only shows the message and not the other parts of the logrus payload.

tummychow: ah of course, that explains why we were using the infof to begin with. I'll put it back

tummychow:
put back the infof / warnf for messages that might be read without logrus metadata

bin/p2-schedule/main.go
```diff
@@ -53,12 +53,14 @@ func main() {
 			log.Fatalf("Could not read manifest at %s: %s\n", manifestPath, err)
 		}
 		path := kp.IntentPath(*nodeName, manifest.ID())
-		if *hookTypeName != "" {
+		if *hookTypeName == "global" {
```
anthonybishopric: How about renaming the flag to `--hook`. If the value is the empty string, then we know the schedule command was `p2-schedule --hook`, and know to schedule it as a global hook. This way there doesn't need to be a sentinel value for global hooks, and will make migrating to an all-global hook world easier

tummychow: idk if kingpin lets you distinguish "specified flag without value (goes to default)" from "did not specify flag at all". I'll try it

anthonybishopric: The pointer to `hookTypeName` will be nil if you don't specify it iirc

tummychow: If that were true then you would get a panic every time you used p2-schedule without a `hook-type`. I don't think kingpin supports this behavior so I propose a compromise:

- keep the existing `hook-type` flag and add a `global` setting so that it can be used to schedule global hooks
- add a hidden boolean option for just `hook` that always treats the manifest as a global hook, regardless of `hook-type`.

In our scripts we can start using the `--hook` option without showing it in `p2-schedule --help`. When we're satisfied with the global hook implementation, we kill the `hook-type` flag and unhide the `hook` flag. Thoughts?

anthonybishopric: :+1:

tummychow:
added a hidden --hook flag to p2-schedule, which is boolean and treats the hook as global

PR #257, opened 2015-09-18
ConsulApplicator: split by "/" instead of using path

petertseng:
This deals with <id> containing "/" in it. Without this commit, an <id>
of, for example, "a/b" (resulting in consul key "labels/pod/a/b") will
cause the type to be inferred as "a" instead of "pod" and ID to be
inferred as "b" instead of "a/b"


tummychow:
looks good

PR #258, opened 2015-09-18
Add hooks ExitUnlessEvent

tummychow:
I should have added this to the global hooks PR, but I only just realized it would be useful.


anthonybishopric:
:ship:

PR #259, opened 2015-09-21
Remove reality polling from health monitor

tummychow:
see commit message. Tests pass on this branch. The semantics of the change "feel" straightforward so my instincts say it should work. I'll see if I can give it a test run.


anthonybishopric:
:+1:

PR #260, opened 2015-09-21
Enable/disable errors should not flip success

tummychow:
see commit message. This was a fairly non-obvious bug to trace (in fact the change that caused the issue is in a completely separate package from the diff where I've chosen to fix the issue), so if the commit message seems unclear to you, let me know so I can improve it.

e: btw I'm probably also going to add tests for this


anthonybishopric:
:+1:

PR #261, opened 2015-09-21
Add more tests to pkg/hoist

tummychow:
We should also have tests for the key exported methods, Launch and Halt. Unfortunately I think there's a lot of lifecycle code in Pod that doesn't have test coverage. Still thinking about how to attack those.


anthonybishopric:
:+1:

PR #262, opened 2015-09-22
Capture output from enable/disable

tummychow:
Pretty self-explanatory. The error return from os/exec is rarely useful (`exit status 1`) so the `Error()` function returns the output instead.

### DIFF
--- a/pkg/hoist/hoist_launchable.go
+++ b/pkg/hoist/hoist_launchable.go
@@ -39,10 +39,12 @@ type StartError error
 type StopError error
 
 func (hl *Launchable) Halt(serviceBuilder *runit.ServiceBuilder, sv *runit.SV) error {
-	// probably want to do something with output at some point
-	_, err := hl.disable()
+	// the error return from os/exec.Run is almost always meaningless
+	// ("exit status 1")
+	// since the output is more useful to the user, that's what we'll preserve
+	out, err := hl.disable()
 	if err != nil {
-		return DisableError(err)
+		return DisableError(util.Errorf("%s", out))
 	}
 
 	// probably want to do something with output at some point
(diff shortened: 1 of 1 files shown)


pkg/hoist/hoist_launchable.go
```diff
@@ -98,7 +112,7 @@ func (hl *Launchable) enable() (string, error) {
 
 	// providing an enable script is optional, ignore those errors
 	if err != nil && !os.IsNotExist(err) {
-		return output, util.Errorf("Could not enable %s: %s", hl.Id, err)
```
anthonybishopric: Worth keeping for the 1) stack trace 2) launchable ID

tummychow: launchable ID will always be logged at a higher layer. I wanted to avoid redundant log messages like

```json
{"err":"Could not enable foo: exit status 1","launchable":"foo","message":"Enable failed"}
```

I'll see what I can do about the stack trace but I think in this case it's not especially meaningful. An EnableError is obviously coming from the Enable function, and with the type information, all of our logging around enable/disable should be easier to trace.

Philosophical tangent: I think stack traces in error messages are a bandage for errors being too aggressively stringified or lacking in context. If this was an exception-based language then obviously they'd be unavoidable, but more effective structuring of error values should make stack traces in errors mostly obsolete, especially in short functions like this one. End tangent.

anthonybishopric: I like stack traces for our errors because it makes it easy for https://getsentry.com/welcome/ to de duplicate them. Otherwise sure, it's pretty obvious where it's originating from.

tummychow: good point about sentry, I'll add them back

tummychow:
alternative implementation that discards the (mostly useless) error return and preserves stack traces

anthonybishopric:
:+1:

PR #263, opened 2015-09-22
add before_launch hook type

anthonybishopric:
Some actions must be taken after we have turned off the old version of the app

### DIFF
--- a/pkg/hooks/hooks.go
+++ b/pkg/hooks/hooks.go
@@ -34,8 +34,9 @@ func (hookType HookType) String() string {
 
 var (
 	BEFORE_INSTALL   = HookType("before_install")
-	AFTER_INSTALL    = HookType("after_install")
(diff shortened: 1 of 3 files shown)


petertseng:
OK, seems good to me. Will it be necessary to add this to the switch/case in `RunHookType` in `orchestrate_test` ? Guess it's not strictly necessary, since we don't test each particular hook type, but perhaps for uniformity...?

anthonybishopric:
Sure, I can do that

bwester:
I'm glad you added extra comments describing the differences between the hooks; it's quite non-intuitive.  That said, it's so non-intuitive that it makes me question my beliefs about when _any_ of those hooks get run.

These hooks have become complex enough that it's time for a formal state diagram to explain them.

anthonybishopric:
@bwester agreed, I can start working on that in a bit.

PR #267, opened 2015-09-25
Add p2-restart command

anthonybishopric:
This is a useful on-host binary for gracefully stopping and starting a service. Also a few commits for small issues I noticed.

### DIFF
--- a/bin/p2-restart/main.go
+++ b/bin/p2-restart/main.go
@@ -8,6 +8,7 @@ import (
 	"github.com/square/p2/Godeps/_workspace/src/github.com/Sirupsen/logrus"
 	"github.com/square/p2/Godeps/_workspace/src/gopkg.in/alecthomas/kingpin.v1"
 	"github.com/square/p2/pkg/pods"
+	"github.com/square/p2/pkg/runit"
 	"github.com/square/p2/pkg/version"
 )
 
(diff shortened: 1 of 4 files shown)


tummychow:
will review code tomorrow, but an immediate thought that comes to mind: we can implement `p2-stop` with this as well, and create a symlink on host so that p2-stop aliases to this command.

bin/p2-restart/main.go
```diff
@@ -0,0 +1,80 @@
...
+	var path string
+	if *podName != "" {
+		path = filepath.Join(pods.DEFAULT_PATH, *podName)
+	} else {
+		path = *podDir
+	}
+
+	pod, err := pods.ExistingPod(path)
+	if err != nil {
+		logger.NoFields().Fatalln(err)
```
tummychow: should probably use logger.WithError instead

bin/p2-restart/main.go
```diff
@@ -0,0 +1,80 @@
...
+	}
+
+	logger.WithField("pod", pod.Id).Infoln("Halting pod")
+
+	ok, err := pod.Halt(manifest)
+	if err != nil {
+		logger.WithError(err).Fatalln("Could not halt pod")
+	} else if !ok {
+		logger.NoFields().Warningln("Had to forcibly kill some services")
```
tummychow: Not sure about the log message here, sigkilling does not cause sv to error. The most likely cause for !ok is that an sv stop failed altogether (eg service directory didn't exist). In any case it's not something the user can diagnose. I'd make the message more vague, actually.

bin/p2-restart/main.go
```diff
@@ -0,0 +1,80 @@
...
+	}
+
+	logger.WithField("pod", pod.Id).Infoln("Starting pod")
+
+	ok, err = pod.Launch(manifest)
+	if err != nil {
+		logger.WithError(err).Fatalln("Could not start pod")
+	} else if !ok {
+		logger.NoFields().Warningln("Some services did not come up quickly")
```
tummychow: again not sure about log message, it's not (usually) an issue of time. Enable failures do not flip ok to false, so this would probably indicate an sv-level error that can't easily be recovered by the user.

tummychow:
pursuant to my comments, it would be good to see sample output with various failure scenarios. If we intend for this tool to be user-facing, then the log messages have to be either actionable, or vague enough that the user clearly understands they need someone else's help. Let's also remember that the pod itself logs some things and we don't want those messages to be redundant with the ones outputted by this utility.

Functionally though, this is good to go :+1:

PR #268, opened 2015-09-28
Add params package

bwester:
Adds a new "params" package which provides a less-formal way to specify
configuration parameters.  Each parameter can be declared in the package where
it's needed, and a single call to `params.Parse()` will set all defined
parameters.

This is mainly intented to permit some configurability for values that would
otherwise be declared as a package constant, or where there are too many layers
between a program's main configuration parser and the use of that configuration
value.

### DIFF
--- a/pkg/preparer/setup.go
+++ b/pkg/preparer/setup.go
@@ -19,6 +19,7 @@ import (
 	"github.com/square/p2/pkg/logging"
 	"github.com/square/p2/pkg/pods"
 	"github.com/square/p2/pkg/util"
+	"github.com/square/p2/pkg/util/param"
 )
 
 // DefaultConsulAddress is the default location for Consul when none is configured.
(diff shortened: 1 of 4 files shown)


pkg/preparer/setup.go
```diff
@@ -52,6 +52,10 @@ type PreparerConfig struct {
 	WriteKVHealth        bool                   `yaml:"write_kv_health,omitempty"`
 	UseSessionHealth     bool                   `yaml:"use_session_health,omitempty"`
 	LogLevel             string                 `yaml:"log_level,omitempty"`
+
+	// Params defines a collection of miscellaneous runtime parameters defined throughout the
+	// source files.
+	Params map[string]string `yaml:"params"`
```
anthonybishopric: by itself I'm concerned that a user of this might be tempted to just read from `config.Params` directly instead of using the `params` package. Perhaps typedef to `params.Map` or something to make it clear that it's used in conjunction with that package?

anthonybishopric:
Overall this is a nice lightweight solution to const setting. :star:

pkg/preparer/setup.go
bwester: That makes sense.

PR #269, opened 2015-09-28
Add --workdir flag to p2-exec to change working directory

bwester:
Adds a new --workdir (short: -w) flag to the p2-exec command that allows it to
change the working directory before exec-ing its main command.


anthonybishopric:
Is this so you can run commands like `p2-exec -w /foo/bar my_command.sh --file my_file.txt`?

bwester:
That would work with this patch.  The immediate reason is that `runc` is unfortunately sensitive to the working directory it's launched in.

tummychow:
nice, this is a useful change in general :+1: code looks good as well

bwester:
> code looks good as well

All 7 lines of it ;)

PR #270, opened 2015-09-28
Fix ZipResultSets conditions

tummychow:
Luckily this panic is only encountered when trying to DELETE a manifest (since intent has to be shorter than reality to cause the panic, vice versa does not cause it) so it won't be triggered in production conditions any time soon.

I'm going to try and write some fuzz tests for this before merging, because this is the third time I've had to modify a logically straightforward function...

### DIFF
--- a/pkg/preparer/result_set_test.go
+++ b/pkg/preparer/result_set_test.go
@@ -1,6 +1,8 @@
 package preparer
 
 import (
+	"math/rand"
+	"strconv"
 	"testing"
 
 	. "github.com/square/p2/Godeps/_workspace/src/github.com/anthonybishopric/gotcha"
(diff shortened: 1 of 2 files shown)


bwester:
Looks good to me
:ship:

tummychow:
added fuzzer, confirmed that it would have caught the panic. might want to give that a quick once-over as well

pkg/preparer/result_set_test.go
```diff
@@ -24,36 +26,68 @@ func TestZipBasic(t *testing.T) {
...
+			}
+		}
+		Assert(t).IsTrue(pair.Reality == nil || realityFound, "should have found at least one manifest in the source reality list")
 	}
 }
+
+func TestZipFuzz(t *testing.T) {
+	for i := 0; i < 500; i++ {
+		intent := generateRandomManifestList(100)
+		reality := generateRandomManifestList(100)
```
bwester: Minor nit: `generateRandomManifestList()` is very unlikely to generate two lists that have any overlap.

tummychow: hm? it's the integers 0-99 shuffled, with up to half taken out. To get no overlap, you would have to remove 50 integers (the maximum) in both calls to generateRandom, and furthermore, the pieces that got discarded would have to be completely disjoint. That seems pretty unlikely to me. I was logging every list when writing the test, and most of the lists were mostly overlapping.

PR #271, opened 2015-09-29
Replication Controllers label themselves

petertseng:
Let's talk a bit about how this should work. What labels should the RC put on itself?

Currently just pod_id. Should we apply the same labels as in the `podLabels` set? I think we earlier decided that the answer to that question should be "no". Should it be a separate parameter?

Perhaps we should also remove all labels when the RC is deleted?

Should that be coupled with the RC store?

By this I mean: should there be a `func (*rc replicationController) Delete()` that does `rc.rcStore.Delete(rc.RC)` and also `rc.DeleteLabels()`? My own suspicion is that there should be, to prevent someone from forgetting

If the answer to the previous question is yes, does that seem inconsistent given that `New()` doesn't perform `rcStore.Create`? It's this way because `New()` could be passed the results of `rcStore.Create` or `rcStore.Get`. Creating a new replication controller doesn't always mean creating one in the store.

### DIFF
--- a/pkg/kp/rcstore/consul_store.go
+++ b/pkg/kp/rcstore/consul_store.go
@@ -32,16 +32,19 @@ type consulKV interface {
 }
 
 type consulStore struct {
-	kv     consulKV
-	logger logging.Logger
+	applicator labels.Applicator
+	kv         consulKV
+	logger     logging.Logger
 }
 
 var _ Store = &consulStore{}
 
 func NewConsul(client *api.Client) *consulStore {
 	return &consulStore{
-		kv:     client.KV(),
-		logger: logging.DefaultLogger,
+		// TODO: Should the number of retries be configurable?
+		applicator: labels.NewConsulApplicator(client, 3),
+		kv:         client.KV(),
+		logger:     logging.DefaultLogger,
 	}
 }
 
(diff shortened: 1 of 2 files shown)


pkg/rc/replication_controller.go
```diff
@@ -42,7 +43,13 @@ func New(
 	rcStore rcstore.Store,
 	scheduler Scheduler,
 	podApplicator labels.Applicator,
```
petertseng: someone may mix up the order of `podApplicator` and `rcApplicator` and it will suck. Any way to deal with this? I didn't want them to necessarily have to be backed by the same kind of applicator (though in both our tests and our real environments they will be...). So I guess I could just make them a single argument...

anthonybishopric:
> By this I mean: should there be a func (*rc replicationController) Delete() that does rc.rcStore.Delete(rc.RC) and also rc.DeleteLabels()? My own suspicion is that there should be, to prevent someone from forgetting.

Agreed.

>  Creating a new replication controller doesn't always mean creating one in the store.

Only creating one in the store should automatically trigger label setting.

anthonybishopric:
Overall :+1: with deletion I think

petertseng:
rcstores should have an rc applicator. it is the rcstore's responsibility to add/remove the labels resp. when creating/deleting rcs. a get from an rcstore doesn't touch labels (but it might read labels to know what rcs to get! future improvement)

pkg/rc/replication_controller.go
petertseng: resolution will be: this job belongs to the rcstore.

mpuncel: @petertseng you could also do something like this to split up PodApplicator and RCApplicator interfaces: https://github.com/kubernetes/kubernetes/blob/master/pkg/runtime/interfaces.go#L144

petertseng:
To implement the above-mentioned changes, the code has changed significantly since previous review. Please review again.

mpuncel:
lgtm :+1:

pkg/kp/rcstore/consul_store.go
```diff
@@ -123,26 +136,25 @@ func (s *consulStore) SetDesiredReplicas(id fields.ID, n int) error {
...
-	if i != 0 {
-		return fmt.Errorf("Replication controller %s has %d desired replicas, must be 0 before can be deleted", id, i)
-	}
-
-	_, err = s.kv.DeleteTree(idPrefix(id), nil)
-	return err
+	// TODO: If this fails, then we have some dangling labels.
+	// Perhaps they can be cleaned up later.
+	return s.forEachLabel(rc, func(id, k, _ string) error {
+		return s.applicator.RemoveLabel(labels.POD, id, k)
```
petertseng: NOPE! Gotta be `RC` here. fixing.

PR #272, opened 2015-09-29
Refactor hook install

tummychow:
Moved the most logically complex part of hook installation (writing hook scripts) into the hooks package, and created a corresponding binary that uses it. Will be useful for distributing hooks without hitting everything.

### DIFF
--- a/pkg/hooks/install.go
+++ b/pkg/hooks/install.go
@@ -1,0 +1,72 @@
+package hooks
+
+import (
+	"fmt"
+	"os"
+	"path/filepath"
+
+	"github.com/square/p2/Godeps/_workspace/src/github.com/Sirupsen/logrus"
+	"github.com/square/p2/pkg/logging"
+	"github.com/square/p2/pkg/pods"
+	"github.com/square/p2/pkg/runit"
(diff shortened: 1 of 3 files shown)


pkg/hooks/install.go
```diff
@@ -0,0 +1,69 @@
...
+// Populates the given directory with executor scripts for each launch script of
+// the given pod, which must be installed. Any orphaned executor scripts (from a
+// past install, but no longer present in this pod) will be cleaned out.
+func InstallHookScripts(dir string, hookPod *pods.Pod, manifest pods.Manifest, logger logging.Logger) error {
+	err := os.MkdirAll(dir, 0755)
+	if err != nil {
+		return err
+	}
+
+	rmPattern := filepath.Join(dir, fmt.Sprintf("%s__*", hookPod.Id))
```
anthonybishopric: Still think this is worth leaving a refactoring note about changing the wildcard glob

anthonybishopric:
:+1:

pkg/hooks/install.go
tummychow: right sorry, I deleted that by accident, will restore

PR #273, opened 2015-09-30
replication_controller: Add interface

petertseng:
We wouldn't want to make our clients have to deal with references to
unexported types, since then all they can do is something like `myrc :=
rc.New(...)`. They can't refer to the type to use it as a function
argument or return type, for instance. Too restrictive. So we add an
interface.

This also adds documentation, and now we need to export an Id as well


tummychow:
yep, this is what i was looking for :+1:

PR #275, opened 2015-10-01
Correctly fail installation on artifact transfer errors

bwester:
Bug: when a network transfer failed, the tar extractor would (highly likely)
fail due to an unexpected EOF and report its error correctly. However, the
partially-extracted artifact wasn't cleaned up correctly, and the _next_
attempt to install would use the bad install directory.

This commit fixes the bug by downloading to a temporary file first, then
extracting files. Cleanup is easy: just delete the file.

### DIFF
--- a/pkg/hoist/hoist_launchable.go
+++ b/pkg/hoist/hoist_launchable.go
@@ -7,6 +7,7 @@ import (
 	"io/ioutil"
 	"os"
 	"os/exec"
+	"path"
 	"path/filepath"
 	"time"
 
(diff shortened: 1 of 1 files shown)


tummychow:
I wonder if this makes installation of large tarballs significantly slower. probably not too bad in practice. Digest verification is the real solution here but this is still a good defensive measure, so :+1:

bwester:
Yes, it's going to be a little slower.  But large artifacts are more likely to fail too.

We probably should move verification steps into the "install" phase.

PR #277, opened 2015-10-02
Orchestration-related refactors part 1

tummychow:
None of this code has been tested yet, requesting review for concepts and structure.

There are a few goals being covered here:
- RCs should use their ID label to find their "current" nodes, rather than their set of pod labels. Otherwise it becomes impossible for two RCs to have the same set of pod labels.
- rcstore should be capable of taking locks on replication controllers. To avoid collisions, we move intent-tree locks to `/lock/intent/<node>/<pod>` and put the new locks under `/lock/replication_controller`. I would have also liked to do some refactors to lock and session management to make the API better (essentially I would have created a new package `store/lockstore` for it), but I don't have a good idea of how long that would take right now and I'm still getting familiar with the sessionmanager api.
- rcstore should be able to watch for new replication controllers. This inevitably requires a list operation so it might as well return the entire fields.RC object.

I think these changes are the key requirements to implement the rc farm (if nobody has any suggestions I'm going to call it the `RCManager`) which is in progress.

### DIFF
--- a/pkg/kp/rcstore/consul_store.go
+++ b/pkg/kp/rcstore/consul_store.go
@@ -121,6 +121,42 @@ func (s *consulStore) List() ([]fields.RC, error) {
 	return rcs, nil
 }
 
+func (s *consulStore) WatchNew(quit <-chan struct{}) (<-chan []fields.RC, <-chan error) {
+	outCh := make(chan []fields.RC)
+	errCh := make(chan error)
+
+	go func() {
+		defer close(outCh)
+		defer close(errCh)
+		var currentIndex uint64 = 0
+
+		for {
+			select {
+			case <-quit:
+				return
+			case <-time.After(1 * time.Second):
+				listed, meta, err := s.kv.List(kp.RC_TREE, &api.QueryOptions{
+					WaitIndex: currentIndex,
+				})
+				if err != nil {
+					// TODO: wrap error types for security
+					errCh <- err
+				} else {
+					currentIndex = meta.LastIndex
+					rcMap := s.kvpsToRcs(listed)
+					rcs := make([]fields.RC, 0, len(rcMap))
+					for _, rc := range rcMap {
+						rcs = append(rcs, *rc)
+					}
+					outCh <- rcs
+				}
+			}
+		}
+	}()
+
+	return outCh, errCh
+}
+
 // TODO: refactor pkg/kp/lock.go and pkg/kp/session.go into their own package,
 // and use that instead of c+ping it here
 // we are intentionally not using kp.Lock, because kp.Lock manages its own
--- a/pkg/kp/rcstore/consul_store.go
+++ b/pkg/kp/rcstore/consul_store.go
@@ -10,6 +10,7 @@ import (
 	"github.com/square/p2/Godeps/_workspace/src/github.com/hashicorp/consul/api"
 	"github.com/square/p2/Godeps/_workspace/src/github.com/pborman/uuid"
 
+	"github.com/square/p2/pkg/kp"
 	"github.com/square/p2/pkg/labels"
 	"github.com/square/p2/pkg/logging"
 	"github.com/square/p2/pkg/pods"
@@ -26,6 +27,7 @@ type consulKV interface {
 	Keys(prefix string, separator string, opts *api.QueryOptions) ([]string, *api.QueryMeta, error)
 	List(prefix string, opts *api.QueryOptions) (api.KVPairs, *api.QueryMeta, error)
 	Put(pair *api.KVPair, opts *api.WriteOptions) (*api.WriteMeta, error)
+	Acquire(pair *api.KVPair, opts *api.WriteOptions) (bool, *api.WriteMeta, error)
 	DeleteTree(prefix string, opts *api.WriteOptions) (*api.WriteMeta, error)
 }
 
@@ -117,6 +119,19 @@ func (s *consulStore) List() ([]fields.RC, error) {
 	}
 
 	return rcs, nil
+}
+
+// TODO: refactor pkg/kp/lock.go and pkg/kp/session.go into their own package,
+// and use that instead of c+ping it here
+// we are intentionally not using kp.Lock, because kp.Lock manages its own
+// session and therefore it cannot cooperate with kp.ConsulSessionManager
+func (s *consulStore) Lock(id fields.ID, session string) (bool, error) {
+	success, _, err := s.kv.Acquire(&api.KVPair{
+		Key:     kp.LockPath(kp.RCPath(id.String())),
+		Value:   []byte(session),
+		Session: session,
+	}, nil)
+	return success, err
 }
 
 func (s *consulStore) Disable(id fields.ID) error {
--- a/pkg/kp/rcstore/consul_store.go
+++ b/pkg/kp/rcstore/consul_store.go
@@ -16,8 +16,6 @@ import (
 	"github.com/square/p2/pkg/rc/fields"
 )
 
-const kvPrefix = "replication_controllers"
-
 type kvPair struct {
 	key   string
 	value []byte
@@ -90,7 +88,7 @@ func (s *consulStore) Create(manifest pods.Manifest, nodeSelector labels.Selecto
 }
 
 func (s *consulStore) Get(id fields.ID) (fields.RC, error) {
-	listed, _, err := s.kv.List(idPrefix(id), nil)
+	listed, _, err := s.kv.List(kp.RCPath(id.String()), nil)
 	if err != nil {
 		return fields.RC{}, nil
 	}
@@ -106,7 +104,7 @@ func (s *consulStore) Get(id fields.ID) (fields.RC, error) {
 }
 
 func (s *consulStore) List() ([]fields.RC, error) {
-	listed, _, err := s.kv.List(kvPrefix, nil)
+	listed, _, err := s.kv.List(kp.RC_TREE, nil)
 	if err != nil {
 		return []fields.RC{}, nil
 	}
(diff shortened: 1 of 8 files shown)


pkg/kp/rcstore/consul_store.go
```diff
@@ -10,13 +10,20 @@ import (
...
+	"github.com/square/p2/pkg/kp"
 	"github.com/square/p2/pkg/labels"
 	"github.com/square/p2/pkg/logging"
 	"github.com/square/p2/pkg/pods"
 	"github.com/square/p2/pkg/rc/fields"
 )
 
-const kvPrefix = "replication_controllers"
+const (
+	RC_TREE string = "intent"
```
anthonybishopric: typo

petertseng:
> rcstore should be able to watch for new replication controllers. This inevitably requires a list operation so it might as well return the entire fields.RC object.

Theoretically... if you *were* able to use the `Keys` operation instead of a `List` operation (assuming `Keys` took a `ModifyIndex`), would you want to?

pkg/kp/rcstore/consul_store.go
```diff
@@ -121,6 +129,55 @@ func (s *consulStore) List() ([]fields.RC, error) {
 	return rcs, nil
 }
 
+func (s *consulStore) WatchNew(quit <-chan struct{}) (<-chan []fields.RC, <-chan error) {
+	outCh := make(chan []fields.RC)
+	errCh := make(chan error)
+	defer close(outCh)
```
petertseng: Does this immediately close them upon return from this function? I think you want to close them upon return from the goroutine instead.

pkg/kp/rcstore/consul_store.go
tummychow: yeah whoopsies. finally having something separate from kp kinda highlights how there are too many things in kp

tummychow:
> Theoretically... if you were able to use the Keys operation instead of a List operation (assuming Keys took a ModifyIndex), would you want to?

Nah. If you do end up claiming an RC you have to get its fields one way or another. Either you get them all up front (and skip most of them because they're already locked by someone) or you get just the keys up front and get the whole field set for each one you lock. I think one larger request will be slightly better than two small requests in practice.

pkg/kp/rcstore/consul_store.go
tummychow: argh you're right, will fix

anthonybishopric:
Seems good :+1:

petertseng:
Yeah seems good from where I'm sitting

tummychow:
will clean up commit order and test functionality before merging. Might also deal with lock refactor on this PR, we'll see.

mpuncel:
keep in mind there are some changes to the lock API in my other PR (like the library handling renewals for you)

pkg/kp/rcstore/consul_store.go
```diff
@@ -10,13 +10,20 @@ import (
...
+	"github.com/square/p2/pkg/kp"
 	"github.com/square/p2/pkg/labels"
 	"github.com/square/p2/pkg/logging"
 	"github.com/square/p2/pkg/pods"
 	"github.com/square/p2/pkg/rc/fields"
 )
 
-const kvPrefix = "replication_controllers"
+const (
+	RC_TREE string = "replication_controller"
```
mpuncel: since the number of these are growing i wonder if they should be centralized somewhere and imported

tummychow: My dream would be to fragment kp into several packages by type (pods, health, sessions/locks, rcs, orchestrator-pairs) and the top level package would only contain the constants and the general-purpose consul client initialization code. However that change is probably too big to pursue atm. For now I think a good policy is to only add new "tree types" as subpackages (as was done with rcstore) and slowly migrate other trees to subpackages as the opportunity arises (hopefully I can do that for locks). If we don't add any more trees to the main package then eventually it'll shrink to something more manageable.

As for whether RC_TREE should be in rcstore or kp/constants.go, I'm pretty indifferent. Originally I put it here because rcstore didn't need to import kp, but now that it uses LockPath, that's no longer relevant. I'll probably move it back to constants.go since it makes sense for them to be in the same place.

tummychow:
> keep in mind there are some changes to the lock API in my other PR (like the library handling renewals for you)

#229? I don't actually see any changes to `lock.go` or `session.go` there. Clarification on what I want: Right now `lock.go` owns a consul session and the routines that keep the session alive. This API is easy to use when the lock is:
- short lived (process will exit fairly quickly and release the lock)
- necessary to survive (losing the lock is probably too damaging to be worth the effort of recovering automatically)
- not shared (only one user of the lock at a time, therefore there is no need to coordinate lock creation across goroutines)

However a long-lived service that depends on holding locks for a very long time needs more granular control over lock renewal and recovery. We already have the ConsulSessionManager in in `session.go` for this (which was made for the preparer's health check sessions, so it's well suited to the problem). I still like the functions exposed by the kp.Lock object, but because it owns its own session, it can't work with the session manager.

So my refactoring plan goes like this:
- kp/lock.go no longer manages its own session. You simply construct it by passing it a session ID and it does not renew the session. This is treated as the responsibility of the caller. For the most part I don't expect the API of the kp.Lock object to change.
- pkg/replication uses a ConsulSessionManager to generate the initial session for its lock rather than letting the lock refresh itself.
- move lock.go and session.go into their own separate packages

Now I don't think any of this needs to be prioritized, since it doesn't block me. I think it would be a good change in general though.

pkg/rc/replication_controller.go
```diff
@@ -184,7 +184,7 @@ func (rc *replicationController) eligibleNodes() ([]string, error) {
 }
 
 func (rc *replicationController) CurrentNodes() ([]string, error) {
-	selector := rc.PodLabels.AsSelector()
+	selector := labels.Everything().Add(rcIdLabel, labels.EqualsOperator, []string{rc.Id().String()})
```
petertseng: ah that's right, before I forget... could you change the doc comments of `rcstore.Store.Create` accordingly? The comments say the pod labels are used, now they're not.

mpuncel:
> #229? I don't actually see any changes to lock.go or session.go there.

Yeah that's what I meant although I forgot that I had split out the lock changes and they're already merged.

pkg/kp/rcstore/consul_store.go
```diff
@@ -10,13 +10,20 @@ import (
...
+)
+
+func RcPath(args ...string) string {
```
bwester: nit: golang typically uses all-caps with abbreviations, and we use "RC" elsewhere too.  Can this be `RCPath()`?

pkg/replication/replication.go
```diff
@@ -39,7 +39,7 @@ func (r Replicator) CheckPreparers() error {
 // wish to lock
 func (r Replicator) LockHosts(lock kp.Lock, overrideLock bool) error {
 	for _, host := range r.Nodes {
-		lockPath := kp.LockPath(host, r.Manifest.ID())
+		lockPath := kp.LockPath(kp.IntentPath(host, r.Manifest.ID()))
```
bwester: With this modification, any binary compiled before this change will not recognize the locks created by binaries compiled after this change, and vice versa.  Do you see this being a problem in production?  (E.g., how many things do we need to simultaneously redeploy and restart for this to not cause problems?)

pkg/kp/rcstore/store.go
```diff
@@ -21,6 +21,19 @@ type Store interface {
...
+	// terminate. Callers must drain the returned channels, including the error
+	// channel.
+	WatchNew(quit <-chan struct{}) (<-chan []fields.RC, <-chan error)
+
+	// Use the given session string (which uniquely identifies the lock holder)
+	// to acquire a lock on the specified replication controller. If the lock
+	// was successfully claimed, return (true, nil). If the request completed
+	// but the lock was already held, return (false, nil). Otherwise return an
+	// appropriate error.
+	Lock(id fields.ID, session string) (bool, error)
```
bwester: Are there any guidelines that callers should follow regarding the session used to acquire the lock?  Any guidelines for what happens if the session is terminated?

pkg/kp/rcstore/fake_store.go
```diff
@@ -72,6 +73,25 @@ func (s *fakeStore) List() ([]fields.RC, error) {
 	return results, nil
 }
 
+func (s *fakeStore) WatchNew(quit <-chan struct{}) (<-chan []fields.RC, <-chan error) {
+	go func() {
```
bwester: This goroutine isn't needed.

tummychow: yeah you're right, will fix. I wasn't sure what the fake's watchnew should look like but it probably doesn't matter

pkg/kp/rcstore/store.go
tummychow: I want the function to be as decoupled from the session as possible; therefore the caller instantiates the session and only they know what happens (ie keys are either deleted or released) when the session expires. I'll mention that the Lock key is ephemeral, therefore delete-on-expire sessions are safe to use.

pkg/replication/replication.go
tummychow: In practice I very much doubt it's a problem. So far, it looks to me like lock conflicts in prod usually come from a deploy getting stuck or cancelled. The actual case of two people independently, unknowingly kicking off a deploy for the same pod and same node (which is where locking would save them from each other) seems quite rare.

It's also worth mentioning that replication locks are only held for the duration of the replication, so it's not like we have long-lived sessions in prod that would make the path change difficult. I don't think it'll end up being an issue.

pkg/kp/rcstore/consul_store.go
tummychow: yeah will fix

pkg/rc/replication_controller.go
tummychow: will fix

tummychow:
cleaned up the branch.

one more thought to address before merging: I notice that a lot of consul errors in rcstore/consul_store.go are silenced. Should we be using kp.KVError there instead?

tummychow:
rebased over mpuncel's replication pr

tummychow:
rebased again

petertseng:
still looking good from my point of view, still approve

PR #278, opened 2015-10-02
Fix bugs in callers of WatchPods()

mpuncel:
communicate an empty []ManifestResult

### DIFF
--- a/bin/p2-watch/main.go
+++ b/bin/p2-watch/main.go
@@ -64,9 +64,13 @@ func main() {
 	for {
 		select {
 		case results := <-podCh:
-			for _, result := range results {
-				fmt.Println("")
-				result.Manifest.Write(os.Stdout)
+			if len(results) == 0 {
+				fmt.Println(fmt.Sprintf("No manifest exists at key %s (it may have been deleted)", path))
+			} else {
+				for _, result := range results {
+					fmt.Println("")
+					result.Manifest.Write(os.Stdout)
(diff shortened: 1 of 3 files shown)


bwester:
super nit: Please use [standard](https://github.com/blog/926-shiny-new-commit-styles) [git](http://tbaggery.com/2008/04/19/a-note-about-git-commit-messages.html) [formatting](http://chris.beams.io/posts/git-commit/) in your commit message, and take the opportunity to [explain why](https://robots.thoughtbot.com/5-useful-tips-for-a-better-commit-message) your diff does what it does.

bwester:
In isolation, it seems like highly desirable behavior that if a caller wants to watch an object for changes, it get notified when the object is deleted.  Why do we want to mask this?

mpuncel:
I was restoring the behavior of `WatchPods()` prior to https://github.com/square/p2/pull/248/files. It is useful to be notified of deletions though like you said, i'll look at changing consumers of this function to properly handle an empty result slice.

mpuncel:
@petertseng I'm not even sure how to handle the situation where length > 1. Which manifest would I use if there is more than one? bubble up an error and block for probably forever?

tummychow:
yeah let's log something if the length is greater than 1, other than that looks fine

mpuncel:
will merge once green

bwester:
:+1:

PR #279, opened 2015-10-02
Add launch.Launchable interface

bwester:
Extracts `hoist.Launchable` methods into an interface `launch.Launchable` in
preparation for creating new types of launchables.

The interface is placed into its own package to avoid circular or unintuitive
dependencies. The package name "launch" was chosen to avoid having to rewrite
every "launchable" identifier used as a local variable.

It explicitly does *not* attempt to improve upon the basic set of methods
exposed by `hoist.Launchable`, though with much further thought and another
implementation, there may be ample room for improvement.


mpuncel:
:+1:

pkg/pods/pod.go
```diff
@@ -100,12 +101,12 @@ func (pod *Pod) Halt(manifest Manifest) (bool, error) {
...
-		case hoist.DisableError:
+		case launch.DisableError:
 			// do not set success to false on a disable error
-			pod.logLaunchableWarning(launchable.Id, err, "Could not disable launchable")
+			pod.logLaunchableWarning(launchable.GetID(), err, "Could not disable launchable")
 		default:
-			// this case intentionally includes hoist.StopError
-			pod.logLaunchableError(launchable.Id, err, "Could not halt launchable")
+			// this case intentionally includes launch.StopError
+			pod.logLaunchableError(launchable.GetID(), err, "Could not halt launchable")
```
anthonybishopric: Consider `ID()` to be consistent with `Pod` and the other getters on the struct?

anthonybishopric:
:+1: except for nit

pkg/pods/pod.go
bwester: There's a naming conflict between `host.Launchable` having an `Id` field and an `ID()` method.  It's not a strict conflict in this case, but it's too close for my comfort.

But you reminded me that we do have a convention of interfaces with `ID()`, so I'm more inclined now to rename the field.

petertseng:
On an unrelated note, it may be appropriate for `fields.RC` 's `Id()` method to be renamed to `ID()` ... separate PR of course

tummychow:
> On an unrelated note, it may be appropriate for fields.RC 's Id() method to be renamed to ID() ... separate PR of course

I'll take care of it in refactors part 2

pkg/pods/pod.go
bwester: Naming getters without the "Get" is a an explicit [language convention](https://golang.org/doc/effective_go.html#Getters), but it's less clear what to do when you really want a public field with that name too. Any strong opinions on naming a public field `ID_` and its getter `ID()`?

bwester: Never mind.  I should probably just do the right thing and not have any of these fields be exported.

PR #280, opened 2015-10-02
Orchestrator-related refactors round 2

tummychow:
Replication controllers are stored in a single key now with CAS mutation. The similarity to the pkg/labels API is remarkable.

Might have other stuff incoming on this PR. Have not tested yet, but am about to do so.

### DIFF
--- a/pkg/kp/rcstore/consul_store.go
+++ b/pkg/kp/rcstore/consul_store.go
@@ -58,7 +58,7 @@ func (s *consulStore) Create(manifest pods.Manifest, nodeSelector labels.Selecto
 	}
 
 	rc := fields.RC{
-		Id:              id,
+		ID:              id,
 		Manifest:        manifest,
 		NodeSelector:    nodeSelector,
 		PodLabels:       podLabels,
@@ -220,7 +220,7 @@ func (s *consulStore) Watch(rc *fields.RC, quit <-chan struct{}) (<-chan struct{
 			case <-quit:
 				return
 			case <-time.After(1 * time.Second):
-				pairs, meta, err := s.kv.List(kp.RCPath(rc.Id.String()), &api.QueryOptions{
+				pairs, meta, err := s.kv.List(kp.RCPath(rc.ID.String()), &api.QueryOptions{
 					WaitIndex: curIndex,
 				})
 				if err != nil {
@@ -230,7 +230,7 @@ func (s *consulStore) Watch(rc *fields.RC, quit <-chan struct{}) (<-chan struct{
 
 					rcMap := s.kvpsToRcs(pairs)
 
-					if newRc, ok := rcMap[rc.Id]; ok {
+					if newRc, ok := rcMap[rc.ID]; ok {
 						*rc = *newRc
 						updated <- struct{}{}
 					}
@@ -344,7 +344,7 @@ func (s *consulStore) kvpsToRcs(kvps api.KVPairs) map[fields.ID]*fields.RC {
 // If forEachLabel encounters any error applying the function, it returns that error immediately.
 // The function is not further applied to subsequent labels on an error.
 func (s *consulStore) forEachLabel(rc fields.RC, f func(id, k, v string) error) error {
-	id := rc.Id.String()
+	id := rc.ID.String()
 	// As of this writing the only label we want is the pod ID.
 	// There may be more in the future.
 	return f(id, "pod_id", rc.Manifest.ID())
--- a/pkg/kp/rcstore/consul_store.go
+++ b/pkg/kp/rcstore/consul_store.go
@@ -1,10 +1,8 @@
 package rcstore
 
 import (
-	"bytes"
+	"encoding/json"
 	"fmt"
-	"strconv"
-	"strings"
 	"time"
 
 	"github.com/square/p2/Godeps/_workspace/src/github.com/hashicorp/consul/api"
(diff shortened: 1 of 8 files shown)


pkg/kp/rcstore/consul_store.go
```diff
@@ -26,39 +24,66 @@ type consulKV interface {
...
 	kv         consulKV
 	logger     logging.Logger
+	retries    int
+}
+
+// TODO: combine with similar CASError type in pkg/labels
+type CASError string
+
+func (e CASError) Error() string {
+	return fmt.Sprintf("Could not compare-and-set key %q", string(e))
```
mpuncel: super nit i'm used to this being called "check-and-set" and seems like that's the terminology consul's documentation uses

tummychow: I can fix the text in both places (here and pkg/labels) if you prefer

pkg/kp/rcstore/consul_store.go
```diff
@@ -26,39 +24,66 @@ type consulKV interface {
...
+		}
+	}
+	if err != nil {
+		return fields.RC{}, err
+	}
 
-	buf := bytes.Buffer{}
-	err := manifest.Write(&buf)
+	// labels do not need to be retried, consul applicator does that itself
+	err = s.forEachLabel(rc, func(id, k, v string) error {
```
mpuncel: I don't feel strongly but maybe consider retrying here instead of within the applicator? Depends on how many things are going to use the applicator. Probably good the way it is

pkg/kp/rcstore/consul_store.go
```diff
@@ -66,59 +91,42 @@ func (s *consulStore) Create(manifest pods.Manifest, nodeSelector labels.Selecto
...
 
-	err = s.forEachLabel(rc, func(id, k, v string) error {
-		return s.applicator.SetLabel(labels.RC, id, k, v)
-	})
-
-	// TODO: If the `put` operations fail, we have already labeled the RC,
-	// yet the RC will not exist in the backing Consul KV store.
-	// The labels may have to be cleaned up.
-
-	err = s.put(id, []kvPair{
```
mpuncel: where did this stuff go

tummychow: made redundant, keys are now marshaled into a single JSON object. see changes in fields.go

pkg/kp/rcstore/consul_store.go
```diff
@@ -171,39 +197,92 @@ func (s *consulStore) Lock(id fields.ID, session string) (bool, error) {
...
+		newRC := mutator(rc)
+		b, err := json.Marshal(newRC)
+		if err != nil {
+			return err
+		}
+		newKVP.Value = b
+		success, _, err = s.kv.CAS(newKVP, nil)
+	} else {
+		// special logic for the delete case
+		if rc.ReplicasDesired != 0 {
```
mpuncel: at some point we will also want to make sure that the replication controller has converged on its desired replicas before deleting

tummychow: yeah, I'm probably going to figure out how to write the "block until RC is healthy" function and append that to this PR. However while detecting the state of newly added nodes is easy, detecting that removed nodes are actually gone is a lot less obvious. Still thinking about how to implement.

pkg/kp/rcstore/consul_store.go
petertseng: Apparently there's compare and swap, compare and set, check and set, test and set, compare and exchange...

Anyway I'm not super fussed, so whatever y'all prefer. Making us consistent with Consul's terminology seems like a good rationale to change it.

pkg/rc/fields/fields.go
```diff
@@ -12,10 +14,64 @@ func (id ID) String() string {
...
+	ID              ID
+	Manifest        []byte
+	NodeSelector    string
+	PodLabels       labels.Set
+	ReplicasDesired int
+	Disabled        bool
+}
+
+// implements encoding/json.{Marshaler,Unmarshaler}
+// Unmarshaler is especially problematic because the pods.Manifest API is
```
mpuncel: what's the problematic part?

mpuncel:
lgtm once CI passes

pkg/rc/fields/fields.go
tummychow: if you implement `UnmarshalJSON` on `pods.Manifest`, you can use that method to mutate the object. I think it feels kinda icky to put that there since it was intentionally designed to be immutable through that interface. At the same time, this is also gross. wasn't really sure which option was grosser

pkg/kp/rcstore/consul_store.go
```diff
@@ -26,39 +24,66 @@ type consulKV interface {
...
+	}
 
-	buf := bytes.Buffer{}
-	err := manifest.Write(&buf)
+	// labels do not need to be retried, consul applicator does that itself
+	err = s.forEachLabel(rc, func(id, k, v string) error {
+		return s.applicator.SetLabel(labels.RC, rc.ID.String(), k, v)
+	})
 	if err != nil {
 		return fields.RC{}, err
```
petertseng: while we're thinking about this, if you have a good solution for the problem of "RC successfully gets written into Consul, but labels fail for some reason" (now the RC watcher will see the new RC and presumably make it run, but it will not have labels which reduces visibility on it), I'd be interested to hear.

I think what this comment boils down to is "sometimes I wish Consul had transactions"???

petertseng: actually now we have a decision to make: If we have to pick an ordering of the two (apply labels, create RC) and have a failure occur between the two, which scenario would be less bad? My thinking was a labeled-but-not-actually-existent RC would be less bad than an existing-but-unlabeled RC so I did apply labels then create RC, whereas now this code is doing create RC then apply labels.

I'm not dead set on one option over the other, but we should figure out which we want.

tummychow: tell me about it, most of this PR would be unnecessary if we had transactions

I think for now, all we can do is look for unlabeled RCs (or labels for nonexistent RCs) and prune them semiregularly. Hopefully it'll be rare in practice since labels retry on CAS failure.

mpuncel: i don't think there's anything we can do that isn't at least partially crazy.

1) can implement transactions ourselves, unapplying any successful steps after a later step fails. Although realistically if something failed we're not going to be able to recover from it either
2) write the desired state of the RC in consul somewhere and have another process repeatedly make it so forever. we use this design for deploys and will be using this design for RC-style rollouts, but probably way way overkill for something like this.

Realistically, a sentry alert and human intervention will probably suffice, and the frequency of something like this should be very low

pkg/kp/rcstore/consul_store.go
```diff
@@ -66,59 +91,42 @@ func (s *consulStore) Create(manifest pods.Manifest, nodeSelector labels.Selecto
...
-		kvPair{key: "node_selector", value: []byte(nodeSelector.String())},
-		kvPair{key: "pod_labels", value: []byte(podLabels.String())},
-		kvPair{key: "pod_manifest", value: buf.Bytes()},
-		kvPair{key: "replicas_desired", value: []byte("0")},
-	})
-
+	jsonRC, err := json.Marshal(rc)
 	if err != nil {
-		return fields.RC{}, err
+		return fields.RC{}, nil
```
petertseng: so this is deliberately not returning the error, when there is one. is it intentional?

tummychow: that's a typo will fix

pkg/kp/rcstore/consul_store.go
```diff
@@ -66,59 +91,42 @@ func (s *consulStore) Create(manifest pods.Manifest, nodeSelector labels.Selecto
...
 	if err != nil {
-		return []fields.RC{}, nil
-	}
-
-	rcMap := s.kvpsToRcs(listed)
-	rcs := make([]fields.RC, 0)
-
-	for _, rc := range rcMap {
-		rcs = append(rcs, *rc)
+		return nil, nil
```
petertseng: here also no error is being returned when there is one. was it also unintentional?

tummychow: that's a typo that i inherited from the singular version of the function which also has a typo in it, all of which i'll fix

pkg/kp/rcstore/consul_store.go
```diff
@@ -171,39 +197,92 @@ func (s *consulStore) Lock(id fields.ID, session string) (bool, error) {
...
+		if _, ok := err.(CASError); ok {
+			err = s.mutateRc(id, mutator)
+		} else {
+			break
+		}
+	}
+	return err
+}
+
+// performs a safe (ie compare-and-set) mutation of the rc with the given id,
```
petertseng: since we were talking about CAS, don't forget this one

pkg/kp/rcstore/consul_store.go
```diff
@@ -171,39 +197,92 @@ func (s *consulStore) Lock(id fields.ID, session string) (bool, error) {
...
+
+		// TODO: If this fails, then we have some dangling labels.
+		// Perhaps they can be cleaned up later.
+		err = s.forEachLabel(rc, func(id, k, _ string) error {
+			return s.applicator.RemoveLabel(labels.RC, id, k)
+		})
+		if err != nil {
+			return err
+		}
+		success, _, err = s.kv.CAS(newKVP, nil)
```
petertseng: sure about `CAS` rather than `DeleteCAS`?

tummychow: damn I'm making typos everywhere today

pkg/kp/rcstore/consul_store.go
```diff
@@ -220,7 +299,7 @@ func (s *consulStore) Watch(rc *fields.RC, quit <-chan struct{}) (<-chan struct{
 			case <-quit:
 				return
 			case <-time.After(1 * time.Second):
-				pairs, meta, err := s.kv.List(kp.RCPath(rc.Id.String()), &api.QueryOptions{
+				pairs, meta, err := s.kv.List(kp.RCPath(rc.ID.String()), &api.QueryOptions{
```
petertseng: Now that all fields of an RC reside under a single key, is it possible we can use `Get` rather than `List`?

tummychow: ah yes good catch, i changed this for `store.Get` but not here

tummychow:
cleaned up commit order

petertseng:
The code looks good. I want to test but if you don't want to wait for me I suppose I wouldn't blame you

petertseng:
It works (Assuming `RC_TREE` is re-pluralized), this is good to go from my perspective after that.

petertseng:
OK never mind, I was on an old version (one without 698174e414467c4b3a4b27d4c8d9532016022df2 , notably) - you're already working on that though. I'll take a look again when you have that figured out

petertseng:
Tested f915cd7e9ef290f72a149c346b6d1691e94d7afb (pretty much testing IN PROD) and it worked so I say this is good

PR #281, opened 2015-10-05
Add RC manager

tummychow:
code mostly speaks for itself

### DIFF
--- a/bin/p2-rcfarm/main.go
+++ b/bin/p2-rcfarm/main.go
@@ -1,0 +1,55 @@
+package main
+
+import (
+	"net/http"
+	"time"
+
+	"github.com/square/p2/Godeps/_workspace/src/github.com/Sirupsen/logrus"
+	"github.com/square/p2/Godeps/_workspace/src/github.com/hashicorp/consul/api"
+	"github.com/square/p2/Godeps/_workspace/src/gopkg.in/alecthomas/kingpin.v1"
+
+	"github.com/square/p2/pkg/kp"
+	"github.com/square/p2/pkg/kp/rcstore"
+	"github.com/square/p2/pkg/labels"
+	"github.com/square/p2/pkg/logging"
+	"github.com/square/p2/pkg/rc"
+	"github.com/square/p2/pkg/util/net"
+	"github.com/square/p2/pkg/version"
+)
+
+var (
+	consulUrl   = kingpin.Flag("consul", "The hostname and port of a consul agent in the p2 cluster. Defaults to 0.0.0.0:8500.").String()
+	consulToken = kingpin.Flag("token", "The consul ACL token to use. Empty by default.").String()
+	headers     = kingpin.Flag("header", "An HTTP header to add to requests, in KEY=VALUE form. Can be specified multiple times.").StringMap()
+	https       = kingpin.Flag("https", "Use HTTPS").Bool()
+)
+
+func main() {
+	kingpin.Version(version.VERSION)
+	kingpin.Parse()
+
+	opts := kp.Options{
+		Address: *consulUrl,
+		Token:   *consulToken,
+		Client:  net.NewHeaderClient(*headers, http.DefaultTransport),
+		HTTPS:   *https,
+	}
+	kpStore := kp.NewConsulStore(opts)
+	client := kp.NewConsulClient(opts)
+	rcStore := rcstore.NewConsul(client, 3, logging.DefaultLogger)
+	labeler := labels.NewConsulApplicator(client, 3)
+	sched := rc.NewApplicatorScheduler(labeler)
+
+	sessionCh := make(chan string)
+	quitCh := make(chan struct{})
+	logging.DefaultLogger.Logger.Level = logrus.DebugLevel
+
+	go kp.ConsulSessionManager(api.SessionEntry{
+		LockDelay: 1 * time.Nanosecond,
+		Behavior:  api.SessionBehaviorDelete,
+		TTL:       "15s",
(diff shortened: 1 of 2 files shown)


pkg/rc/manager.go
```diff
@@ -0,0 +1,183 @@
...
+	"github.com/square/p2/pkg/logging"
+	"github.com/square/p2/pkg/rc/fields"
+)
+
+// The RCManager is responsible for spawning and reaping replication controllers
+// as they are added to and deleted from Consul. Multiple managers can exist
+// simultaneously, but each one must hold a different Consul session. This
+// ensures that the managers do not instantiate the same replication controller
+// multiple times.
+type RCManager struct {
```
anthonybishopric: Personal beef: I hate the word "manager" and would prefer to limit its use as much as possible. The only reason I was okay with "SessionManager" is because it has precedent in other frameworks. I prefer your term "RCFarm" honestly.

tummychow: I sympathize with that and I am also okay with the name `RCFarm`. thoughts from anyone else? otherwise I'll change it

pkg/rc/manager.go
```diff
@@ -0,0 +1,183 @@
...
+				// TODO: restart the watch so that you get updates right away?
+			}
+		case err := <-rcErr:
+			rcm.logger.WithError(err).Errorln("Could not read consul replication controllers")
+		case rcFields := <-rcWatch:
+			rcm.logger.WithField("n", len(rcFields)).Debugln("Received replication controller update")
+			if session == "" {
+				// we can't claim new nodes because our session is invalidated.
+				// raise an error and ignore this update
+				rcm.logger.NoFields().Infoln("Received replication controller update, but do not have session to acquire locks")
```
anthonybishopric: Should this be a `Warnln`? Your comment suggests it should be an error case.

pkg/rc/manager.go
```diff
@@ -0,0 +1,183 @@
...
+				continue
```
anthonybishopric: Can we recover if we throw away results from `rcWatch` when we don't have a session at the time when they're received? What if we get a session on the next iteration of this loop?

pkg/rc/manager.go
tummychow: hm, well the error was already logged earlier when the session expired, but probably infoln is too kind. yeah I'll bump up to warn

pkg/rc/manager.go
tummychow: sure, this doesn't kill the loop. you'll just have to wait for the next batch of results come in, and then you'll start acquiring locks as usual again. The only problem is that it might take a while for the next batch if there are no changes to trigger the watch (you would have to wait for it to timeout). See my TODO comment about restarting the watch when a new session is obtained.

pkg/rc/manager.go
```diff
@@ -0,0 +1,183 @@
...
+			}
+
+			// track which children were found in the returned set
+			foundChildren := make(map[fields.ID]struct{})
```
petertseng: you could use `bool` rather than `struct{}` and take advantage of the fact that the zero value is `false` (when checking) if you prefer that

pkg/rc/manager.go
petertseng: what about replication controller controller

ok I jest, rcfarm seems fine

mpuncel: :+1:

mpuncel:
lgtm

pkg/rc/manager.go
```diff
@@ -0,0 +1,183 @@
...
+				}
+			}
+		}
+	}
+}
+
+// close one child
+func (rcm *RCManager) releaseChild(id fields.ID) {
+	rcm.logger.WithField("rc_id", id).Infoln("Releasing replication controller")
+	close(rcm.children[id].quit)
```
anthonybishopric: I wish I knew looking at a channel's close operation whether it's appropriate for the clients using the channel. For instance, I can't tell if the other side is doing a `for.. range` over the channel or if it's expecting to read a quit value and/or if it's going to send an ack.

pkg/rc/manager.go
```diff
@@ -0,0 +1,183 @@
...
+					rcm.releaseChild(id)
+				}
+			}
+		}
+	}
+}
+
+// close one child
+func (rcm *RCManager) releaseChild(id fields.ID) {
+	rcm.logger.WithField("rc_id", id).Infoln("Releasing replication controller")
```
anthonybishopric: Probably want to verify that the channel child is valid.

anthonybishopric:
Overall :+1:

pkg/rc/manager.go
tummychow: that's a valid question in general, but i think the answer is:

- channels used primarily for quit should almost never be sensitive to the value being sent (and in our code they're all `struct{}` so by definition there is only one value that can be sent)
- range loop is terminated by close
- channels that are used in both directions should be very rare, and very obviously named

pkg/rc/manager.go
tummychow: clarify?

anthonybishopric: As in, `id` is actually managed by this farm

tummychow: I know this is a copout answer, but it should be guaranteed by construction - elements only get added to this map after we obtain the lock on them, and if we ever lose a lock or a key gets deleted, then we also remove from the map. There's not much else to check without writing a more fluent lock API - which I will probably do in the near future, but not yet

tummychow:
renamed to farm, upgraded one infoln to warnln - should be good for merge

PR #282, opened 2015-10-06
Add p2-rctl

tummychow:
Supports create, delete, set replicas, list and get. I decided not to support disable because there's no enable method right now. Might add shortly. Other suggestions for what to add are also welcome.

e: before I forget, I should also adjust the rcstore api to support force-deletion (ignoring open replica count) and implement that here as well

### DIFF
--- a/bin/p2-rctl/main.go
+++ b/bin/p2-rctl/main.go
@@ -1,0 +1,179 @@
+package main
+
+import (
+	"encoding/json"
+	"fmt"
+	"net/http"
+
+	"github.com/square/p2/Godeps/_workspace/src/github.com/Sirupsen/logrus"
+	"github.com/square/p2/Godeps/_workspace/src/gopkg.in/alecthomas/kingpin.v1"
+
+	"github.com/square/p2/pkg/kp"
+	"github.com/square/p2/pkg/kp/rcstore"
+	"github.com/square/p2/pkg/labels"
+	"github.com/square/p2/pkg/logging"
+	"github.com/square/p2/pkg/pods"
+	"github.com/square/p2/pkg/rc/fields"
+	"github.com/square/p2/pkg/util/net"
+	"github.com/square/p2/pkg/version"
+)
+
+const (
+	CMD_CREATE   = "create"
+	CMD_DELETE   = "delete"
+	CMD_REPLICAS = "set-replicas"
+	CMD_LIST     = "list"
+	CMD_GET      = "get"
+)
+
+var (
+	consulUrl   = kingpin.Flag("consul", "The hostname and port of a consul agent in the p2 cluster. Defaults to 0.0.0.0:8500.").String()
+	consulToken = kingpin.Flag("token", "The consul ACL token to use. Empty by default.").String()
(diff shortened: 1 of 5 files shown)


bin/p2-rctl/main.go
```diff
@@ -0,0 +1,179 @@
...
+)
+
+var (
+	consulUrl   = kingpin.Flag("consul", "The hostname and port of a consul agent in the p2 cluster. Defaults to 0.0.0.0:8500.").String()
+	consulToken = kingpin.Flag("token", "The consul ACL token to use. Empty by default.").String()
+	headers     = kingpin.Flag("header", "An HTTP header to add to requests, in KEY=VALUE form. Can be specified multiple times.").StringMap()
+	https       = kingpin.Flag("https", "Use HTTPS").Bool()
+
+	cmdCreate       = kingpin.Command(CMD_CREATE, "Create a new replication controller")
+	createManifest  = cmdCreate.Flag("manifest", "manifest file to use for this replication controller").Short('m').Required().String()
```
anthonybishopric: This is a pod manifest right

bin/p2-rctl/main.go
```diff
@@ -0,0 +1,179 @@
...
+	CMD_REPLICAS = "set-replicas"
+	CMD_LIST     = "list"
+	CMD_GET      = "get"
+)
+
+var (
+	consulUrl   = kingpin.Flag("consul", "The hostname and port of a consul agent in the p2 cluster. Defaults to 0.0.0.0:8500.").String()
+	consulToken = kingpin.Flag("token", "The consul ACL token to use. Empty by default.").String()
+	headers     = kingpin.Flag("header", "An HTTP header to add to requests, in KEY=VALUE form. Can be specified multiple times.").StringMap()
+	https       = kingpin.Flag("https", "Use HTTPS").Bool()
```
anthonybishopric: Should these just go into some shared global flags file?

bin/p2-rctl/main.go
tummychow: yeah

bin/p2-rctl/main.go
```diff
@@ -0,0 +1,208 @@
...
+	client := kp.NewConsulClient(opts)
+	rcStore := rcstore.NewConsul(client, 3, logger)
+	rctl := RCtl{
+		rcs:    rcStore,
+		logger: logger,
+	}
+
+	switch cmd {
+	case CMD_CREATE:
+		rctl.Create(*createManifest, *createNodeSel, *createPodLabels)
```
anthonybishopric: Since we have multiple subcommands we can't use Kingpin's `.Required()` directive for flags. Need parameter validation here.

bin/p2-rctl/main.go
tummychow: I was thinking about that, however nightmares of glog flags are convincing me that the answer is no. If kingpin had a way to make flags "silent" and then attach to the application, then I would consider it

bin/p2-rctl/main.go
tummychow: hm? The required directive on a given kingpin command only kicks in if that command is invoked.

anthonybishopric: Oh derp, you are totally right.

PR #283, opened 2015-10-07
Export InvokeBinScript() on the hoist.Launchable struct.

mpuncel:
This enables custom automation that makes use of various scripts not
directly related to the deploy lifecycle.


anthonybishopric:
:+1:

PR #284, opened 2015-10-07
Make hoist.Launchable.fetcher an unexported field

mpuncel:
This allows an interface conversion from launch.Launchable to
hoist.Launchable. The naming clash between the Fetcher() function on the
launch.Launchable() interface and the field on hoist.Launchable
prevented this previously.

### DIFF
--- a/pkg/pods/pod.go
+++ b/pkg/pods/pod.go
@@ -548,13 +548,13 @@ func (pod *Pod) getLaunchable(launchableStanza LaunchableStanza, runAsUser strin
 			Id:               launchableId,
 			RunAs:            runAsUser,
 			ConfigDir:        pod.EnvDir(),
-			Fetcher:          uri.DefaultFetcher,
 			RootDir:          launchableRootDir,
 			P2exec:           pod.P2exec,
 			RestartTimeout:   restartTimeout,
 			CgroupConfig:     launchableStanza.CgroupConfig,
 			CgroupConfigName: launchableStanza.LaunchableId,
 		}
+		ret.SetFetcher(uri.DefaultFetcher)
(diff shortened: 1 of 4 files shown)


pkg/pods/pod.go
```diff
@@ -548,13 +548,13 @@ func (pod *Pod) getLaunchable(launchableStanza LaunchableStanza, runAsUser strin
...
 			RunAs:            runAsUser,
 			ConfigDir:        pod.EnvDir(),
-			Fetcher:          uri.DefaultFetcher,
 			RootDir:          launchableRootDir,
 			P2exec:           pod.P2exec,
 			RestartTimeout:   restartTimeout,
 			CgroupConfig:     launchableStanza.CgroupConfig,
 			CgroupConfigName: launchableStanza.LaunchableId,
 		}
+		ret.SetFetcher(uri.DefaultFetcher)
```
mpuncel: I would have preferred renaming the field to having to make a method for this but couldn't think of a better name. Should this function be on the `launch.Launchable` interface if i keep it?

petertseng: I was kinda considering `hoist.NewLaunchable(...)` that takes these as args, including the fetcher. Then you wouldn't need to have `SetFetcher`, right? but I realize that `NewLaunchable` would have a huge number of args. I'm not sure if that's something we want...

anthonybishopric: I think I'd prefer we just rename the interface method to GetFetcher or something that removes the collision.

mpuncel: works for me, not idiomatic Go but that's okay

mpuncel:
closing in favor of a different workaround (renaming the interface method)

PR #286, opened 2015-10-08
Unexport the fields of hoist.Launchable

mpuncel:
The original intent of this change is to allow an interface conversion
from launch.Launchable to hoist.Launchable.


mpuncel:
discussed in person, not going to do this

PR #287, opened 2015-10-08
Let tgz extractor preserve tar ownership info

bwester:
Changes the ExtractTarGz() function to create each file with its original
owner, as recorded in the TAR file, when no owner is specified. This will
be used when extracting container filesystems.

### DIFF
--- a/pkg/gzip/gzip.go
+++ b/pkg/gzip/gzip.go
@@ -11,37 +11,32 @@ import (
 	"github.com/square/p2/pkg/util"
 )
 
-func ExtractTarGz(owner string, fp io.Reader, fpName string, dest string) (err error) {
+// ExtractTarGz reads a gzipped tar stream and extracts all files to the destination
+// directory. If an owner name is specified, all files will be created to be owned by that
+// user; otherwise, the tar specifies ownership.
+func ExtractTarGz(owner string, fp io.Reader, dest string) (err error) {
 	fz, err := gzip.NewReader(fp)
 	if err != nil {
-		return util.Errorf("Unable to create gzip reader: %s", err)
+		return util.Errorf("error reading gzip data: %s", err)
 	}
 	defer fz.Close()
-
 	tr := tar.NewReader(fz)
-	uid, gid, err := user.IDs(owner)
-	if err != nil {
-		return err
+
+	var ownerUID, ownerGID int
+	if owner != "" {
+		ownerUID, ownerGID, err = user.IDs(owner)
+		if err != nil {
+			return err
+		}
 	}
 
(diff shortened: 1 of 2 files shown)


pkg/gzip/gzip.go
```diff
@@ -11,37 +11,32 @@ import (
...
-		return util.Errorf(
-			"Unable to create root directory %s when unpacking %s: %s",
-			dest,
-			fpName,
-			err,
-		)
+		return util.Errorf("error creating root directory %s: %s", dest, err)
 	}
-	err = os.Chown(dest, uid, gid)
+	err = os.Chown(dest, ownerUID, ownerGID)
```
anthonybishopric: What is the behavior of passing `dest,0,0` here?

pkg/gzip/gzip.go
```diff
@@ -50,32 +45,40 @@ func ExtractTarGz(owner string, fp io.Reader, fpName string, dest string) (err e
 			break
 		}
 		if err != nil {
-			return util.Errorf("Unable to read %s: %s", fpName, err)
+			return util.Errorf("read error: %s", err)
```
anthonybishopric: Why limit the message?

pkg/gzip/gzip.go
bwester: The UID and GID are set to 0, e.g., `root:root`.

pkg/gzip/gzip.go
bwester: not sure what you mean by "limit."  The only reason for `fpName` to exist as a parameters is so it can be added to error messages.  That seemed wrong to me, so I moved the file-name-printing out of the method.

anthonybishopric: Got it, that makes sense.

anthonybishopric:
:ship:

tummychow:
does this impact the file ownership of the tarballs we currently use in prod? I'm not sure our CI builds them with all files owned by the app's user (although, if it doesn't, we should also go and fix that)

bwester:
> does this impact the file ownership of the tarballs we currently use in prod

It shouldn't. For "hoist" launchables, `ExtractTarGz()` always uses the manifest's `RunAsUser()` value. However, since you mentioned it, I went audited where that value comes from, and there many be some places where it can be set to an empty string. Those cases should be rejected.  I'll do that in another PR.

For other as-yet-uncommitted launchable types, they'll require a different build process anyway.

tummychow:
> For "hoist" launchables, ExtractTarGz() always uses the manifest's RunAsUser() value. 

Whoops, yeah, I didn't read your code closely enough. looks good :+1:

PR #289, opened 2015-10-08
Minor RC refactors part 3

tummychow:
intermediate PR for things that RU depends on, but are not themselves part of RU.

### DIFF
--- a/pkg/kp/rcstore/consul_store.go
+++ b/pkg/kp/rcstore/consul_store.go
@@ -284,6 +284,10 @@ func (s *consulStore) mutateRc(id fields.ID, mutator func(fields.RC) (fields.RC,
 	if newRC.ID.String() == "" {
 		// TODO: If this fails, then we have some dangling labels.
 		// Perhaps they can be cleaned up later.
+		// note that if the CAS fails afterwards, we will have still deleted
+		// the labels, and then we will retry, which will involve deleting them
+		// again
+		// really the only way to solve this is a transaction
 		err = s.forEachLabel(rc, func(id, k, _ string) error {
 			return s.applicator.RemoveLabel(labels.RC, id, k)
 		})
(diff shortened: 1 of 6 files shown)


pkg/kp/rcstore/consul_store.go
```diff
@@ -220,6 +220,16 @@ func (s *consulStore) SetDesiredReplicas(id fields.ID, n int) error {
 	})
 }
 
+func (s *consulStore) MutateDesiredReplicas(id fields.ID, n int) error {
```
mpuncel: what about "Add" instead of "Mutate". I thought the second argument was the desired value rather than an operand

mpuncel:
:+1:  just that one nit on function name

pkg/kp/rcstore/consul_store.go
tummychow: yeah that's a good change. anyone else have opinions on this name? otherwise I'll use AddDesiredReplicas instead

petertseng: yup, I would prefer Add rather than Mutate

pkg/kp/rcstore/store.go
```diff
@@ -33,7 +33,10 @@ type Store interface {
 	// session that deletes its keys on invalidation.
 	Lock(id fields.ID, session string) (bool, error)
 
+	// Set the desired replica count for the given RC to the given integer.
 	SetDesiredReplicas(fields.ID, int) error
```
petertseng: I take it that you feel that both `Set` and `Add` are still necessary and worth keeping?

petertseng:
Also 👍

pkg/kp/rcstore/store.go
tummychow: yes I think both are useful. Add happened to be an efficiency win for the code I'm writing at the moment, but I think Set is a more obvious interface for humans (rctl), and it's also the only way to clear an RC to zero in a single call.

petertseng: seems reasonable

PR #290, opened 2015-10-08
Check validity of input manifests

bwester:
Adds a check that validates the input data of a manifest when it is being read.
Invalid manifests will be rejected.

Currently, the validity checks only verify that required fields are provided.

### DIFF
--- a/pkg/pods/manifest.go
+++ b/pkg/pods/manifest.go
@@ -7,6 +7,7 @@ package pods
 import (
 	"crypto/sha256"
 	"encoding/hex"
+	"fmt"
 	"io"
 	"io/ioutil"
 	"os"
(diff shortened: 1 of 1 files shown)


pkg/pods/manifest.go
```diff
@@ -320,3 +324,22 @@ func (m manifest) SignatureData() (plaintext, signature []byte) {
...
+// data is inconsistent or "nil" otherwise.
+func ValidManifest(m Manifest) error {
+	if m.ID() == "" {
+		return fmt.Errorf("manifest must contain an 'id'")
+	}
+	for key, stanza := range m.GetLaunchableStanzas() {
+		switch {
+		case stanza.LaunchableType == "":
+			return fmt.Errorf("'%s': launchable must contain a 'launchable_type'", key)
+		case stanza.LaunchableId == "":
```
anthonybishopric: I would like to remove `LaunchableId` and favor the use of the map key. Perhaps unrelated to this change, but something I'd like to ensure works later.

bwester: Currently, the key is ignored and `LaunchableId` is used everywhere.  Once that changes, we can alter the validation routine here.

anthonybishopric: :+1:

anthonybishopric:
:ship:

PR #291, opened 2015-10-08
Rolling updates

tummychow:
Includes the core implementation with DeletePods=false (DeletePods=true is still subject to at least one race condition and potential other bugs), and a p2-rctl command so you can try it for yourself.


pkg/roll/run_update.go
```diff
@@ -0,0 +1,305 @@
...
+	"github.com/square/p2/pkg/kp"
+	"github.com/square/p2/pkg/kp/rcstore"
+	"github.com/square/p2/pkg/labels"
+	"github.com/square/p2/pkg/logging"
+	"github.com/square/p2/pkg/rc"
+	rcf "github.com/square/p2/pkg/rc/fields"
+	"github.com/square/p2/pkg/roll/fields"
+)
+
+type update struct {
```
mpuncel: nit: order these fields the same as the constructor

tummychow: whoopsies, added them kinda haphazardly. will fix

pkg/roll/run_update.go
```diff
@@ -0,0 +1,305 @@
...
+
+func (u update) lock(session string) error {
+	if success, err := u.rcs.LockWrite(u.NewRC, session); err != nil {
+		return err
+	} else if !success {
+		return fmt.Errorf("could not lock new %s", u.NewRC)
+	}
+
+	if success, err := u.rcs.LockWrite(u.OldRC, session); err != nil {
+		return err
```
mpuncel: should we attempt to release the lock on the NewRC in this case?

tummychow: yeah there are a few situations where locks have to be released (not just here, in rcfarm too iirc) and aren't. for now I'm writing it off as "blow the session away and let the lock die on its own". a dedicated lock API (shared across all lock-taking packages) would have a method to delete a lock key and thereby release the lock. which I might write shortly before doing the RU farm since that's where it would become important. This would also come with things like ConsulSessionManager deleting its session when terminated.

For the record, if `err != nil`, releasing the lock will probably also fail (connectivity error). Should definitely release on `!success` though.

bin/p2-rctl/main.go
```diff
@@ -200,9 +256,61 @@ func (r RCtl) Enable(id string) {
...
+	err := r.rcs.Disable(rc_fields.ID(id))
 	if err != nil {
 		r.logger.WithError(err).Fatalln("Could not disable replication controller in Consul")
 	}
 	r.logger.WithField("id", id).Infoln("Disabled replication controller")
 }
+
+func (r RCtl) RollingUpdate(oldID, newID string, want, need int, deletes bool) {
+	sessions := make(chan string)
+	go kp.ConsulSessionManager(api.SessionEntry{
```
mpuncel: what stops this goroutine

pkg/roll/run_update.go
mpuncel: probably you just want the session to end on any failure actually

bin/p2-rctl/main.go
mpuncel: oh you're running this in a CLI so everything will die at the end correct?

tummychow: program termination - it's in a binary, so by definition it's short-lived. I didn't see the point of adding an explicit quit channel here when it would never be used. Maybe if it needed to cleanup before termination then it would be useful, but right now that's not the case.

pkg/roll/run_update.go
tummychow: Well if session renewal fails, the ConsulSessionManager will immediately send a `""` to indicate that it can no longer guarantee the session's liveness. What i"m most interested in is if a program holding a session gracefully exits - it should destroy its session to free the locks it was holding immediately rather than waiting for up to 15s TTL.

Although in practice that doesn't matter, because nobody would try to claim the lock until their watches expired.

bin/p2-rcfarm/main.go
```diff
@@ -1,55 +0,0 @@
...
-)
-
-var (
-	consulUrl   = kingpin.Flag("consul", "The hostname and port of a consul agent in the p2 cluster. Defaults to 0.0.0.0:8500.").String()
-	consulToken = kingpin.Flag("token", "The consul ACL token to use. Empty by default.").String()
-	headers     = kingpin.Flag("header", "An HTTP header to add to requests, in KEY=VALUE form. Can be specified multiple times.").StringMap()
-	https       = kingpin.Flag("https", "Use HTTPS").Bool()
-)
-
-func main() {
```
anthonybishopric: Not into this anymore?

tummychow: moved it into p2-rctl instead: `p2-rctl farm`

anthonybishopric: got it :+1:

pkg/roll/run_update.go
```diff
@@ -0,0 +1,305 @@
...
+// - need: the number of nodes that must always be up (ie the minimum)
+// given these four arguments, rollAlgorithm returns the number of nodes to add
+// to the new RC (and, if DeletePods=true, to delete from the old)
+//
+// returns zero under the following circumstances:
+// - new >= want (the update is done)
+// - old+new <= need (at or below the minimum, update has to block)
+func algorithm(old, new, want, need int) int {
+	// how much "headroom" do we have between the number of nodes that are
+	// currently healthy, and the number that must be healthy?
```
mpuncel: this comment is confusing to me. don't `old` and `new` just signify the number desired rather than the number that are healthy

tummychow: so - it's both. in Run() you'll see that, if healthy<desired, it blocks and does not execute the algorithm again. It waits until they're equal before proceeding. The idea here is that if healthy<desired, then I'm probably still waiting for my most recent update to finish, so I shouldn't run the algorithm again and go for a new update.

I agree it's not obvious. It took me a few tries to figure out what to pass to the algorithm in the first place. I'm open to any improvements on how to make the comment clearer.

mpuncel: ah okay, I think that was the source of the confusion (comment said healthy but desired was being passed)

bin/p2-watch/main.go
```diff
@@ -22,6 +22,7 @@ var (
 	consulToken   = kingpin.Flag("token", "The ACL to use for accessing consul.").String()
 	headers       = kingpin.Flag("header", "An HTTP header to add to requests, in KEY=VALUE form. Can be specified multiple times.").StringMap()
 	https         = kingpin.Flag("https", "Use HTTPS").Bool()
+	waitTime      = kingpin.Flag("wait", "Duration to wait for Consul changes. Useful for proxy services that time out on long-polling requests.").Default("30s").Duration()
```
anthonybishopric: Probably worth indicating that this won't cancel the command in 30 seconds, just that it resets the watch.

pkg/kp/rcstore/store.go
```diff
@@ -31,7 +31,13 @@ type Store interface {
 	//
 	// The lock will be taken against an ephemeral key, so it is safe to use a
 	// session that deletes its keys on invalidation.
-	Lock(id fields.ID, session string) (bool, error)
+	//
+	// The separate locks are for one reader (who watches the RC and acts when
+	// it changes) and one writer (who mutates the RC to cause those changes).
+	// Each lock may be held by one, and only one, party (readers cannot share
+	// the read lock).
```
anthonybishopric: This is different from standard R/W locking semantics. Why wouldn't you allow multiple readers?

bin/p2-watch/main.go
tummychow: sure, lemme think of a better help text to put there

pkg/kp/rcstore/store.go
tummychow: I'm going to rewrite this section in the next PR, but it's still important to understand why so let me explain.

There are two entities interested in an RC: the entity that reads the RC and schedules/unschedules pods, and the entity that writes to the RC as part of some higher-layer abstraction. Obviously there can only be one writer. But the reader is also performing mutation on the layer of abstraction below the RC (preparers), and therefore there can only be one of those as well. If you had multiple goroutines running WatchDesires, you would get oscillation as each one tries to schedule over the other's changes. That's why there has to be mutual exclusion for the entity that reads the RC.

anthonybishopric: Why isn't that just a function of a separate locking mechanism over the preparers themselves? It seems orthogonal to the task of locking an RC itself.

anthonybishopric: Discussed in person - will be renaming these to reflect the clients that will use these locks. Something along the lines of "RCLock" and "RULockForRCs" or better.

pkg/roll/fields/update.go
```diff
@@ -0,0 +1,57 @@
...
+	// When the new RC's replica count equals this number, the Update is
+	// considered to be complete. This should be at least equal to the minimum.
+	DesiredReplicas int
+	// During the course of the update, pods may be restarted or removed. This
+	// number specifies the minimum number of replicas that must be alive
+	// during the entire Update.
+	//
+	// The acceptable value of Minimum is subject to several constraints:
+	//   - the update will not even start until the minimum is satisfied, so if
+	//     the current number of live nodes is than the minimum, it will block
```
anthonybishopric: > the current number of live nodes is than the minimum, it will block forever

Probably want some regular logging to happen when this is the case.

tummychow: yep there's a debug log around line 150 of run_update.go to indicate that it's blocked

bin/p2-watch/main.go
tummychow: done, improved help text is: `Maximum duration for Consul watches, before resetting and starting again.`

pkg/roll/fields/update.go
```diff
@@ -0,0 +1,57 @@
...
+	// disabled until the update is finished.
+	//
+	// DeletePods=false is a good choice when scheduling a new pod on a node will
+	// also stop the old one (eg hoist artifacts). In that case, deleting the
+	// old pod is not necessary and may cause unnecessary slowdowns. Using
+	// DeletePods=true is more suited to pods that are independent of each
+	// other (eg linux containers), where it is desired to remove the old pods
+	// throughout the rollout.
+	//
+	// Be certain that the delete command cannot delete a newly scheduled pod.
```
anthonybishopric: Clarify how to prevent this from happening

tummychow:
rebased. Concerns about locks will be addressed in the next pr. Any final comments?

anthonybishopric:
I think I'm :+1:

PR #292, opened 2015-10-08
Make the -n flag to p2-exec configurable via hoist.Launchable struct

mpuncel:
This allows code utilizing the hoist.Launchable struct as a library to
determine whether p2-exec commands run with the -n (no-limit) flag. This
is useful because the -n flag requires root privileges.

### DIFF
--- a/pkg/hoist/hoist_launchable.go
+++ b/pkg/hoist/hoist_launchable.go
@@ -15,6 +15,7 @@ import (
 	"github.com/square/p2/pkg/cgroups"
 	"github.com/square/p2/pkg/gzip"
 	"github.com/square/p2/pkg/launch"
+	"github.com/square/p2/pkg/p2exec"
 	"github.com/square/p2/pkg/runit"
 	"github.com/square/p2/pkg/uri"
 	"github.com/square/p2/pkg/user"
@@ -29,7 +30,8 @@ type Launchable struct {
 	ConfigDir        string         // The value for chpst -e. See http://smarden.org/runit/chpst.8.html
 	Fetcher          uri.Fetcher    // Callback that downloads the file from the remote location.
 	RootDir          string         // The root directory of the launchable, containing N:N>=1 installs.
-	P2exec           string         // The path to p2-exec
+	P2Exec           string         // Struct that can be used to build a p2-exec invocation with appropriate flags
+	ExecNoLimit      bool           // If set, execute with the -n (--no-limit) argument to p2-exec
 	CgroupConfig     cgroups.Config // Cgroup parameters to use with p2-exec
 	CgroupConfigName string         // The string in PLATFORM_CONFIG to pass to p2-exec
 	RestartTimeout   time.Duration  // How long to wait when restarting the services in this launchable.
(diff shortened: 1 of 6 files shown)


pkg/hoist/hoist_launchable.go
```diff
@@ -223,19 +215,10 @@ func (hl *Launchable) Executables(
...
-			hl.ConfigDir,
-			"-l",
-			hl.CgroupConfigName,
-			"-c",
-			hl.Id,
-			filepath.Join(serviceDir, service.Name()),
-		}
+		execCmd := append(
+			[]string{hl.P2exec},
+			append(hl.p2ExecArgs(), filepath.Join(serviceDir, service.Name()))...,
```
mpuncel: lmk if i should break this up into multiple lines or if i'm missing some special Go way of doing this

pkg/pods/pod.go
```diff
@@ -551,6 +551,7 @@ func (pod *Pod) getLaunchable(launchableStanza LaunchableStanza, runAsUser strin
 			Fetcher:          uri.DefaultFetcher,
 			RootDir:          launchableRootDir,
 			P2exec:           pod.P2exec,
+			ExecNoLimit:      true,
```
mpuncel: i figured this isn't worth the trouble to bubble up to preparer config but LMK if you disagree

pkg/hoist/hoist_launchable.go
```diff
@@ -359,3 +342,21 @@ func (hl *Launchable) InstallDir() string {
 	launchableName := hl.Version()
 	return filepath.Join(hl.RootDir, "installs", launchableName)
 }
+
+func (hl *Launchable) p2ExecArgs() []string {
```
bwester: Considering how this method is used, I think it makes sense to have it build the entire command.  Build the slice with `hl.P2exec` prepended, and add a `command string` parameter that gets appended. Then there will be no more need for the funky `append()` calls at the method call sites.

bwester:
After thinking about this for a while, I think it would be useful and very clean to abstract the construction of this command into a builder type where the various parameters of p2-exec can be specified.  Something like the following usage:

    p2exec := P2Exec{
        Bin: hl.P2exec,
        EnvDir: hl.ConfigDir,
        Command: filepath.Join(...),
    }
    if hl.ExecNoLimit {
        execCmd.NoLimits = true
    }
    cmd := exec.Command(p2exec.Build()...)

Building commands based on p2-exec will be important for any kind of launchable, but not all launchables will need the same mix of flags.

Thoughts?

anthonybishopric:
@bwester :+1: good util

mpuncel:
@bwester good suggestion, it's now implemented.

pkg/hoist/hoist_launchable.go
```diff
@@ -133,19 +134,15 @@ func (hl *Launchable) InvokeBinScript(script string) (string, error) {
...
-	)
+	p2ExecArgs := util.P2ExecArgs{
+		Command:          cmdPath,
+		User:             hl.RunAs,
+		EnvDir:           hl.ConfigDir,
+		NoLimits:         hl.ExecNoLimit,
+		CgroupConfigName: hl.CgroupConfigName,
+		CgroupName:       hl.Id,
+	}
+	cmd := exec.Command(hl.P2Exec.Bin, hl.P2Exec.BuildWithArgs(p2ExecArgs)...)
```
bwester: For this use, it's unfortunate that you have to specify the first argument separately.

pkg/util/p2_exec.go
```diff
@@ -0,0 +1,41 @@
+package util
+
+type P2Exec struct {
+	Bin string
```
bwester: Is this type really needed?  You don't seem to be using it for anything.

pkg/util/p2_exec.go
```diff
@@ -0,0 +1,41 @@
...
+type P2ExecArgs struct {
+	User             string
+	EnvDir           string
+	NoLimits         bool
+	CgroupName       string
+	CgroupConfigName string
+	Command          string
+}
+
+func (p P2Exec) BuildWithArgs(args P2ExecArgs) []string {
```
bwester: nit: Can you shorten this to `Build()`?  I think the `ActionWithThing1(t)` naming pattern is more appropriate when you're trying to distinguish it from a default action (e.g., `Action()`) or different different styles of arguments (e.g., `ActionWithThing2(t)`).  That doesn't apply here.

Or, name the method for what it produces: `CommandLine()`

pkg/util/p2_exec.go
```diff
@@ -0,0 +1,37 @@
+package util
```
anthonybishopric: Please move to `pkg/exec` or `pkg/p2exec`. I'd like to remove the `util` package altogether.

anthonybishopric:
:+1: pending move

PR #293, opened 2015-10-09
Preliminary OpenContainer support

bwester:
This is definitely experimental code, but it kind of works!

I'd like to merge this to master at this preliminary stage so that I can commit the work I've made so far. Not sure when I'll get back to it.  By default, the new type is disabled and fairly self-contained, so it shouldn't pose any risk.

### DIFF
--- a/pkg/opencontainer/containers.go
+++ b/pkg/opencontainer/containers.go
@@ -1,0 +1,294 @@
+// The "opencontainer" package implements support for launching serivces packaged in an
+// OpenContainer image. Containers can be used by specifying "type: opencontainer" in a
+// launchable's configuration in a pod manifest.
+//
+// P2 support for OpenContainer images is EXPERIMENTAL, even beyond the fact that the
+// OpenContainer spec has not released a stable version and is in constant flux.
+package opencontainer
+
+import (
+	"encoding/json"
+	"io/ioutil"
+	"os"
+	"os/exec"
+	"path/filepath"
+	"time"
+
+	"github.com/square/p2/pkg/cgroups"
+	"github.com/square/p2/pkg/gzip"
+	"github.com/square/p2/pkg/launch"
+	"github.com/square/p2/pkg/p2exec"
+	"github.com/square/p2/pkg/runit"
+	"github.com/square/p2/pkg/uri"
+	"github.com/square/p2/pkg/user"
+	"github.com/square/p2/pkg/util"
+	"github.com/square/p2/pkg/util/param"
+)
+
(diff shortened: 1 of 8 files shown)


pkg/opencontainer/default.go
```diff
@@ -0,0 +1,150 @@
+package opencontainer
+
+// DefaultRuntimeSpec is the default template for running Linux containers.
+var DefaultRuntimeSpec = LinuxRuntimeSpec{
```
mpuncel: is this copied from upstream as well? (just curious)

mpuncel:
lgtm!

pkg/opencontainer/default.go
bwester: The defaults don't exist in the upstream specs repository.  Presumably, something like this is encoded in _runc_'s source code.  The configuration here is based on the output of the `runc spec` command.

bwester:
I haven't gotten much feedback on this PR.  Does anyone strongly _object_ to me merging gated, experimental code?

pkg/opencontainer/containers.go
```diff
@@ -0,0 +1,294 @@
...
+		return nil
+	}
+
+	data, err := uri.DefaultFetcher.Open(l.Location)
+	if err != nil {
+		return err
+	}
+	defer data.Close()
+	defer func() {
+		if returnedError != nil {
```
anthonybishopric: What assigns `returnedError`? Is this useful?

bwester: The return value is named only so that it can be inspected in this one location.  There are no explicit assignments to `returnedError`.

anthonybishopric: Got it, I didn't realize calling `return err` would implicitly assign the error to the variable named by the return clause.

pkg/opencontainer/specs.go
```diff
@@ -0,0 +1,437 @@
+package opencontainer
+
+import (
+	"os"
+)
+
+// These struct definitions are copied verbatim from "github.com/opencontainers/specs".
+// The original source package includes Linux-specific structures configured to be
+// compiled only on Linux systems, making it unnecessarily difficult to develop P2 from
+// other environments.
```
anthonybishopric: Can you enumerate what they were? I'm slightly worried about maintaining this file.

bwester: Any structure with "Linux" in the name.  Seriously, this is a giant concatenation of 4 files with their headers aggregated.  Upgrading this file isn't going to be a problem.

anthonybishopric:
:ship:

PR #294, opened 2015-10-12
Only change user in p2-exec when necessary.

mpuncel:
If the user we want to run p2-exec as is the current user, there is no
need to set uid or gid. This allows scripts to run p2-exec as their own
user without having to do checks of their own to leave out the -u
argument


tummychow:
:+1: looks straightforward enough

PR #295, opened 2015-10-12
Rolling update farms

tummychow:
Fixes various lock-related things in rolling updates and RCs. Implements farm for rolling updates and associated p2-rctl commands. Verified in staging.

### DIFF
--- a/pkg/kp/constants.go
+++ b/pkg/kp/constants.go
@@ -10,6 +10,7 @@ const (
 	HOOK_TREE    string = "hooks"
 	LOCK_TREE    string = "lock"
 	RC_TREE      string = "replication_controllers"
+	ROLL_TREE    string = "rolls"
 )
 
 func IntentPath(args ...string) string {
@@ -30,4 +31,8 @@ func LockPath(args ...string) string {
 
 func RCPath(args ...string) string {
 	return strings.Join(append([]string{RC_TREE}, args...), "/")
+}
+
+func RollPath(args ...string) string {
+	return strings.Join(append([]string{ROLL_TREE}, args...), "/")
 }
--- a/pkg/rc/farm.go
+++ b/pkg/rc/farm.go
@@ -63,59 +63,59 @@ func NewFarm(
 //
 // Start is not safe for concurrent execution. Do not execute multiple
 // concurrent instances of Start.
-func (rcm *Farm) Start(quit <-chan struct{}) {
+func (rcf *Farm) Start(quit <-chan struct{}) {
 	subQuit := make(chan struct{})
 	defer close(subQuit)
-	rcWatch, rcErr := rcm.rcStore.WatchNew(subQuit)
+	rcWatch, rcErr := rcf.rcStore.WatchNew(subQuit)
 
 START_LOOP:
 	for {
 		select {
 		case <-quit:
-			rcm.logger.NoFields().Infoln("Halt requested, releasing replication controllers")
-			rcm.releaseChildren()
+			rcf.logger.NoFields().Infoln("Halt requested, releasing replication controllers")
+			rcf.releaseChildren()
 			return
-		case session := <-rcm.sessions:
+		case session := <-rcf.sessions:
 			if session == "" {
 				// our session has expired, we must assume our locked children
 				// have all been released and that someone else may have
 				// claimed them by now
-				rcm.logger.NoFields().Errorln("Session expired, releasing replication controllers")
-				rcm.lock = nil
-				rcm.releaseChildren()
+				rcf.logger.NoFields().Errorln("Session expired, releasing replication controllers")
+				rcf.lock = nil
+				rcf.releaseChildren()
 			} else {
 				// a new session has been acquired - only happens after an
 				// expiration message, so len(children)==0
-				rcm.logger.WithField("session", session).Infoln("Acquired new session")
-				lock := rcm.kpStore.NewUnmanagedLock(session, "")
-				rcm.lock = &lock
+				rcf.logger.WithField("session", session).Infoln("Acquired new session")
+				lock := rcf.kpStore.NewUnmanagedLock(session, "")
+				rcf.lock = &lock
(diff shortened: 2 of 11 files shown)


pkg/kp/constants.go
```diff
@@ -31,3 +32,6 @@ func LockPath(args ...string) string {
 func RCPath(args ...string) string {
 	return strings.Join(append([]string{RC_TREE}, args...), "/")
 }
```
mpuncel: nit: newline

tummychow: whoopsies, will fix

pkg/rc/farm.go
```diff
@@ -108,37 +105,32 @@ func (rcm *Farm) Start(quit <-chan struct{}) {
...
+				if _, ok := err.(kp.AlreadyLockedError); ok {
 					// someone else must have gotten it first - log and move to
 					// the next one
-					rcm.logger.WithFields(logrus.Fields{
-						"rc_id":   rcField.ID,
-						"session": session,
-					}).Debugln("Lock on replication controller was denied")
+					rcLogger.NoFields().Debugln("Lock on replication controller was denied")
 					continue
+				} else if err != nil {
```
mpuncel: in this case (if it happened on all servers running the rc farm) is there any mechanism by which this orphaned replication controller will get a parent? any retrying anywhere?

I guess WatchNew returns the whole list so a newly created replication controller should trigger a retry?

tummychow: yeah, eventually the watch will roll over and the farms will iterate over all the RCs again. If a locking error is transient then it will be retried eventually. Unfortunately there's no way to "kick" the watch if the system recovers, so you have to wait until it times out naturally or gets triggered by another change.

pkg/rc/farm.go
```diff
@@ -108,37 +105,32 @@ func (rcm *Farm) Start(quit <-chan struct{}) {
...
-				if err != nil {
-					// log and break out of the loop - we can't continue if
-					// we had a real failure, our session probably disappeared
-					rcm.logger.WithErrorAndFields(err, logrus.Fields{
-						"rc_id":   rcField.ID,
-						"session": session,
-					}).Errorln("Got error while locking replication controller - session may be expired")
-					break // TODO: should we be pruning after this?
-				} else if !success {
+				err := rcm.lock.Lock(kp.LockPath(kp.RCPath(rcField.ID.String())))
```
anthonybishopric: Should this variable be renamed `rcFarm`? It wasn't obvious to me at first that it was that and suspect it's a holdover.

pkg/roll/farm.go
```diff
@@ -0,0 +1,203 @@
...
+	}
+}
+
+// Start is a blocking function that monitors Consul for updates. The Farm will
+// attempt to claim updates as they appear and, if successful, will start
+// goroutines for those updatesto do their job. Closing the quit channel will
+// cause this function to return, releasing all locks it holds.
+//
+// Start is not safe for concurrent execution. Do not execute multiple
+// concurrent instances of Start.
```
anthonybishopric: You could actually protect against the double starting case with a `chan(struct{},1)`:

```
select {
   case startCh <- struct{}{}: {}
     // no op
   default:  {
        rlf.logger.Errorln("or fatalln, whichever")
   }
}

Comment (anthonybishopric) [inline pkg/roll/farm.go]: You could also do it with some boolean flag but it wouldn't necessarily be safe in a concurrent context.

File: pkg/roll/farm.go [hunk 5]
```diff
@@ -0,0 +1,203 @@
...
+					// someone else must have gotten it first - log and move to
+					// the next one
+					rlLogger.NoFields().Debugln("Lock on update was denied")
+					continue
+				} else if err != nil {
+					rlLogger.NoFields().Errorln("Got error while locking update - session may be expired")
+					// stop processing this update and go back to the select
+					// chances are this error is a network problem or session
+					// expiry, and all the others in this update would also fail
+					continue START_LOOP
```
anthonybishopric: So, lets assume the session manager is imperfect and sends multiple session IDs without sending the `""` expiry message. Both the rc farm and the roll farm seem like they'll get stuck forever trying to use the same locks. Should we explicitly release the lock for children if we enter this error case?

tummychow: Explicitly releasing the lock (ie deleting the lock key) would probably also result in an error if it was a network problem, and if the session actually is expired then all the locks are also lost. Either way I don't think any action in response to this error is especially meaningful.

One thing we could do is panic in the session-tracking clause if we get a new session while our previous one is believed to be active.

pkg/roll/farm.go
tummychow: I don't think enforcing the single-start is important, I just wanted to document it because there's shared state in the Farm object itself. Personally, I think the entire function is already approaching its maximum allowable complexity, and there's not much value to be gained by checking the single-start-ness.

pkg/rc/farm.go
tummychow: yeah the M is a leftover from when it was called RCManager, I'll add a commit that changes the names

pkg/roll/farm.go
```diff
@@ -0,0 +1,203 @@
...
+				newChild := rlf.factory.New(rlField, rlLogger, *rlf.lock)
+				childQuit := make(chan struct{})
+				rlf.children[rlField.NewRC] = childRU{ru: newChild, quit: childQuit}
+				foundChildren[rlField.NewRC] = struct{}{}
+
+				go func(id fields.ID) {
+					err := newChild.Prepare()
+					if err != nil {
+						rlLogger.WithError(err).Errorln("Could not prepare update")
+					} else {
```
anthonybishopric: return instead of `} else {` ing, hard to read this arrow code

tummychow: oh yep, somehow I must have forgotten this was in a closure

anthonybishopric:
I say :ship:

PR #297, opened 2015-10-14
Fixes to pkg/runit

anthonybishopric:
Previously the runit package did not account for when services were
in a down state. This would have resulted in pkg/runit returning an
error instead of stat information.

Exposes the child status field, which is necessary for the runit
monitor to accurately ascertain if the service is up and healthy.


tummychow:
:+1:

PR #298, opened 2015-10-14
Fix monitor port parsing

tummychow:


petertseng:
Yup :+1: 

Does this imply we were not using either of these things?

PR #300, opened 2015-10-17
rm seamus.tar.gz

petertseng:
This got added in 9214eeb319aa9b175c5d350fe238aea776115d7d but I can't
see where it's used. `git grep seamus` has no result. The only reason it
might be necessary is if we wildcard in any .tar.gz files, but I don't
see this happening either. So I think it's safe to delete.


petertseng:
note for me - travis failed with not only the common "cpu subsystem not available" but also an unexpected lock failure in unit tests (I got a 2-for-1 deal on sporadic test failures today) - maybe useful to look at https://travis-ci.org/square/p2/builds/85853891

anthonybishopric:
:+1:

PR #301, opened 2015-10-20
Mpuncel/fix session renewal

mpuncel:


tummychow:
:+1: looks straightforward enough

PR #302, opened 2015-10-20
Extract status server, enable socket status reporting

anthonybishopric:
Allows the preparer to publish its status on a socket instead of a port.

### DIFF
--- a/pkg/preparer/status.go
+++ b/pkg/preparer/status.go
@@ -1,0 +1,91 @@
+package preparer
+
+import (
+	"fmt"
+	"net"
+	"net/http"
+	"os"
+
+	"github.com/square/p2/pkg/logging"
+)
+
+type StatusServer struct {
+	listener net.Listener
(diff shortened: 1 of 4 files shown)


pkg/preparer/status.go
```diff
@@ -0,0 +1,91 @@
...
+	mux := http.NewServeMux()
+	mux.HandleFunc("/_status", func(w http.ResponseWriter, r *http.Request) {
+		fmt.Fprintf(w, "p2-preparer OK")
+	})
+
+	s.server.Handler = mux
+	err := s.server.Serve(s.listener)
+	s.logger.WithError(err).Warnln("Status server exited!")
+	s.Exit <- err
+	close(s.Exit)
```
anthonybishopric: Currently unused but could be retrieved and used to kill the preparer on failure.

mpuncel:
:+1:

PR #303, opened 2015-10-20
Release RC write locks on Run() termination

tummychow:
If a rolling update is deleted from Consul, the farm automatically
releases the lock that it took on that update, under /locks/rolls/id.
But the update itself also claimed locks on the two RCs it was using.
Those locks also have to be released, or otherwise the update cannot
be Prepare()d by someone else.

To fix this, we make Update.Run() a long-lived function with an error
channel return (instead of bailing on the first error). Then, when Run()
terminates, we release the locks claimed by Prepare(). The caller would
have to call Prepare() again before resuming Run().

### DIFF
--- a/bin/p2-rctl/main.go
+++ b/bin/p2-rctl/main.go
@@ -4,6 +4,9 @@ import (
 	"encoding/json"
 	"fmt"
 	"net/url"
+	"os"
+	"os/signal"
+	"syscall"
 	"time"
 
 	"github.com/square/p2/Godeps/_workspace/src/github.com/Sirupsen/logrus"
@@ -268,12 +271,20 @@ func (r RCtl) Disable(id string) {
 }
 
 func (r RCtl) RollingUpdate(oldID, newID string, want, need int, deletes bool) {
+	if want < need {
+		r.logger.WithFields(logrus.Fields{
+			"want": want,
+			"need": need,
+		}).Fatalln("Cannot run update with desired replicas less than minimum replicas")
+	}
 	sessions := make(chan string)
+	quit := make(chan struct{})
+
 	go kp.ConsulSessionManager(api.SessionEntry{
 		LockDelay: 1 * time.Nanosecond,
 		Behavior:  api.SessionBehaviorDelete,
 		TTL:       "15s",
-	}, r.baseClient, sessions, nil, r.logger)
+	}, r.baseClient, sessions, quit, r.logger)
 
 	session := <-sessions
 	if session == "" {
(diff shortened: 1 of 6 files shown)


tummychow:
addressed both of the previous comments

tummychow:
never mind, not ready for merge. Consider this a WIP PR while I fix things coming up in p2-master.

tummychow:
travis being a pain, this is ready for review

bin/p2-rctl/main.go
```diff
@@ -281,32 +290,35 @@ func (r RCtl) RollingUpdate(oldID, newID string, want, need int, deletes bool) {
...
-			r.logger.WithError(err).Errorln("Could not run update")
+	}, r.kps, r.rcs, r.hcheck, r.labeler, r.sched, r.logger, lock).Run(quit)
+
+LOOP:
+	for {
+		select {
+		case <-signals:
+			// try to clean up locks on ^C
+			close(quit)
+			r.logger.NoFields().Fatalln("Got signal (manual cleanup of RC locks may be required)")
```
anthonybishopric: What happens when the goroutine running `main` exits before cleanup happens? Does this program exit?

tummychow: Basically the Run() loop is deferring a lock cleanup so that other farms/RUs can claim the lock later. If the program exits too fast the locks might not be released, hence the warning message.

Now that you mention it though, the solution is to not exit right away, but instead to drain the error or session channels to be sure that the child goroutines have exited. Will fix

tummychow: and related to that, farms should probably return a channel that lets them acknowledge termination.. I'll append that to this PR momentarily, just need to run some tests with rctl

pkg/roll/run_update.go
```diff
@@ -57,65 +58,57 @@ func NewUpdate(
...
-}
 
-func (u update) Run(quit <-chan struct{}) error {
+	u.logger.NoFields().Debugln("Launching health watch")
 	newFields, err := u.rcs.Get(u.NewRC)
-	if err != nil {
-		return err
+	for ; err != nil; newFields, err = u.rcs.Get(u.NewRC) {
+		errs <- err
+		time.Sleep(1 * time.Second)
```
anthonybishopric: Noting that there's no way to respond to a quit message in this loop

tummychow: ugh I knew I forgot something, I'll fix this and find a way to make the retry loops less repetitive

pkg/roll/run_update.go
```diff
@@ -256,17 +258,30 @@ func (u update) countHealthy(id rcf.ID, checks map[string]health.Result) (rcNode
...
 
 	nodes, err := rc.New(rcFields, u.kps, u.rcs, u.sched, u.labeler, u.logger).CurrentNodes()
 	if err != nil {
 		return ret, err
 	}
-	ret.current = len(nodes)
+	ret.Current = len(nodes)
 
 	for _, node := range nodes {
+		// TODO: is reality checking an rc-layer concern?
```
anthonybishopric: It is for pkg/replication, so I'd argue it is for RCs as well.

anthonybishopric:
Overall this code looks okay, but given how much growth in the RC/RU world you've introduced in the last few weeks, I'd say it's worth a bigger review. For now, :+1: to merge, worth reviewing more holistically.

pkg/roll/run_update.go
tummychow: I might end up moving it to RC anyway, but I'm not sure I agree with your argument - the functionality of pkg/replication is partially replaced by RUs and partially by RCs. RCs alone aren't equivalent to pkg/replication and in particular they have no safety guarantees when their replica count is changed. That's why I put this in RUs originally (because I felt that the safety-aware code belonged there).

petertseng:
yup :+1: pending the planned fixes around quitting in retry loops and unlocks when rctl exits. my review is mostly on a "everything that *is* in this PR is good, but I'm unsure of whether there is anything that *isn't* in this PR that should be" basis

tummychow:
okay, I have other changes to make regarding blocking semantics, but the code here is mergeable

PR #304, opened 2015-10-22
Use params to control Consul detailed errors

bwester:
The `showConsulErrors` flag was created before the "param" package was written,
and it used the tool it had available at the time: command line flags. These
have been awkward to work with, so this commit migrates variable from being a
Kingpin flag to being a parameter.

With this change, the library functionality of P2 (e.g., nearly everything in
"pkg") no longer depends on any particular CLI framework.


anthonybishopric:
:+1:

mpuncel:
:+1:

PR #305, opened 2015-10-23
Fix error formatting for p2-sum

bwester:
Installs a standard usage message and fixes a couple of missing newlines.


petertseng:
Seems good. (Test failure unrelated, should be able to just get it to rerun)

PR #306, opened 2015-10-26
remove deadlock case

anthonybishopric:
When receiving an error reading from reality in certain cases, we would block forever. This explains why during a Consul reset, a number of preparers would suddenly hang forever. Follow up work should alter the status checker to perform a liveness check of the various routines we care about.


mpuncel:
:+1:

tummychow:
:+1:

PR #308, opened 2015-10-27
Handle errors inside Update.Run

tummychow:
See commit message. This simplifies the semantics of the function a bit and avoids potential races against program exit (where the program might live long enough to accidentally delete an update that wasn't finished yet, if asked to quit the update on termination). Particularly cc @bwester since I went over this code with you and you might be better positioned to review it.

### DIFF
--- a/bin/p2-rctl/main.go
+++ b/bin/p2-rctl/main.go
@@ -292,13 +292,17 @@ func (r RCtl) RollingUpdate(oldID, newID string, want, need int, deletes bool) {
 	}
 	lock := r.kps.NewUnmanagedLock(session, "")
 
-	errs := roll.NewUpdate(roll_fields.Update{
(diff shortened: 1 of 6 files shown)


anthonybishopric:
:+1: but will defer to @bwester for ultimate approval

tummychow:
rebase

tummychow:
- rebase
- export RetryOrQuit, I ended up rewriting it in our internal implementation of this so let's just share it around

bwester:
Other than the above nits, it looks fine.  I'll keep looking at the test case.

tummychow:
good eye on the label names, I actually never noticed that the convention for constants was the same as other exports. will fix both.

bwester:
ok, i still don't fully understand the test, but your explanation of the change seems sound

Just ship it

PR #309, opened 2015-10-27
Add a EnvPath() function to HookEnv

mpuncel:
This makes it easier for hooks to locate a pod's env directory via
environment variables. It is essential for hooks to compute a pod's env
directory in this manner so that hooks will maintain compatibility with
a preparer change that moves the env directory

### DIFF
--- a/pkg/hooks/hook_env.go
+++ b/pkg/hooks/hook_env.go
@@ -18,7 +18,7 @@ func CurrentEnv() *HookEnv {
 type HookEnv struct{}
 
 func (h *HookEnv) Manifest() (pods.Manifest, error) {
-	path := os.Getenv("HOOKED_POD_MANIFEST")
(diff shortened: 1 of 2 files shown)


pkg/hooks/hook_env.go
```diff
@@ -46,6 +46,10 @@ func (h *HookEnv) Event() (HookType, error) {
 	return AsHookType(os.Getenv("HOOK_EVENT"))
 }
 
+func (h *HookEnv) EnvPath() string {
+	return os.Getenv("HOOKED_ENV_PATH")
```
anthonybishopric: Is the variable already available?

mpuncel: it is, constant-ified it for future proofing if we change the variable name

mpuncel: although we probably shouldn't change those because it'd be hard to preserve compatibility with hooks

anthonybishopric:
:+1:

PR #311, opened 2015-10-28
Return an error if removing the pod's home directory fails on uninsta…

mpuncel:
…llation


petertseng:
Oh. yeah seems super easy. seems good.

petertseng:
though I suspect github is telling us the first line is too long? not sure what we want

mpuncel:
shortened commit message

PR #312, opened 2015-10-29
Use "cgroup.procs" for cgroup membership

bwester:
This commit changes p2-exec to use a different method for changing its cgroup
membership: it will write "0" to the cgroup's "cgroup.procs" file. Writing to
this file will change membership for the current thread group (e.g., the entire
process) instead of an individual thread. This may be a fix for a bug wherein a
launched service is observed to be running in the root cgroup.

### DIFF
--- a/pkg/cgroups/cgroups.go
+++ b/pkg/cgroups/cgroups.go
@@ -173,11 +173,11 @@ func (subsys Subsystems) Write(config Config) error {
 }
 
 func (subsys Subsystems) AddPID(name string, pid int) error {
-	err := appendIntToFile(filepath.Join(subsys.Memory, name, "tasks"), pid)
+	err := appendIntToFile(filepath.Join(subsys.Memory, name, "cgroup.procs"), pid)
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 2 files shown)


pkg/cgroups/cgroups.go
```diff
@@ -180,6 +180,14 @@ func (subsys Subsystems) AddPID(name string, pid int) error {
 	return appendIntToFile(filepath.Join(subsys.CPU, name, "tasks"), pid)
```
tummychow: github won't let me comment in the hidden part of the diff, but I would remove or replace `AddPID` as part of this change. Right now it still uses `tasks`.

bwester: :+1:

pkg/cgroups/cgroups.go
```diff
@@ -173,11 +173,11 @@ func (subsys Subsystems) Write(config Config) error {
 }
 
 func (subsys Subsystems) AddPID(name string, pid int) error {
```
tummychow: :+1: but remember to change the caller to use `0` instead of `os.Getpid()`

tummychow:
also needs rebase, btw

tummychow:
:+1:

PR #314, opened 2015-11-02
Add pod field to roll farm logs

tummychow:
Necessary to attach a pod field to the logs produced by RUs.


anthonybishopric:
:+1:

PR #316, opened 2015-11-03
Mpuncel/restart options

mpuncel:
See commit messages


tummychow:
code wise i think this looks pretty good :+1:

PR #317, opened 2015-11-05
Move post-activate output to logrus field

tummychow:
This shouldn't be in the logrus message in the first place; parametric data should be in fields.


tummychow:
test failure is flake - probably failed to connect to the server and the response was nil or malformed. Will fix tomorrow

rudle:
:+1:

mpuncel:
:+1:

tummychow:
fixed flake

petertseng:
yup :+1:

PR #318, opened 2015-11-09
Documents pkg/rc/fields.RC and specifies its JSON format

bwester:
Adds documentation to the main Resource Controller data structure. Also
provides a concrete specification of its JSON-encoded format, which will be
used to store the structure into Consull.


bwester:
@tummychow: let me know if I've got the field descriptions right. I'm only starting to really dig into all this.

tummychow:
:+1:

PR #319, opened 2015-11-12
p2 log bridge

rudle:
A generic bridge from an io.Reader to an io.Writer. This also allows us to configure a log executable on a per pod basis. Internally, we'll use this to catch logs from stdout and route them into our logging infrastructure.

Please pay special attention to style, logging convention and testing convention. This is my first PR into this repo.

### DIFF
--- a/bin/p2-log-bridge/main.go
+++ b/bin/p2-log-bridge/main.go
@@ -1,0 +1,50 @@
+package main
+
+import (
+	"io"
+	"os"
+	"os/exec"
+	"os/signal"
+
+	"github.com/square/p2/Godeps/_workspace/src/gopkg.in/alecthomas/kingpin.v2"
+	"github.com/square/p2/pkg/logbridge"
+	"github.com/square/p2/pkg/logging"
+	"github.com/square/p2/pkg/version"
+)
+
+var (
+	durableLogger = kingpin.Arg("exec", "An executable that logbridge will log to without dropping messages. If a write to STDIN of this program blocks, logbridge will block.").Required().String()
+)
+
+func main() {
+	kingpin.Version(version.VERSION)
+	kingpin.Parse()
+
+	loggerCmd := exec.Command(*durableLogger)
+	durablePipe, err := loggerCmd.StdinPipe()
+	if err != nil {
+		logging.DefaultLogger.WithError(err).Errorln("failed during logbridge setup")
+		os.Exit(1)
+	}
+
+	go handleSignals(durablePipe)
+
+	go func(r io.Reader, durableWriter io.Writer, nonDurableWriter io.Writer, logger logging.Logger) {
+		logbridge.Tee(r, durableWriter, nonDurableWriter, logger)
+	}(os.Stdin, durablePipe, os.Stdout, logging.DefaultLogger)
+
+	loggerCmd.Start()
+	err = loggerCmd.Wait()
+	if err != nil {
+		logging.DefaultLogger.WithError(err)
+		os.Exit(1)
+	}
+}
+
+// This is slightly naive, it could be more narrow in the signals it cares about
+func handleSignals(c io.Closer) {
+	signals := make(chan os.Signal, 2)
+	signal.Notify(signals)
+	<-signals
+	c.Close()
+}
--- a/pkg/logbridge/doc.go
+++ b/pkg/logbridge/doc.go
@@ -1,0 +1,11 @@
+/*
+This package implements a log bridge. It is a bridge in the literal sense - it
+will carry logs from one side to the other until it reaches its capacity. When
+the bridge is at its capacity, it will drop new entries instead of blocking.
+This behaviour is desirable because it makes this package safe to attach
+to the logging end of an executable
+
+A typical use for this package is to attach it to stdout of an executable
+and passing the results into your system's proprietary logging backend.
+*/
+package logbridge
--- a/pkg/logbridge/logbridge.go
+++ b/pkg/logbridge/logbridge.go
@@ -1,0 +1,109 @@
+package logbridge
+
+import (
+	"bufio"
+	"io"
+	"time"
+
+	"github.com/square/p2/pkg/logging"
+)
+
+// Copy implements a buffered copy operation between dest and src.
+// It returns the number of dropped messages as a result of insufficient
+// capacity
+func LossyCopy(dest io.Writer, src io.Reader, capacity int, logger logging.Logger) {
+	lines := make(chan []byte, capacity)
+
+	go lossyCopy(src, lines, logger)
+
+	var n int
+	var err error
+	for line := range lines {
+		n, err = writeWithRetry(dest, line, logger)
+		if err != nil {
+			logger.WithError(err).WithField("dropped line", line).WithField("retried", isRetriable(err)).WithField("bytes written", n).Errorln("Encountered a non-recoverable error. Proceeding.")
+		}
+	}
+}
+
+// This function will scan lines from src and send them on the lines channel,
+// except when the channel is full in which case it will skip the line
+func lossyCopy(src io.Reader, lines chan []byte, logger logging.Logger) {
+	defer close(lines)
+
+	droppedLines := 0
+	scanner := bufio.NewScanner(src)
+	var line []byte
+	for scanner.Scan() {
+		line = scanner.Bytes() // consume a line regardless of the state of the writer
+		select {
+		case lines <- line:
+		default:
+			droppedLines++
+
+			warningMessage := "Dropped was dropped due to full capacity. If this occurs frequently, consider increasing the capacity of this logbridge."
+			logger.WithField("dropped line", line).Errorln(warningMessage)
+			if droppedLines%10 == 0 {
+				select {
+				case lines <- []byte(warningMessage):
+				case <-time.After(100 * time.Millisecond):
+					// best effort warning of dropped messages. If this doesn't suceed expediently, forget it and get back to work
+				}
+			}
+		}
+	}
+	if err := scanner.Err(); err != nil {
+		logger.WithError(err).Errorln("Encountered error while reading from src. Proceeding.")
+	}
+}
+
+// Tee will copy to faithfulWriter without dropping messages, it will copy
+// through a buffer to better handle mismatched latencies. Lines written to
+// lossyWriter will be copied in a best effort way with respect to latency and
+// buffered through a go channel.
+func Tee(r io.Reader, durableWriter io.Writer, lossyWriter io.Writer, logger logging.Logger) {
+	tr := io.TeeReader(r, bufio.NewWriterSize(durableWriter, 1<<10))
+
+	LossyCopy(lossyWriter, tr, 1<<10, logger)
+}
+
+// This is an error wrapper type that may be used to denote an error is retriable
+// RetriableError is exported so clients of this package can express their
+// error semantics to this package
+type retriableError struct {
+	err error
+}
+
+func NewRetriableError(err error) retriableError {
+	return retriableError{err}
+}
+
+func (r retriableError) Error() string {
+	return r.err.Error()
+}
+
+func isRetriable(err error) bool {
+	_, ok := err.(retriableError)
+	return ok
+}
+
+var backoff = func(i int) time.Duration {
+	return time.Duration(1 << uint(i) * time.Second)
+}
+
+func writeWithRetry(w io.Writer, line []byte, logger logging.Logger) (int, error) {
+	var err error
+	var n int
+	totalAttempts := 5
+
+	for attempt := 1; attempt <= totalAttempts; attempt++ {
+		n, err = w.Write(line)
+		if err == nil || !isRetriable(err) {
+			return n, err
+		}
+		logger.WithError(err).Errorf("Retriable error, retry %d of %d", attempt, totalAttempts)
+		time.Sleep(backoff(attempt))
+	}
+
+	return n, err
+}
--- a/pkg/logbridge/logbridge_test.go
+++ b/pkg/logbridge/logbridge_test.go
@@ -1,0 +1,180 @@
+package logbridge
+
+import (
+	"bytes"
+	"errors"
+	"testing"
+	"time"
+
+	"github.com/square/p2/pkg/logging"
+)
+
+const newLine = byte(10)
+
+type TrackingWriter struct {
+	numWrites int
+}
+
+func (tw *TrackingWriter) Write(p []byte) (int, error) {
+	tw.numWrites++
+	return len(p), nil
+}
+
+func TestLogBridge(t *testing.T) {
+	type testCase struct {
+		inputSize      int
+		bridgeCapacity int
+		expected       int
+	}
+
+	testCases := []testCase{
+		{inputSize: 0, bridgeCapacity: 0, expected: 0},
+		{inputSize: 10, bridgeCapacity: 10, expected: 10},
+		{inputSize: 100, bridgeCapacity: 10, expected: 10},
+		{inputSize: 10, bridgeCapacity: 100, expected: 10},
+	}
+
+	for i, testCase := range testCases {
+		t.Logf("test case %d", i)
+		input := make([]byte, 0, 2*testCase.inputSize)
+		for i := 0; i < testCase.inputSize; i++ {
+			input = append(input, byte('a'), newLine)
+		}
+
+		reader := bytes.NewReader(input)
+		writer := &TrackingWriter{}
+
+		LossyCopy(writer, reader, testCase.bridgeCapacity, logging.DefaultLogger)
+		if writer.numWrites < testCase.expected {
+			t.Errorf("Writer did not receive enough writes, got %d expected: %d", writer.numWrites, testCase.inputSize)
+		}
+	}
+}
+
+// This writer becomes latent after receiving capacity writes.
+type LatentWriter struct {
+	capacity int
+	writes   int
+}
+
+func (sw *LatentWriter) Write(p []byte) (n int, err error) {
+	if sw.writes > sw.capacity {
+		time.Sleep(10 * time.Millisecond)
+	}
+	sw.writes++
+
+	return len(p), nil
+}
+
+func TestLogBridgeLogDrop(t *testing.T) {
+	bridgeCapacity := 3
+	input := []byte{
+		byte('a'), newLine,
+		byte('b'), newLine,
+		byte('c'), newLine,
+		byte('d'), newLine,
+		byte('e'), newLine,
+		byte('f'), newLine}
+	reader := bytes.NewReader(input)
+	writer := &LatentWriter{}
+
+	LossyCopy(writer, reader, bridgeCapacity, logging.DefaultLogger)
+
+	if writer.writes < bridgeCapacity {
+		t.Errorf("Expected at least %d messages to succeed under writer latency, got %d.", bridgeCapacity, writer.writes)
+	}
+}
+
+// This writer returns errors of type (*ErrorWriter).err after numSuccess calls to Write
+type errorWriter struct {
+	numSuccess int
+	writes     int
+	bytes      []byte
+}
+
+type transientErrorWriter struct {
+	numSuccess int
+	writes     int
+	bytes      []byte
+}
+
+func (ew *errorWriter) Write(p []byte) (n int, err error) {
+	if ew.writes > ew.numSuccess {
+		return 0, errors.New("boom")
+	}
+
+	ew.writes++
+
+	ew.bytes = append(ew.bytes, p...)
+
+	return len(p), nil
+}
+
+func (tew *transientErrorWriter) Write(p []byte) (n int, err error) {
+	tew.writes++
+
+	if tew.writes > tew.numSuccess && tew.writes%2 == 0 {
+		return 0, NewRetriableError(errors.New("boom"))
+	}
+
+	tew.bytes = append(tew.bytes, p...)
+	return len(p), nil
+}
+
+func TestErrorCases(t *testing.T) {
+	input := []byte{
+		byte('a'), newLine,
+		byte('b'), newLine,
+		byte('c'), newLine,
+		byte('d'), newLine,
+		byte('e'), newLine,
+		byte('f'), newLine}
+	bridgeCapacity := 6
+
+	// we're testing errors and want the tests to be fast
+	backoff = func(_ int) time.Duration { return time.Duration(0 * time.Millisecond) }
+
+	retriableErrorWriter := &transientErrorWriter{
+		numSuccess: 2,
+	}
+
+	errorWriter := &errorWriter{
+		numSuccess: 4,
+	}
+
+	reader := bytes.NewReader(input)
+	LossyCopy(errorWriter, reader, bridgeCapacity, logging.DefaultLogger)
+
+	if len(errorWriter.bytes) >= len(input) {
+		t.Errorf("Expected non-retriable error to cause line to be skipped.")
+	}
+
+	reader = bytes.NewReader(input)
+	LossyCopy(retriableErrorWriter, reader, bridgeCapacity, logging.DefaultLogger)
+
+	if len(retriableErrorWriter.bytes) != bridgeCapacity {
+		t.Errorf("Expected bridge to successfully retry writes that result in error. Expected %d Got %d", bridgeCapacity, len(retriableErrorWriter.bytes))
+	}
+}
+
+func TestIsRetriable(t *testing.T) {
+	type testCase struct {
+		err         error
+		expectation bool
+	}
+
+	testCases := []testCase{
+		{errors.New("an error"), false},
+		{errors.New("an error"), false},
+		{&retriableError{errors.New("an error")}, false},
+		{retriableError{errors.New("an error")}, true},
+		{NewRetriableError(errors.New("an error")), true},
+	}
+
+	for i, testCase := range testCases {
+		actual := isRetriable(testCase.err)
+		if actual != testCase.expectation {
+			t.Errorf("test case %d: expected %b got %b", i, testCase.expectation, actual)
+		}
+	}
+}
--- a/pkg/pods/manifest_test.go
+++ b/pkg/pods/manifest_test.go
@@ -105,7 +105,7 @@ func TestPodManifestCanBeWritten(t *testing.T) {
 
 func TestPodManifestCanWriteItsConfigStanzaSeparately(t *testing.T) {
 	config := testPod()
-	manifest, err := ManifestFromBytes(bytes.NewBufferString(config).Bytes())
+	manifest, err := ManifestFromBytes([]byte(config))
 	Assert(t).IsNil(err, "should not have erred when building manifest")
 
 	buff := bytes.Buffer{}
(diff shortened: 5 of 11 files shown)


pkg/logbridge/logbridge.go
```diff
@@ -0,0 +1,88 @@
...
+	for {
+		select {
+		case buf, moreLines = <-b.lines:
+			if !moreLines {
+				return nil
+			}
+			_, err := b.dest.Write(buf)
+			if err != nil {
+				logging.DefaultLogger.WithError(err).Error("logbridge: encountered fatal error during write")
+				b.teardown()
```
tummychow: If you get an error here, you'll close the lines channel from underneath the blockingRead goroutine and it'll panic on send. It looks like what you really want is an explicit quit channel as an argument to blockingRead.

pkg/logbridge/logbridge.go
```diff
@@ -0,0 +1,88 @@
...
+				return err
+			}
+		default:
```
tummychow: Doesn't this make the loop spin? range loop over b.lines should suffice.

pkg/logbridge/logbridge.go
```diff
@@ -0,0 +1,88 @@
...
+		dest:          dest,
+		lines:         make(chan []byte, capacity),
+		needsTeardown: true,
+	}
+}
+
+// Run is responsible for reading from src and writing to dst until src is
+// consumed or an error is encountered. The caller is responsible for closing
+// and flushing the reader and writer.
+func (b *Bridge) Run() error {
```
tummychow: One thing I regret about the API of farms which you might want to consider here: if Run and New are separate actions, then it should be safe to call Run repeatedly on a single bridge or otherwise enforce invariants so that it can't be done. I'm not sure it is in this design (since the lines channel is scoped to the structure and not to an instance of Run).

In this case I don't see much value to separating the two functions. I prefer the design of ConsulSessionManager where there is no struct at all and the function just spins out a goroutine.

tummychow: Another thing to note is that this API provides no means of termination (eg quit channel), but since it'll probably be at the level of main, maybe that's not a big deal.

pkg/logbridge/logbridge.go
```diff
@@ -0,0 +1,88 @@
...
+
+	for {
+		select {
+		case buf, moreLines = <-b.lines:
+			if !moreLines {
+				return nil
+			}
+			_, err := b.dest.Write(buf)
+			if err != nil {
+				logging.DefaultLogger.WithError(err).Error("logbridge: encountered fatal error during write")
```
tummychow: avoid default logger outside tests and `package main`. For example if I want to add custom fields to the logger, I'd rather not mutate the default logger global to do so.

pkg/logbridge/logbridge.go
```diff
@@ -0,0 +1,88 @@
...
+			logging.DefaultLogger.NoFields().Debugln("Overflow encountered. Dropping this message.")
+			continue
+		}
+	}
+
+	b.teardown()
+}
+
+func (b *Bridge) teardown() {
+	if b.needsTeardown {
```
tummychow: b.needsTeardown never gets set to false. Although frankly I would ditch this function altogether. I recommend not sharing the send side of a channel across scopes or goroutines anyway, unless very carefully coordinated to avoid panic.

pkg/logbridge/logbridge_test.go
```diff
@@ -0,0 +1,50 @@
...
+	for i, testCase := range testCases {
+		r := rand.New(rand.NewSource(time.Now().Unix()))
+		input := make([]byte, 0, 2*testCase.inputSize)
+		for i := 0; i < testCase.inputSize; i++ {
+			input = append(input, byte(r.Intn(256)), byte(10))
+		}
+
+		reader := bytes.NewReader(input)
+		buf := new(bytes.Buffer)
+		writer := bufio.NewWriter(buf)
```
tummychow: What's the bufio.Writer for? bytes.Buffer is already a writer; my instinctive impression (not saying it's correct, but this is what I think on first reading) is that the bufio wrapper is unnecessary. Does it impact test behavior?

pkg/logbridge/logbridge.go
```diff
@@ -0,0 +1,88 @@
...
+
+	lines         chan []byte
+	needsTeardown bool
+}
+
+// New takes two files src and dest and returns a composite object of the
+// two.
+func New(src io.Reader, dest io.Writer, capacity int) Bridge {
+	if src == nil {
+		panic("Argument Error: Reader cannot be nil")
```
tummychow: Exported panicking is bad. I would document the nil bound in the comment of this function, and either add an error return for it or let the language panic when it calls a method on a nil interface.

pkg/logbridge/logbridge.go
```diff
@@ -0,0 +1,88 @@
...
+		if err := scanner.Err(); err != nil {
+			logging.DefaultLogger.WithError(err).Errorln("Encountered error while reading from src. Proceeding.")
+			break
+		}
+
+		select {
+		case b.lines <- scanner.Bytes():
+		default:
+			logging.DefaultLogger.NoFields().Debugln("Overflow encountered. Dropping this message.")
+			continue
```
tummychow: continue is unnecessary here, it's already at the end of the loop

pkg/logbridge/logbridge.go
```diff
@@ -0,0 +1,88 @@
...
+	for scanner.Scan() {
+		if err := scanner.Err(); err != nil {
+			logging.DefaultLogger.WithError(err).Errorln("Encountered error while reading from src. Proceeding.")
+			break
+		}
+
+		select {
+		case b.lines <- scanner.Bytes():
+		default:
+			logging.DefaultLogger.NoFields().Debugln("Overflow encountered. Dropping this message.")
```
tummychow: output dropped message as a logrus field? (also makes me think of an interesting general question - where do the logs of _this_ application go?)

pkg/logbridge/logbridge.go
anthonybishopric: +1, it's almost always preferable to return `(Bridge, error)` and force clients to decide whether they care. Not panicking is conventional in P2 and is generally reserved for fatal program problems. From [Effective Go](https://golang.org/doc/effective_go.html#panic):

> This is only an example but real library functions should avoid panic. If the problem can be masked or worked around, it's always better to let things continue to run rather than taking down the whole program. One possible counterexample is during initialization: if the library truly cannot set itself up, it might be reasonable to panic, so to speak.

pkg/logbridge/logbridge.go
anthonybishopric: Consider injecting the logger you want in the constructor of `Bridge`

pkg/pods/manifest_test.go
```diff
@@ -134,6 +134,21 @@ func TestRunAs(t *testing.T) {
 	Assert(t).AreEqual(manifest.RunAsUser(), "specialuser", "RunAsUser() didn't match expectations")
 }
 
+func TestLogExecutable(t *testing.T) {
+	config := testPod()
+	manifest, err := ManifestFromBytes(bytes.NewBufferString(config).Bytes())
```
bwester: `string` and `[]byte` have [built-in conversions](https://golang.org/ref/spec#Conversions):

    ManifestFromBytes([]byte(config))

pkg/logbridge/logbridge.go
anthonybishopric: +1 spin loops are bad

pkg/logbridge/doc.go
```diff
@@ -0,0 +1,12 @@
+package logbridge
+
+/*
```
bwester: Package-level documentation needs to go _above_ the `package` declaration for [godoc](http://blog.golang.org/godoc-documenting-go-code) to recognize it.

pkg/logbridge/logbridge_test.go
```diff
@@ -0,0 +1,50 @@
...
+
+func TestLogBridge(t *testing.T) {
+	type testCase struct {
+		inputSize      int
+		bridgeCapacity int
+		expected       int
+	}
+
+	testCases := []testCase{
+		testCase{inputSize: 0, bridgeCapacity: 0, expected: 0},
```
anthonybishopric: you may be able to just write this as:
```
+ testCases := []testCase{
+	        {inputSize: 0, bridgeCapacity: 0, expected: 0},
+		{inputSize: 10, bridgeCapacity: 10, expected: 10},
+		{inputSize: 100, bridgeCapacity: 10, expected: 10},
+		{inputSize: 10, bridgeCapacity: 100, expected: 10},
+	}
```

pkg/logbridge/logbridge_test.go
```diff
@@ -0,0 +1,50 @@
...
+		if err := bridge.Run(); err != nil {
+			t.Fatal(err)
+		}
+
+		if err := writer.Flush(); err != nil {
+			t.Fatal(err)
+		}
+
+		if len(buf.String()) != testCase.expected {
+			t.Errorf("Failure in test case %d. Bridge received %d bytes when expecting %d. \n buf: %v", i, len(buf.String()), testCase.bridgeCapacity, buf)
```
anthonybishopric: Won't this mean the test will continue to pass, even if error messages are being logged? I'm pretty sure if you never call `t.Fail` it won't even print it.

pkg/logbridge/logbridge.go
```diff
@@ -0,0 +1,88 @@
+package logbridge
+
+import (
+	"bufio"
+	"io"
+
+	"github.com/square/p2/pkg/logging"
+)
+
+// This type is agnostic of pipes. os.File conforms to io.Reader and io.Writer
```
bwester: After a quick reading through the types and methods to get an overview of what's happening here, I'm still not quite sure what this comment is referring to.  Could you use this comment section to describe what a `Bridge` is or what it does?  A discussion of `os.File` might be relevant, but not without some context.

pkg/pods/pod.go
```diff
@@ -163,7 +163,7 @@ func (pod *Pod) Launch(manifest Manifest) (bool, error) {
 		}
 	}
 
-	err = pod.buildRunitServices(launchables, manifest.GetRestartPolicy())
```
anthonybishopric: intentional?

anthonybishopric: nevermind, i see it now

pkg/pods/pod.go
```diff
@@ -229,11 +229,12 @@ func (pod *Pod) buildRunitServices(launchables []launch.Launchable, restartPolic
 				return util.Errorf("Duplicate executable %q for launchable %q", executable.Service.Name, launchable.ID())
 			}
 			sbTemplate[executable.Service.Name] = runit.ServiceTemplate{
+				Log: newManifest.GetLogExecutable(),
```
anthonybishopric: So, this approach will require all pod manifests to specify their hook into the `log-bridge`. It seems like it might be an easier deployment story if we make the default log program something configured in the preparer's configuration and then just apply that to all pods.

pkg/logbridge/logbridge.go
```diff
@@ -0,0 +1,88 @@
...
+// so it should not be difficult to migrate
+type Bridge struct {
+	src  io.Reader
+	dest io.Writer
+
+	lines         chan []byte
+	needsTeardown bool
+}
+
+// New takes two files src and dest and returns a composite object of the
```
bwester: It would be useful to make it explicit that `New` is a constructor for a `Bridge`.  What kinds of properties do `src` and `dest` need to provide?  Why must they be files?  How are the files composed?

pkg/logbridge/logbridge.go
bwester: I don't see cancellation as a necessary feature of this method.  It can return when it runs out of work to do.

pkg/logbridge/logbridge.go
```diff
@@ -0,0 +1,88 @@
...
+import (
+	"bufio"
+	"io"
+
+	"github.com/square/p2/pkg/logging"
+)
+
+// This type is agnostic of pipes. os.File conforms to io.Reader and io.Writer
+// so it should not be difficult to migrate
+type Bridge struct {
```
bwester: I don't think it makes sense to package this procedure into its own struct type.  There is only one public method, and you can only call it once.  Instead of this:

    b := logbridge.New(src, dest, 1000)
    b.Run()

what would you think of:

  logbridge.Copy(src, dest, 1000)

pkg/logbridge/logbridge.go
```diff
@@ -0,0 +1,88 @@
...
+	src  io.Reader
+	dest io.Writer
+
+	lines         chan []byte
+	needsTeardown bool
+}
+
+// New takes two files src and dest and returns a composite object of the
+// two.
+func New(src io.Reader, dest io.Writer, capacity int) Bridge {
```
bwester: The functionality of the logbridge is very similar to [`io.Copy()`](https://golang.org/pkg/io/#Copy), but that function takes its arguments in the other order:  `func Copy(dst Writer, src Reader) (written int64, err error)`.  Can you change the order of your arguments to match the standard library?

pkg/logbridge/logbridge.go
```diff
@@ -0,0 +1,88 @@
...
+		}
+	}
+
+	return nil
+}
+
+func (b *Bridge) blockingRead() {
+	scanner := bufio.NewScanner(b.src)
+	for scanner.Scan() {
+		if err := scanner.Err(); err != nil {
```
bwester: Judging from the [`Scanner.Scan()`](https://golang.org/pkg/bufio/#Scanner.Scan) documentation, `Err()` should be used _after_ `Scan()` has already returned false.

pkg/logbridge/logbridge.go
```diff
@@ -0,0 +1,88 @@
...
+				return err
+			}
+		default:
+		}
+	}
+
+	return nil
+}
+
+func (b *Bridge) blockingRead() {
```
bwester: This method is the single place where messages are being sent on `b.lines`, so it in effect _owns_ the send side of that channel.  When it will no longer send anything more on that channel, this method (and this alone) should be responsible for closing it.

The canonical way to ensure channel closure is with a deferred closure before anything else happens. So, this should be the first line of the method:

    defer close(b.lines)

pkg/logbridge/logbridge.go
bwester: :dizzy_face:  :skull: :ghost:

pkg/logbridge/logbridge_test.go
rudle: :scissors:

pkg/logbridge/logbridge_test.go
rudle: The docs say: `Errorf is equivalent to Logf followed by Fail.`

I like to use `Errorf` instead of `Fatalf` exclusively because it runs all my test cases every time giving me extra feedback.

anthonybishopric: Interesting, I thought `Fail` terminated the test immediately. Carry on!

pkg/logbridge/logbridge.go
rudle: Fair point, I think that comment helped me in the moment more than it helps the future versions of us. Will remove

pkg/pods/pod.go
rudle: I agree with you - we need to think through the configuration a bit more. Is this code path the right place to express a default log exec? For gradual rollout and general configurability I want the pod manifest to override any default that P2 supplies.

Fully with you that the kafka logbridge is the right thing for 99% of cases and it will eventually become the default.

anthonybishopric: So maybe the answer you're looking for is 1) preserve the `log_exec` section of the pod manifest 2) also add a config section for the preparer to supply a `default_log_exec` (see `pkg/preparer/setup.go`) 3) use the pod's log exec as an override to the preparer's default.

rudle: After a discussion offline: we decided that the gradual rollout will be managed by the Square linkage to this code. p2-log-bridge will be the logger for all applications henceforth.

This means that all regular STDOUT logs will be unnecessarily buffered  (in a potentially leaky way) in the open source impl.

rudle:
Friends, PTAL. I've addressed the outstanding comments and cleaned a couple of things up along the way.

pkg/logbridge/logbridge.go
```diff
@@ -0,0 +1,116 @@
...
+		if err := scanner.Err(); err != nil {
+			logger.WithError(err).Errorln("Encountered error while reading from src. Proceeding.")
+		}
+	}(dest, src, lines)
+
+	go func(droppedMessages chan []byte) {
+		defer close(droppedMessages)
+
+		numDropped := 0
+		for range droppedMessages {
```
tummychow: This goroutine is ranging over a channel that it itself is responsible for closing; it'll never terminate. I'd let another one of these handle closing this channel (and remember, you have to guarantee both senders have finished before doing so).

pkg/logbridge/logbridge.go
```diff
@@ -0,0 +1,116 @@
...
+				}
+				logger.WithField("dropped line", line).Errorln("Dropped was dropped due to full capacity. If this occurs frequently, consider increasing the capacity of this logbridge.")
+			}
+		}
+		if err := scanner.Err(); err != nil {
+			logger.WithError(err).Errorln("Encountered error while reading from src. Proceeding.")
+		}
+	}(dest, src, lines)
+
+	go func(droppedMessages chan []byte) {
```
tummychow: broadly speaking, I don't see a lot of value in handling dropped messages via a separate goroutine. If counting the number of dropped messages is important, you can implement it locklessly via an int from sync/atomic. The thing about handling dropped messages via a channel is that the dropped messages channel might itself be so backed up that you have to drop the message another way. And if you're going to have another way to drop messages without using channels, why not just use that way all the time?

pkg/logbridge/logbridge.go
```diff
@@ -0,0 +1,116 @@
...
+		for scanner.Scan() {
+			line = scanner.Bytes() // consume a line regardless of the state of the writer
+			select {
+			case lines <- line:
+			default:
+				select {
+				case droppedMessages <- line:
+				default:
+				}
+				logger.WithField("dropped line", line).Errorln("Dropped was dropped due to full capacity. If this occurs frequently, consider increasing the capacity of this logbridge.")
```
tummychow: Looks like this log message is intended for the case where droppedMessages is blocking. It should probably be under the default case of the inner select?

pkg/logbridge/logbridge.go
rudle: Right, I'm thinking that a `defer close(droppedMessages)` at the top of `Copy(…)` will do the trick.

pkg/logbridge/logbridge.go
```diff
@@ -0,0 +1,116 @@
...
+// error semantics to this package
+type RetriableError struct {
+	err error
+}
+
+func (r RetriableError) Error() string {
+	return r.err.Error()
+}
+
+func retriableError(err error) bool {
```
tummychow: nit: this name should be styled after similar functions from os, eg `isRetriable`.

pkg/logbridge/logbridge.go
```diff
@@ -0,0 +1,116 @@
...
+			logger.WithError(err).WithField("dropped line", line).WithField("retried", retriableError(err)).WithField("bytes written", n).Errorln("Encountered a non-recoverable error. Proceeding.")
+		}
+	}
+}
+
+// This is an error wrapper type that may be used to denote an error is retriable
+// RetriableError is exported so clients of this package can express their
+// error semantics to this package
+type RetriableError struct {
+	err error
```
tummychow: if it has only one member, just make it a type alias: `type RetriableError error`. This also means you don't have to rewrite the Error() function.

pkg/logbridge/logbridge.go
```diff
@@ -0,0 +1,116 @@
...
+}
+
+func (r RetriableError) Error() string {
+	return r.err.Error()
+}
+
+func retriableError(err error) bool {
+	_, ok := err.(*RetriableError)
```
tummychow: hm, I feel like this assertion should be to a value. Pointers to error types feel off to me and it results in surprising behavior for an implementer. (eg `RetriableError{err}` would fail this assertion because it needed an &)

pkg/logbridge/doc.go
```diff
@@ -0,0 +1,12 @@
...
+This package implements a log bridge. It is a bridge in the literal sense - it
+will carry logs from one side to the other until it reaches its capacity. When
+the bridge is at its capacity, it will drop new entries instead of blocking.
+This behaviour is desirable because it makes this package safe to attach
+to the logging end of an executable
+
+A typical use for this package is to attach it to stdout of an executable
+and passing the results into your system's proprietary logging backend.
+*/
+
```
bwester: Would you believe that this blank line actually prevents `go doc` from realizing that the above comment applies to the package?

bin/p2-log-bridge/main.go
```diff
@@ -0,0 +1,12 @@
+package main
+
+import (
+	"os"
+
+	"github.com/square/p2/pkg/logbridge"
+	"github.com/square/p2/pkg/logging"
+)
+
+func main() {
```
bwester: I know it's a simple function, but it would be nice to at least support `--help` and give a short description of what the program is for.

pkg/logbridge/logbridge.go
```diff
@@ -0,0 +1,116 @@
...
+			return n, err
+		}
+		logger.WithError(err).Errorf("Retriable error, retry %d of %d", attempt, totalAttempts)
+		time.Sleep(backoff(attempt))
+	}
+
+	return n, err
+}
+
+func writeDroppedMessageWarning(w io.Writer) error {
```
tummychow: Why isn't this implemented in terms of writeWithRetry? Is it intentional that this function ignores retriability of errors?

pkg/logbridge/logbridge_test.go
```diff
@@ -0,0 +1,161 @@
...
+const newLine = byte(10)
+
+type TrackingWriter struct {
+	bytes     []byte
+	numWrites int
+}
+
+func (tw *TrackingWriter) Write(p []byte) (int, error) {
+	tw.numWrites++
+	return len(p), nil
```
tummychow: this function doesn't appear to actually update the byte array inside the writer, intentional?

pkg/logbridge/logbridge_test.go
```diff
@@ -0,0 +1,161 @@
...
+	Bytes    []byte
+}
+
+func (sw *LatentWriter) Write(p []byte) (n int, err error) {
+	if sw.writes > sw.capacity {
+		time.Sleep(10 * time.Millisecond)
+	}
+	sw.writes++
+
+	sw.Bytes = append(sw.Bytes, p...)
```
tummychow: Like I mentioned above, might be worth implementing this as a wrapper around another writer. Then you only need to do the sleep here, and the tracking is provided by trackingwriter, and the bytes comes from bytes.Buffer.

pkg/logbridge/logbridge.go
bwester: Normally yes, but `error` is an interface, and we don't want [all errors to be RetriableErrors](http://play.golang.org/p/4MRykLr4Lc).

tummychow: bleh we'll have to fix a few other places where we already use this pattern. Sometimes I really wish go had strongly typed type aliases.

pkg/logbridge/logbridge.go
```diff
@@ -0,0 +1,116 @@
...
+	"io"
+	"time"
+
+	"github.com/square/p2/pkg/logging"
+)
+
+// Copy implements a buffered copy operation between dest and src.
+// It returns the number of dropped messages as a result of insufficient
+// capacity
+// The caller is responsible for closing & flushing the reader and writer.
```
bwester: > The caller is responsible for closing & flushing the reader and writer.

The caller is busy executing this method and won't be able to do anything to the reader or writer.  I think you're trying to say that something else should take care of making sure the reader and writer have enough data.  Don't worry about it.

pkg/logbridge/logbridge.go
```diff
@@ -0,0 +1,116 @@
...
+func Copy(dest io.Writer, src io.Reader, capacity int, logger logging.Logger) {
+	lines := make(chan []byte, capacity)
+	droppedMessages := make(chan []byte, capacity)
+	go func(dest io.Writer, src io.Reader, lines chan []byte) {
```
bwester: Considering the length of `Copy()`, the file as a whole would read better if you split these async functions into their own top-level functions.  This one right here would make a nice self-contained function instead of an anonymous lambda.  That would also let you document precisely what this routine is doing.

pkg/preparer/setup.go
```diff
@@ -36,15 +36,16 @@ type LogDestination struct {
...
 type Preparer struct {
-	node         string
-	store        Store
-	hooks        Hooks
-	hookListener HookListener
-	Logger       logging.Logger
-	podRoot      string
-	caFile       string
-	authPolicy   auth.Policy
-	consulHealth bool
```
tummychow: probably needs rebase, mpuncel removed the consul health options

pkg/logbridge/logbridge.go
```diff
@@ -0,0 +1,116 @@
...
+
+		scanner := bufio.NewScanner(src)
+		var line []byte
+		for scanner.Scan() {
+			line = scanner.Bytes() // consume a line regardless of the state of the writer
+			select {
+			case lines <- line:
+			default:
+				select {
+				case droppedMessages <- line:
```
bwester: I'm with @tummychow: not sure why you want to offload processing a line not sent to another goroutine.  _Right here_ seems like a nice place to do the handling.  Bonus: you don't need to worry about concurrently updating a counter if this one goroutine is all that writes to it.

pkg/logbridge/logbridge.go
tummychow: btw, the member error needs to be exported as well.

pkg/logbridge/logbridge_test.go
```diff
@@ -0,0 +1,180 @@
...
+		t.Logf("test case %d", i)
+		input := make([]byte, 0, 2*testCase.inputSize)
+		for i := 0; i < testCase.inputSize; i++ {
+			input = append(input, byte('a'), newLine)
+		}
+
+		reader := bytes.NewReader(input)
+		writer := &TrackingWriter{}
+
+		LossyCopy(writer, reader, testCase.bridgeCapacity, logging.DefaultLogger)
```
anthonybishopric: Doesn't this only verify that the synchronous retried copy worked? Is there any way to verify the `go lossyCopy()` part of the `LossyCopy` function?

bin/p2-log-bridge/main.go
```diff
@@ -0,0 +1,52 @@
...
+
+	"gopkg.in/alecthomas/kingpin.v2"
+
+	"github.com/square/p2/pkg/logbridge"
+	"github.com/square/p2/pkg/logging"
+	"github.com/square/p2/pkg/version"
+)
+
+var (
+	durableLogger = kingpin.Arg("exec", "An executable that logbridge will log to with durability").Required().String()
```
anthonybishopric: > with durability

Worth clarifying what that means

rudle: WDYT 
"An executable that logbridge will log to without dropping messages. If a write to STDIN of this program blocks, logbridge will block."

bin/p2-log-bridge/main.go
```diff
@@ -0,0 +1,52 @@
...
+	if err != nil {
+		logging.DefaultLogger.WithError(err)
+		os.Exit(1)
+	}
+}
+
+// This is slightly naive, it could be more narrow in the signals it cares about
+func handleSignals(c io.Closer) {
+	signals := make(chan os.Signal)
+	signal.Notify(signals)
```
anthonybishopric: Check out the main for `bin/p2-preparer` -> signal handling is a little funky in that you can create nasty deadlocks if you don't buffer the channel.

bin/p2-log-bridge/main.go
anthonybishopric: :star:

anthonybishopric:
Overall I think this is a pretty good start, code is well organized and tested :ship:

rudle:
Thanks @anthonybishopric, I'm going to rewrite some of these commits & merge.

PR #320, opened 2015-11-12
Mpuncel/remove use session health

mpuncel:


tummychow:
on first glance all these changes look good to me so i'm going to :+1:, but I bet we can be even more aggressive. My guess is we could also kill the following:

- kp.RegisterService
- checker.fetchNodeHealth
- checker.consulCheckToResult
- various other code paths in checker that branch on which type of health is being used (eg checker.selectResult and checker.Service can probably be simplified)

anyway great change, it's satisfying to see dead code go

anthonybishopric:
:scissors: :scissors: :ship:

pkg/health/checker/checker.go
```diff
@@ -42,8 +46,7 @@ func NewConsulHealthChecker(opts kp.Options) ConsulHealthChecker {
 	client := kp.NewConsulClient(opts)
 	return consulHealthChecker{
 		client:      client,
-		health:      client.Health(),
-		consulStore: kp.NewConsulStore(opts),
+		consulStore: kp.NewConsulStore(client),
 		WaitTime:    1 * time.Minute,
```
tummychow: with these changes, is the `client` field still necessary? Also similar to what you did for kp.Store, i'd change this to consume the client directly (if it even still needs it). WaitTime argument should also be safe to remove.

pkg/health/checker/checker.go
```diff
@@ -57,32 +60,17 @@ func (c consulHealthChecker) WatchNodeService(
...
 			return
 		case <-time.After(1 * time.Second):
-			var err error
-			var checks []*api.HealthCheck
-			checks, curIndex, err = c.fetchNodeHealth(nodename, curIndex)
+			kvCheck, err := c.consulStore.GetHealth(serviceID, nodename)
 			if err != nil {
 				errCh <- err
+			} else {
+				resultCh <- consulWatchToResult(kvCheck)
```
tummychow: blergh polling... luckily only pkg/replication uses this function, and technically it's not _worse_ than it was before, so we can punt on improving this.

tummychow:
:+1: that commit also looks good with the exception of the fields that looked redundant.

pkg/kp/fixtures_test.go
```diff
@@ -10,9 +10,10 @@ import (
 
 type ConsulTestFixture struct {
 	TestServer *testutil.TestServer
-	Store      *consulStore
-	Client     *api.Client
-	T          *testing.T
+	// Store      *consulStore
```
bwester: Delete, don't comment out.

mpuncel: good catch

petertseng:
Think I'm good with my nits now. Just need to fix up those calls to `NewConsulHealthChecker`

PR #321, opened 2015-11-12
Generic prefix watcher

bwester:
There are already several implementations of code that watches Consul for a list of keys and does something with the changed ones.  Each implementation has its own quirks in how it polls Consul (unnecessary) and its own way of processing the results into specific object types (necessary).  The intent of this PR was to write a more uniform, robust implementation of this procedure that could be re-used by other methods inside `pkg/kp/...` and still allow them to add type-specific behavior.

The changes to `"pkg/kp".consulStore.WatchPods()` is illustrative of the kind of change that I intent to make on other stores.  It doesn't reduce the line count, but I think the result is more robust and has cleaner separation into generic and type-specific parts.

The PR isn't quite complete in that I haven't used the new methods to replace all the users I wanted to replace—it turns out that they are quite complicated.  Still, it'll be useful to get some feedback on this code, which seems like it's fairly clean and stable.

### DIFF
--- a/pkg/kp/consulutil/watch.go
+++ b/pkg/kp/consulutil/watch.go
@@ -1,0 +1,165 @@
+package consulutil
+
+import (
+	"time"
+
+	"github.com/square/p2/Godeps/_workspace/src/github.com/hashicorp/consul/api"
+)
+
+// WatchPrefix watches a Consul prefix for changes to any keys that have the prefix. When
+// anything changes, all Key/Value pairs having that prefix will be written to the
+// provided channel.
+//
+// Errors will sent on the given output channel but do not otherwise affect execution. The
+// given output stream will become owned by this function call, and this call will close
+// it when the function ends. This function will run until explicitly canceled by closing
+// the "done" channel. Data is written to the output channel synchronously, so readers
+// must consume the data or this method will block.
+func WatchPrefix(
+	prefix string,
+	clientKV ConsulLister,
+	outPairs chan<- api.KVPairs,
+	done <-chan struct{},
+	outErrors chan<- error,
+) {
+	defer close(outPairs)
+	var currentIndex uint64
+	timer := time.NewTimer(time.Duration(0))
+
+	for {
+		select {
+		case <-done:
+			return
+		case <-timer.C:
+		}
+		timer.Reset(250 * time.Millisecond) // upper bound on request rate
+		pairs, queryMeta, err := SafeList(clientKV, done, prefix, &api.QueryOptions{
+			WaitIndex: currentIndex,
+		})
+		switch err {
+		case CanceledError:
+			return
+		case nil:
+			currentIndex = queryMeta.LastIndex
+			select {
+			case <-done:
+			case outPairs <- pairs:
+			}
+		default:
+			select {
+			case <-done:
+			case outErrors <- err:
+			}
+			timer.Reset(2 * time.Second) // backoff
+		}
+	}
+}
+
+type NewKeyHandler func(key string) chan<- *api.KVPair
+
+type keyMeta struct {
+	created    uint64
+	modified   uint64
+	subscriber chan<- *api.KVPair
+}
+
+// WatchNewKeys watches for changes to a list of Key/Value pairs and lets each key be
+// handled individually though a subscription-like interface.
+//
+// This function models a key's lifetime in the following way. When a key is first seen,
+// the given NewKeyHandler function will be run, which may return a channel. When the
+// key's value changes, new K/V updates are sent to the key's notification channel. When
+// the key is deleted, `nil` is sent. After being deleted or if the watcher is asked to
+// exit, a key's channel will be closed, to notify the receiver that no further updates
+// are coming.
+//
+// WatchNewKeys doesn't watch a prefix itself--the caller should arrange a suitable input
+// stream of K/V pairs, probably from WatchPrefix(). This function runs until the input
+// stream closes. Closing "done" will asynchronously cancel the watch and cause it to
+// eventually exit.
+func WatchNewKeys(pairsChan <-chan api.KVPairs, onNewKey NewKeyHandler, done <-chan struct{}) {
+	keys := make(map[string]*keyMeta)
+
+	defer func() {
+		for _, keyMeta := range keys {
+			if keyMeta.subscriber != nil {
+				close(keyMeta.subscriber)
+			}
+		}
+	}()
+
+	for {
+		var pairs api.KVPairs
+		var ok bool
+		select {
+		case <-done:
+			return
+		case pairs, ok = <-pairsChan:
+			if !ok {
+				return
+			}
+		}
+
+		visited := make(map[string]bool)
+
+		// Scan for new and changed keys
+		for _, pair := range pairs {
+			visited[pair.Key] = true
+			if keyMeta, ok := keys[pair.Key]; ok {
+				if keyMeta.created == pair.CreateIndex {
+					// Existing key that was seen before
+					if keyMeta.subscriber != nil && keyMeta.modified != pair.ModifyIndex {
+						// It's changed!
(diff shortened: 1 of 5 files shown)


pkg/kp/consulutil/watch.go
```diff
@@ -0,0 +1,165 @@
...
+
+// WatchPrefix watches a Consul prefix for changes to any keys that have the prefix. When
+// anything changes, all Key/Value pairs having that prefix will be written to a provided
+// channel.
+//
+// Errors will sent on the given output channel but do not otherwise affect execution. The
+// provided output stream will become owned by this function call, and this call will
+// close it when the function ends. This function will run until explicitly canceled by
+// closing the "done" channel. Data is written to the output channel synchronously, so
+// readers must consume the data to this method will block.
```
rudle: I think you meant to say "consume the data _or_ this method will block"

pkg/kp/consulutil/watch.go
```diff
@@ -0,0 +1,165 @@
...
+		case nil:
+			currentIndex = queryMeta.LastIndex
+			select {
+			case <-done:
+			case outPairs <- pairs:
+			}
+		default:
+			select {
+			case <-done:
+			case outErrors <- err:
```
tummychow: I think genericized watch is a great opportunity to clean up our error handling and error channels have often been the wrong way to do it, so I'd like to see something different here. I have two (mutually exclusive) suggestions:

1. couple errors and meaningful results together. Instead of having a channel for kvpairs and a channel for errors, have a single channel that outputs a (kvpairs, error) struct. That way callers cannot possibly leave the errors unconsumed - by definition they would have to consume errors and results together. In addition this couples the error to the kvpairs that it was generated on (although probably kvpairs would be nil so that's of questionable value).
2. do not propagate errors out at all and just log them.

I've been reading a lot of rust documentation lately so things like haskell either and rust result are really on my mind. I think if we still want to represent errors via channels, a tuple for both the result and the error is the best approach. (Sadly go doesn't have good sum types but a tuple is still better than separate channels imo)

pkg/kp/consulutil/watch.go
```diff
@@ -0,0 +1,165 @@
...
+// key's value changes, new K/V updates are sent to the key's notification channel. When
+// the key is deleted, `nil` is sent. After being deleted or if the watcher is asked to
+// exit, a key's channel will be closed, to notify the receiver that no further updates
+// are coming.
+//
+// WatchNewKeys doesn't watch a prefix itself--the caller should arrange a suitable input
+// stream of K/V pairs, probably from WatchPrefix(). This function runs until the input
+// stream closes. Closing "done" will asynchronously cancel the watch and cause it to
+// eventually exit.
+func WatchNewKeys(pairsChan <-chan api.KVPairs, onNewKey NewKeyHandler, done <-chan struct{}) {
```
tummychow: Interesting, is this a generic farm API? I like it already. However speaking from past experience - it is valuable to be able to block on the termination of this function _and_ all its children. Not implementing that soundly was a mistake in my original design which made teardown of farms a bit hairy at program exit. Might want to work in a sync.WaitGroup here.

pkg/kp/kv.go
```diff
@@ -307,32 +308,38 @@ func (c consulStore) ListPods(keyPrefix string) ([]ManifestResult, time.Duration
...
+			select {
+			case <-quitChan:
+				return
+			case errChan <- NewKVError("list", keyPrefix, err):
+			}
+		}
+	}()
+
+	kvPairsChan := make(chan api.KVPairs)
+	go consulutil.WatchPrefix(keyPrefix, c.client.KV(), kvPairsChan, quitChan, unsafeErrors)
```
tummychow: _hell yes_

tummychow:
need more time to review generic farm and tests in detail, but this is already magnificent and should pave the way for some insane code deduplication

pkg/kp/consulutil/watch.go
```diff
@@ -0,0 +1,165 @@
...
+						close(keyMeta.subscriber)
+					}
+					// Fall through to re-create this key
+				}
+			}
+			// Found a new key
+			keyMeta := &keyMeta{
+				created:    pair.CreateIndex,
+				modified:   pair.ModifyIndex,
+				subscriber: onNewKey(pair.Key),
```
tummychow: I'd pass the entire kvp to this function; that way it doesn't have to block for the first value before instantiating whatever it is it's operating on.

bwester: My first versions of this method did just that.  It doesn't interact well with middle layers that have to parse these raw values into concrete types and might fail.

Imagine that someone is manually modifying, say, resource controller JSON definitions and creates a key that doesn't parse correctly.  If a client picks that instance to start watching the rcstore, what will the store deliver to the client?  Delay the new key notification until a clean value arrives?  Send the notification without a value?  The former adds a lot of complexity to the _type_-store, the latter adds complexity to the consumers.

There is less total complexity when the new key notification is separate from the value stream itself.  And there won't be that much blocking going on: sending to the returned channel happens a few lines later.

pkg/kp/consulutil/watch.go
bwester: :+1: Will fix this.

pkg/kp/consulutil/watch.go
tummychow: hm okay, that's a good point.

tummychow:
All of this looks great to me. I have outstanding comments about handling errors without separate channels and blocking for child termination, but both of those things are probably not as important as just getting this into the wild and migrating other parts of our code onto it. If you would rather skip or punt on my remaining comments, then :shipit:

pkg/kp/consulutil/watch.go
bwester: **Short answer**: Let's save it for another diff.

**Long answer:**

If only Golang had lightweight (syntax-wise) anonymous tuples.

I too have been thinking a lot about how we handle errors, and I actually like both approaches in different circumstances.

(1) works well when the errors are meaningful and fatal and you really need a higher layer to handle them. Imagine something that produces a sequence of values and then hits an error that prevents it from going any further. This would be the most direct way to build something that treats channels like an asynchronous iterator/stream with exception-like behavior (e.g., a Golang version of [Rx](http://reactivex.io/) in Golang or perhaps an Exception Monad in Haskell).

(2) works well when the errors really don't matter.  The method is designed to just keep going in the presence of most errors.  But for debugging and metrics, you still want to know, somehow, that the errors were occurring, so you log them.

Getting back to your question, (2) seems more suitable for this kind of watcher, and my early versions threaded a `logging.Logger` through many of the methods.  I don't like the error channel.  I think it's an anti-pattern that we've repeated everywhere.  But once I started trying to integrate this into our existing code, I took out the loggers and went back to the error channels.  That pattern is too deep, and this change will be big enough already.  Changing all `chan<- error` arguments into `logging.Logger` needs its own PR.

pkg/kp/consulutil/watch.go
bwester: It's the beginnings of a generic farm routine.  You've written two farms already—there's also one in the health checker that does its own thing.  But the common part is watching for a list of keys, diffing them, and processing changes, which this provides.

As for being able to block on the termination of children, I think it doesn't matter if this low-level code terminates or not.  It modifies no shared state and produces no side effects.  If the value stream consumers need to be waited on, that's up to the NewKeyHandler to arrange for this.

This seems reasonable to me, but I don't have your context on this topic. What do you think?

tummychow: hm delegating it to NewKeyHandler seems reasonable. They would have to close over a waitgroup or something, but in practice you'd probably write it like that anyway. Yeah okay, I'm fine with leaving this as is.

bwester:
@tummychow: I'll give others a chance to review this too ;)

PR #323, opened 2015-11-16
Remove typed hooks

tummychow:
hooray for code deletion

### DIFF
--- a/bin/p2-schedule/main.go
+++ b/bin/p2-schedule/main.go
@@ -5,7 +5,6 @@ import (
 	"os"
 
 	"github.com/square/p2/Godeps/_workspace/src/gopkg.in/alecthomas/kingpin.v2"
-	"github.com/square/p2/pkg/hooks"
 	"github.com/square/p2/pkg/kp"
 	"github.com/square/p2/pkg/kp/flags"
 	"github.com/square/p2/pkg/pods"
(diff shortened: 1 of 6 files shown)


bwester:
Looks great! Red diffs are awesome.

:ship:

anthonybishopric:
:shipit:

bin/p2-schedule/main.go
```diff
@@ -13,11 +12,9 @@ import (
...
 
 var (
-	manifests    = kingpin.Arg("manifests", "one or more manifest files to schedule in the intent store").Strings()
-	nodeName     = kingpin.Flag("node", "The node to do the scheduling on. Uses the hostname by default.").String()
-	hookTypeName = kingpin.Flag("hook-type", "Schedule as a hook, not an intended pod, as the given hook type. Can be one of the hooks listed in hooks.go, or the word \"global\"").String()
-	// unhide this flag when typed hooks are removed
-	hookGlobal = kingpin.Flag("hook", "Schedule as a global hook.").Hidden().Bool()
+	manifests  = kingpin.Arg("manifests", "one or more manifest files to schedule in the intent store").Strings()
+	nodeName   = kingpin.Flag("node", "The node to do the scheduling on. Uses the hostname by default.").String()
+	hookGlobal = kingpin.Flag("hook", "Schedule as a global hook.").Bool()
```
petertseng: Do I understand correctly that all hooks are global now? In that case, I might suggest removing "global" so that people don't wonder "So what's a non-global hook?"

PR #324, opened 2015-11-16
Move roll/rc stores to generic watch API

tummychow:
These are probably the lowest-impact watches in the repo, so it shouldn't hurt much if this doesn't work.


mpuncel:
:+1:

pkg/kp/rcstore/consul_store.go
```diff
@@ -138,31 +138,24 @@ func (s *consulStore) List() ([]fields.RC, error) {
 func (s *consulStore) WatchNew(quit <-chan struct{}) (<-chan []fields.RC, <-chan error) {
 	outCh := make(chan []fields.RC)
 	errCh := make(chan error)
+	inCh := make(chan api.KVPairs)
+
+	go consulutil.WatchPrefix(kp.RC_TREE, s.kv, inCh, quit, errCh)
```
bwester: It's a little more complex than this unfortunately.  This watcher might still be writing errors to the error channel when the channel is closed below.

tummychow: good point - how about blocking on the goroutine exit until the kvp channel is closed?

pkg/kp/rcstore/consul_store.go
```diff
@@ -138,31 +138,24 @@ func (s *consulStore) List() ([]fields.RC, error) {
...
 
 		for {
 			select {
 			case <-quit:
 				return
-			case <-time.After(1 * time.Second):
-				listed, meta, err := s.kv.List(kp.RC_TREE, &api.QueryOptions{
-					WaitIndex: currentIndex,
-				})
+			case listed := <-inCh:
```
bwester: You could simplify this to a `for listed := range inCh {` loop.  It would additionally fix the ordering of writes to the error channel.

pkg/kp/rcstore/consul_store.go
```diff
@@ -315,21 +311,12 @@ func (s *consulStore) Watch(rc *fields.RC, quit <-chan struct{}) (<-chan struct{
...
-				})
-				if err != nil {
-					errors <- err
-				} else {
-					curIndex = meta.LastIndex
-					if kvp == nil {
-						// seems this RC got deleted from under us. quitting
-						// would be unexpected, so we'll just wait for it to
-						// reappear in consul
+			case kvps := <-input:
```
bwester: Ditto here.

tummychow:
refactored to range loop as aptly suggested by @bwester

pkg/kp/rcstore/consul_store.go
```diff
@@ -305,38 +299,37 @@ func (s *consulStore) mutateRc(id fields.ID, mutator func(fields.RC) (fields.RC,
 func (s *consulStore) Watch(rc *fields.RC, quit <-chan struct{}) (<-chan struct{}, <-chan error) {
 	updated := make(chan struct{})
 	errors := make(chan error)
-	var curIndex uint64 = 0
+	input := make(chan api.KVPairs)
+
+	targetPath := kp.RCPath(rc.ID.String())
+	go consulutil.WatchPrefix(targetPath, s.kv, input, quit, errors)
```
bwester: I think you really do want a "get" operation here, not a "list."

bwester:
Other than the "get" -> "list" that I'd like you to revert, this looks good.

There is one additional problem that I'm addressing with a new PR which I will submit shortly.

pkg/kp/rcstore/consul_store.go
tummychow: Bleh. There's one other place in p2 that uses watch on get instead of list (checker.watchnodeservice). Should we also add a generic watchkey function for that?

bwester: Yes

tummychow:
added a watch function for get and used that for single RC watches

pkg/kp/consulutil/safe.go
```diff
@@ -106,3 +106,33 @@ func SafeList(
...
+// Like SafeList, but for a single key instead of a list.
+func SafeGet(
+	clientKV ConsulGetter,
+	done <-chan struct{},
+	key string,
+	options *api.QueryOptions,
+) (*api.KVPair, *api.QueryMeta, error) {
+	resultChan := make(chan getReply, 1)
+	go func() {
+		kvp, queryMeta, err := clientKV.Get(key, options)
```
bwester: Wrap the error with `NewKVError`

tummychow: whoopsies, just fixed

pkg/kp/consulutil/watch.go
```diff
@@ -55,6 +55,48 @@ func WatchPrefix(
 	}
 }
 
+// WatchSingle has the same semantics as WatchPrefix, but for a single key in
+// Consul.
```
bwester: This method won't necessarily signal that a key has been deleted if it was quickly recreated afterwards.  For the application of this routine, that doesn't matter, but could you document that as a known issue?

tummychow: improved comments as discussed

PR #325, opened 2015-11-17
Emit KVErrors directly from SafeList()

bwester:
Changes SafeList() to only return errors that have already been wrapped in the
KVError type. This greatly simplifies error handling around the WatchPrefix()
function.

To do this, I had to move the KVError type into the "consulutil" package to avoid
a circular import.

### DIFF
--- a/pkg/kp/consulutil/safe.go
+++ b/pkg/kp/consulutil/safe.go
@@ -2,9 +2,71 @@ package consulutil
 
 import (
 	"errors"
+	"fmt"
+	"path/filepath"
+	"runtime"
 
 	"github.com/square/p2/Godeps/_workspace/src/github.com/hashicorp/consul/api"
+
+	"github.com/square/p2/pkg/util/param"
 )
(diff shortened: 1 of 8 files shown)


pkg/kp/consulutil/safe.go
```diff
@@ -2,10 +2,71 @@ package consulutil
...
+type KVError struct {
+	Op          string
+	Key         string
+	UnsafeError error
+	filename    string
+	function    string
+	lineNumber  int
+}
+
+var _ util.CallsiteError = KVError{}
```
bwester: Actually, I'm going to get rid of this assertion.  `"util"` isn't otherwise included.

tummychow:
:+1: good to see kverror moved to a general package

pkg/kp/consulutil/safe.go
tummychow: might want a comment to that effect then?

PR #326, opened 2015-11-18
Use wrapper structs around typed errors

tummychow:
as observed by @bwester , the type aliasing doesn't work for interface type assertion. We need real structs to wrap around the errors. (cough sum type cough)


bwester:
Yea, oops.

:shipit:

tummychow:
fixed tests to confirm that error types can't assert to one another

PR #327, opened 2015-11-18
Add "list" format for p2-inspect

bwester:
Adds a new flag "--format list" to the options for p2-inspect that lets it
print its data in a new format. The old format has been named "tree" because it
resembles a nested tree where app instances are grouped first by app then
labeled by hostname:

```
{
  "mypod": {
    "myhost": {
      "keys": "values", ...
    },
    ...
  },
  ...
}
```

The new "list" output format prints a flatter JSON structure that has a list of
app instances, one object for each host:

```
[
  {
    "node": "myhost",
    "pod": "mypod",
    "keys": values", ...
  },
  ...
]
```

This output format allows for easier processing by filters, e.g., using the
"jq" command.


tummychow:
:+1:

PR #328, opened 2015-11-23
Update to go1.5.1

tummychow:
Silences a new linker warning for the -X flag:

```
$ rake install
go install -a -ldflags "-X github.com/square/p2/pkg/version.VERSION $(git describe --tags)" ./...
# github.com/square/p2/bin/p2-exec
link: warning: option -X github.com/square/p2/pkg/version.VERSION 0.0.10-49-gc46c850 may not work in future releases; use -X github.com/square/p2/pkg/version.VERSION=0.0.10-49-gc46c850
```

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,7 @@
 language: go
 
 go:
-  - 1.4.2
+  - 1.5.1
 
 before_install:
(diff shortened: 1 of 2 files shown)


mpuncel:
:+1:

PR #329, opened 2015-11-24
Watch one reality pod in replication

tummychow:


pkg/kp/kv.go
```diff
@@ -240,6 +241,35 @@ func (c consulStore) ListPods(keyPrefix string) ([]ManifestResult, time.Duration
...
 
+// WatchPod is like WatchPods, but for a single key only. The output channel
+// may contain nil manifests, if the target key does not exist.
+func (c consulStore) WatchPod(key string, quitChan <-chan struct{}, errChan chan<- error, podChan chan<- ManifestResult) {
+	defer close(podChan)
+
+	kvpChan := make(chan *api.KVPair)
+	go consulutil.WatchSingle(key, c.client.KV(), kvpChan, quitChan, errChan)
+	for pair := range kvpChan {
+		out := ManifestResult{Path: pair.Key}
```
mpuncel: you're doing a nil check after referencing pair.Key, potential panic?

tummychow: whoops good eye, it's supposed to be checking for pair.manifest == nil. will fix

mpuncel:
:ship:

pkg/kp/kv.go
```diff
@@ -240,6 +241,35 @@ func (c consulStore) ListPods(keyPrefix string) ([]ManifestResult, time.Duration
...
+	go consulutil.WatchSingle(key, c.client.KV(), kvpChan, quitChan, errChan)
+	for pair := range kvpChan {
+		out := ManifestResult{Path: key}
+		if pair != nil {
+			manifest, err := pods.ManifestFromBytes(pair.Value)
+			if err != nil {
+				select {
+				case <-quitChan:
+					return
+				case errChan <- util.Errorf("Could not parse pod manifest at %s: %s. Content follows: \n%s", pair.Key, err, pair.Value):
```
tummychow: actually - should there be a `continue` after this?

mpuncel: the old behavior was to keep going. that's how we recover when watches are broken by connections being killed, etc

mpuncel: but maybe there should be a select on the quit channel instead of a `for` over `kvpChan`

pkg/kp/kv.go
```diff
@@ -240,6 +241,35 @@ func (c consulStore) ListPods(keyPrefix string) ([]ManifestResult, time.Duration
...
 }
 
+// WatchPod is like WatchPods, but for a single key only. The output channel
+// may contain nil manifests, if the target key does not exist.
+func (c consulStore) WatchPod(key string, quitChan <-chan struct{}, errChan chan<- error, podChan chan<- ManifestResult) {
+	defer close(podChan)
+
+	kvpChan := make(chan *api.KVPair)
+	go consulutil.WatchSingle(key, c.client.KV(), kvpChan, quitChan, errChan)
+	for pair := range kvpChan {
```
mpuncel: this can't be canceled and probably should be

pkg/kp/kv.go
tummychow: well - the continue would loop back to blocking on kvpChan again. You don't need to select on quit because WatchSingle is doing that for you and will terminate the range loop when it finishes.

pkg/kp/kv.go
tummychow: see below - the termination condition is that when quit closes, WatchSingle gets the quit and closes kvpChan, which then terminates the range loop. No need to select on quit ourselves.

anthonybishopric:
:+1:

tummychow:
rebase

PR #330, opened 2015-11-25
write an 'env' dir in the launchable root with LAUNCHABLE_HOME var

mpuncel:
this will allow us to export an environment variable to launched
processes that will allow them to find the root of the launchable
easily.

### DIFF
--- a/pkg/pods/pod.go
+++ b/pkg/pods/pod.go
@@ -362,14 +362,6 @@ func (pod *Pod) Install(manifest Manifest) error {
 		return util.Errorf("Could not create pod home: %s", err)
 	}
 
-	// we may need to write config files to a unique directory per pod version, depending on restart semantics. Need
-	// to think about this more.
-	err = pod.setupConfig(manifest)
-	if err != nil {
-		pod.logError(err, "Could not setup config")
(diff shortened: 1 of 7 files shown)


mpuncel:
this is currently not used, a separate PR to p2-exec will be made to make it take multiple `-e` arguments, and once that is rolled out a PR to write this env dir to the runit services

pkg/pods/pod.go
```diff
@@ -423,12 +423,26 @@ func (pod *Pod) Verify(manifest Manifest, authPolicy auth.Policy) error {
...
+// processes started by all launchables (as described in
+// http://smarden.org/runit/chpst.8.html, with the -e option), including
+// CONFIG_PATH
+//
+// 3) writes an "env" directory for each launchable. The "env" directory
+// contains environment files specific to a launchable (such as
+// LAUNCHABLE_HOME)
+//
+// We may wish to provide a "config" directory per launchable at some point as
+// well, so that launchables can have different config namespaces
```
anthonybishopric: :star:

anthonybishopric:
:shipit: after renaming `launchable.ConfigDir`, since it was already poorly named and now is very unclear relative to the new `EnvDir`

PR #331, opened 2015-11-25
Allow multiple --env arguments to be passed to p2-exec

mpuncel:
This will allow us to provide environment variables from multiple
directories to a process. The immediate motivation is to combine
pod-wide environment variables with launchable-specific environment
variables.

### DIFF
--- a/bin/p2-exec/main.go
+++ b/bin/p2-exec/main.go
@@ -20,13 +20,15 @@ import (
 )
 
 var (
-	username       = kingpin.Flag("user", "the user to execute as").Short('u').String()
-	envDir         = kingpin.Flag("env", "a directory of env files to add to the environment").Short('e').String()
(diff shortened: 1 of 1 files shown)


mpuncel:
Should this have any protection against conflicting environment variables being set in the various directories?

bin/p2-exec/main.go
```diff
@@ -39,8 +39,8 @@ func main() {
 		os.Clearenv()
 	}
 
-	if *envDir != "" {
-		err := loadEnvDir(*envDir)
+	for _, dir := range *envDir {
```
anthonybishopric: Small request: It'd be nice if the arg description said in what order the env directories are processed - thinking about situations where a launchable needs to override a global value or something.

anthonybishopric:
I think it's desirable to have overlapping entries, it just needs to be clearly stated how the invocation would work.

mpuncel:
updated to show in help text that the directory appearing last will win

anthonybishopric:
:+1:

PR #332, opened 2015-11-25
Pass multiple env dirs to p2-exec when launching a launchable.

mpuncel:
This allows us to utilize the launchable-specific env directory and
export the variables within it to the launched process(es)


anthonybishopric:
Assuming that the ordering is correct, :shipit:

mpuncel:
rebasing

mpuncel:
rebased again :tongue:

PR #333, opened 2015-12-01
Modify the LAUNCHABLE_HOME variable to be LAUNCHABLE_ROOT

mpuncel:
The variable used to point to the top level directory for a launchable,
e.g. /data/pods/<pod>/<launchable> in a default preparer configuration.
It's more useful to have a variable that tells the launchable where it
has been installed, e.g.
/data/pods/<pod>/<launchable>/installs/<launchable>_<version>/

### DIFF
--- a/pkg/opencontainer/containers.go
+++ b/pkg/opencontainer/containers.go
@@ -92,10 +92,6 @@ func (l *Launchable) EnvDir() string {
 	return filepath.Join(l.RootDir, "env")
 }
 
-func (l *Launchable) Path() string {
-	return l.RootDir
-}
-
 // InstallDir is the directory where this launchable should be installed.
 func (l *Launchable) InstallDir() string {
 	launchableName := l.Version()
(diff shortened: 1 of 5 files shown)


petertseng:
I think your commit message has lost some `$VARS`?

Sure nobody is using `LAUNCHABLE_HOME`? 

If we're sure it's OK to remove, could probably delete the code instead of commenting.

mpuncel:
the commit message is fine if you click on Commits, github somehow interpreted it weirdly when displaying.

Nobody at Square is using it, and given we're on v0 I think this sort of change is okay, @anthonybishopric do you agree?

didn't mean to leave it commented, will fix

pkg/opencontainer/containers.go
```diff
@@ -92,9 +92,9 @@ func (l *Launchable) EnvDir() string {
 	return filepath.Join(l.RootDir, "env")
 }
 
-func (l *Launchable) Path() string {
-	return l.RootDir
-}
+// func (l *Launchable) Path() string {
```
anthonybishopric: Delete this

anthonybishopric:
I think this change is fine, I believe nobody is using it. :+1:

petertseng:
Oh it's the angle brackets okay

PR #334, opened 2015-12-01
Fix double-launch bug

bwester:
Adds a quick fix for a bug in which app will be installed twice. More details
are in the patch itself.

Tested manually by watching scheduling a manifest, waiting for the preparer to
pick it up, then scheduling the same thing again. This procedure is sufficient
to get the earlier version to install twice. With this change, it doesn't.

### DIFF
--- a/pkg/preparer/orchestrate.go
+++ b/pkg/preparer/orchestrate.go
@@ -31,6 +31,7 @@ type Hooks interface {
 type Store interface {
 	ListPods(keyPrefix string) ([]kp.ManifestResult, time.Duration, error)
 	SetPod(string, pods.Manifest) (time.Duration, error)
+	Pod(key string) (pods.Manifest, time.Duration, error)
 	DeletePod(key string) (time.Duration, error)
 	WatchPods(string, <-chan struct{}, chan<- error, chan<- []kp.ManifestResult)
 }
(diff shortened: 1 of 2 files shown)


pkg/preparer/orchestrate.go
```diff
@@ -168,6 +169,29 @@ func (p *Preparer) handlePods(podChan <-chan ManifestPair, quit <-chan struct{})
...
+				//
+				// This is a quick fix to ensure that the reality value being used is
+				// up-to-date. The de-bouncing logic in this method should ensure that the
+				// intent value is fresh (to the extent that Consul is timely). Fetching
+				// the reality value again ensures its freshness too.
+				reality, _, err := p.store.Pod(kp.RealityPath(p.node, nextLaunch.ID))
+				if err == pods.NoCurrentManifest {
+					nextLaunch.Reality = nil
+				} else if err != nil {
+					manifestLogger.WithError(err).Errorln("Error getting reality manifest")
```
anthonybishopric: Should we `break` at this point? It seems like we can't know if the pair is valid and the installation would be attempted again anyway when connectivity is restored

bwester: That should work.

bwester:
Updated to add a `break` and to rebase onto master.

tummychow:
:+1:

PR #335, opened 2015-12-01
add byte string converter

anthonybishopric:
This will be used for parsing cgroup memory sizes as well as configuring the p2-preparer with a max `installs` directory size per pod.

### DIFF
--- a/pkg/util/size/bytes.go
+++ b/pkg/util/size/bytes.go
@@ -1,0 +1,70 @@
+// Package size provides a converter between a string representation of a size and
+// a number of bytes, and reverse.
+package size
+
+import (
+	"fmt"
+	"regexp"
+	"strconv"
+	"strings"
+)
+
+type ByteCount float64
+
+const (
+	Byte ByteCount = 1 << (10 * iota)
+	Kibibyte
+	Mebibyte
+	Gibibyte
+	Tebibyte
+)
+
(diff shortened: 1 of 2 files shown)


tummychow:
https://godoc.org/github.com/dustin/go-humanize#ParseBigBytes ?

anthonybishopric:
I didn't want to introduce a dependency to do this since it's pretty straightforward.

tummychow:
okay I'll :+1: since the tests pass, but I think you'd be surprised by the intricacies of this format if you really want to be rigorous about it, which is why I'd rather we _didn't_ write it ourselves. Not to mention there's no real standard for these abbreviations that I know of. For example: 

- the power-of-two prefixes are typically indicated by a lowercase i, eg KiB is 1024 bytes and KB is 1000
- lowercase b is often used to signify bit instead of byte, eg Mb/s and MB/s mean very different things

bwester:
As far as I can find, the kernel tends to export numbers without any suffix, so `strconv.ParseInt()` should suffice.

anthonybishopric:
Good points. Flexible input format for callers is probably the main reason why it's written this way. I really don't want to have to explain why `KiB` is parsed differently from `KB`. Same thing when someone uses `mb` instead of `MB`.

bwester:
What, you don't think developers will give their memory sizes in millibits?  ;)

PR #336, opened 2015-12-02
Add flags to p2-launch to support user auth policy

mpuncel:
Adds --auth-type, --keyring, and --deploy-policy flags.

If launching a signed manifest from an untrusted source, a user of
p2-launch may wish to verify the signature of the manifest against a
deploy policy just as the preparer does.

### DIFF
--- a/bin/p2-launch/main.go
+++ b/bin/p2-launch/main.go
@@ -3,14 +3,26 @@ package main
 import (
 	"log"
 
-	"github.com/square/p2/Godeps/_workspace/src/gopkg.in/alecthomas/kingpin.v2"
+	"github.com/square/p2/pkg/auth"
+	"github.com/square/p2/pkg/logging"
 	"github.com/square/p2/pkg/pods"
+	"github.com/square/p2/pkg/preparer"
+	"github.com/square/p2/pkg/util"
 	"github.com/square/p2/pkg/version"
+
+	"github.com/square/p2/Godeps/_workspace/src/github.com/Sirupsen/logrus"
+	"github.com/square/p2/Godeps/_workspace/src/gopkg.in/alecthomas/kingpin.v2"
 )
 
 var (
-	manifestURI = kingpin.Arg("manifest", "a path or url to a pod manifest that will be installed and launched immediately.").String()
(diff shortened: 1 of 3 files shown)


bin/p2-launch/main.go
```diff
@@ -36,3 +53,44 @@ func main() {
...
 }
+
+func authorize(manifest pods.Manifest) error {
+	var policy auth.Policy
+	var err error
+	switch *authType {
+	case auth.NULL_POLICY:
+		return nil
+	case auth.KEYRING_POLICY:
+		return util.Errorf("--auth-type of '%s' currently not supported by p2-launch", *authType)
```
mpuncel: i didn't support it because you have to provide a list of authorized users and wasn't sure how best to do that from the command line

pkg/auth/policy.go
```diff
@@ -15,6 +15,14 @@ import (
 	"github.com/square/p2/pkg/util"
 )
 
+// These string constants are used to determine the requested auth policy type
+// both in the "auth" section of preparer config and in p2-launch flags
+const (
+	NULL_POLICY    = "none"
```
bwester: Use camel case: `NullPolicy`, etc.

bin/p2-launch/main.go
bwester: Add another flag for specifying them?

bin/p2-launch/main.go
```diff
@@ -36,3 +53,44 @@ func main() {
...
+	case auth.USER_POLICY:
+		if *keyring == "" {
```
bwester: You should also check `deployPolicy` similarly.

pkg/auth/policy.go
mpuncel: that's a name conflict with the `struct` type (for `NullPolicy` at least)

mpuncel: maybe i'll call these `*Type`

mpuncel:
@bwester how's this looking?

bwester:
Did you mean to push an update?

mpuncel:
ah i think i did my PR from the wrong branch, will open another

PR #337, opened 2015-12-02
Prune launchable directories over specified size limit.

anthonybishopric:
This change prevents launchable directories from exceeding a
configurable size limit by removing cached installations from oldest
to newest until the limits have been overcome. This pruning facility
ignores directories pointed at by the `current` and `last` symlinks.

### DIFF
--- a/pkg/hoist/prune.go
+++ b/pkg/hoist/prune.go
@@ -1,0 +1,100 @@
+package hoist
+
+import (
+	"io/ioutil"
+	"os"
+	"path/filepath"
+	"sort"
+
+	"github.com/square/p2/pkg/util/size"
+)
+
+// Prune will attempt to get the total size of the launchable's installs
+// directory to be equal to or less than the maximum specified number of
+// bytes. This method will preserve any directories that are pointed to
+// by the `current` or `last` symlinks. Installations will be removed from
+// oldest to newest.
+func (hl *Launchable) Prune(maxSize size.ByteCount) error {
+	curTarget, err := os.Readlink(hl.CurrentDir())
+	if os.IsNotExist(err) {
+		curTarget = ""
+	} else if err != nil {
+		return err
+	}
+	curTarget = filepath.Base(curTarget)
+
+	lastTarget, err := os.Readlink(hl.LastDir())
+	if os.IsNotExist(err) {
+		lastTarget = ""
+	} else if err != nil {
+		return err
+	}
+	lastTarget = filepath.Base(lastTarget)
+
+	installs, err := ioutil.ReadDir(hl.AllInstallsDir())
+	if os.IsNotExist(err) {
+		return nil
+	} else if err != nil {
+		return err
+	}
+
+	var totalSize size.ByteCount = 0
+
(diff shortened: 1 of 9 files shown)


pkg/hoist/prune.go
```diff
@@ -0,0 +1,101 @@
...
+			return err
+		}
+		totalSize += size.ByteCount(installSize)
+		installSizes[i.Name()] = installSize
+	}
+
+	oldestFirst := installsByAge(installs)
+	sort.Sort(oldestFirst)
+
+	// remove all the installs that are older than the threshold
```
petertseng: comment seems not quite accurate

pkg/pods/pod.go
```diff
@@ -38,6 +46,7 @@ var DefaultP2Exec = "/usr/local/bin/p2-exec"
 
 func init() {
 	Log = logging.NewLogger(logrus.Fields{})
+
```
petertseng: this seems stray

pkg/pods/pod.go
```diff
@@ -30,6 +31,13 @@ var (
 	// ExperimentalOpencontainer permits the use of the experimental "opencontainer"
 	// launcahble type.
 	ExperimentalOpencontainer = param.Bool("experimental_opencontainer", false)
+
+	// MaxLaunchableSpaceStr is a string representation of size.ByteCount that will be
+	// passed to launchables as part of pruning.
+	MaxLaunchableSpaceStr = param.String("max_launchable_space", (size.Gibibyte * 10).String())
+
+	// Default,
+	MaxLaunchableSpace = size.Gibibyte * 10
```
petertseng: is this being used? why both this and `DefaultAllowableDiskUsage`

pkg/pods/pod.go
```diff
@@ -193,6 +202,21 @@ func (pod *Pod) Launch(manifest Manifest) (bool, error) {
...
+		return err
+	}
+	for _, l := range launchables {
+		err = l.Prune(max)
+		if err != nil {
+			pod.logLaunchableError(l.ID(), err, "Could not prune directory")
+			// Don't return here. We want to prune other launchables if possible.
+		}
+	}
+	return err
```
petertseng: if I have three launchables and they returned in order no error, error, no error, this function would return no error, is that OK? I have a feeling it should return the error sent by the second prune.

(It will be logged correctly of course, so that's fine, but should probably return the error too)

pkg/pods/pod.go
anthonybishopric: spurious! will remove

pkg/pods/pod.go
anthonybishopric: will remove

pkg/hoist/prune.go
anthonybishopric: yep, wrong

pkg/pods/pod.go
anthonybishopric: Good catch

petertseng:
:+1:

PR #338, opened 2015-12-04
Add flags to p2-launch to support user auth policy

mpuncel:
Adds --auth-type, --keyring, and --deploy-policy flags.

If launching a signed manifest from an untrusted source, a user of
p2-launch may wish to verify the signature of the manifest against a
deploy policy just as the preparer does.

### DIFF
--- a/bin/p2-launch/main.go
+++ b/bin/p2-launch/main.go
@@ -3,14 +3,27 @@ package main
 import (
 	"log"
 
-	"github.com/square/p2/Godeps/_workspace/src/gopkg.in/alecthomas/kingpin.v2"
+	"github.com/square/p2/pkg/auth"
+	"github.com/square/p2/pkg/logging"
 	"github.com/square/p2/pkg/pods"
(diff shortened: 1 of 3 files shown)


bin/p2-launch/main.go
```diff
@@ -36,3 +54,73 @@ func main() {
 		log.Fatalln("Unsuccessful launch of one or more things in the manifest")
 	}
 }
+
+func authorize(manifest pods.Manifest) error {
```
anthonybishopric: Is this copied from `setup.go`? Seems like it should be provided by something in the auth package.

mpuncel: it is, but I don't know what I can put in `auth` that isn't some weird function like `func GetAuthPolicy(authType string, keyringPath *string, deployPolicy *string)` and callers would be providing `nil` values all around

anthonybishopric:
:+1:

mpuncel:
rebased

PR #339, opened 2015-12-05
Allow shorthand for memory limits.

anthonybishopric:
This change allows manifest authors to specify a shorthand for
the byte limits of their cgroups. Previously, all `memory` entries
for the `cgroup` stanza had to be in the form of an integer
number of bytes. Authors may now use a shorthand as defined by
the package `github.com/square/p2/pkg/util/size`.

An author wishing to limit their launchable to a maximum resident
memory limit of 4 Gibibytes may use the shorthand `4G` instead of
`4294967296`.

The old form of specifying an integer limit of bytes is also supported.

### DIFF
--- a/pkg/cgroups/cgroups.go
+++ b/pkg/cgroups/cgroups.go
@@ -169,7 +169,11 @@ func (subsys Subsystems) Write(config Config) error {
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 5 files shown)


bwester:
Looks fine to me. I'm not sure that the test failure is related, but it's worth investigating.

PR #341, opened 2015-12-07
add pod to the healthchecker logger

anthonybishopric:
Small change, makes it easier to identify healthchecking in logs.


mpuncel:
:+1:

PR #342, opened 2015-12-08
Preserve cgroup serialization

bwester:
Changes the cgroup.Config struct so that it preserves its old serialization.
For example, an input of "4K" is valid and parses to 4096, but regardless of
the input format, it will always serialize as 4096. This will allow older tools
to continue to work with the update input format, and it will ensure that there
is a single canonical representation of any manifest.

### DIFF
--- a/pkg/cgroups/config_test.go
+++ b/pkg/cgroups/config_test.go
@@ -12,17 +12,15 @@ import (
 func TestMarshalIntegerByteCount(t *testing.T) {
 	integeredMemory := []byte(`memory: 500`)
 	config := Config{}
-	yaml.Unmarshal(integeredMemory, &config)
-	memory, err := config.MemoryByteCount()
-	Assert(t).IsNil(err, "Should not have erred getting the byte count")
-	Assert(t).AreEqual(memory, size.ByteCount(500), "Should have unmarshaled the integer representation of bytes")
+	err := yaml.Unmarshal(integeredMemory, &config)
+	Assert(t).IsNil(err, "Should not have erred unmarshaling")
+	Assert(t).AreEqual(config.Memory, size.ByteCount(500), "Should have unmarshaled the integer representation of bytes")
 }
 
 func TestMarshalStringByteCount(t *testing.T) {
 	integeredMemory := []byte(`memory: 500G`)
 	config := Config{}
-	yaml.Unmarshal(integeredMemory, &config)
-	memory, err := config.MemoryByteCount()
-	Assert(t).IsNil(err, "Should not have erred getting the byte count")
-	Assert(t).AreEqual(memory, 500*size.Gibibyte, "Should have unmarshaled the integer representation of bytes")
+	err := yaml.Unmarshal(integeredMemory, &config)
+	Assert(t).IsNil(err, "Should not have erred unmarshaling")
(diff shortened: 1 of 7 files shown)


pkg/util/size/bytes.go
```diff
@@ -68,3 +68,31 @@ func Parse(sizeStr string) (ByteCount, error) {
...
+func (b ByteCount) MarshalYAML() (interface{}, error) {
+	return uint64(b), nil
+}
+
+// UnmarshalYAML unserializes a YAML representation of the ByteCount.
+func (b *ByteCount) UnmarshalYAML(unmarshal func(interface{}) error) error {
+	var intVal uint64
+	err := unmarshal(&intVal)
+	if err == nil {
+		*b = ByteCount(intVal)
```
bwester: Huh... forgot a `return` and all the unit tests still pass.

anthonybishopric: If there was an error it could be in the string form, right?

anthonybishopric: Ignore me, yeah there should be a `return` here. The YAML -> string unmarshaling of an integer value passes, which is why tests aren't erring.

bwester: Well, I _intended_ for there to be a `return`.  But if the value "`4`" can parse into a string, it has the right behavior anyway, so the first extraction into an integer isn't needed at all. I would say now that it _shouldn't_ be there, to keep the code simpler.

bwester:
Updated:
- Removed unnecessary parsing logic
- Broke apart unit tests

pkg/cgroups/config_test.go
```diff
@@ -12,17 +12,15 @@ import (
...
 
 func TestMarshalStringByteCount(t *testing.T) {
 	integeredMemory := []byte(`memory: 500G`)
 	config := Config{}
-	yaml.Unmarshal(integeredMemory, &config)
-	memory, err := config.MemoryByteCount()
-	Assert(t).IsNil(err, "Should not have erred getting the byte count")
-	Assert(t).AreEqual(memory, 500*size.Gibibyte, "Should have unmarshaled the integer representation of bytes")
+	err := yaml.Unmarshal(integeredMemory, &config)
+	Assert(t).IsNil(err, "Should not have erred unmarhsaling")
```
tummychow: nit: typo

also while I'm nitpicking, I probably should have noticed this earlier, but both these tests are poorly named - we should probably call them `Unmarshal<foo>ByteCount`.

I also wonder if we want a roundtripping test for the Config struct?

tummychow:
just nits looks sound :+1:

pkg/cgroups/config_test.go
bwester: I'll fix the typo. The method names are close enough I'd rather not touch them.  The round-trip test would be useful—I'll add.

bwester: Actually, no, I think it's reasonable to trust that the YAML module can serialize and parse a struct type where there isn't any custom parsing for the type as a whole.

bwester:
Updated:
- Fixed typo
- One more unit test for `ByteCount`

tummychow:
:ship:

PR #344, opened 2015-12-10
Do not return partial lists from WatchPods

tummychow:
Fixes the bug where, if a pod gets an unparseable manifest, the preparer will attempt to delete it. Basically we just don't send the list if any of its elements failed to parse. We might miss updates that are relevant in the process, but right now the structure of the code makes it difficult to handle both a parse error in one manifest and a desired update in another. The long-term fix is to rewrite the orchestrate loop around NewKeyWatcher.

### DIFF
--- a/pkg/kp/kv.go
+++ b/pkg/kp/kv.go
@@ -284,6 +284,7 @@ func (c consulStore) WatchPods(keyPrefix string, quitChan <-chan struct{}, errCh
 
 	kvPairsChan := make(chan api.KVPairs)
 	go consulutil.WatchPrefix(keyPrefix, c.client.KV(), kvPairsChan, quitChan, errChan)
+WATCH_LOOP:
 	for kvPairs := range kvPairsChan {
 		manifests := make([]ManifestResult, 0, len(kvPairs))
 		for _, pair := range kvPairs {
@@ -294,6 +295,11 @@ func (c consulStore) WatchPods(keyPrefix string, quitChan <-chan struct{}, errCh
 					return
 				case errChan <- util.Errorf("Could not parse pod manifest at %s: %s. Content follows: \n%s", pair.Key, err, pair.Value):
 				}
+				// do not send the partially built list, since we know it's incomplete
+				// and some callers may care about elements being removed
+				// TODO(sjung): those callers should probably be refactored to
+				// use NewKeyWatcher instead
+				continue WATCH_LOOP
 			} else {
 				manifests = append(manifests, ManifestResult{manifest, pair.Key})
 			}


pkg/kp/kv.go
```diff
@@ -294,6 +295,11 @@ func (c consulStore) WatchPods(keyPrefix string, quitChan <-chan struct{}, errCh
 					return
 				case errChan <- util.Errorf("Could not parse pod manifest at %s: %s. Content follows: \n%s", pair.Key, err, pair.Value):
 				}
+				// do not send the partially built list, since we know it's incomplete
+				// and some callers may care about elements being removed
+				// TODO(sjung): those callers should probably be refactored to
+				// use NewKeyWatcher instead
+				continue WATCH_LOOP
```
anthonybishopric: This means an invalid manifest would cause a whole host to stop doing deploys. I'd argue that is a worse situation to be in, especially in cases where we may accidentally already have invalid data in `/intent`. This change seems dangerous to me.

tummychow: That's exactly correct, one bad manifest will lock down the host until it's excised. But I am not convinced it's worse:

- the bad manifest was probably put there by a deploy which means someone is expecting something to happen, so they will investigate when it goes wrong. Yes I can think of cases where a person is not going to pay attention, but the worst case scenario there is nothing happens.
- the alternative is _deleting the pod from the host_ which takes down the service itself. I would much rather have a deploy outage than an outage for the service being deployed. With the former, developers can't push new code; with the latter, you get customer visible impact.

The long term solution is to rewrite the loop altogether with `NewKeyWatcher`/`WatchNewKeys` so that the manifests are parsed after being handed out to child goroutines, but that's going to take more time to get right.

bwester: > they will investigate when it goes wrong

In general, manual intervention doesn't scale, and it makes the system more difficult to operate. We should never fall back to "a human will fix it" unless we truly can't avoid it. Humans can spend many hours failing to notice an "obvious" problem.

> a person is not going to pay attention

Is my deploy taking a long time, or is the process blocked?  I don't think the users of this system will necessarily be able to distinguish the two without resorting to a timeout, even if they are paying attention.  And until this hypothetical user asks for manual intervention, the host is effectively down for _all_ users.  Failure isolation is a very useful property to have.

> the alternative is [do nothing].

We can brainstorm for real alternatives offline.

> outage for the service being deployed

A service outage is not necessarily the outcome from this bug being triggered, especially because our standard deployment tools rely on a node coming up healthy before moving on.

> The long term solution is [...], but that's going to take more time to get right.

It's worth the effort.

anthonybishopric:
Going to close this for now. I agree we need to do something about this issue but I think we can do better by addressing this with upfront linting.

PR #345, opened 2015-12-10
Do not panic on nonexistent update

tummychow:
really wish 404 was an error case rather than a nil kvp


anthonybishopric:
:ship:

bwester:
:shipit:

PR #346, opened 2015-12-12
handle kills correctly

anthonybishopric:
Fixes two bugs involving forcible termination of runit services. In both cases, a reliance on SIGKILL to terminate a process would incorrectly prevent other services from being killed or started.


bwester:
Looks fine to me.  :shipit: once the flaky unit tests have a good run.

PR #347, opened 2015-12-15
Always release RC locks in a rolling update

bwester:
When a rolling update executes, it acquires locks on the RCs involved in the
change, and those locks should be released when the update is finished.
However, the existing code never verifies that the locks are actually released.
Transient errors could allow the updater to leak a locked RC, preventing
anything from modifying that RC until the updater's Consul session is
restarted.

This change reduces the capacity for deadlocks among rolling updates by
retrying all unlock operations. As long as Consul is accessible, the unlocks
should eventually succeed. During prolonged Consul outages, the updater's
session could expire, also dropping the locks and allowing the rolling update
to restart.

### DIFF
--- a/pkg/roll/run_update.go
+++ b/pkg/roll/run_update.go
@@ -2,6 +2,7 @@ package roll
 
 import (
 	"fmt"
+	"sync"
 	"time"
 
 	"github.com/square/p2/Godeps/_workspace/src/github.com/Sirupsen/logrus"
@@ -87,24 +88,15 @@ func RetryOrQuit(f func() error, quit <-chan struct{}, logger logging.Logger, er
 func (u update) Run(quit <-chan struct{}) (ret bool) {
 	u.logger.NoFields().Debugln("Locking")
 	// TODO: implement API for blocking locks and use that instead of retrying
-	if !RetryOrQuit(u.lockRCs, quit, u.logger, "Could not lock update") {
+	if !RetryOrQuit(
+		func() error { return u.lockRCs(quit) },
+		quit,
+		u.logger,
+		"Could not lock update",
+	) {
 		return
 	}
-
-	// release both locks on termination
-	defer func() {
-		// failure to unlock probably means:
-		// - session was already invalidated, lock was already released when
-		//   that happened
-		// - network error, session will probably time out soon
-		// either way no point retrying
-		if err := u.lock.Unlock(u.lockPath(u.OldRC)); err != nil {
-			u.logger.WithError(err).Warnln("Could not unlock old RC for updates")
-		}
-		if err := u.lock.Unlock(u.lockPath(u.NewRC)); err != nil {
-			u.logger.WithError(err).Warnln("Could not unlock new RC for updates")
-		}
-	}()
+	defer u.unlockRCs(quit)
 
 	u.logger.NoFields().Debugln("Enabling")
 	if !RetryOrQuit(u.enable, quit, u.logger, "Could not enable/disable RCs") {
@@ -226,7 +218,7 @@ func (u update) lockPath(id rcf.ID) string {
 	return kp.LockPath(kp.RCPath(id.String(), "update"))
 }
 
-func (u update) lockRCs() error {
+func (u update) lockRCs(done <-chan struct{}) error {
 	newPath := u.lockPath(u.NewRC)
 	oldPath := u.lockPath(u.OldRC)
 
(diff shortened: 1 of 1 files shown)


pkg/roll/run_update.go
```diff
@@ -238,10 +230,16 @@ func (u update) lockRCs() error {
 	}
 
 	err = u.lock.Lock(oldPath)
```
mpuncel: couldn't you just put the `RetryOrQuit` here?

pkg/roll/run_update.go
```diff
@@ -238,10 +230,16 @@ func (u update) lockRCs() error {
...
+	if err != nil {
+		// The second key couldn't be locked, so release the first key before retrying.
+		RetryOrQuit(
+			func() error { return u.lock.Unlock(newPath) },
+			done,
+			u.logger,
+			fmt.Sprintf("unlocking %s", newPath),
+		)
+	}
 	if _, ok := err.(kp.AlreadyLockedError); ok {
```
mpuncel: `err` is guaranteed to be `nil` here

mpuncel: actually no i'm confused

pkg/roll/run_update.go
```diff
@@ -87,24 +88,15 @@ func RetryOrQuit(f func() error, quit <-chan struct{}, logger logging.Logger, er
...
-		// - network error, session will probably time out soon
-		// either way no point retrying
-		if err := u.lock.Unlock(u.lockPath(u.OldRC)); err != nil {
-			u.logger.WithError(err).Warnln("Could not unlock old RC for updates")
-		}
-		if err := u.lock.Unlock(u.lockPath(u.NewRC)); err != nil {
-			u.logger.WithError(err).Warnln("Could not unlock new RC for updates")
-		}
-	}()
+	defer u.unlockRCs(quit)
```
mpuncel: under what conditions does `quit` fire? what if releasing locks hasn't succeeded by the time that happens?

bwester: The `quit` channel is signaled up in [`Farm.releaseChild()`](https://github.com/square/p2/blob/master/pkg/roll/farm.go#L197).  As to when this "child" update is released, I see three triggers:

1. The session is expiring—all locks have been or will soon be released automatically, and you just want to exit all pending operations as fast as possible.
2. The whole Farm is being asked to quit—the whole server is exiting, and you can treat this like No. 1.
3. The Update object has been removed from Consul—this is the mechanism by which a rolling update can be stopped mid-update, so I'd expect it to happen frequently relative to Nos. 1-2.

The problem is that there isn't a way to distinguish why the update is exiting.  Adding that signal means a lot more complexity here.  Is it worth the extra development time and effort?  I don't know.

I can only point out that `RetryOrQuit()` always performs 1 attempt whether or not `quit` is signaled.

pkg/roll/run_update.go
bwester: I will try to structure these conditionals better to make it more clear what's going on.

anthonybishopric: Is an unlock an error when there was no lock acquired in the first place?

anthonybishopric: (should be under line 236)

pkg/roll/run_update.go
```diff
@@ -250,6 +248,26 @@ func (u update) lockRCs() error {
...
+// unlockRCs releases the locks on the old and new RCs. To avoid a system-wide deadlock in
+// RCs,  this method ensures that the locks are always released, either by retying until
+// individual releases are successful or until the session is reset.
+func (u update) unlockRCs(done <-chan struct{}) {
+	wg := sync.WaitGroup{}
+	for _, path := range []string{u.lockPath(u.NewRC), u.lockPath(u.OldRC)} {
+		wg.Add(1)
+		go func(path string) {
+			RetryOrQuit(
+				func() error { return u.lock.Unlock(path) },
```
anthonybishopric: Same question here

pkg/roll/run_update.go
bwester: If no lock was acquired in the first place, control flow shouldn't come here :)

But if you're asking, "Does `Unlock()` return an error if the lock isn't held?" the answer is maybe.  If the key is not held by any session, an attempt to unlock it will succeed.  If the lock is held by a different session, it's an error.  If control flow makes it here and the lock is held by a different sessions, the locker's session must have been lost.

anthonybishopric: I'm mostly concerned that there are cases where the `RetryOrQuit(unlock)` will go on forever. Particularly, if `lock.Lock` fails to acquire a lock, we get an `err`, and we enter this loop wherein `Unlock` will return an error every time we try to unlock our nonexistent lock (perhaps because our session expired between lock(newPath) and lock(oldPath)).

pkg/roll/run_update.go
mpuncel: to me it looks like `RetryOrQuit()` will try forever until either `quit` is signaled or the operation succeeds

bwester: > RetryOrQuit() will try forever until either quit is signaled or the operation succeeds

Yes, I agree.  In particular, if `quit` has already been signaled before `RetryOrQuit()` is called, it still performs the operation once.

I don't understand what you're trying to point out.

pkg/roll/run_update.go
bwester: The short explanation is that if a lock succeeds, I assume the key will stay assigned to the client's session and will only change for two reasons: the client explicitly unlocked it or the session expired.  Until one of those two things happens, a repeated attempt to unlock the session will eventually succeed.  If it succeeds, the method exits.  If the session expires, this will be noticed outside the process and the `done` channel will be signaled, and the method exits.

pkg/roll/run_update.go
mpuncel: I was concerned about it running forever but you've addressed that in another thread :+1:

mpuncel:
:ship:

anthonybishopric:
:shipit:

PR #348, opened 2015-12-15
Limit healthchecks to 5 seconds

anthonybishopric:
Services should always be able to respond to a healthcheck within
five seconds of the connection initiation. If longer-running status
checks are required, they should be backgrounded and cached.

### DIFF
--- a/pkg/preparer/setup.go
+++ b/pkg/preparer/setup.go
@@ -133,8 +133,8 @@ func loadToken(path string) (string, error) {
 	return strings.TrimSpace(string(token)), nil
 }
 
-// getTLSClient constructs an HTTP client that uses keys/certificates in the given files.
-func getTLSClient(certFile, keyFile, caFile string) (*http.Client, error) {
+// getTLSConfig constructs a tls.Config that uses keys/certificates in the given files.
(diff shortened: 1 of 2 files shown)


pkg/preparer/setup.go
```diff
@@ -161,7 +161,7 @@ func getTLSClient(certFile, keyFile, caFile string) (*http.Client, error) {
 	}
 
 	if len(certs) == 0 && cas == nil {
-		return http.DefaultClient, nil
+		return &tls.Config{}, nil
```
bwester: `nil` is a perfectly fine value to return to get a [default config](https://golang.org/pkg/net/http/#Transport).

anthonybishopric: :+1: will change

pkg/watch/health.go
```diff
@@ -19,6 +19,9 @@ import (
 // Duration between health checks
 const HEALTHCHECK_INTERVAL = 1 * time.Second
 
+// Maximum allowed time for a single check
+const HEALTHCHECK_TIMEOUT = 5 * time.Second
```
bwester: `"./pkg/util/param".Int`?

rudle:
:+1:

PR #349, opened 2015-12-15
Always delete pods in a rolling update

bwester:
Changes the rolling update execution logic to always delete pods from the old
RC. The existing use case, "when scheduling a new pod on a node will also stop
the old one (eg hoist artifacts)," turns out to be a bad example. If the old RC
isn't kept up-to-date and later becomes re-enabled (e.g., by reversing the
update), it will instantly re-deploy all its nodes, probably violating the
update's minimum replicas constraint. It's better just to update all the time.

Other resulting changes: remove the "DeletePods" field from the Update struct
and the "-delete" flags from the p2-rctl command.

### DIFF
--- a/pkg/roll/run_update.go
+++ b/pkg/roll/run_update.go
@@ -174,12 +174,10 @@ ROLL_LOOP:
 					"new":  newNodes,
 					"next": next,
 				}).Infoln("Undergoing next update")
-				if u.DeletePods {
-					err = u.rcs.AddDesiredReplicas(u.OldRC, -next)
-					if err != nil {
-						u.logger.WithError(err).Errorln("Could not decrement old replica count")
-						break
-					}
+				err = u.rcs.AddDesiredReplicas(u.OldRC, -next)
+				if err != nil {
+					u.logger.WithError(err).Errorln("Could not decrement old replica count")
+					break
 				}
 				err = u.rcs.AddDesiredReplicas(u.NewRC, next)
 				if err != nil {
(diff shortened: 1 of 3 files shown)


pkg/roll/run_update.go
```diff
@@ -182,12 +182,10 @@ ROLL_LOOP:
 					"new":  newNodes,
 					"next": next,
 				}).Infoln("Undergoing next update")
-				if u.DeletePods {
-					err = u.rcs.AddDesiredReplicas(u.OldRC, -next)
-					if err != nil {
-						u.logger.WithError(err).Errorln("Could not decrement old replica count")
-						break
-					}
+				err = u.rcs.AddDesiredReplicas(u.OldRC, -next)
```
anthonybishopric: So this works because when the old RC is having its count reduced, it is also disabled, correct? Otherwise there's a risk that it could unschedule one of its active, not yet deployed-over instances.

anthonybishopric: Or fight with the new RC over ownership of a host, which may be worse

bwester: "Works" is a relative term right now, as RCs are generally too weak to guarantee much of anything.  But yes, it is critical that the old RC is disabled during this transformation.  And it will (almost certainly) be disabled.

anthonybishopric:
I still question whether it's even a good idea to let RCs be health-agnostic. Eventually, it seems like the goal should be to make RCs health-aware, so they can redeploy nodes that are terminated. Something to discuss later. Otherwise, :+1:

bwester:
Health checking is one thing.  It'd be nice right now if RCs kept track of their children at all!

All I'm trying to do is get rolling updates working (with high probability, for some reasonable definition of "working") as fast as possible, which means deferring on fixing some of the more foundational issues with RCs.

PR #350, opened 2015-12-15
Give p2-rctl time to clean up

bwester:
Increases the "LockDelay" timeout in the Consul session used in p2-rctl rolling
update operations. This change gives p2-rctl a little time to detect that its
lease on the RCs has expired and cease its modifications before another p2-rctl
(in Farm mode) thinks it has exclusive access. It doesn't technically improve
the correctness of the rolling update, but hopefully it makes it harder for a
bad execution to occur.


anthonybishopric:
Was just wondering about this, :+1:

PR #352, opened 2015-12-16
p2-rctl: Name new Consul sessions

bwester:
Adds a node-identifying description to Sessions created by the p2-rctl command.
Without this information, it's difficult to trace back and query which nodes
has opened which session.

### DIFF
--- a/bin/p2-rctl/main.go
+++ b/bin/p2-rctl/main.go
@@ -153,6 +153,17 @@ func main() {
 	}
 }
 
+// SessionName returns a node identifier for use when creating Consul sessions.
+func SessionName() string {
+	hostname, err := os.Hostname()
+	if err != nil {
+		hostname = "unknown hostname"
+	}
+	// Current time of "Jan 2, 2006, 15:04:05" turns into "2006-01-02-15-04-05"
+	timeStr := time.Now().Format("2006-01-02-15-04-05")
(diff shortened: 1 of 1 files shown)


bin/p2-rctl/main.go
```diff
@@ -153,6 +153,15 @@ func main() {
 	}
 }
 
+// SessionName returns a node identifier for use when creating Consul sessions.
+func SessionName() string {
+	hostname, err := os.Hostname()
+	if err != nil {
+		hostname = "unknown hostname"
+	}
+	return fmt.Sprintf("p2-rctl:%s:%s", hostname, time.Now().Format("0102-150405"))
```
rudle: This date formatting is hard to sort lexicographically (ie. YYYY-mm-dd would be more amenable to sorting). Is that use case relevant here?

rudle:
:+1: lgtm pending my dumb question about date formats

bin/p2-rctl/main.go
bwester: Not useful IMHO. Note that the hostname comes first, and Consul exposes a `CreateIndex` property on the session itself to order by creation (logical) time.

mpuncel:
i also thought date format looked strange but :+1:

bwester:
I'll change the date :)

mpuncel:
:+1:

PR #353, opened 2015-12-18
Fix ordering violation in stream package

bwester:
A bug in the stream package was found while running unit tests for an unrelated
change. The test failure was:

    --- FAIL: TestUnsubscribeIndependence (0.00s)
            publisher_test.go:254: sub2 got unexpected sequence: [foo]

A subscriber should have received the current value of the StringValuePublisher
as of the time of the Subscribe() call; in this execution, that initial value
was skipped.

Root cause: when the subscriber unsubscribes, it closes its "canceled" channel
to signal the publisher that it shouldn't wait when sending to the subscriber.
Then, in the presence of an active reader, it is possible for sendFirst() to
non-deterministically choose to abort its send while the main publisher read()
loop non-deterministically chooses to send a subsequent value.

This commit fixes the ordering violation by ensuring that sendFirst() always
finishes executing before read() can publish a value, then by checking if the
subscription has been canceled before attempting to send a new value.

Tested with ~300,000 executions of the unit test without any failures.

### DIFF
--- a/pkg/util/stream/publisher.go
+++ b/pkg/util/stream/publisher.go
@@ -44,12 +44,13 @@ func (p *StringValuePublisher) read() {
 		subs := p.subscribers
 		p.mutex.Unlock()
 		for _, s := range subs {
-			s.waitForReady()
-			select {
-			case s.values <- val:
-				// This send might block if the subscriber has quit receiving from its
-				// values channel.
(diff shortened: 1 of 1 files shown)


rudle:
:+1:

anthonybishopric:
:+1:

PR #355, opened 2015-12-23
Sync hooks once before starting normal pod management

mpuncel:
This avoids a race condition where simultaneous pod updates and hook
updates are scheduled while the preparer is stopped. We want to
guarantee that when it is started, hooks are processed before any other
pods so that th epod will benefit from any hooks being installed

### DIFF
--- a/pkg/preparer/listener.go
+++ b/pkg/preparer/listener.go
@@ -2,6 +2,7 @@ package preparer
 
 import (
 	"path/filepath"
+	"time"
 
 	"github.com/square/p2/Godeps/_workspace/src/github.com/Sirupsen/logrus"
 	"github.com/square/p2/pkg/auth"
@@ -13,6 +14,7 @@ import (
 
 type IntentStore interface {
 	WatchPods(watchPath string, quit <-chan struct{}, errCh chan<- error, manifests chan<- []kp.ManifestResult)
+	ListPods(keyPrefix string) ([]kp.ManifestResult, time.Duration, error)
 }
 
 type HookListener struct {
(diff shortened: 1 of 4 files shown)


pkg/preparer/listener.go
```diff
@@ -60,6 +62,24 @@ func (l *HookListener) Sync(quit <-chan struct{}, errCh chan<- error) {
 	}
 }
 
+// Sync the hooks just once. This is useful at preparer startup so that hooks
+// will be synced before normal pod management begins
+func (l *HookListener) SyncOnce(errCh chan<- error) {
+	pods, _, err := l.Intent.ListPods(l.HookPrefix)
```
anthonybishopric: Could probably just return `error` from this function

pkg/preparer/orchestrate.go
```diff
@@ -56,6 +56,18 @@ func (p *Preparer) WatchForHooks(quit chan struct{}) {
 	}
 }
 
+// Sync the hooks just once. This is useful at preparer startup so that hooks
+// will be synced before normal pod management begins
+func (p *Preparer) SyncHooksOnce() {
+	hookErrCh := make(chan error)
```
anthonybishopric: Probably a good idea to just return this `error` directly, since there will only be one opportunity

mpuncel: any number of hooks can fail to be installed, i thought we should mimic the behavior of the normal loop

anthonybishopric:
:ship:

mpuncel:
fixed existing tests/added a test

PR #356, opened 2015-12-23
Add watches for label selectors

anthonybishopric:
This change introduces watches for label selectors. This is especially useful for forming abstractions around label selectors, such as Kubernetes-style Services.

This implementation attempts to cache the result of the label tree, with a watch in place to get updates over time. Watches share the same cache result. See function documentation for details.

### DIFF
--- a/pkg/labels/applicator.go
+++ b/pkg/labels/applicator.go
@@ -61,4 +61,14 @@ type Applicator interface {
 
 	// Return all objects of the given type that match the given selector
 	GetMatches(selector labels.Selector, labelType Type) ([]Labeled, error)
+
+	// Watch a label selector of a give type and see updates to that set
+	// of Labeled over time. If an error occurs, the Applicator implementation
+	// is responsible for handling it and recovering gracefully.
+	//
+	// The watch may be terminated by the underlying implementation without signaling on
+	// the quit channel - this will be indicated by the closing of the result channel. For
+	// this reason, the safest way to process the result of this channel is to use
+	// it in a for loop.
+	WatchMatches(selector labels.Selector, labelType Type, quitCh chan struct{}) chan []Labeled
 }
--- a/pkg/labels/applicator.go
+++ b/pkg/labels/applicator.go
@@ -62,13 +62,13 @@ type Applicator interface {
 	// Return all objects of the given type that match the given selector
 	GetMatches(selector labels.Selector, labelType Type) ([]Labeled, error)
 
-	// Watch a label selector of a give type and see updates to that set
+	// Watch a label selector of a given type and see updates to that set
 	// of Labeled over time. If an error occurs, the Applicator implementation
 	// is responsible for handling it and recovering gracefully.
 	//
 	// The watch may be terminated by the underlying implementation without signaling on
 	// the quit channel - this will be indicated by the closing of the result channel. For
 	// this reason, the safest way to process the result of this channel is to use
-	// it in a for loop.
-	WatchMatches(selector labels.Selector, labelType Type, quitCh chan struct{}) chan []Labeled
+	// it in a for loop. Alternatively, you may check for nullity of the result []Labeled
+	WatchMatches(selector labels.Selector, labelType Type, quitCh chan struct{}) chan *[]Labeled
 }
--- a/pkg/labels/consul_aggregator.go
+++ b/pkg/labels/consul_aggregator.go
@@ -1,0 +1,174 @@
+package labels
+
+import (
+	"sync"
+	"time"
+
+	"github.com/square/p2/pkg/kp/consulutil"
+	"github.com/square/p2/pkg/logging"
+
+	"github.com/square/p2/Godeps/_workspace/src/github.com/Sirupsen/logrus"
+	"github.com/square/p2/Godeps/_workspace/src/github.com/hashicorp/consul/api"
+	"github.com/square/p2/Godeps/_workspace/src/k8s.io/kubernetes/pkg/labels"
+)
+
+// minimum required time between retrievals of label subtrees.
+var AggregationRateCap = 10 * time.Second
+
+// linked list of watches with control channels
+type selectorWatch struct {
+	selector labels.Selector
+	resultCh chan []Labeled
+	canceled chan struct{}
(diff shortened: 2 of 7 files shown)


pkg/labels/applicator.go
```diff
@@ -61,4 +61,14 @@ type Applicator interface {
 
 	// Return all objects of the given type that match the given selector
 	GetMatches(selector labels.Selector, labelType Type) ([]Labeled, error)
+
+	// Watch a label selector of a give type and see updates to that set
```
mpuncel: given

pkg/labels/consul_aggregator.go
```diff
@@ -0,0 +1,174 @@
...
+	resultCh chan []Labeled
+	canceled chan struct{}
+	next     *selectorWatch
+}
+
+func (s *selectorWatch) append(w *selectorWatch) {
+	if w == nil {
+		return
+	}
+	if s.next != nil {
```
mpuncel: is there a race between two `append` calls?

mpuncel: this comment should have been on the `else`

mpuncel: ah you have this handled below

pkg/labels/consul_aggregator.go
```diff
@@ -0,0 +1,174 @@
...
+	}()
+	return watch.resultCh
+}
+
+func (c *consulAggregator) removeWatch(watch *selectorWatch) {
+	close(watch.canceled)
+	close(watch.resultCh)
+	c.watcherLock.Lock()
+	defer c.watcherLock.Unlock()
+	if c.watchers == watch {
```
mpuncel: could `delete()` handle this case by doing `s = s.next`?

pkg/labels/consul_aggregator_test.go
```diff
@@ -0,0 +1,133 @@
...
+	"testing"
+	"time"
+
+	"github.com/square/p2/pkg/logging"
+
+	. "github.com/square/p2/Godeps/_workspace/src/github.com/anthonybishopric/gotcha"
+	"github.com/square/p2/Godeps/_workspace/src/k8s.io/kubernetes/pkg/labels"
+)
+
+// defer the _result_ of this function to change the time for the duration
```
mpuncel: is this comment relevant? if so i'm confused what it means

pkg/labels/consul_aggregator_test.go
```diff
@@ -0,0 +1,133 @@
...
+	go func() {
+		for _ = range res {
+		}
+		success <- struct{}{}
+	}()
+
+	select {
+	case <-success:
+	case <-time.After(time.Second):
+		t.Fatal("Should still be waiting or processing results after a second")
```
mpuncel: should not?

mpuncel:
ship it!

pkg/labels/consul_aggregator.go
anthonybishopric: That would reassign the pointer `s` within the `delete` function but wouldn't change the pointer of `c.watchers`

pkg/labels/consul_aggregator.go
```diff
@@ -0,0 +1,174 @@
...
+	"github.com/square/p2/Godeps/_workspace/src/k8s.io/kubernetes/pkg/labels"
+)
+
+// minimum required time between retrievals of label subtrees.
+var AggregationRateCap = 10 * time.Second
+
+// linked list of watches with control channels
+type selectorWatch struct {
+	selector labels.Selector
+	resultCh chan *[]Labeled
```
bwester: As in the other PR, why `*[]Labeled` instead of `[]Labeled`?

pkg/labels/consul_aggregator.go
```diff
@@ -0,0 +1,174 @@
...
+	canceled chan struct{}
+	next     *selectorWatch
+}
+
+func (s *selectorWatch) append(w *selectorWatch) {
+	if w == nil {
+		return
+	}
+	if s.next != nil {
+		s.next.append(w)
```
bwester: Here and in `delete()`, do you know if the Go compiler will apply any tail call optimization to limit the size of the stack while traversing data structures?

pkg/labels/consul_applicator.go
```diff
@@ -188,5 +194,28 @@ func convertLabeledToKVP(l Labeled) (*api.KVPair, error) {
...
+// Due to the possibility that this tree might change quite frequently in environments
+// with lots of concurrent deployments, the aggregated result from Consul will only be queried
+// by a maximum frequency of once per LabelAggregationCap.
+//
+// Preparers should not use the consulApplicator's implementation of WatchMatches directly due
+// to the cost of querying for this subtree on any sizeable fleet of machines. Instead, preparers should
+// use the httpApplicator from a server that exposes the results of this (or another)
+// implementation's watch.
+func (c *consulApplicator) WatchMatches(selector labels.Selector, labelType Type, quitCh chan struct{}) chan *[]Labeled {
+	aggregator, ok := c.aggregators[labelType]
```
bwester: You'll need to wrap the singleton lookup/instantiation code inside a mutex to prevent a data race.

pkg/labels/consul_aggregator.go
anthonybishopric: I don't think so - I can change this to a non-recursive implementation if this becomes a concern.

pkg/labels/consul_applicator.go
anthonybishopric: Good catch

pkg/labels/consul_aggregator.go
anthonybishopric: It's to distinguish the case when a label selector yields nothing versus the case when the channel has been closed.

bwester: Isn't `[]Labeled{}` vs `nil` already enough for that?

pkg/labels/consul_aggregator.go
anthonybishopric: It is, I'm not sure what I was thinking. Will change

PR #357, opened 2015-12-23
Instantiate empty config map in GetConfig() if necessary

mpuncel:
This is for convenience so that callers modifying a manifest's config do
not need to check if it is nil before doing so


anthonybishopric:
:+1:

bwester:
:ship:

PR #358, opened 2015-12-24
Use cache on first adding a watcher

anthonybishopric:
This change allows new watchers to immediately use the cached label
result. This is important in cases where the label tree is relatively
unchanging and the ongoing Consul watch will take a long time to
return new results. A new watcher added immediately after such a
Consul watch completed might have to wait for minutes before a new
result was returned.

### DIFF
--- a/pkg/labels/consul_aggregator.go
+++ b/pkg/labels/consul_aggregator.go
@@ -18,7 +18,7 @@ var AggregationRateCap = 10 * time.Second
 // linked list of watches with control channels
 type selectorWatch struct {
 	selector labels.Selector
-	resultCh chan *[]Labeled
+	resultCh chan WatchResult
 	canceled chan struct{}
 	next     *selectorWatch
 }
@@ -52,7 +52,7 @@ type consulAggregator struct {
 	path           string
 	kv             consulutil.ConsulLister
 	watchers       *selectorWatch
-	labeledCache   *[]Labeled // cached contents of the label subtree
+	labeledCache   []Labeled // cached contents of the label subtree
 	aggregatorQuit chan struct{}
 }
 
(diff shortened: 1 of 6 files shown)


pkg/labels/consul_aggregator.go
```diff
@@ -52,7 +52,7 @@ type consulAggregator struct {
 	path           string
 	kv             consulutil.ConsulLister
 	watchers       *selectorWatch
-	labeledCache   []Labeled // cached contents of the label subtree
+	labeledCache   *[]Labeled // cached contents of the label subtree
```
bwester: Since `[]T` is itself nullable, what benefit does `*[]T` provide?

anthonybishopric: Yeah, not sure what I was thinking, setting to `[]T`

anthonybishopric: So, elsewhere in this code I use `chan *[]Labeled`. This is primarily to signal to clients when there are 0 matches for a query instead of the watch being terminated for whatever reason. See http://play.golang.org/p/t8tlbBF9jO

On reflection, it's probably worth altering the chan to be `chan LabelWatchResult` and make that 0 vs nil distinction explicit.

pkg/labels/consul_aggregator.go
```diff
@@ -89,6 +89,11 @@ func (c *consulAggregator) Watch(selector labels.Selector, quitCh chan struct{})
 	} else {
 		c.watchers.append(watch)
 	}
+	// send first value immediately to take advantage of the cache.
+	if c.labeledCache != nil {
+		c.doMatch(watch)
```
bwester: This introduces a deadlock because nothing is reading `resCh` yet.

pkg/labels/consul_applicator.go
```diff
@@ -208,7 +210,9 @@ func convertLabeledToKVP(l Labeled) (*api.KVPair, error) {
 // to the cost of querying for this subtree on any sizeable fleet of machines. Instead, preparers should
 // use the httpApplicator from a server that exposes the results of this (or another)
 // implementation's watch.
-func (c *consulApplicator) WatchMatches(selector labels.Selector, labelType Type, quitCh chan struct{}) chan *[]Labeled {
+func (c *consulApplicator) WatchMatches(selector labels.Selector, labelType Type, quitCh chan struct{}) chan WatchResult {
+	c.aggregatorMux.Lock()
+	defer c.aggregatorMux.Unlock()
```
bwester: I'm a little nervous about releasing this lock after calling `Wait()` given the complexity of that method. Can you release the lock before? `Wait()` can handle the extra concurrency.

pkg/labels/consul_aggregator.go
anthonybishopric: That is definitely true. Will rework this PR

bwester:
I still don't see the benefit of `*[]Labeled` or `WatchResult` over `[]Labeled`. Let's chat offline.

PR #360, opened 2016-01-06
pod: Write LAUNCHABLE_ID

petertseng:


petertseng:
Please confirm that what is desired is for example `p2-preparer__preparer` (for our preparer manifest). If `preparer` was desired instead, should say so now.

mpuncel:
I would think it should be `preparer`, i.e. the key of the launchable in the launchables stanza.

For example:

```
launchables:
  foo:
    type: hoist
    location: somewhere.com/foo
```
would yield `foo`

bwester:
Please add a description to your git commit that explains what "LAUNCHABLE_ID" is and **why** you want it to be written.  Assume that someone will be looking at your commit message without necessarily looking through the entire diff itself.

bwester:
I expect behavior like @mpuncel specified.  If you have a particular use case in mind that needs a specific behavior, let's discuss it.

petertseng:
Given that we want the behavior specified in https://github.com/square/p2/pull/360#issuecomment-169474668 , let's talk a little about the ID field of a `hoist.Launchable`. Currently, it stores `"#{podID}__#{launchableID}`, to make it unique from other launchables with the same ID but under different pods.

If we want to just have just `foo` not `podID__foo`, I probably need to store the launchable ID (`foo`) as a separate field, while avoiding confusion with the existing ID.

I am thinking of three ways to deal with this:
1) rename current ID to `RunitID` or `FullID` (or anything else you suggest) and then create a new `ID` that just has the launchable ID
2) keep current `ID` the same, add something like `LaunchableID` (stutters because now it's `launchable.LaunchableID()` or `IDSuffix` or `Name`
3) keep everything as it is, and simply strip out everything before `__` returned by current `ID()` when writing `LAUNCHABLE_ID`

I'll think about this and am open to suggestions.

As for why we want this, need to ask our users who request it

mpuncel:
I think 1) is the best option. We should be more specific about the way the current ID is used, and it seems like it's only for runit name spacing currently. RunitID or ServiceID or something would be good in my opinion

rudle:
Option 1 seems ok to me also. The namespacing required for runit is a bit of a kludge and I don't think it should be our ID for launchables.

petertseng:
Great, option 1 has been implemented (I took `ServiceID`), with much detail in individual commit messages.

petertseng:
Some of the uses of `ServiceID` could possibly be `ID` instead, mostly the ones in error messages. However, 

1) Unless error messages already include the pod ID (which might be true if it's a logrus field we set), it is useful to have `ServiceID` there, it seems like it would be useful to know what pod is erroring
2) I wanted to change as little existing behavior as possible.

Now, the places where we most definitely need to use `ServiceID` instead of `ID` are the runit service namd and the cgroup name. I think that's all of them.

We can discuss whether to change the others to `ID`, but I'm in favor of taking fewer risks here

mpuncel:
:ship:

PR #361, opened 2016-01-08
Test using Consul agent in dev mode

bwester:
This PR updates the Consul dependency to v0.6.1 and uses its new "dev mode" feature to run unit tests using an in-process in-memory instance of Consul.  It no longer requires testers to locally install a specific version of Consul, and I expect that failing unit tests shouldn't get stuck on Travis CI due to child process cleanup issues.  Startup of each unit test is considerably faster too, enough that this could potentially be used for all unit tests that go through a storage system (to be considered in subsequent work).

Due to the size of the last commit in this stack, it will be impractical to leave inline comments on the PR proper. Please leave inline feedback on individual commits.


mpuncel:
looks great :+1:

petertseng:
I grepped for instances of TestServer to see whether there were any left since https://github.com/bwester/p2/commit/f96effd524edfc3e2fc6e1b2c391e2a922d43058 claims it will go away. All usages of it do go away but the `consultest` package comment still refers to it. I would have left a comment on the relevant line but it is too far away from any change for me to comment on.

bwester:
> the consultest package comment still refers to [TestServer]
will fix.

petertseng:
OK, looking good with just the two comment changes that have already been described.

bwester:
Updated to move a comment and remove a reference to TestServer

petertseng:
:+1:

PR #362, opened 2016-01-11
Make manifest.GetConfig() return a deep copy of the config map

mpuncel:
This avoids modification of config maps returned by GetConfig() from
mutating the underlying manifest or eachother.

Copying is done by marshaling to yaml and then unmarshaling into a new
map. This introduces a chance for an error, but by doing error checking
when configs are set we can guarantee that in practice these errors
should not occur, and if they do then something is very wrong and thus
we panic.

### DIFF
--- a/pkg/pods/manifest.go
+++ b/pkg/pods/manifest.go
@@ -33,7 +33,7 @@ type LaunchableStanza struct {
 type ManifestBuilder interface {
 	GetManifest() Manifest
 	SetID(string)
-	SetConfig(config map[interface{}]interface{})
+	SetConfig(config map[interface{}]interface{}) error
 	SetRunAsUser(user string)
 	SetStatusPort(port int)
 	SetStatusHTTP(statusHTTP bool)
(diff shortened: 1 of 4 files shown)


pkg/pods/manifest.go
```diff
@@ -125,14 +125,38 @@ func (manifest *manifest) SetLaunchables(launchableStanzas map[string]Launchable
...
 }
 
-func (m manifestBuilder) SetConfig(config map[interface{}]interface{}) {
+func (m manifestBuilder) SetConfig(config map[interface{}]interface{}) error {
+	// Confirm that the data passed in can be successfully serialized as YAML
+	_, err := yaml.Marshal(config)
+	if err != nil {
+		return err
+	}
 	m.Config = config
```
mpuncel: I should also copy here, will update

pkg/pods/manifest_test.go
```diff
@@ -230,3 +233,44 @@ func TestGetConfigInitializesIfEmpty(t *testing.T) {
...
+
+type CantMarshal struct{}
+
+var _ yaml.Marshaler = CantMarshal{}
+
+func (CantMarshal) MarshalYAML() (interface{}, error) {
+	return nil, errors.New("Can't yaml marshal this type")
+}
+
+func TestSetConfigErrsIfBadYAML(t *testing.T) {
```
bwester: :+1:

pkg/pods/manifest_test.go
```diff
@@ -230,3 +233,44 @@ func TestGetConfigInitializesIfEmpty(t *testing.T) {
...
+	cantMarshalConfig := make(map[interface{}]interface{})
+	cantMarshalConfig["foo"] = CantMarshal{}
+
+	builder := NewManifestBuilder()
+	err := builder.SetConfig(cantMarshalConfig)
+	Assert(t).IsNotNil(err, "Should have erred setting config with a type that cannot be marshaled as YAML")
+}
+
+func TestSetConfigCopies(t *testing.T) {
```
bwester: :+1:

bwester:
Overall: :ship:

PR #363, opened 2016-01-11
Fix flake in replication test by reporting unhealthy for all nodes

mpuncel:
In TestStopsIfLockDestroyed, the fake health checker would report
healthy for the first node in the replication, but then assert that the
second node is not deployed. This is incorrect behavior and creates a
race where replication might move onto the second node before the test
is performed. Reporting unhealthy for all nodes properly creates the
behavior where replication will only be applied to a single node


mpuncel:
example failure: https://travis-ci.org/square/p2/builds/101702438

bwester:
Fine by me.
:ship:

PR #364, opened 2016-01-13
Add support to kv.go for deleting a health entry for a service/node

mpuncel:
This is useful for clearing out consul data for a host when it is being
removed from a p2 cluster


anthonybishopric:
@bwester what happens when the preparer session terminates? Shouldn't the health entries be eliminated as a result?

bwester:
All health checks are acquired using the preparer's session and will be deleted automatically when the preparer exits. While the preparer is running, the health monitor watches the node's intent store to determine what checks need to run, adding and deleting them as needed. Really, you should be able to treat the health monitor almost as a disjoint system from the rest of the preparer. It takes care of itself.

What's the benefit in eagerly deleting health checks?

mpuncel:
after reading your comment, I believe there are none :)

PR #365, opened 2016-01-20
Update p2 log bridge path

rudle:
When https://github.com/square/p2/pull/319 was merged, we were not sure how log-bridge would be installed on our systems. It turns out that it fits naturally as a secondary launchable for p2-preparer.

### DIFF
--- a/bin/p2-log-bridge/main.go
+++ b/bin/p2-log-bridge/main.go
@@ -1,11 +1,13 @@
 package main
 
 import (
+	"fmt"
 	"io"
 	"os"
 	"os/exec"
-	"os/signal"
+	"sync"
 
+	"github.com/square/p2/Godeps/_workspace/src/golang.org/x/sys/unix"
 	"github.com/square/p2/Godeps/_workspace/src/gopkg.in/alecthomas/kingpin.v2"
 	"github.com/square/p2/pkg/logbridge"
 	"github.com/square/p2/pkg/logging"
@@ -20,31 +22,61 @@ func main() {
 	kingpin.Version(version.VERSION)
 	kingpin.Parse()
 
+	if err := setSTDINToBlock(); err != nil {
+		logging.DefaultLogger.WithError(err).Error("fatal error setting STDIN to block")
+		os.Exit(1)
+	}
+
+	var wg sync.WaitGroup
 	loggerCmd := exec.Command(*durableLogger)
 	durablePipe, err := loggerCmd.StdinPipe()
 	if err != nil {
-		logging.DefaultLogger.WithError(err).Errorln("failed during logbridge setup")
-		os.Exit(1)
+		logging.DefaultLogger.WithError(err).Error("fatal error during configuration of subordinate log command")
 	}
 
-	go handleSignals(durablePipe)
+	wg.Add(1)
+	go func(loggerCmd exec.Cmd) {
+		defer wg.Done()
 
-	go func(r io.Reader, durableWriter io.Writer, nonDurableWriter io.Writer, logger logging.Logger) {
-		logbridge.Tee(r, durableWriter, nonDurableWriter, logger)
+		if err := loggerCmd.Start(); err != nil {
+			logging.DefaultLogger.WithError(err).Errorln("fatal error during execution of subordinate log command")
+			os.Exit(1)
+		}
+
+		err := loggerCmd.Wait()
+		if err != nil {
+			logging.DefaultLogger.WithError(err).Errorln("fatal error in subordinate log command")
+			os.Exit(1)
+		}
+	}(*loggerCmd)
+
+	wg.Add(1)
+	go func(r io.Reader, durableWriter, lossyWriter io.Writer, logger logging.Logger) {
+		defer wg.Done()
+
+		logbridge.Tee(r, durableWriter, lossyWriter, logger)
+		logging.DefaultLogger.NoFields().Infoln("logbridge Tee returned. Shutting down subordinate log command.")
+		durablePipe.Close()
 	}(os.Stdin, durablePipe, os.Stdout, logging.DefaultLogger)
 
-	loggerCmd.Start()
-	err = loggerCmd.Wait()
-	if err != nil {
-		logging.DefaultLogger.WithError(err)
-		os.Exit(1)
-	}
+	logging.DefaultLogger.NoFields().Info("logging running in background…")
+	wg.Wait()
 }
 
-// This is slightly naive, it could be more narrow in the signals it cares about
-func handleSignals(c io.Closer) {
-	signals := make(chan os.Signal, 2)
-	signal.Notify(signals)
-	<-signals
-	c.Close()
+// In environments where svlogd is used, the pipe that becomes STDIN of this
+// program can be non-blocking. Go's File implementation does not play well
+// with non-blocking pipes, in particular it does not recover from an EAGAIN
+// error from read(2).
+// This function defensively sets its 0th file descriptor to be blocking so
+// that we do not have to handle EAGAIN errors.
+func setSTDINToBlock() error {
+	oldflags, _, errno := unix.Syscall(unix.SYS_FCNTL, 0, unix.F_GETFL, 0)
+	if errno != 0 {
+		return fmt.Errorf("unix.FCNTL F_GETFL errno: %d", errno)
+	}
+	_, _, errno = unix.Syscall(unix.SYS_FCNTL, 0, unix.F_SETFL, oldflags&^unix.O_NONBLOCK)
+	if errno != 0 {
+		return fmt.Errorf("unix.FCNTL F_SETFL errno: %d", errno)
+	}
+	return nil
 }
--- a/pkg/pods/pod.go
+++ b/pkg/pods/pod.go
@@ -66,6 +66,7 @@ func NewPod(id string, path string) *Pod {
 		ServiceBuilder: runit.DefaultBuilder,
 		P2Exec:         DefaultP2Exec,
 		DefaultTimeout: 60 * time.Second,
+		LogExec:        defaultLogExec,
 	}
 }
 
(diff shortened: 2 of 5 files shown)


pkg/pods/pod.go
```diff
@@ -651,3 +651,20 @@ func (p *Pod) logExec() runit.LogExec {
 	}
 	return p.LogExec
 }
+
+// LogBridgeExec returns the full path to successfully execute LogBridge
+// This slice is eventually serialized into YAML, so we're careful about escaping
+func (p *Pod) LogBridgeExec() []string {
+	return []string{
+		"/usr/bin/chpst",
```
anthonybishopric: We should use `p2-exec` here

pkg/pods/pod.go
```diff
@@ -651,3 +651,20 @@ func (p *Pod) logExec() runit.LogExec {
...
+		"-u",
+		"nobody",
+		"-e",
+		p.EnvDir(),
+		"--",
+		filepath.Join(DEFAULT_PATH, "p2-preparer", "p2-log-bridge", "current", "bin", "p2-log-bridge"),
+		"bridge",
+		"-logExec",
+		"/usr/bin/svlogd -tt ./main ", // trailing space forces single-quotes by YAML serializer
```
anthonybishopric: Nit: extract `svlogd` location to a constant

pkg/pods/pod.go
```diff
@@ -651,3 +651,20 @@ func (p *Pod) logExec() runit.LogExec {
...
+// This slice is eventually serialized into YAML, so we're careful about escaping
+func (p *Pod) LogBridgeExec() []string {
+	return []string{
+		"/usr/bin/chpst",
+		"-u",
+		"nobody",
+		"-e",
+		p.EnvDir(),
+		"--",
+		filepath.Join(DEFAULT_PATH, "p2-preparer", "p2-log-bridge", "current", "bin", "p2-log-bridge"),
```
anthonybishopric: The pods package has generally been decoupled from the preparer (a specific implementation of a thing that launches pods). You can imagine that this would break clients who don't use a preparer, ie folks that just rely on `p2-launch` or some alternative.

Instead of hardcoding preparer values here, I'd suggest putting them in `pkg/preparer` and injecting the logexec command into a field in `pods.Pod`. Additionally, it'd be cleaner to read the pod home from the environment instead of assuming the default path - that makes it a lot easier to do per-pod testing of regular `svlogd` vs `p2-log-bridge` as well.

rudle:
@anthonybishopric I've addressed your comments, PTAL

anthonybishopric:
:ship:

bin/p2-log-bridge/main.go
```diff
@@ -20,31 +21,57 @@ func main() {
 	kingpin.Version(version.VERSION)
 	kingpin.Parse()
 
+	if err := setSTDINToBlock(); err != nil {
+		logging.DefaultLogger.WithError(err).Error("fatal error setting STDIN to block")
+		os.Exit(1)
+	}
+
+	var wg sync.WaitGroup
+	fmt.Printf("durableLogger: %v", *durableLogger)
```
bwester: Did you mean for this `Printf()` call to be here?

bin/p2-log-bridge/main.go
```diff
@@ -20,31 +21,57 @@ func main() {
...
 
-	loggerCmd.Start()
-	err = loggerCmd.Wait()
-	if err != nil {
-		logging.DefaultLogger.WithError(err)
-		os.Exit(1)
-	}
+		err := loggerCmd.Wait()
+		if err != nil {
+			logging.DefaultLogger.WithError(err)
```
bwester: I don't think this line has any effect.

pkg/preparer/orchestrate.go
```diff
@@ -218,6 +220,22 @@ func (p *Preparer) handlePods(podChan <-chan ManifestPair, quit <-chan struct{})
...
+		pod.P2Exec,
+		"-u",
+		"nobody",
+		"-e",
+		pod.EnvDir(),
+		"--",
+		filepath.Join(os.Getenv("POD_HOME"), "p2-log-bridge", "current", "bin", "p2-log-bridge"),
+		"bridge",
+		"-logExec",
+		svlogdExec + " ", // trailing space forces single-quotes by YAML serializer
```
bwester: Why do we need to play games with the YAML serializer?

bin/p2-log-bridge/main.go
```diff
@@ -20,31 +21,57 @@ func main() {
...
+			os.Exit(1)
+		}
+	}(*loggerCmd, wg)
+
+	logbridge.Tee(os.Stdin, durablePipe, os.Stdout, logging.DefaultLogger)
+	logging.DefaultLogger.NoFields().Info("logbridge Tee returned. Shutting down subordinate log command.\n")
+	durablePipe.Close()
+
+	logging.DefaultLogger.NoFields().Info("logging running in background…")
```
bwester: Typographical correctness aside, I think it's safer to avoid the whole issue of non-ASCII output and prefer `...` to `…`.

bin/p2-log-bridge/main.go
```diff
@@ -1,10 +1,11 @@
 package main
 
 import (
-	"io"
+	"fmt"
 	"os"
 	"os/exec"
-	"os/signal"
+	"sync"
+	"syscall"
```
bwester: Should new code use [`golang.org/x/sys/unix`](https://godoc.org/golang.org/x/sys/unix)? I don't think it changes much in the code below?

bin/p2-log-bridge/main.go
```diff
@@ -20,31 +21,57 @@ func main() {
...
 	durablePipe, err := loggerCmd.StdinPipe()
 	if err != nil {
-		logging.DefaultLogger.WithError(err).Errorln("failed during logbridge setup")
-		os.Exit(1)
+		logging.DefaultLogger.WithError(err).Error("fatal error during configuration of subordinate log command")
 	}
 
-	go handleSignals(durablePipe)
+	go func(loggerCmd exec.Cmd, wg sync.WaitGroup) {
+		wg.Add(1)
```
bwester: The `Add(1)` should execute before the goroutine starts.

pkg/preparer/orchestrate.go
rudle: Looks like we don't. I had assumed we needed to say `'svlogd -tt ./main'` where the single quotes were necessary. I've just tested this hypothesis and found the quotes to be unnecessary. Said another way the following is valid configuration for our implementation of servicebuilder.

```
…
  log:
  - /usr/local/bin/p2-exec
  - -u
  - nobody
  - -e
  - /data/pods/slug/env
  - --
  - /data/pods/p2-preparer/p2-log-bridge/current/bin/p2-log-bridge
  - bridge
  - -logExec
  - svlogd -tt ./main
```

bin/p2-log-bridge/main.go
```diff
@@ -20,31 +22,56 @@ func main() {
...
-		os.Exit(1)
-	}
+		err := loggerCmd.Wait()
+		if err != nil {
+			logging.DefaultLogger.WithError(err).Errorln("fatal error in subordinate log command")
+			os.Exit(1)
+		}
+	}(*loggerCmd, wg)
+
+	logbridge.Tee(os.Stdin, durablePipe, os.Stdout, logging.DefaultLogger)
```
bwester: Shouldn't stdout be the non-lossy output?

rudle: Well, the way it's written this command could be wired up via shell pipe to a potentially slow log forwarder. 

I'm thinking of this script as an example of how to use the logbridge package rather than a reference implementation. I did find it a bit difficult to emulate our lossy samsa writer without pulling in a lot of dependencies. 

If you feel strongly, I could add a second argument to allow a lossy cmd to be specified on the command line here. That would definitely illustrate how we use logbridge internally more clearly at the cost of a bit more development time.

bwester: The command has two outputs formats—one lossy, one not—and it has two output channels—one via a command, one through stdout—so perhaps it doesn't matter much how those two pairs match up.

bwester:
I know it's beyond the scope of this PR, but I just came across a problem: the lossy output channel doesn't print newlines correctly. An input of:

    hello
    world

will be printed as:

    helloworld

bin/p2-log-bridge/main.go
```diff
@@ -1,10 +1,12 @@
...
 
 import (
-	"io"
+	"fmt"
 	"os"
 	"os/exec"
-	"os/signal"
+	"sync"
+
+	"golang.org/x/sys/unix"
```
bwester: How about referencing the vendored version directly?

```
	"github.com/square/p2/Godeps/_workspace/src/golang.org/x/sys/unix"
```

bin/p2-log-bridge/main.go
```diff
@@ -20,31 +21,56 @@ func main() {
...
 	durablePipe, err := loggerCmd.StdinPipe()
 	if err != nil {
-		logging.DefaultLogger.WithError(err).Errorln("failed during logbridge setup")
-		os.Exit(1)
+		logging.DefaultLogger.WithError(err).Error("fatal error during configuration of subordinate log command")
 	}
 
-	go handleSignals(durablePipe)
+	wg.Add(1)
+	go func(loggerCmd exec.Cmd, wg sync.WaitGroup) {
```
bwester: This goroutine needs a reference to the same `sync.WaitGroup` value that's defined in line 29, not its own copy.

bwester:
What benefit is "wrapping tee in goroutine"?

rudle:
Backgrounding the Tee lets us print a message like "things have started correctly" to the logs. It also feels more natural to me to wrap the long-running call to logbridge.Tee() in a goroutine instead of doing it in the main thread of control.

FWIW, our implementation does the same thing.

bwester:
Merge your commits and :shipit:!

PR #366, opened 2016-01-22
logbridge: Preserve newlines and copy data

bwester:
The output delivered to the lossy logger is stripped of its newlines by the
default scanner. This change adds a new splitter that preserves newlines so
that the output stream of the lossy logger is the same as stdin.

---

After digging into the implementation of bufio.Scanner, it turns out that it
maintains one internal array as a large input buffer, and all the tokens it
returns are slices from that shared array. This works fine if the application
fully-handles each token as it's encountered, but logbridge asynchronously
sends them to the lossy writer. If the lossy writer is slow, there is the
potential for the array's data to have been overwritten by the Scanner by the
time a slow lossy writer gets to the slice. The data handled would not be the
same as the data sent.

This commit copies the contents of each line into a large append-only buffer.
When full, a new large buffer is allocated. If this code turns out to be a
throughput bottleneck, we could instead pursue writing a new Scanner that
doesn't need a copy.

### DIFF
--- a/pkg/logbridge/logbridge.go
+++ b/pkg/logbridge/logbridge.go
@@ -2,6 +2,7 @@ package logbridge
 
 import (
 	"bufio"
+	"bytes"
 	"io"
 	"time"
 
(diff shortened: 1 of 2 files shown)


rudle:
Cool fix! I'm wondering now if we really desire trailing new lines for all types of lossyWriters. In our case in which the contents of the lossyWriter are eventually viewed on a webpage the newline character would actually be detrimental to the user experience. That is, no one cares to see "application booted\n" vs "application booted" - the newline is implied instead of being expressed. Of course, the front-end code could trim the trailing newline before presenting it to the user.

I'm presuming you're of the opinion that log-bridge should preserve as much of the input data as possible and that the presentation of the log lines is a separate concern. I tend to agree with you, but I'd like to see how this change affects the code implementing our use case before I can offer full support.

bwester:
Yes, my opinion is that `logbridge.Tee()` should send the same bytes to both writers, though by design, the lossy writer will get its writes chunked into separate lines.  This requires fewer assumptions on the part of the consumers and it allows the consumers to be more generic.  If the consumer doesn't want to display newlines, it should strip the trailing newline itself. I don't expect that this will be a large burden.

bwester:
Updated to fix failing unit test

anthonybishopric:
:ship:

rudle:
:+1:

PR #367, opened 2016-01-25
If ReplicasDesired is decreased on an RC, prefer unscheduling ineligi…

mpuncel:
…ble nodes

Previously a replication controller would retrieve the list of eligible
nodes only when ReplicasDesired is increased. The list should also be
checked when ReplicasDesired is increased, and nodes that are no longer
eligible should be preferred for unscheduling.

Future work:
Converge to eligible nodes more often (not only when ReplicasDesired
changes, but also when any current nodes become ineligible).

### DIFF
--- a/pkg/rc/replication_controller_test.go
+++ b/pkg/rc/replication_controller_test.go
@@ -1,6 +1,7 @@
 package rc
 
 import (
+	"fmt"
 	"testing"
 	"time"
 
@@ -126,6 +127,7 @@ func TestSchedule(t *testing.T) {
 	Assert(t).IsNil(err, "expected no error labeling node2")
 
 	quit := make(chan struct{})
+	defer close(quit)
 	rc.WatchDesires(quit)
 
 	rcStore.SetDesiredReplicas(rc.ID(), 1)
@@ -183,6 +185,7 @@ func TestScheduleTwice(t *testing.T) {
 	Assert(t).IsNil(err, "expected no error labeling node2")
 
 	quit := make(chan struct{})
+	defer close(quit)
 	rc.WatchDesires(quit)
 
 	rcStore.SetDesiredReplicas(rc.ID(), 1)
@@ -219,6 +222,7 @@ func TestUnschedule(t *testing.T) {
 	Assert(t).IsNil(err, "expected no error labeling node2")
 
 	quit := make(chan struct{})
+	defer close(quit)
 	rc.WatchDesires(quit)
 
 	rcStore.SetDesiredReplicas(rc.ID(), 1)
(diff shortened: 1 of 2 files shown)


pkg/rc/replication_controller_test.go
```diff
@@ -237,3 +237,39 @@ func TestUnschedule(t *testing.T) {
...
+
+	rcStore.SetDesiredReplicas(rc.ID(), 3)
+	numNodes := waitForNodes(t, rc, 3)
+	Assert(t).AreEqual(numNodes, 3, "took too long to schedule")
+
+	scheduled := scheduledPods(t, applicator)
+	Assert(t).AreEqual(len(scheduled), 3, "expected a pod to have been labeled")
+	Assert(t).AreEqual(len(kp.manifests), 3, "expected a manifest to have been scheduled")
+
+	// Make node2 ineligible, so that it will be preferred for unscheduling
```
petertseng: I'm wondering a bit - if `node3` would have been the preferred node in either case, then this test will be inconclusive. So basically, depending on the order of `PopAny`, this test might falsely declare bad code good.

But I don't want to suggest that we have one test with `node2` ineligible and one test with `node3` ineligible. So I'm not sure we can solve that problem. So maybe we just have to live with it? Tests can't prove the absence of bugs, only their presence, or some expression to that effect?

mpuncel: yeah I think it's inevitable that bad code might get lucky. If it makes you feel better I marked node3 as bad once and my test failed

bwester: If you want a tighter bound on the probability of accidentally picking the right node, just add more nodes.  I'd feel comfortable with 1000.

pkg/rc/replication_controller_test.go
```diff
@@ -237,3 +237,39 @@ func TestUnschedule(t *testing.T) {
...
+func TestPreferUnscheduleIneligible(t *testing.T) {
+	rcStore, kp, applicator, rc := setup(t)
+	err := applicator.SetLabel(labels.NODE, "node1", "nodeQuality", "good")
+	Assert(t).IsNil(err, "expected no error labeling node1")
+	err = applicator.SetLabel(labels.NODE, "node2", "nodeQuality", "good")
+	Assert(t).IsNil(err, "expected no error labeling node2")
+	err = applicator.SetLabel(labels.NODE, "node3", "nodeQuality", "good")
+	Assert(t).IsNil(err, "expected no error labeling node3")
+
+	quit := make(chan struct{})
```
bwester: Close `quit` somewhere?

bwester:
Except for my two earlier comments, the rest of this looks good!

bwester:
:+1:

One final nit: fix your commit message so that GitHub doesn't truncate the first line. The limit is 70 chars.

mpuncel:
rebased

PR #368, opened 2016-01-27
Always list keys with an extra "/"

bwester:
Changes all callers of api.KV.List(), consulutil.SafeList(), or
consulutil.WatchPrefix() to append an extra "/" to the end of the watched
string.  Also changes kp.Store.ListPods() and kp.Store.WatchPods() to accept a
"path" to a Consul subtree, not an arbitrary "prefix." All existing callers of
of these two methods already pass subtree paths, e.g., `kp.*_TREE` or
`kp.*Path(hostname)`, so this shouldn't change any existing behavior.

These changes are an extra safeguard to make sure that a List operation returns
the right results in the presence of path components that are subsets of each
other. For instance, if two nodes were named "node1" and "node10", listing all
of node1's pods (e.g., List("intent/node1")) would previously have returned
node10's pods too (e.g., "intent/node10/p2-preparer").

### DIFF
--- a/pkg/kp/kv.go
+++ b/pkg/kp/kv.go
@@ -32,9 +32,9 @@ type Store interface {
 	GetHealth(service, node string) (WatchResult, error)
 	GetServiceHealth(service string) (map[string]WatchResult, error)
 	WatchPod(key string, quitChan <-chan struct{}, errChan chan<- error, podChan chan<- ManifestResult)
-	WatchPods(keyPrefix string, quitChan <-chan struct{}, errChan chan<- error, podChan chan<- []ManifestResult)
+	WatchPods(path string, quitChan <-chan struct{}, errChan chan<- error, podChan chan<- []ManifestResult)
 	Ping() error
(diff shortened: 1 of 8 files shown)


petertseng:
In case anyone wonders, what happens when there are too many trailing slash (Such as `reality//`) is that you get a 301 to direct to the version with only one trailing slash. So as long as our things follow the 301 we should be OK even if we have too many slashes. So, better to have too many slashes than too few.

bwester:
@petertseng, that's pretty interesting. I hadn't tried that before.  It seems you also get this redirect behavior by trying to GET or PUT a key named `a//b//c`. It will redirect you to `a/b/c`.  I'm surprised that Consul will reject/rewrite your keys.

BTW, I wouldn't still expect a double-"/" to be submitted in the first place.

petertseng:
Looks OK to me, looked for all call sites to `consulutil.WatchPrefix` and `kv.List`. I don't suppose `consulutil.WatchSingle` needs any changes given its name...

PR #369, opened 2016-01-28
create an ID type in pods package for pod IDs

mpuncel:
This will increase type safety when passing around pod IDs from package
to package. For instance, the kp package can provide some safety around
constructing keys by differentiating the pod id type from the hostname
type.

### DIFF
--- a/bin/p2-bootstrap/bootstrap.go
+++ b/bin/p2-bootstrap/bootstrap.go
@@ -8,6 +8,7 @@ import (
 	"github.com/square/p2/Godeps/_workspace/src/gopkg.in/alecthomas/kingpin.v2"
 	"github.com/square/p2/pkg/kp"
 	"github.com/square/p2/pkg/pods"
+	"github.com/square/p2/pkg/types"
 	"github.com/square/p2/pkg/util"
 	"github.com/square/p2/pkg/version"
 )
@@ -128,7 +129,7 @@ func VerifyConsulUp(timeout string) error {
 	}
 }
 
-func VerifyReality(waitTime time.Duration, consulID, agentID string) error {
+func VerifyReality(waitTime time.Duration, consulID types.PodID, agentID types.PodID) error {
 	quit := make(chan struct{})
 	defer close(quit)
 	store := kp.NewConsulStore(kp.NewConsulClient(kp.Options{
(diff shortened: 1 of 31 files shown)


mpuncel:
I intend to follow this up with a PR that makes various `pkg/kp` operations use this type, e.g. `IntentPath` and `SetPod`. The PR was huge though so I decided to break it up

bin/p2-bootstrap/bootstrap.go
```diff
@@ -39,7 +40,7 @@ func main() {
 		if err != nil {
 			log.Fatalf("Could not get consul manifest: %s", err)
 		}
-		consulPod = pods.NewPod(consulManifest.ID(), pods.PodPath(*podRoot, consulManifest.ID()))
+		consulPod = pods.NewPod(string(consulManifest.ID()), pods.PodPath(*podRoot, consulManifest.ID()))
```
anthonybishopric: Make `NewPod` take a `types.PodID`?

bin/p2-bootstrap/bootstrap.go
```diff
@@ -128,7 +129,7 @@ func VerifyConsulUp(timeout string) error {
 	}
 }
 
-func VerifyReality(waitTime time.Duration, consulID, agentID string) error {
+func VerifyReality(waitTime time.Duration, consulID types.PodID, agentID types.PodID) error {
```
petertseng: can say `types.PodID` just once, unless this was deliberate?

mpuncel: I personally don't really like reusing the type, happy to put it back if people disagree

bin/p2-bootstrap/bootstrap.go
mpuncel: The call sites for NewPod are pretty evenly split between callers wanting to pass a string vs an id pulled from some other manifest. I felt like it's philosophically better to make NewPod do the cast for you, but if you feel strongly I can change it

anthonybishopric: That makes sense, however I'd argue that it's weird to cast back to `string` if you do have a manifest object though. I'd still be in favor of making `NewPod` take a `types.PodID`

mpuncel: okay i'll make the change

bin/p2-bootstrap/bootstrap.go
petertseng: OK. I don't care strongly either way

bwester:
:ship:

anthonybishopric:
:+1: pending comments

PR #370, opened 2016-01-29
Add type safety when computing consul key paths

mpuncel:
Previously the kp package provided functions (e.g.
kp.IntentPath(string...)) that would simply join the passed strings
with a "/" in between and return a consul path that would be used. This
means all callers need to know the structure of consul keys and do
sanity checking against bad things happening.

These functions are replaced with functions that are named more
explicitly to indicate what the caller wants.

### DIFF
--- a/bin/p2-schedule/main.go
+++ b/bin/p2-schedule/main.go
@@ -41,11 +41,11 @@ func main() {
 		if err != nil {
 			log.Fatalf("Could not read manifest at %s: %s\n", manifestPath, err)
 		}
-		path := kp.IntentPath(*nodeName, string(manifest.ID()))
+		podPrefix := kp.INTENT_TREE
 		if *hookGlobal {
-			path = kp.HookPath(string(manifest.ID()))
+			podPrefix = kp.HOOK_TREE
 		}
-		duration, err := store.SetPod(path, manifest)
+		duration, err := store.SetPod(podPrefix, *nodeName, manifest)
 		if err != nil {
 			log.Fatalf("Could not write manifest %s to intent store: %s\n", manifest.ID(), err)
 		}
--- a/pkg/kp/constants.go
+++ b/pkg/kp/constants.go
@@ -1,38 +1,59 @@
 package kp
 
 import (
-	"strings"
+	"path"
+
+	"github.com/square/p2/pkg/types"
+	"github.com/square/p2/pkg/util"
 )
+
+// PodPrefix represents the top level of a subtree whose leaves are expected to
+// be pods, e.g. "hooks", "intent", "reality"
+type PodPrefix string
 
 const (
-	INTENT_TREE  string = "intent"
-	REALITY_TREE string = "reality"
-	HOOK_TREE    string = "hooks"
-	LOCK_TREE    string = "lock"
-	RC_TREE      string = "replication_controllers"
-	ROLL_TREE    string = "rolls"
+	INTENT_TREE  PodPrefix = "intent"
+	REALITY_TREE PodPrefix = "reality"
+	HOOK_TREE    PodPrefix = "hooks"
+	LOCK_TREE    string    = "lock"
 )
 
-func IntentPath(args ...string) string {
-	return strings.Join(append([]string{INTENT_TREE}, args...), "/")
+func nodePath(podPrefix PodPrefix, nodeName string) (string, error) {
+	// hook tree is an exception to the rule because they are not scheduled
+	// by host, and it is valid to want to watch for them agnostic to pod
+	// id. There are plans to deploy hooks by host at which time this
+	// exception can be removed
+	if podPrefix == HOOK_TREE {
+		nodeName = ""
+	} else {
+		if nodeName == "" {
+			return "", util.Errorf("nodeName not specified when computing host path")
+		}
+	}
+
+	return path.Join(string(podPrefix), nodeName), nil
 }
 
-func RealityPath(args ...string) string {
-	return strings.Join(append([]string{REALITY_TREE}, args...), "/")
+func podPath(podPrefix PodPrefix, nodeName string, podId types.PodID) (string, error) {
+	nodePath, err := nodePath(podPrefix, nodeName)
+	if err != nil {
+		return "", err
+	}
+
+	if podId == "" {
+		return "", util.Errorf("pod id not specified when computing pod path")
+	}
+
+	return path.Join(nodePath, string(podId)), nil
 }
 
-func HookPath(args ...string) string {
-	return strings.Join(append([]string{HOOK_TREE}, args...), "/")
-}
+// Returns the consul path to use when intending to lock a pod, e.g.
+// lock/intent/some_host/some_pod
+func PodLockPath(podPrefix PodPrefix, nodeName string, podId types.PodID) (string, error) {
+	subPodPath, err := podPath(podPrefix, nodeName, podId)
+	if err != nil {
+		return "", err
+	}
 
-func LockPath(args ...string) string {
-	return strings.Join(append([]string{LOCK_TREE}, args...), "/")
-}
-
-func RCPath(args ...string) string {
-	return strings.Join(append([]string{RC_TREE}, args...), "/")
-}
-
-func RollPath(args ...string) string {
-	return strings.Join(append([]string{ROLL_TREE}, args...), "/")
+	return path.Join(LOCK_TREE, subPodPath), nil
 }
--- a/pkg/kp/kv.go
+++ b/pkg/kp/kv.go
@@ -26,16 +26,17 @@ type ManifestResult struct {
 }
 
 type Store interface {
-	SetPod(key string, manifest pods.Manifest) (time.Duration, error)
-	Pod(key string) (pods.Manifest, time.Duration, error)
-	DeletePod(key string) (time.Duration, error)
+	SetPod(podPrefix PodPrefix, nodename string, manifest pods.Manifest) (time.Duration, error)
+	Pod(podPrefix PodPrefix, nodename string, podId types.PodID) (pods.Manifest, time.Duration, error)
+	DeletePod(podPrefix PodPrefix, nodename string, podId types.PodID) (time.Duration, error)
 	PutHealth(res WatchResult) (time.Time, time.Duration, error)
 	GetHealth(service, node string) (WatchResult, error)
 	GetServiceHealth(service string) (map[string]WatchResult, error)
-	WatchPod(key string, quitChan <-chan struct{}, errChan chan<- error, podChan chan<- ManifestResult)
-	WatchPods(path string, quitChan <-chan struct{}, errChan chan<- error, podChan chan<- []ManifestResult)
+	WatchPod(podPrefix PodPrefix, nodename string, podId types.PodID, quitChan <-chan struct{}, errChan chan<- error, podChan chan<- ManifestResult)
+	WatchPods(podPrefix PodPrefix, nodename string, quitChan <-chan struct{}, errChan chan<- error, podChan chan<- []ManifestResult)
 	Ping() error
-	ListPods(path string) ([]ManifestResult, time.Duration, error)
+	ListPods(podPrefix PodPrefix, nodename string) ([]ManifestResult, time.Duration, error)
+	AllPods(podPrefix PodPrefix) ([]ManifestResult, time.Duration, error)
 	LockHolder(key string) (string, string, error)
 	DestroyLockHolder(id string) error
 	NewLock(name string, renewalCh <-chan time.Time) (Lock, chan error, error)
@@ -171,11 +172,15 @@ func (c consulStore) GetServiceHealth(service string) (map[string]WatchResult, e
 	return healthRes, nil
 }
 
-// SetPod writes a pod manifest into the consul key-value store. The key should
-// not have a leading or trailing slash.
-func (c consulStore) SetPod(key string, manifest pods.Manifest) (time.Duration, error) {
+// SetPod writes a pod manifest into the consul key-value store.
+func (c consulStore) SetPod(podPrefix PodPrefix, nodename string, manifest pods.Manifest) (time.Duration, error) {
 	buf := bytes.Buffer{}
 	err := manifest.Write(&buf)
+	if err != nil {
+		return 0, err
+	}
+
+	key, err := podPath(podPrefix, nodename, manifest.ID())
 	if err != nil {
 		return 0, err
 	}
@@ -197,7 +202,12 @@ func (c consulStore) SetPod(key string, manifest pods.Manifest) (time.Duration, 
 
 // DeletePod deletes a pod manifest from the key-value store. No error will be
 // returned if the key didn't exist.
-func (c consulStore) DeletePod(key string) (time.Duration, error) {
+func (c consulStore) DeletePod(podPrefix PodPrefix, nodename string, podId types.PodID) (time.Duration, error) {
+	key, err := podPath(podPrefix, nodename, podId)
+	if err != nil {
+		return 0, err
+	}
+
 	writeMeta, err := c.client.KV().Delete(key, nil)
 	if err != nil {
 		return 0, consulutil.NewKVError("delete", key, err)
@@ -208,7 +218,12 @@ func (c consulStore) DeletePod(key string) (time.Duration, error) {
 // Pod reads a pod manifest from the key-value store. If the given key does not
 // exist, a nil *PodManifest will be returned, along with a pods.NoCurrentManifest
 // error.
-func (c consulStore) Pod(key string) (pods.Manifest, time.Duration, error) {
+func (c consulStore) Pod(podPrefix PodPrefix, nodename string, podId types.PodID) (pods.Manifest, time.Duration, error) {
+	key, err := podPath(podPrefix, nodename, podId)
+	if err != nil {
+		return nil, 0, err
+	}
+
 	kvPair, writeMeta, err := c.client.KV().Get(key, nil)
 	if err != nil {
 		return nil, 0, consulutil.NewKVError("get", key, err)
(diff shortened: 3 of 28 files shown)


anthonybishopric:
Thanks for doing this mechanical, but very important change. :ship:

bin/p2-schedule/main.go
```diff
@@ -41,11 +41,11 @@ func main() {
...
 			log.Fatalf("Could not read manifest at %s: %s\n", manifestPath, err)
 		}
-		path := kp.IntentPath(*nodeName, string(manifest.ID()))
+		podPrefix := kp.INTENT_TREE
 		if *hookGlobal {
-			path = kp.HookPath(string(manifest.ID()))
+			podPrefix = kp.HOOK_TREE
 		}
-		duration, err := store.SetPod(path, manifest)
+		duration, err := store.SetPod(podPrefix, *nodeName, manifest)
```
bwester: It looks like a hook pod named `H` will now be stored  under `hooks/mylaptop/H` instead of `hooks/H`. What is the impact of this change?

bwester: Ah, now I get it.

mpuncel: it actually won't because hostname is ignored in the `nodePath` function in `pkg/kp/constants.go`

mpuncel: see `TestHookPathIsExceptionToNodeNameRule`

pkg/kp/constants.go
```diff
@@ -1,38 +1,147 @@
...
+	INTENT_TREE  PodPrefix = "intent"
+	REALITY_TREE PodPrefix = "reality"
+	HOOK_TREE    PodPrefix = "hooks"
+	LOCK_TREE    TreeType  = "lock"
+	RC_TREE      TreeType  = "replication_controllers"
+	ROLL_TREE    TreeType  = "rolls"
+)
+
+var (
+	VALID_POD_PREFIXES = []PodPrefix{
```
bwester: It seems awkward to have to list each prefix three times to establish its full type. I suspect you could make this a lot cleaner if you expressed the concepts of "this is a tree prefix" and "this is a tree prefix that contains pods" using interfaces.

bwester: Actually, I'm not sure why you need `TreeType`. It's unused.  And from a type theoretic perspective, its values are not interchangeable: there is no reason to call any of them the same "type" as any other.

In fact, you could probably make `LOCK_TREE`, `RC_TREE`, and `ROLL_TREE` private constants, though you'd have to move the latter two out of the `kp` package (which is probably where they belong anyway).

mpuncel: good suggestions :+1:

pkg/kp/constants.go
```diff
@@ -1,38 +1,84 @@
...
+		HOOK_TREE,
+	}
+)
 
-func RealityPath(args ...string) string {
-	return strings.Join(append([]string{REALITY_TREE}, args...), "/")
-}
+func nodePath(podPrefix PodPrefix, nodeName string) (string, error) {
+	validPrefix := false
+	for _, prefix := range VALID_POD_PREFIXES {
```
bwester: You don't need `VALID_POD_PREFIXES` anymore. The type system should catch the mistakes.

pkg/kp/constants.go
```diff
@@ -1,38 +1,84 @@
...
+	// exception can be removed
+	if podPrefix == HOOK_TREE {
+		nodeName = ""
+	} else {
+		if nodeName == "" {
+			return "", util.Errorf("nodeName not specified when computing host path")
+		}
+	}
+
+	if nodeName == "" {
```
bwester: ?

pkg/kp/constants.go
```diff
@@ -1,38 +1,84 @@
...
+		return "", util.Errorf("pod id not specified when computing pod path")
+	}
+
+	return path.Join(nodePath, string(podId)), nil
 }
 
-func RollPath(args ...string) string {
-	return strings.Join(append([]string{ROLL_TREE}, args...), "/")
+// Returns the consul path to use when intending to lock a pod, e.g.
+// /lock/intent/some_host/some_pod
```
bwester: There isn't a `/` at the front of the keys

pkg/kp/constants.go
```diff
@@ -1,38 +1,84 @@
...
+func PodLockPath(podPrefix PodPrefix, nodeName string, podId types.PodID) (string, error) {
+	subPodPath, err := podPath(podPrefix, nodeName, podId)
+	if err != nil {
+		return "", err
+	}
+
+	return path.Join(string(LOCK_TREE), subPodPath), nil
```
bwester: `LOCK_TREE` doesn't need the cast now

pkg/kp/constants.go
mpuncel: I'm pretty sure a caller can pass `PodPrefix("foobar")` otherwise? is there a way to prevent that?

mpuncel: for example kubernetes checks for invalid `Operator` values in the labels package: https://github.com/kubernetes/kubernetes/blob/master/pkg/labels/selector.go#L129

bwester: It depends what we want out of your modification. If we want to make it reasonably harder to operate on the wrong tree, then it's probably fine to create a few constants with well-defined values and then don't create any more. As a result, these constants can be implemented in a straightforward way. If you want to make it _impossible_ to use anything outside a couple constants, you can, but it complicates your design.

Personally, I'd shoot for the "reasonably hard to screw it up," and my comment was made to that end.

And If you really want to make it impossible to pick the wrong tree, you shouldn't make `VALID_POD_PREFIXES` public. I would also expect there to be a way to get the type system to prevent the construction of non-zero values of a public type.

mpuncel: fair enough, removing it

pkg/kp/rcstore/consul_store.go
```diff
@@ -301,7 +312,17 @@ func (s *consulStore) Watch(rc *fields.RC, quit <-chan struct{}) (<-chan struct{
...
 
-	go consulutil.WatchSingle(kp.RCPath(rc.ID.String()), s.kv, input, quit, errors)
+	go func() {
+		rcp, err := RCPath(rc.ID)
+		if err != nil {
+			select {
+			case <-quit:
+			case errors <- err:
+			}
+			return
```
bwester: `WatchSingle()` would run until `quit` is closed, then it would close `input` to stop the other goroutine. If you want your error handling to be done inside this goroutine, it should provide the same semantics. Alternatively, you could do the error handling before spawning goroutines (that you know will never produce anything) and provide the expected behavior to the caller of `Watch()` directly.

pkg/kp/rcstore/store.go
```diff
@@ -1,12 +1,18 @@
...
+
 	"github.com/square/p2/Godeps/_workspace/src/k8s.io/kubernetes/pkg/labels"
 
+	"github.com/square/p2/pkg/kp"
 	"github.com/square/p2/pkg/pods"
 	"github.com/square/p2/pkg/rc/fields"
+	"github.com/square/p2/pkg/util"
 )
 
+const rc_tree string = "replication_controllers"
```
bwester: nit: `rcTree`

pkg/kp/rcstore/store.go
```diff
@@ -46,3 +52,33 @@ type Store interface {
 	// The two output channels will be closed in response.
 	Watch(rc *fields.RC, quit <-chan struct{}) (<-chan struct{}, <-chan error)
 }
+
+func RCPath(rcId fields.ID) (string, error) {
+	if rcId == "" {
+		return "", util.Errorf("rcId not specified when computing rc path")
+	}
+	return path.Join(string(rc_tree), string(rcId)), nil
```
bwester: It's already a `string`

pkg/kp/rcstore/store.go
```diff
@@ -46,3 +52,33 @@ type Store interface {
 	// The two output channels will be closed in response.
 	Watch(rc *fields.RC, quit <-chan struct{}) (<-chan struct{}, <-chan error)
 }
+
+func RCPath(rcId fields.ID) (string, error) {
```
bwester: Unless I missed something, this doesn't need to be public.  And it runs against your goals to have a public function that returns a string path.

pkg/kp/rcstore/store.go
```diff
@@ -46,3 +52,33 @@ type Store interface {
...
+	if rcId == "" {
+		return "", util.Errorf("rcId not specified when computing rc path")
+	}
+	return path.Join(string(rc_tree), string(rcId)), nil
+}
+
+func RCLockPath(rcId fields.ID) (string, error) {
```
bwester: Same comment as for `RCPath()`.

pkg/kp/rcstore/store.go
```diff
@@ -46,3 +52,33 @@ type Store interface {
...
+
+	return path.Join(kp.LOCK_TREE, subRCPath), nil
+}
+
+// Replication controllers have potential for two locks:
+// 1) the key itself, held by a replication controller goroutine that is
+// responsible for making sure its desires are met
+// 2) an "update" key under the replication controller key, held by the
+// goroutine running a rolling update that will mutate a replication controller
+func RCUpdateLockPath(rcId fields.ID) (string, error) {
```
bwester: Ditto.

pkg/kp/rollstore/store.go
```diff
@@ -10,8 +11,11 @@ import (
 	"github.com/square/p2/pkg/kp/consulutil"
 	rcf "github.com/square/p2/pkg/rc/fields"
 	rollf "github.com/square/p2/pkg/roll/fields"
+	"github.com/square/p2/pkg/util"
 )
 
+const roll_tree string = "rolls"
```
bwester: nit: `rollTree`

pkg/kp/rcstore/store.go
bwester: You know, don't bother with this. What you've got now is safer than what came before.

pkg/kp/constants.go
```diff
@@ -1,38 +1,59 @@
...
+// lock/intent/some_host/some_pod
+func PodLockPath(podPrefix PodPrefix, nodeName string, podId types.PodID) (string, error) {
+	subPodPath, err := podPath(podPrefix, nodeName, podId)
+	if err != nil {
+		return "", err
+	}
 
-func RollPath(args ...string) string {
-	return strings.Join(append([]string{ROLL_TREE}, args...), "/")
+	return path.Join(string(LOCK_TREE), subPodPath), nil
```
petertseng: not strictly necessary since `LOCK_TREE` is a string already

mpuncel: i've been doing fixes in the second commit, i'll squash

mpuncel: oh second commit has the problem

pkg/kp/rollstore/store.go
```diff
@@ -108,7 +128,7 @@ func (s consulStore) Watch(quit <-chan struct{}) (<-chan []rollf.Update, <-chan
 	errCh := make(chan error)
 	inCh := make(chan api.KVPairs)
 
-	go consulutil.WatchPrefix(kp.ROLL_TREE+"/", s.kv, inCh, quit, errCh)
+	go consulutil.WatchPrefix(string(roll_tree)+"/", s.kv, inCh, quit, errCh)
```
petertseng: `roll_tree` is already a string

pkg/kp/rcstore/consul_store.go
mpuncel: @bwester what do you think about making this function return an `error` in addition to the other two return values. I know that's a really weird signature, but really the inability to compute this path is unrecoverable and the whole operation should be abandoned.

that would propagate to pkg/rc/farm.glo calling `WatchDesires` on a child. If an error is returned it would release its lock and allow another farm to pick up the replication controller.

bwester: Could you get the watcher to move on without requiring another error handler?  Could the same thing be accomplished by returning a closed channel for `updated`?

mpuncel: that looks like it may be awkward as well: see https://github.com/square/p2/blob/master/pkg/rc/farm.go#L150

If the err channel is closed, the goroutine there exits and no mechanism removes the replication controller as a child of the farm. Currently only deleting the replication controller from consul will accomplish that.

I could add another `rcf.releaseChild()` call at the end of the goroutine, but that can be reached for two reasons:

1) childQuit is closed. If that's the case, we don't want to call releaseChild again (or we have to make it idempotent which would entail a mutex on `rcf.children`)
2) err channel closes. We definitely want to release the child here.

what do you think is the best option here?

bwester: I suppose I'm not concerned with releasing the RC, because if this agent has trouble handling the RC, I have no faith that another host will do any better.  Bubbling the error up to the main farm loop could conceivably allow that loop to be re-written so that the RC could be ignored, but doing so sounds like more work than should go in this PR.

I think you should ignore this for now. The simplest way forward is to accept that invalid controllers will not be handled—you just don't want them to impact availability for the correct controllers.  I think you can get that easily by returning a closed `updated` and `errors`.

mpuncel: sounds good :+1: i'll leave a comment

pkg/kp/rcstore/store.go
mpuncel: this is used by `pkg/roll`

pkg/kp/kv.go
```diff
@@ -245,9 +275,26 @@ func (c consulStore) ListPods(path string) ([]ManifestResult, time.Duration, err
...
 	defer close(podChan)
 
+	key, err := podPath(podPrefix, nodename, podId)
+	if err != nil {
+		select {
+		case <-quitChan:
+		case errChan <- err:
+		}
+		return
+	}
```
bwester: :+1:

pkg/kp/kv.go
```diff
@@ -273,18 +320,32 @@ func (c consulStore) WatchPod(key string, quitChan <-chan struct{}, errChan chan
...
 	defer close(podChan)
 
+	keyPrefix, err := nodePath(podPrefix, nodename)
+	if err != nil {
+		select {
+		case <-quitChan:
+		case errChan <- err:
+		}
+		return
+	}
```
bwester: :+1:

pkg/kp/rcstore/consul_store.go
```diff
@@ -301,34 +312,48 @@ func (s *consulStore) Watch(rc *fields.RC, quit <-chan struct{}) (<-chan struct{
...
-			newRC, err := s.kvpToRC(kvp)
-			if err != nil {
-				select {
-				case errors <- err:
-				case <-quit:
+			close(errors)
+			close(updated)
+		}()
+	} else {
+		go func() {
```
bwester: nit: Simplify `go func() { F() }()` to `go F()`.

mpuncel: :/

pkg/kp/rcstore/consul_store.go
```diff
@@ -301,34 +312,48 @@ func (s *consulStore) Watch(rc *fields.RC, quit <-chan struct{}) (<-chan struct{
...
-
-		for kvp := range input {
-			if kvp == nil {
-				// seems this RC got deleted from under us. quitting
-				// would be unexpected, so we'll just wait for it to
-				// reappear in consul
-				continue
+	rcp, err := rcPath(rc.ID)
+	if err != nil {
+		go func() {
```
bwester: There's an interesting pattern you see when working with futures that applies here too.  Even though the method is returning a placeholder that would otherwise be filled asynchronously, if the method already has the value, just return a placeholder that already contains the value. As it applies here:

    errors := make(chan error, 1)
    errors <- err
    close(errors)
    return updated, errors

mpuncel: :+1:

pkg/rc/replication_controller_test.go
```diff
@@ -18,12 +20,14 @@ type fakeKpStore struct {
 	manifests map[string]pods.Manifest
 }
 
-func (s *fakeKpStore) SetPod(key string, manifest pods.Manifest) (time.Duration, error) {
+func (s *fakeKpStore) SetPod(podPrefix kp.PodPrefix, nodeName string, manifest pods.Manifest) (time.Duration, error) {
+	key := fmt.Sprintf("%s/%s/%s", podPrefix, nodeName, manifest.ID())
```
bwester: nit: `path.Join()`

pkg/roll/run_update.go
```diff
@@ -251,17 +250,23 @@ func (u update) lockRCs(done <-chan struct{}) error {
...
 	wg := sync.WaitGroup{}
-	for _, path := range []string{u.lockPath(u.NewRC), u.lockPath(u.OldRC)} {
+	for _, rcID := range []rcf.ID{u.NewRC, u.OldRC} {
 		wg.Add(1)
-		go func(path string) {
+		go func(rcID rcf.ID) {
 			RetryOrQuit(
-				func() error { return u.lock.Unlock(path) },
+				func() error {
+					path, err := rcstore.RCUpdateLockPath(rcID)
```
bwester: You're going to have to make the path outside the `RetryOrQuit()` loop, which keeps running the function until it returns a `nil` error.  As-is, an invalid path results in an infinite loop.

pkg/kp/rcstore/consul_store.go
```diff
@@ -298,10 +309,18 @@ func (s *consulStore) mutateRc(id fields.ID, mutator func(fields.RC) (fields.RC,
 
 func (s *consulStore) Watch(rc *fields.RC, quit <-chan struct{}) (<-chan struct{}, <-chan error) {
 	updated := make(chan struct{})
+
+	rcp, err := rcPath(rc.ID)
+	if err != nil {
+		errors := make(chan error, 1)
+		errors <- err
+		close(errors)
+		return updated, errors
```
bwester: Still need to close `updated` too!

bwester:
:shipit:

mpuncel:
thanks for the thorough review @bwester !!

PR #371, opened 2016-02-02
Check for 0 byte manifest in pkg/rc/fields.UnmarshalJSON

mpuncel:
This allows unmarshaling of structs that contain an empty fields.RC.
Previously an error would be hit due to manifest validation in
pods.Manifest.ManifestFromBytes()

### DIFF
--- a/pkg/kp/rcstore/consul_store.go
+++ b/pkg/kp/rcstore/consul_store.go
@@ -13,6 +13,7 @@ import (
 	"github.com/square/p2/pkg/labels"
 	"github.com/square/p2/pkg/pods"
 	"github.com/square/p2/pkg/rc/fields"
+	"github.com/square/p2/pkg/util"
 )
 
 const (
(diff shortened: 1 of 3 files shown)


bwester:
This code seems good.

Have you looked at the callers of `UnmarshalJSON()` to check whether they are capable of dealing with a "valid" `*RC` that has a null `Manifest` field?

mpuncel:
sadly they'll be hard to find because it'll be an unmarshal call into a struct that has a `fields.RC` field, i'll do my best

pkg/kp/rcstore/consul_store.go
```diff
@@ -168,6 +169,10 @@ func (s *consulStore) WatchNew(quit <-chan struct{}) (<-chan []fields.RC, <-chan
 func (s *consulStore) kvpToRC(kvp *api.KVPair) (fields.RC, error) {
 	rc := fields.RC{}
 	err := json.Unmarshal(kvp.Value, &rc)
+
+	if rc.Manifest == nil {
```
mpuncel: i'm not sure if this is the right fix actually, it can cause a `List()` operation to fail if one rc has a nil manifest. Probably better to address line 343, but i'll have to look at the callers of that

bwester:
There are references at the following locations that assume `Manifest` is not nil:
* `bin/p2-rctl/main.go:252`
* `pkg/rc/farm.go:110`
* `pkg/roll/farm.go:138`
* `pkg/roll/run_update.go:120,309`

And the `rc.replicationController` type generally assumes that its `Manifest` field isn't nil, so its constructors should assume the same:
* `pkg/roll/run_update.go:301`
* `pkg/rc/farm.go:137`

bwester:
What is the motivation for this PR?  Why do we want to be able to deserialize RCs that don't have a manifest?  What creates RCs without a manifest?

PR #372, opened 2016-02-03
Change comment about DesiredReplicas on fields.RC

mpuncel:
It stated that DesiredReplicas should be at least equal to the value for
MinimumReplicas. However, in a use case where only part of a cluster
should be updated, it may be desired to deploy only to a subset while
maintaining a level of health greater than the number of nodes being
updated.


anthonybishopric:
:+1:

PR #373, opened 2016-02-03
gzip: Create parents

petertseng:
In our tar extraction code, if we were to encounter a tarball with a
file `a/b` but not the directory `a/`, we would attenpt to create `a/b`
without `a/` existing and fail in doing so.

This commit always creates the parent of the item we are about to
extract so that we avoid such problems.


petertseng:
I would like to add a test for this first. Please standby.

It is known that all tests that implicitly test this gzip code do work correctly, but it's probably also wise to test this new functionality.

pkg/gzip/gzip.go
```diff
@@ -55,6 +55,16 @@ func ExtractTarGz(owner string, fp io.Reader, dest string) (err error) {
 			uid, gid = ownerUID, ownerGID
 		}
 
+		parent := filepath.Dir(fpath)
+		if err := os.MkdirAll(parent, 0755); err != nil {
```
anthonybishopric: You'll almost certainly want to use ` util.MkdirChownAll(dest, ownerUID, ownerGID, 0755)`

petertseng:
Now using `MkdirChownAll`, and a test with a tarball with just "a/b" in it.

anthonybishopric:
:ship:

bwester:
:ship:

PR #374, opened 2016-02-04
gzip: Error on paths above destination

petertseng:
If a tarball contains a path of the form `../a`, we wouldn't want to
allow this to escape from the intended destination directory. Such
behavior is unsafe, as it might allow modification of system files.

Behavior of BSD tar: Forbids extracting such a tarball unless
specifically directed to do so with the `-P` flag.

Behavior of GNU tar: Strips the leading `../` from the path, such that
you get `./a` extracted instead.

We propose that silently changing the extraction target might surprise
the user more than simply raising an error, so we prefer to raise the
error.


pkg/gzip/gzip.go
```diff
@@ -48,6 +49,7 @@ func ExtractTarGz(owner string, fp io.Reader, dest string) (err error) {
 			return util.Errorf("read error: %s", err)
 		}
 		fpath := filepath.Join(dest, hdr.Name)
+
```
petertseng: removing newline that snuck in

pkg/gzip/gzip_test.go
```diff
@@ -25,12 +28,35 @@ func TestFileWithoutDir(t *testing.T) {
...
 	err = ExtractTarGz(user.Username, file, dest)
-	Assert(t).IsNil(err, "expected no error extracting tarball")
 
-	_, err = os.Stat(filepath.Join(dest, "a", "b"))
-	Assert(t).IsNil(err, "expected no error statting extracted file")
+	check(err, dest)
+}
+
+func testExtraction(t *testing.T, tarfile string, check func(error, string)) {
+	testExtractionWithSetup(t, tarfile, func(string) error { return nil }, check)
```
petertseng: You may notice nothing calls `testExtractionWithSetup` with an actual setup function... I was going to use that in another test for upcoming functionality I haven't finished yet. I guess I can remove it from this PR

pkg/gzip/gzip.go
```diff
@@ -56,6 +57,12 @@ func ExtractTarGz(owner string, fp io.Reader, dest string) (err error) {
 		}
 
 		parent := filepath.Dir(fpath)
+
+		// Ignore all paths that would end up outside the destination directory.
+		if !strings.HasPrefix(parent, dest) {
+			continue
```
petertseng: desired behavior? error? ignore?

bwester: My own ranked preferences are:
1. Call the archive "invalid" and reject it
2. Extract into the destination dir: `/my/jail` + `../../bin/bash` => `/my/jail/bin/bash`
3. Ignore the file
4. Silently ignore file  <= this will bite us eventually

petertseng:
It is now an error to try to extract paths that would end up outside the destination directory.

The test is updated to expect this error (as well as the file's nonexistence).

The documentation for `ExtractTarGz` is updated to warn about the error behavior.

bwester:
:shipit:

savarin:
Hey guys,

Can you please remove me from this mailing list? I don't why I'm receiving
these emails.

Thanks,

On Fri, Feb 5, 2016 at 12:11 AM, Peter Tseng <notifications@github.com>
wrote:

> Merged #374 <https://github.com/square/p2/pull/374>.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/square/p2/pull/374#event-539949019>.
>



-- 


*Alexandre Schweitzer*Senior Media Analyst Manager
Cell: +1 (415) 244-1071
Skype: alexandre.schweitzer_hitfox
LinkedIn: Alexandre Schweitzer
<http://www.linkedin.com/in/alexschweitzer/en>

*AppLift Inc.*
143 2nd St. 4th floor
San Francisco, CA 94105
www.AppLift.com <http://www.applift.com/>

xaviershay:
@savarin at some point did you "watch" this project? Click through to the PR, click "unwatch" at top right. (We can't do this on your behalf.)

PR #376, opened 2016-02-04
gzip: Error on escaping symlinks

petertseng:
Consider a tarball with `a/b` being extracted to `/my/jail`. Now let's
consider the case where `/my/jail/a` is a symlink pointing to `..` or
even `/usr` (this may happen through a previous entry in the same
tarball, a previous tarball, or any variety of reasons).

If we do not check symlinks, we may blindly extract `b` to a path
outside of `/my/jail`. We should disallow this, as this sort of path
traversal can result in unsafe behavior (modification of system files).

Behavior of BSD tar: Forbids extracting such a tarball unless
specifically directed to do so with the `-P` flag.

Behavior of GNU tar: Extracts normally, such that you get `/my/b`
or `/usr/b` extracted, escaping the jail.

We will err on the side of caution because we may extract files as a
privileged user. Therefore we will raise an error in this case.


pkg/gzip/gzip.go
```diff
@@ -43,6 +46,11 @@ func ExtractTarGz(owner string, fp io.Reader, dest string) (err error) {
 		return util.Errorf("error setting ownership of root directory %s: %s", dest, err)
 	}
 
+	realDest, err := filepath.EvalSymlinks(dest)
+	if err != nil {
+		return util.Errorf("error evaling symlink of root directory %s: %s", dest, err)
```
petertseng: "evaling" here, "evaluating" below. please be consistent

petertseng:
commit message a bit too abstract. making it concrete.

pkg/gzip/gzip_test.go
```diff
@@ -52,3 +60,54 @@ func TestPathWithParent(t *testing.T) {
...
+		Assert(t).IsNil(tarErr, "expected no error extracting tarball")
+
+		_, err := os.Stat(filepath.Join(dest, "b", "c"))
+		Assert(t).IsNil(err, "expected no error statting extracted file")
+	})
+}
+
+func TestEscapingAbsoluteSymlink(t *testing.T) {
+	testEscapingSymlink(t, func(dest string) string {
+		return filepath.Join(dest, "..")
```
petertseng: `filepath.Dir` ?

pkg/gzip/gzip.go
```diff
@@ -69,7 +77,29 @@ func ExtractTarGz(owner string, fp io.Reader, dest string) (err error) {
 			)
 		}
 
+		existing := firstExistingParent(fpath)
+		if realParent, err := filepath.EvalSymlinks(existing); err == nil {
+			if !strings.HasPrefix(realParent, realDest) {
```
petertseng: the test above: `!strings.HasPrefix(fpath, dest)` is subsumed by this, so it could be deleted. the only difference is the error message (the error above doesn't mention symlinks, this error does). think I would like to present a different error message in the two cases, so maybe I'll leave it.

petertseng: there, that should do it... combined the two.

petertseng: I made the tests evaluate the symlinks in the `tmpdir` paths, so now the tests seem good enough to verify that everything is good.

petertseng:
OK, final answer. Ready for your review

petertseng:
assuming https://travis-ci.org/square/p2/builds/107139225 is transient, let's see

petertseng:
rebased on master

pkg/gzip/gzip.go
```diff
@@ -165,3 +191,11 @@ func ExtractTarGz(owner string, fp io.Reader, dest string) (err error) {
 	}
 	return nil
 }
+
+func firstExistingParent(path string) string {
+	for _, err := os.Stat(path); os.IsNotExist(err); {
+		path = filepath.Dir(path)
+		_, err = os.Stat(path)
```
rudle: I can't see where this err gets checked or returned.

petertseng: above in `os.IsNotExist(err)`.

I could move this update to be after the second semicolon to make that more clear that it's the same `err`. A bit undecided.

mpuncel: what if you got an error that was not `os.IsNotExist`? you would just return a garbage path right?

rudle: If I'm following things, as it's written this code does the Stat operation twice for the same value of `path`. I think the intent of this code would be clearer if it were written as a for expression like you suggest.

petertseng: > what if you got an error that was not `os.IsNotExist`? you would just return a garbage path right?

OK, that's bad. I think I'd better check that `err == nil` then.

petertseng:
Updated:

* rebase on master
* put `_, err = os.Stat(path)` in the post-statement of the for loop to make it clearer that it's the same `err` as the one in the init-statement. Ideally I'd put the `path =` update in there too but can I actually put both of them there?
* check for any error, not just `isNotExist`, so that getting a different error won't give us a bogus path.

pkg/gzip/gzip.go
```diff
@@ -165,3 +191,10 @@ func ExtractTarGz(owner string, fp io.Reader, dest string) (err error) {
 	}
 	return nil
 }
+
+func firstExistingParent(path string) string {
+	for _, err := os.Stat(path); err != nil; _, err = os.Stat(path) {
```
mpuncel: my feeble mind can't parse this

what if `err` is `os.TheStatFailedIDontKnowWhatsUp`?

petertseng: then it will not be the case that `err != nil`, thus the loop will not terminate, thus we will go one path higher up. Is it possible for stat to fail without knowing what's up... I'm not sure.

If stat could fail intermittently, I'm a bit concerned because it could falsely allow a bad tarball through. If tarball has `dir/passwd`, but statting `dir` fails for whatever reason, even though `dir` links to `/etc`, then we could overwrite `/etc/passwd`.

mpuncel: I could envision a scenario where the filesystem is just unavailable somehow and all `stat` calls fail and this becomes an infinite loop

petertseng: I assumed this wouldn't happen and that we'd at least not get an error on the root destination directory, but I've changed my mind and agree with your concern. I think I should deal with this by breaking the loop if `dirname(path) == path`

petertseng:
Based on discussion:

* It's agreed that the issue fixed by the PR is still important to fix
* But this PR is not a complete fix: race condition because the time of check is distinct from the time of extract - someone acting between these two times would evade the check
* We should extract as the target user. Then anyone who attempts to escape via a symlink or a tarball with relative paths will just get a permission error.

So I close this PR and will work on extraction as the target user.

petertseng:
In loving memory.

Replaced by #739.

PR #378, opened 2016-02-09
travis-ci: Download consul binary from Hashicorp

bwester:
The old download location quit working. At the time of this commit, the
official Consul website lists the included URI as the location to find
consul-0.5.2.


mpuncel:
:ship:

PR #379, opened 2016-02-09
Add List() to pkg/kp/rollstore.Store interface.

mpuncel:


bwester:
Could you add a description, perhaps discussing why we would want a `List()` operation for rolling updates?

pkg/kp/rollstore/store.go
```diff
@@ -176,3 +186,21 @@ func RollLockPath(rcId rcf.ID) (string, error) {
...
 	return path.Join(kp.LOCK_TREE, subRollPath), nil
 }
+
+func (s consulStore) kvpToRU(kvp *api.KVPair) (rollf.Update, error) {
+	ru := rollf.Update{}
+	err := json.Unmarshal(kvp.Value, &ru)
+	return ru, util.Errorf("Unable to unmarshal value as rolling update: %s", err)
+}
+
+func (s consulStore) kvpsToRUs(kvps []*api.KVPair) ([]rollf.Update, error) {
```
bwester: Is this method useful elsewhere?

Why do you need this to be a method of `consulStore`?

mpuncel: i was copying what's done in rcstore, but you're right it doesn't make sense. removed

pkg/kp/rollstore/store.go
```diff
@@ -176,3 +190,9 @@ func RollLockPath(rcId rcf.ID) (string, error) {
 
 	return path.Join(kp.LOCK_TREE, subRollPath), nil
 }
+
+func kvpToRU(kvp *api.KVPair) (rollf.Update, error) {
+	ru := rollf.Update{}
+	err := json.Unmarshal(kvp.Value, &ru)
+	return ru, util.Errorf("Unable to unmarshal value as rolling update: %s", err)
```
petertseng: Hmm. So the `error` returned by `kvpToRU` will always be non-nil, even if that returned by `json.Umharshal` is non-nil. I'm unsure of this, especially since this will make  at least `List()` always fail

mpuncel:
shame! okay going to write a test

pkg/kp/rollstore/store_test.go
```diff
@@ -52,3 +61,125 @@ func TestRollLockPathErrorNoID(t *testing.T) {
...
+		}
+	}
+
+	if !matched {
+		t.Errorf("Expected to find a roll with NewRC of %s", testRCId)
+	}
+
+	matched = false
+	for _, val := range rolls {
+		if val.NewRC != testRCId2 {
```
anthonybishopric: Should this be `==`?

mpuncel: yup

anthonybishopric:
:shipit:

PR #380, opened 2016-02-10
Add some useful log messages

anthonybishopric:
* Adds messages when we're about to install, launch, or execute a hook.
* Adds messages when a pod has changed health or is flapping.

### DIFF
--- a/pkg/kp/healthmanager.go
+++ b/pkg/kp/healthmanager.go
@@ -207,6 +207,14 @@ func processHealthUpdater(
 			// The local health checker sent a new result
 			if ok {
 				logger.NoFields().Debug("new health status: ", h.Status)
+				if !healthEquiv(localHealth, &h) {
+					msg := fmt.Sprintf("Service %s is now %s", h.Service, h.Status)
+					if health.Passing.Is(h.Status) {
+						logger.NoFields().Infoln(msg)
+					} else {
+						logger.NoFields().Warnln(msg)
+					}
+				}
 				localHealth = &h
 			} else {
 				logger.NoFields().Debug("check stream closed")
@@ -232,7 +240,7 @@ func processHealthUpdater(
 				remoteHealth = result.Health
 				if result.Throttle && throttle == nil {
 					throttle = time.After(time.Duration(*HealthResumeLimit) * bucketRefreshRate)
-					logger.NoFields().Warning("health is flapping; throttling updates")
+					logger.NoFields().Warningf("Service %s health is flapping; throttling updates", result.Health.Service)
 				}
 			}
 		case <-throttle:
--- a/pkg/preparer/orchestrate.go
+++ b/pkg/preparer/orchestrate.go
@@ -293,6 +293,8 @@ func (p *Preparer) resolvePair(pair ManifestPair, pod Pod, logger logging.Logger
 func (p *Preparer) installAndLaunchPod(pair ManifestPair, pod Pod, logger logging.Logger) bool {
 	p.tryRunHooks(hooks.BEFORE_INSTALL, pod, pair.Intent, logger)
 
+	logger.NoFields().Infoln("Installing pod and launchables")
+
 	err := pod.Install(pair.Intent)
 	if err != nil {
 		// install failed, abort and retry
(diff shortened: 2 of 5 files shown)


pkg/kp/healthmanager.go
```diff
@@ -230,9 +230,14 @@ func processHealthUpdater(
 			write = nil
 			if result.OK {
 				remoteHealth = result.Health
+				healthLogger := logger.SubLogger(logrus.Fields{
```
bwester: `logger` already has these fields. See `consulHealthManager.NewUpdater()`.

pkg/kp/healthmanager.go
```diff
@@ -230,9 +230,14 @@ func processHealthUpdater(
...
+					"pod":     result.Health.Id,
+					"service": result.Health.Service,
+				})
+				healthLogger.NoFields().Warningf("Service %s is now %s", result.Health.Status, result.Health.Service)
 				if result.Throttle && throttle == nil {
 					throttle = time.After(time.Duration(*HealthResumeLimit) * bucketRefreshRate)
-					logger.NoFields().Warning("health is flapping; throttling updates")
+					healthLogger.NoFields().Warningf("Service %s health is flapping; throttling updates", result.Health.Service)
```
bwester: Is it not sufficient to have the service already encoded in a field for this log message?

pkg/kp/healthmanager.go
```diff
@@ -230,9 +230,14 @@ func processHealthUpdater(
 			write = nil
 			if result.OK {
 				remoteHealth = result.Health
+				healthLogger := logger.SubLogger(logrus.Fields{
+					"pod":     result.Health.Id,
+					"service": result.Health.Service,
+				})
+				healthLogger.NoFields().Warningf("Service %s is now %s", result.Health.Status, result.Health.Service)
```
bwester: I think it would be more useful to print this message when the local state changes, instead of when the remote state was successfully updated.  Perhaps do this up around line 210, if the old and new states are not equivalent?

pkg/preparer/orchestrate.go
```diff
@@ -322,6 +324,8 @@ func (p *Preparer) installAndLaunchPod(pair ManifestPair, pod Pod, logger loggin
 
 	p.tryRunHooks(hooks.BEFORE_LAUNCH, pod, pair.Intent, logger)
 
+	logger.NoFields().Infoln("Calling enable and setting up runit services")
```
bwester: We set up runit first, then call enable, right?

pkg/kp/healthmanager.go
anthonybishopric: :+1: , removing this sublogger

pkg/kp/healthmanager.go
anthonybishopric: I'm leaving it in because it's the message we will display to users. "Service became unhealthy" without changes to frontends that display this message might not be as friendly without the actual service name in the log message.

pkg/kp/healthmanager.go
anthonybishopric: That does have the added benefit of exposing when health changes aren't successfully being written. Will change.

pkg/preparer/orchestrate.go
anthonybishopric: You are correct, will make the ordering explicit

pkg/kp/healthmanager.go
```diff
@@ -208,6 +208,9 @@ func processHealthUpdater(
 			if ok {
 				logger.NoFields().Debug("new health status: ", h.Status)
 				localHealth = &h
+				if !healthEquiv(localHealth, remoteHealth) {
```
bwester: I expect `checksStream` to receive a value once every second. Of course, the first time there is a transition, `localHealth` won't be equivalent to `remoteHealth`. But if it takes longer than 1s to write that into Consul, due to throttling or Consul slowdowns, the message will be printed multiple times.

Would this still accomplish your goals if it were edge-triggered on `localHealth` changing (without caring about the remote value)?

anthonybishopric: Is your suggestion to add another health variable to this function, a la `previousLocalHealth` and compare to that?

bwester: Isn't that just the relationship between `h` and `localHealth`?

```
case h, ok := <-checksStream:
    if ok {
        if !healthEquiv(&h, localHealth) { logger.NoFields().Infof(...) }
        localHealth = &h
    }
    ...
```

anthonybishopric: Doh, yes it is. Will use that.

bwester:
:ship:

PR #381, opened 2016-02-11
Restart a log agent whenever we restart a service

rudle:
The log process is not bound by the RestartPolicy because we need to be
restarted in case of a crash regardless of what the service does.

### DIFF
--- a/pkg/hoist/hoist_launchable.go
+++ b/pkg/hoist/hoist_launchable.go
@@ -201,6 +201,11 @@ func (hl *Launchable) start(serviceBuilder *runit.ServiceBuilder, sv runit.SV) e
 		if err != nil && err != runit.SuperviseOkMissing && err != runit.Killed {
 			return err
 		}
+
+		if _, err = sv.Restart(&executable.LogAgent, hl.RestartTimeout); err != nil {
+			return err
+		}
+
 	}
 
 	return nil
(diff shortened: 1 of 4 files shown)


pkg/hoist/hoist_launchable.go
```diff
@@ -186,13 +186,13 @@ func (hl *Launchable) stop(serviceBuilder *runit.ServiceBuilder, sv runit.SV) er
 // Start will take a launchable and start every runit service associated with the launchable.
 // All services will attempt to be started.
 func (hl *Launchable) start(serviceBuilder *runit.ServiceBuilder, sv runit.SV) error {
+	var err error
```
mpuncel: why'd you move this?

pkg/hoist/hoist_launchable.go
```diff
@@ -201,6 +201,8 @@ func (hl *Launchable) start(serviceBuilder *runit.ServiceBuilder, sv runit.SV) e
 		if err != nil && err != runit.SuperviseOkMissing && err != runit.Killed {
 			return err
 		}
+
+		_, err = sv.Restart(&executable.LogAgent, hl.RestartTimeout)
```
mpuncel: should you do something with this error?

pkg/hoist/hoist_launchable.go
rudle: I guess it's diff noise at this point. In an earlier version I was recycling this value a bit more.

pkg/hoist/hoist_launchable.go
rudle: d'oh. Good catch.

mpuncel:
other than those comments, LGTM

petertseng:
If you wanted to be super, could make the `RecordingSV` record the service for which each command is issued so we can be sure the `restart` affected the logger and not the app, but I'm not too fussed about that

PR #382, opened 2016-02-11
Move status config into it's own stanza, allow custom path

drcapulet:


pkg/pods/manifest.go
```diff
@@ -167,20 +177,39 @@ func (m manifestBuilder) SetConfig(config map[interface{}]interface{}) error {
...
+func (manifest *manifest) SetStatusHTTP(statusHTTP bool) {
+	manifest.StatusHTTP = false
+	manifest.Status.HTTP = statusHTTP
 }
 
-func (manifest *manifest) GetStatusHTTP() bool {
-	return manifest.StatusHTTP
+func (manifest *manifest) GetStatusPath() string {
+	if manifest.Status.Path != "" {
+		return path.Join("/", manifest.Status.Path)
```
anthonybishopric: Nit: just use a complete string instead of an array

drcapulet: What do you mean? This is meant to normalize the URL so that it has one and only one leading slash.

anthonybishopric: Got it, sorry, I thought `Status.Path` was an array - now I see it's just for normalization :+1:

PR #383, opened 2016-02-12
Reduce restart timeout period for log agent

rudle:
7 seconds is runit's default wait period for stopping processes: http://smarden.org/runit/sv.8.html

We are less friendly to the log agent than the service it is paired with
primarily because we need the log agent to boot before the service does. Not
having a log agent on the reading end of STDOUT, STDERR may result in
indigestion for the service.


anthonybishopric:
:+1:

bwester:
:+1:

rudle:
Thanks for the review!

PR #384, opened 2016-02-13
Launchables may specify environment variables

anthonybishopric:
This is a useful feature that is mostly, but not entirely,
solved by CONFIG_PATH. There are occasions where code
cannot conveniently use the exported YAML file for config,
despite being a more flexible configuration system. This
is especially true for legacy or 3rd party software that
is conventionally configured with environment variables.

A bug I discovered while implementing this: the current
implementation of the opencontainer launchable does not
support pod-supplied variables.

### DIFF
--- a/pkg/hoist/hoist_launchable.go
+++ b/pkg/hoist/hoist_launchable.go
@@ -37,6 +37,7 @@ type Launchable struct {
 	CgroupConfigName string              // The string in PLATFORM_CONFIG to pass to p2-exec
 	RestartTimeout   time.Duration       // How long to wait when restarting the services in this launchable.
 	RestartPolicy    runit.RestartPolicy // Dictates whether the launchable should be automatically restarted upon exit.
+	SuppliedEnvVars  map[string]string   // A map of user-supplied environment variables to be exported for this launchable
 }
 
 // LaunchAdapter adapts a hoist.Launchable to the launch.Launchable interface.
@@ -384,4 +385,8 @@ func (hl *Launchable) AllInstallsDir() string {
 func (hl *Launchable) InstallDir() string {
 	launchableName := hl.Version()
 	return filepath.Join(hl.AllInstallsDir(), launchableName)
+}
+
+func (hl *Launchable) EnvVars() map[string]string {
+	return hl.SuppliedEnvVars
 }
--- a/pkg/opencontainer/containers.go
+++ b/pkg/opencontainer/containers.go
@@ -37,15 +37,16 @@ var RuncPath = param.String("runc_path", "/usr/local/bin/runc")
 
 // Launchable represents an installation of a container.
 type Launchable struct {
-	Location       string              // A URL where we can download the artifact from.
-	ID_            string              // A (pod-wise) unique identifier for this launchable, used to distinguish it from other launchables in the pod
-	ServiceID_     string              // A (host-wise) unique identifier for this launchable, used when creating runit services
-	RunAs          string              // The user to assume when launching the executable
-	RootDir        string              // The root directory of the launchable, containing N:N>=1 installs.
-	P2Exec         string              // The path to p2-exec
-	RestartTimeout time.Duration       // How long to wait when restarting the services in this launchable.
-	RestartPolicy  runit.RestartPolicy // Dictates whether the container should be automatically restarted upon exit.
-	CgroupConfig   cgroups.Config      // Cgroup parameters to use with p2-exec
+	Location        string              // A URL where we can download the artifact from.
+	ID_             string              // A (pod-wise) unique identifier for this launchable, used to distinguish it from other launchables in the pod
+	ServiceID_      string              // A (host-wise) unique identifier for this launchable, used when creating runit services
+	RunAs           string              // The user to assume when launching the executable
+	RootDir         string              // The root directory of the launchable, containing N:N>=1 installs.
+	P2Exec          string              // The path to p2-exec
+	RestartTimeout  time.Duration       // How long to wait when restarting the services in this launchable.
+	RestartPolicy   runit.RestartPolicy // Dictates whether the container should be automatically restarted upon exit.
+	CgroupConfig    cgroups.Config      // Cgroup parameters to use with p2-exec
+	SuppliedEnvVars map[string]string   // User-supplied env variables
 
 	spec *LinuxSpec // The container's "config.json"
 }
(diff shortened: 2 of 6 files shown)


pkg/hoist/hoist_launchable.go
```diff
@@ -24,19 +24,20 @@ import (
...
+	PodEnvDir        string                 // The value for chpst -e. See http://smarden.org/runit/chpst.8.html
+	Fetcher          uri.Fetcher            // Callback that downloads the file from the remote location.
+	RootDir          string                 // The root directory of the launchable, containing N:N>=1 installs.
+	P2Exec           string                 // Struct that can be used to build a p2-exec invocation with appropriate flags
+	ExecNoLimit      bool                   // If set, execute with the -n (--no-limit) argument to p2-exec
+	CgroupConfig     cgroups.Config         // Cgroup parameters to use with p2-exec
+	CgroupConfigName string                 // The string in PLATFORM_CONFIG to pass to p2-exec
+	RestartTimeout   time.Duration          // How long to wait when restarting the services in this launchable.
+	RestartPolicy    runit.RestartPolicy    // Dictates whether the launchable should be automatically restarted upon exit.
+	SuppliedEnvVars  map[string]interface{} // A map of user-supplied environment variables to be expored for this launchable
```
petertseng: s/expored/exported/

pkg/pods/pod.go
```diff
@@ -529,6 +529,12 @@ func (pod *Pod) setupConfig(manifest Manifest, launchables []launch.Launchable)
 	}
 
 	for _, launchable := range launchables {
+		// we need to remove any unset env vars
+		err = os.RemoveAll(launchable.EnvDir())
+		if err != nil && !os.IsNotExist(err) {
```
petertseng: Since `RemoveAll` [returns no error if the path doesn't exist](https://golang.org/pkg/os/#RemoveAll), I suspect the `!os.IsNotExist` is unnecessary, but I see no harm in leaving it in (also don't have time to check what happens if a parent doesn't exist, so maybe it's needed)

pkg/pods/pod_test.go
```diff
@@ -152,6 +156,10 @@ config:
 		launchableRootEnv, err := ioutil.ReadFile(filepath.Join(launchable.EnvDir(), "LAUNCHABLE_ROOT"))
 		Assert(t).IsNil(err, "should not have erred reading the launchable root env file")
 		Assert(t).AreEqual(launchable.InstallDir(), string(launchableRootEnv), "The launchable root path did not match expected")
+
+		enableBlamSetting, err := ioutil.ReadFile(filepath.Join(launchable.EnvDir(), "ENABLED_BLAMS"))
+		Assert(t).IsNil(err, "should not have erred reading custom env var")
+		Assert(t).AreEqual("5", string(enableBlamSetting), "The setting was wrong")
```
petertseng: probably should at least mention that it's a the launchable custom env var instead of a vague setting

petertseng:
just one typo, otherwise I'm :+1:  on it

pkg/hoist/hoist_launchable.go
bwester: Environment variables are definitely strings, so let's have `map[string]string`

---

I agree that these are definitely "user-supplied" variables, but why not keep the name simple: `EnvVars`?

bwester: Never mind, I see that the function takes the name `EnvVars()`. This apparent naming conflict comes up more than I'd like, but I don't see a clear solution.

pkg/opencontainer/containers.go
```diff
@@ -150,6 +155,7 @@ func (l *Launchable) Executables(serviceBuilder *runit.ServiceBuilder) ([]launch
 			p2exec.P2ExecArgs{
 				NoLimits: true,
 				WorkDir:  l.InstallDir(),
+				EnvDirs:  []string{l.EnvDir()}, // TODO: add pod env dir
```
bwester: I'm not at all sure that `runc`'s environment gets propagated inside the container.  Probably the correct way to approach this is to write the environment variables to the container's [`config.json`](https://github.com/opencontainers/specs/blob/master/config.md#process-configuration).

I'd prefer that you leave this out for the time being.

pkg/pods/manifest.go
```diff
@@ -22,13 +22,14 @@ import (
...
-	RestartTimeout          string         `yaml:"restart_timeout,omitempty"`
-	CgroupConfig            cgroups.Config `yaml:"cgroup,omitempty"`
+	LaunchableType          string                 `yaml:"launchable_type"`
+	LaunchableId            string                 `yaml:"launchable_id"`
+	Location                string                 `yaml:"location"`
+	DigestLocation          string                 `yaml:"digest_location,omitempty"`
+	DigestSignatureLocation string                 `yaml:"digest_signature_location,omitempty"`
+	RestartTimeout          string                 `yaml:"restart_timeout,omitempty"`
+	CgroupConfig            cgroups.Config         `yaml:"cgroup,omitempty"`
+	EnvVars                 map[string]interface{} `yaml:"env,omitempty"`
```
bwester: If its JSON name is `env`, can you name the Golang variable `Env`?

pkg/pods/pod.go
```diff
@@ -529,6 +529,12 @@ func (pod *Pod) setupConfig(manifest Manifest, launchables []launch.Launchable)
 	}
 
 	for _, launchable := range launchables {
+		// we need to remove any unset env vars
```
bwester: Why? What happens if you don't do this?

anthonybishopric: If someone elides an environment variable from their pod manifest, it should not be present on the subsequent deploy (ie, we need to actively remove it from the env directory)

pkg/pods/manifest.go
anthonybishopric: :+1:

pkg/opencontainer/containers.go
anthonybishopric: Sure thing, will leave out

pkg/pods/pod.go
bwester: Ah, because we re-use env directories between pods. Could you append "... from a previous pod" to the comment?

pkg/pods/pod.go
anthonybishopric: I'll remove it since I agree it's not strictly necessary. By this point in the process, the parent must exist since it's already been installed at this point.

anthonybishopric:
Comments should all be addressed

bwester:
:+1:

petertseng:
:+1:

PR #385, opened 2016-02-15
Adds RollDelay to rolling updates

anthonybishopric:
RollDelay establishes a required period of inactivity prior to
altering any underlying RCs. See the documentation for more
information.

### DIFF
--- a/pkg/roll/run_update.go
+++ b/pkg/roll/run_update.go
@@ -16,6 +16,7 @@ import (
 	"github.com/square/p2/pkg/rc"
 	rcf "github.com/square/p2/pkg/rc/fields"
 	"github.com/square/p2/pkg/roll/fields"
+	"github.com/square/p2/pkg/util"
 )
 
 type update struct {
@@ -180,22 +181,12 @@ ROLL_LOOP:
 						return
 					}
 
-					// Check health again following the roll delay. If things have gotten
-					// worse since we last looked, or there is an error, we break this iteration.
-					checks, err = u.hcheck.Service(newFields.Manifest.ID().String())
-					afterDelayNew, err := u.countHealthy(u.NewRC, checks)
+					// determine the new value of `next`, which may have changed
+					// following the delay.
+					next, err = u.shouldRollAfterDelay(newFields)
 
 					if err != nil {
-						u.logger.WithError(err).Errorln("Could not count nodes after delay, skipping this iteration")
-						break
-					}
-
-					if afterDelayNew.Healthy < newNodes.Healthy {
-						u.logger.WithFields(logrus.Fields{
-							"beforeDelay": newNodes.Healthy,
-							"afterDelay":  afterDelayNew.Healthy,
-							"delay":       u.RollDelay,
-						}).Errorf("Healthy node count on the new RC dropped after %v roll delay, skipping this iteration", u.RollDelay)
+						u.logger.NoFields().Errorln(err)
 						break
 					}
 				}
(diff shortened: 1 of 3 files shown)


pkg/roll/run_update.go
```diff
@@ -167,13 +167,44 @@ ROLL_LOOP:
 				break ROLL_LOOP
 			}
 
-			next := algorithm(oldNodes.Healthy, newNodes.Healthy, u.DesiredReplicas, u.MinimumReplicas)
+			next := rollAlgorithm(oldNodes.Healthy, newNodes.Healthy, u.DesiredReplicas, u.MinimumReplicas)
 			if next > 0 {
+				// apply the delay only if we've already added to the new RC, since there's
+				// no value in sitting around doing nothing before anything has happened.
+				if newNodes.Desired > 0 && u.RollDelay > time.Duration(0) {
+					u.logger.WithField("delay", u.RollDelay).Infof("Waiting %v before continuing deploy", u.RollDelay)
```
bwester: I'm conflicted about duplicating the `u.RollDelay` value in these log messages. I can understand that we want the user to see "Waiting 30s ...". Why bother with the field?

pkg/roll/run_update.go
```diff
@@ -167,13 +167,44 @@ ROLL_LOOP:
...
+
+					select {
+					case <-time.After(u.RollDelay):
+					case <-quit:
+						return
+					}
+
+					// Check health again following the roll delay. If things have gotten
+					// worse since we last looked, or there is an error, we break this iteration.
+					checks, err = u.hcheck.Service(newFields.Manifest.ID().String())
```
bwester: You probably shouldn't ignore `err`

pkg/roll/run_update.go
anthonybishopric: I don't mind having fully expanded human messages as well as metadata in the logs. It gives us the opportunity in the future to both display the message and also annotate it. For example, I imagine printing this message on our dashboard with a :clock10: adjacent to it, indicating that the system is waiting. It seems reasonable to want a field to indicate that icon should be displayed.

Is it the duplicativeness that bothers you?

pkg/roll/run_update.go
anthonybishopric: Correct, fixing

pkg/roll/run_update.go
```diff
@@ -167,13 +167,44 @@ ROLL_LOOP:
...
+					afterDelayNew, err := u.countHealthy(u.NewRC, checks)
+
+					if err != nil {
+						u.logger.WithError(err).Errorln("Could not count nodes after delay, skipping this iteration")
```
bwester: "this iteration" won't make any sense to a user. I had to go back and review the code to figure out what "iteration" you were referring to.  And I'm not sure that it's relevant that _this_ check happened to occur after a delay.  How about: `Could not count nodes`?

pkg/roll/run_update.go
```diff
@@ -167,13 +167,44 @@ ROLL_LOOP:
...
+
+					if afterDelayNew.Healthy < newNodes.Healthy {
+						u.logger.WithFields(logrus.Fields{
+							"beforeDelay": newNodes.Healthy,
+							"afterDelay":  afterDelayNew.Healthy,
+							"delay":       u.RollDelay,
+						}).Errorf("Healthy node count on the new RC dropped after %v roll delay, skipping this iteration", u.RollDelay)
+						break
+					}
+				}
```
bwester: Do you think it would make sense to re-run `rollAlgorithm()` before continuing, in case the health has further changed.  I know, by this point, health is no worse than it was before waiting, but what if it's better?

bwester:
:ship:

PR #387, opened 2016-02-18
Make replication controllers ensure consistency of their nodes

bwester:
Adds a consistency check to replication controllers so that they will ensure
that their nodes are running the same pod manifest specified in the RC.

### DIFF
--- a/pkg/rc/replication_controller.go
+++ b/pkg/rc/replication_controller.go
@@ -1,6 +1,7 @@
 package rc
 
 import (
+	"fmt"
 	"strings"
 	"time"
 
@@ -35,12 +36,20 @@ type ReplicationController interface {
 	// The error channel will be closed in response.
 	WatchDesires(quit <-chan struct{}) <-chan error
 
-	// CurrentNodes() returns all nodes that this replication controller is currently scheduled on,
-	// by selecting pods labeled with the replication controller's ID.
-	CurrentNodes() ([]string, error)
+	// CurrentPods() returns all pods managed by this replication controller.
+	CurrentPods() (PodLocations, error)
+}
 
-	// Internal: meetDesires synchronously schedules or unschedules pods to meet desired state.
-	meetDesires() error
+type PodLocation struct{ Node, PodID string }
+type PodLocations []PodLocation
+
+// Nodes returns a list of just the locations' nodes.
+func (l PodLocations) Nodes() []string {
+	nodes := make([]string, len(l))
+	for i, pod := range l {
+		nodes[i] = pod.Node
+	}
+	return nodes
 }
 
 // These methods are the same as the methods of the same name in kp.Store.
@@ -51,6 +60,13 @@ type kpStore interface {
 		nodeName string,
 		manifest pods.Manifest,
 	) (time.Duration, error)
+
+	Pod(
+		podPrefix kp.PodPrefix,
+		nodeName string,
+		podId types.PodID,
+	) (pods.Manifest, time.Duration, error)
+
 	DeletePod(podPrefix kp.PodPrefix,
 		nodeName string,
 		manifestID types.PodID,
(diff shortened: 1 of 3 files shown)


pkg/rc/replication_controller.go
```diff
@@ -35,9 +36,9 @@ type ReplicationController interface {
 	// The error channel will be closed in response.
 	WatchDesires(quit <-chan struct{}) <-chan error
 
-	// CurrentNodes() returns all nodes that this replication controller is currently scheduled on,
+	// CurrentPods() returns all nodes that this replication controller is currently scheduled on,
```
mpuncel: not the result of this PR, but this comment is weird. the replication controller isn't scheduled on nodes

pkg/rc/replication_controller.go
```diff
@@ -55,6 +56,12 @@ type kpStore interface {
 		nodeName string,
 		manifestID types.PodID,
 	) (time.Duration, error)
+	Pod(podPrefix kp.PodPrefix,
+		nodename string,
```
mpuncel: nit: the other methods have `nodeName` instead of `nodename`

pkg/rc/replication_controller.go
```diff
@@ -141,97 +148,183 @@ func (rc *replicationController) meetDesires() error {
...
+	current, err := rc.CurrentPods()
 	if err != nil {
 		return err
 	}
+
 	rc.logger.NoFields().Infof("Currently on nodes %s", current)
 
 	if rc.ReplicasDesired > len(current) {
-		eligible, err := rc.eligibleNodes()
+		err := rc.addPods(current)
```
mpuncel: :+1:

pkg/rc/replication_controller.go
bwester: The actual `kp.Store` interface uses `nodename`—I'm not sure why the other methods here want to be different.

pkg/rc/replication_controller.go
```diff
@@ -141,97 +148,183 @@ func (rc *replicationController) meetDesires() error {
...
+	for node, intentManifest := range current {
+		intentSHA, err := intentManifest.SHA()
+		if err != nil {
+			rc.logger.WithError(err).WithField("node", node).Warn("Could not hash manifest to determine consistency of intent")
+			preferred.Insert(node)
+			continue
+		}
+
+		if intentSHA != rcManifestSHA {
+			preferred.Insert(node)
```
mpuncel: are you sure we want to prefer removing a node that has an inconsistent manifest over one that is ineligible? I would think we should unschedule ineligible ones and then let `ensureConsistency` take care of the inconsistencies

mpuncel: ah i'm confused, `ensureConsistency` won't remove a pod, it just overwrites them. It still might be better to prefer ineligible nodes first but less important

bwester: I think the strategy you propose makes sense.

mpuncel:
looks good! my one concern is making the RC unschedule an inconsistent node with equal priority to an ineligible one

pkg/rc/replication_controller.go
```diff
@@ -35,9 +36,9 @@ type ReplicationController interface {
 	// The error channel will be closed in response.
 	WatchDesires(quit <-chan struct{}) <-chan error
 
-	// CurrentNodes() returns all nodes that this replication controller is currently scheduled on,
+	// CurrentPods() returns all nodes that this replication controller is currently scheduled on,
 	// by selecting pods labeled with the replication controller's ID.
-	CurrentNodes() ([]string, error)
+	CurrentPods() (map[string]pods.Manifest, error)
```
bwester: `CurrentPods()` is like `CurrentNodes()`, but it also looks up the nodes' intent manifests. After looking a bit more at this, I think it makes sense to keep `CurrentNodes()` in the public interface. Fetching the manifest involves more network operations, so I'd rather avoid them if possible, and it turns out that many users of `CurrentPods()` don't care about the manifest.

pkg/rc/replication_controller.go
bwester: After sleeping on the problem, it seems like too much added complexity to implement this proposal. @rudle and I only implemented this optimization to the remove-nodes path because it seemed straightforward. Implementing a priority-based set data type kind of blows that argument out of the water. And I agree with you, that removing "ineligible" nodes does take priority.

It's an optimization, in that anything "fixed" here would also be caught and fixed by the consistency check pass, so I'm going to leave out this optimization for now.

mpuncel: :+1:

mpuncel:
:ship:

pkg/rc/replication_controller.go
```diff
@@ -141,72 +157,129 @@ func (rc *replicationController) meetDesires() error {
...
+		intent, _, err := rc.kpStore.Pod(kp.INTENT_TREE, pod.Node, types.PodID(pod.PodID))
+		if err != nil && err != pods.NoCurrentManifest {
+			return err
+		}
+		var intentSHA string
+		if intent != nil {
+			intentSHA, err = intent.SHA()
 			if err != nil {
-				return err
+				rc.logger.WithError(err).WithField("node", pod.Node).Warn("Could not hash manifest to determine consistency of intent")
```
petertseng: I recall in 8edd6cb513acd419d23078a41bdf0f52baaabf60 we decided we couldn't chain calls to `WithError` and `WithField` in this manner. Is that still true? If it is, you'd want to use `WithErrorAndFields` here instead.

pkg/rc/replication_controller.go
```diff
@@ -141,72 +157,129 @@ func (rc *replicationController) meetDesires() error {
...
+			}
+			if intentSHA == manifestSHA {
+				continue
 			}
 		}
-	} else {
-		rc.logger.NoFields().Debugln("Taking no action")
+
+		rc.logger.WithField("node", pod.Node).WithField("intentManifestSHA", intentSHA).Info("Found inconsistency in scheduled manifest")
```
petertseng: possibly a place to use `WithFields`

pkg/rc/replication_controller.go
bwester: I am utterly unable to replicate that problem, even when I check out commit 7181aa8, the parent of 8edd6cb where `WithError()` was introduced. The following program fragment:

```
func main() {
	logging.DefaultLogger.WithError(util.Errorf("This is an error")).
        WithField("foo", "foo").WithField("bar", "bar").Infoln("My message")
}
```

always produces output with keys `"err"`, `"filename"`, `"function"`, `"line_number"`, `"foo"`, `"bar"`, et al., just as I would anticipate. Perhaps @mpuncel could provide more context about what was broken back then.

pkg/rc/replication_controller.go
bwester: I agree that `WithFields()` would work too, but since you point this out a day after the PR was merged, I'm reluctant to change it simply for stylistic reasons.

bwester:
Thanks for the post-merge review @petertseng!

pkg/rc/replication_controller.go
mpuncel: I'm not sure what I was thinking, I confirm that things work as expected

mpuncel: Actually, I think what i meant is that `WithError` only behaves properly if its first, since our Logger has special implementation.

Repro case: 

```package main

import (
	"fmt"
	"os"

	"github.com/square/p2/Godeps/_workspace/src/github.com/Sirupsen/logrus"
)

func main() {
	logger := logging.NewLogger(logrus.Fields{
		"a": 1,
	})
	logger.Logger.Out = os.Stdout

	sublogger := logger.SubLogger(logrus.Fields{
		"b": 2,
	})

	err := util.Errorf("some_error")

	fmt.Println("Using both calls:")
	sublogger.WithField("c", 3).WithError(err).Errorln("using both calls")

	fmt.Println("Using one call:")
	sublogger.WithErrorAndFields(err, logrus.Fields{
		"c": 3,
	},
	).Errorln("using one call")
}
```

output: 
```Using both calls:
{"Counter":0,"PID":33187,"a":1,"b":2,"c":3,"error":"main.go:22: some_error","level":"error","msg":"using both calls","time":"2016-02-22T09:23:59-08:00"}
Using one call:
{"Counter":2,"PID":33187,"a":1,"b":2,"c":3,"err":"main.go:22: some_error","filename":"main.go","function":"main.main","level":"error","line_number":22,"msg":"using one call","time":"2016-02-22T09:23:59-08:00"}
```

bwester: OK, looks like something's going to change...

PR #388, opened 2016-02-19
Add label type for rolling updates so that they may be labeled

mpuncel:
This will allow labels to be applied to rolling updates so they can be
located using label selectors


anthonybishopric:
:+1:

petertseng:
please add to `func AsType(v string) (Type, error)`

PR #389, opened 2016-02-22
logging: Defer most operations to Logrus

bwester:
This change refactors the logging package to be little more than a thin wrapper
around Logrus's native data structures. The wrapper allows P2 to add some extra
functionality to Logrus, like recording more information about errors and
attaching a sequential counter.

With this refactor, a couple bugs are also fixed:
* Use atomic increments in the ProcessCounter type to avoid data races
* Allow WithError() calls to work correctly when given as a second field, e.g.:
  `logger.WithField(...).WithError(e)`


mpuncel:
brilliant :+1:

anthonybishopric:
:ship:

PR #390, opened 2016-02-23
Allow fetching from relative file paths

bwester:
Several command line tools use the generic URI parsing library to fetch their
arguments. These tools are painful to use interactively due to their insistence
on using only absolute file paths. This change allows relative paths to be used
for schemaless URIs (local files) to make commands easier to type.


mpuncel:
:ship:

PR #391, opened 2016-02-23
Prevent replicator from overwriting managed pods

bwester:
Now that replication controllers are actively managing their pods' intent
manifests, the only effective way to change a pod's manifest is by going
through its controller. Simply changing the intent store, like what a
replicator does, will succeed, but the controller will reset the node later.
That probably isn't what was intended.

This commit changes the replicator to check whether its targeted pods are
already managed. If so, the replication is aborted.

### DIFF
--- a/bin/p2-replicate/main.go
+++ b/bin/p2-replicate/main.go
@@ -15,6 +15,7 @@ import (
 	"github.com/square/p2/pkg/health/checker"
 	"github.com/square/p2/pkg/kp"
 	"github.com/square/p2/pkg/kp/flags"
+	"github.com/square/p2/pkg/labels"
 	"github.com/square/p2/pkg/logging"
 	"github.com/square/p2/pkg/pods"
 	"github.com/square/p2/pkg/replication"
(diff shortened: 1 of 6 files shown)


anthonybishopric:
In general I wonder if it makes sense for this to be bypass-able, but I can't come up with a situation where I care just yet.

It would be nice if this solution worked for future types of controllers too, or at least made it easy to extend that check to other controllers. Perhaps the act of testing if an RC already exists for a pod could be centralized?

bwester:
Considering we don't have multiple kinds of controllers and there are no written plans for such, we'd be sparring over no more than dreams.

In my version of this dream, a pod is managed by a single controller (of any kind) at a time. It might make sense to have only a single `controller_id` label attached to a pod, and this code can be easily generalized for that scenario.

bwester:
> I wonder if it makes sense for this to be bypass-able

I hope not, but it doesn't hurt to take precautions. I'll add an off switch. If the locking safety check is any indication, it might be used more than I anticipate (but if that does happen, something's very wrong with controllers).

anthonybishopric:
Hopefully will never need `ignore-controllers`, but i'm glad the option exists. :ship:

PR #393, opened 2016-02-24
Mpuncel/protect against farm panic

mpuncel:


pkg/roll/farm.go
```diff
@@ -134,6 +135,12 @@ START_LOOP:
 					rlLogger.WithError(err).Errorln("Could not read new RC")
 					continue
 				}
+
+				if rcField == nil {
```
anthonybishopric: I think I would have preferred modeling this after the `os` package. 

```
if rcs.IsNotExist(err) {
 ...
}
else if err != nil {
 ... 
}
```

Adding a new pointer makes it possible to test for nil in places, but it also makes it hard to use the zero value of the struct safely. Not a serious enough problem to block this PR, just expressing that the `return nil, nil` case from `rcs.Get` deserves something a bit more structurally meaningful.

pkg/kp/rcstore/store.go
```diff
@@ -82,3 +85,11 @@ func RCUpdateLockPath(rcId fields.ID) (string, error) {
 
 	return path.Join(kp.LOCK_TREE, rcPath, "update"), nil
 }
+
+func IsNotExist(err error) bool {
+	if err == NoReplicationController {
```
petertseng: is this function equivalent to `return err == NoReplicationController` or is there something more subtle going on that prevents this?

mpuncel: nah you're right :) it used to be a type assertion on a type and i didn't clean up all the way

pkg/kp/rcstore/consul_store.go
```diff
@@ -341,7 +341,7 @@ func (s *consulStore) Watch(rc *fields.RC, quit <-chan struct{}) (<-chan struct{
 				case <-quit:
 				}
 			} else {
-				*rc = newRC
+				rc = &newRC
```
petertseng: as I understand it, [this will result in no externally-visible change to `rc`](http://play.golang.org/p/BOoyrK0lTa) - is that OK with you? If it is OK, how do callers of this function get an updated RC (or do they not care about having one?)?

petertseng:
I think I'm happy with this. :+1:

PR #394, opened 2016-02-24
Query health through localhost

### LINKED ISSUE #392: Health checks talk to hostname
Right now, health checks do an HTTP request to the hostname:
https://github.com/square/p2/blob/812016d11fe887b81c746f3b3c8569213d13023f/pkg/watch/health.go#L168

Since p2-preparer runs on the same node, it could just talk to `localhost` instead. This way apps can serve the status port on the loopback interface only. In many cases, it's not necessary to expose it to the outside world. 

(Alternatively, would it make sense to have this be configurable?)

csstaub:
Since p2-preparer runs on the same node, it can talk to the app on
localhost in order to check its health. This way apps can serve the
status port on the loopback interface only. In many cases, it's not
necessary to expose it to the outside world. Closes #392.

### DIFF
--- a/pkg/watch/health.go
+++ b/pkg/watch/health.go
@@ -68,12 +68,6 @@ func MonitorPodHealth(config *preparer.PreparerConfig, logger *logging.Logger, s
 		logger.WithError(err).Fatalln("error creating health monitor KV store")
 	}
 	healthManager := store.NewHealthManager(config.NodeName, *logger)
-	// if GetClient fails it means the certfile/keyfile/cafile were
-	// invalid or did not exist. It makes sense to throw a fatal error
-	client, err := config.GetClient(time.Duration(*HEALTHCHECK_TIMEOUT) * time.Second)
-	if err != nil {
-		logger.WithError(err).Fatalln("failed to get http client for this preparer")
-	}
 
 	node := config.NodeName
 	pods := []PodWatch{}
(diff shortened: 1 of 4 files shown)


anthonybishopric:
Main concern with this PR is that we use TLS status checks by default. This is useful in a small but significant way - it shows that the app can successfully use its secret keys. Given that we use [keywhiz-fs](https://github.com/square/keywhiz-fs) that is a somewhat important quality to establish on deploy. @csstaub is there a way to do this without disabling secure status checks?

pkg/watch/health.go
```diff
@@ -95,7 +89,7 @@ func MonitorPodHealth(config *preparer.PreparerConfig, logger *logging.Logger, s
 			// check if pods have been added or removed
 			// starts monitor routine for new pods
 			// kills monitor routine for removed pods
-			pods = updatePods(healthManager, client, pods, results, node, logger)
+			pods = updatePods(healthManager, config, pods, results, node, logger)
```
anthonybishopric: This change will cause each new watch result to create a new `http.Client`, which is actually super dangerous. Each one caches pools of TCP connections. I'd rather you just init the insecure client outside this loop here and pass it in, both to prevent resource leaks as well as a package cleanliness point of view (we're trying to remove the `pkg/preparer` dependency in logically isolated packages, like health)

anthonybishopric: ping @csstaub

csstaub: Yeah, will fix. Sorry been busy the last few days.

stouset: @csstaub Bump :)

csstaub:
@anthonybishopric Updated, PTAL.

anthonybishopric:
:shipit:

PR #395, opened 2016-02-24
make kp.Lock into an interface so that tests may mock it

mpuncel:
code using pkg/kp as a library will have a much easier time testing
functionality that relies on locking if the implementation can be
mocked. Furthermore, the underlying (consul) api.Client that the kp.Lock
struct contains is also a struct type and cannot be mocked.

### DIFF
--- a/pkg/kp/lock.go
+++ b/pkg/kp/lock.go
@@ -10,7 +10,14 @@ import (
 	"github.com/square/p2/pkg/util"
 )
 
-type Lock struct {
+type Lock interface {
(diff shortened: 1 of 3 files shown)


rudle:
This is a good idea.

I wonder: does this generalization unblock any upcoming work? What made you notice that this type wasn't an interface?

mpuncel:
I wanted it for testing the lock that prevents duplicate RUs from being created

anthonybishopric:
Seems reasonable :ship:

PR #396, opened 2016-02-24
Add p2-rm command

bwester:
This commit adds the "p2-rm" command, which can be used to remove/unschedule a
pod from a node. It will delete the pod's manifest from intent and remove any
labels attached to the pod. If the node pod is being managed by a controller,
it will print a warning message but allow the deletion.

Sample execution:

    $ p2-rm --node mynode mypod
    warning: mynode/mypod is managed by replication controller: 12345678
             To reduce the controller's capacity, use "p2-rctl set-replicas"
    mynode: successfully removed mypod
    $ p2-rm --node mynode unmanagedpod
    mynode: successfully removed unmanagedpod

### DIFF
--- a/bin/p2-rm/main.go
+++ b/bin/p2-rm/main.go
@@ -1,0 +1,84 @@
+// p2-rm is a command line tool for removing a pods and its labels.
+package main
+
+import (
+	"bufio"
+	"fmt"
+	"os"
+	"path"
+	"strings"
+
+	"github.com/square/p2/Godeps/_workspace/src/gopkg.in/alecthomas/kingpin.v2"
+
+	"github.com/square/p2/pkg/kp"
+	"github.com/square/p2/pkg/kp/flags"
+	"github.com/square/p2/pkg/labels"
+	"github.com/square/p2/pkg/rc"
+	"github.com/square/p2/pkg/types"
+	"github.com/square/p2/pkg/version"
+)
+
+// Command line flags
+var (
+	podName  = kingpin.Arg("pod", "The names of the pod to be removed").Required().String()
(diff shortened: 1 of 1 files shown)


bin/p2-rm/main.go
```diff
@@ -0,0 +1,75 @@
...
+		}
+		*nodeName = hostname
+	}
+
+	// If the node is managed by a controller, let the user know. Changes to the
+	// controller might be needed.
+	labelID := path.Join(*nodeName, *podName)
+	podLabels, err := labeler.GetLabels(labels.POD, labelID)
+	if err != nil {
+		fmt.Fprintf(os.Stderr, "unable to check node for lables: %v\n", err)
```
rudle: typo: lables -> labels

bin/p2-rm/main.go
```diff
@@ -0,0 +1,75 @@
...
+		os.Exit(1)
+	}
+	if podLabels.Labels.Has(rc.RCIDLabel) {
+		fmt.Fprintf(
+			os.Stderr,
+			"warning: %s is managed by replication controller: %v\n"+
+				"         To reduce the controller's capacity, use \"p2-rctl set-replicas\"\n",
+			labelID,
+			podLabels.Labels.Get(rc.RCIDLabel),
+		)
```
rudle: should this halt the program? It could instead confirm that the operation is desired given the state of the world.

bwester: OK, a confirmation could be useful.

rudle:
:+1: I don't think the confirmation is extremely important

bwester:
Updated: fixed typo and added confirmation step

PR #397, opened 2016-02-25
Move farm command from p2-rctl to p2-rctl-server

bwester:
Moves the "p2-rctl farm" command into its own binary, "p2-rctl-server". This
separates the resource controller functionality into two distinct parts: CLI
tools and long-running services. This makes it easier to give end users only
the parts that might be relevant to them.


bwester:
To be clear, the ultimate goal of this PR is to allow `p2-rctl` to be distributed to users as one method for controlling their apps.

mpuncel:
:+1:

anthonybishopric:
:+1:

PR #398, opened 2016-02-27
Refactor session locking out of RC/RU farms

bwester:
Adds the `WithSession()` helper method for working with Consul sessions. Launches a function in a goroutine when a session appears; asks the function to exit when the session dies. This method provides a simple model for processes that require sessions but don't want the added complexity of handling session refreshes.

This helper is used to simplify the main "farm" loops for resource controllers and rolling updates so they don't need cases for lock management. It doesn't save on lines of code, but the new farms have a better separation of concerns and `WithSession()`'s session handling is more robust (incidentally implementing the "TODO: restart the watch so that you get updates right away?" functionality).

This was part of a larger planned refactor/generalization of the resource controller and rolling update farms, but it doesn't look like I'll have the time to complete the whole thing. This piece should make sense on its own.

### DIFF
--- a/pkg/rc/farm.go
+++ b/pkg/rc/farm.go
@@ -4,6 +4,7 @@ import (
 	"github.com/square/p2/Godeps/_workspace/src/github.com/Sirupsen/logrus"
 
 	"github.com/square/p2/pkg/kp"
+	"github.com/square/p2/pkg/kp/consulutil"
 	"github.com/square/p2/pkg/kp/rcstore"
 	"github.com/square/p2/pkg/labels"
 	"github.com/square/p2/pkg/logging"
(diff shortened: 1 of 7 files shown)


mpuncel:
looks great! I want to use this in my branch for RU mutual exclusion

pkg/rc/farm.go
```diff
@@ -72,35 +81,14 @@ START_LOOP:
 	for {
 		select {
 		case <-quit:
-			rcf.logger.NoFields().Infoln("Halt requested, releasing replication controllers")
+			rcf.logger.NoFields().Infoln("Session expired, releasing replication controllers")
+			rcf.lock = nil
 			rcf.releaseChildren()
 			return
-		case session := <-rcf.sessions:
```
anthonybishopric: There doesn't appear to be a way to get a new session ID without quitting this `for` loop. Is the implication that this function will be called once for every new session ID instead of forcing this function to handle the new session case?

bwester: Yes, and that's the main point of this change. I think it's easier to express the intended behavior as a single main loop that dies whenever its one session is lost, compared to being a long-running loop that handles sessions coming and going itself.

anthonybishopric:
Looks good, nice refactor :shipit:

PR #400, opened 2016-03-01
Make health checker properly provide WaitIndex

mpuncel:
The health checker aims to use consul watches to reduce QPS on the
consul instance, however the current index was not properly being
updated after successful queries, so they were taking place every
second.


anthonybishopric:
yep :+1:

bwester:
nice find :+1:

PR #402, opened 2016-03-03
Mpuncel/session unlocker

mpuncel:
This paves the way for the work I'm doing to prevent conflicting RUs from being admitted to consul


pkg/rc/farm.go
```diff
@@ -65,9 +66,9 @@ func NewFarm(
 // Start is not safe for concurrent execution. Do not execute multiple
 // concurrent instances of Start.
 func (rcf *Farm) Start(quit <-chan struct{}) {
-	consulutil.WithSession(quit, rcf.sessions, func(sessionQuit <-chan struct{}, session string) {
-		rcf.logger.WithField("session", session).Infoln("Acquired new session")
-		rcf.lock = rcf.kpStore.NewUnmanagedLock(session, "")
+	consulutil.WithSession(quit, rcf.sessions, func(sessionQuit <-chan struct{}, sessionID string) {
+		rcf.logger.WithField("session", sessionID).Infoln("Acquired new session")
+		rcf.session = rcf.kpStore.NewUnmanagedSession(sessionID, "")
```
mpuncel: The terminology gets a little confusing here... `WithSession()` is creating the underlying consul session, we just want a `kp.Session` that uses it. If anyone has suggestions to increase clarity i'm all ears

petertseng:
I have some nits about commit cleanliness, but they are ultimately just nits. A few other small things as well, but I generally like this direction this is going in.

Might it be the case that e.g. `RollLockPath` can be `RCLockPath` can be deleted on landing this?

petertseng:
given that `lock.go` was renamed to `session.go` is it also appropriate to rename `lock_test` to `session_test`?

mpuncel:
`RollLockPath` and `RCLockPath` are still used in their respective farms, I have a branch that is getting rid of that though (at least the RC one)

petertseng:
:+1:

PR #403, opened 2016-03-03
don't warn if log agent is KILLed

anthonybishopric:
This fixes a case where forcibly terminating the log agent would prevent the `enable` part of the Hoist launchable lifecycle from executing. This allows this case to proceed without error.


mpuncel:
:+1:

PR #404, opened 2016-03-04
Only attempt to unlock an RC if the lock is held

mpuncel:
This is a bug introduced by https://github.com/square/p2/pull/402, in
which the library that handles consul locking was changed. Previously
you could call session.Unlock(key) on a key that you may or may not
hold, and it would not panic the process.

Now, however, session.Lock(key) returns a kp.Unlocker object that is
used to unlock a key. The code fixed in this PR had assumed that it was
okay to call unlock on a key that wasn't held, but this resulted in a
panic if the unlocker was nil

### DIFF
--- a/pkg/roll/run_update.go
+++ b/pkg/roll/run_update.go
@@ -283,18 +283,22 @@ func (u update) lockRCs(done <-chan struct{}) error {
 func (u update) unlockRCs(done <-chan struct{}) {
 	wg := sync.WaitGroup{}
 	for _, unlocker := range []kp.Unlocker{u.newRCUnlocker, u.oldRCUnlocker} {
-		wg.Add(1)
-		go func(unlocker kp.Unlocker) {
-			defer wg.Done()
-			RetryOrQuit(
-				func() error {
-					return unlocker.Unlock()
-				},
-				done,
-				u.logger,
-				fmt.Sprintf("unlocking rc: %s", unlocker.Key()),
-			)
-		}(unlocker)
+		// unlockRCs is called whenever Run() exits, so we have to
+		// handle the case where we didn't lock anything yet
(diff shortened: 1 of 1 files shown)


pkg/roll/run_update.go
```diff
@@ -283,18 +283,22 @@ func (u update) lockRCs(done <-chan struct{}) error {
...
-					return unlocker.Unlock()
-				},
-				done,
-				u.logger,
-				fmt.Sprintf("unlocking rc: %s", unlocker.Key()),
-			)
-		}(unlocker)
+		// unlockRCs is called whenever Run() exits, so we have to
+		// handle the case where we didn't lock anything yet
+		if unlocker != nil {
```
anthonybishopric: Is this the only usage of unlocker right now? Are there other locations we should check?

anthonybishopric:
provisional :ship: per the answer to the unlocker usage

rudle:
seems ok for a bugfix.

I'm behind on code review so I'm not sure what situations lead to a null Unlocker, but I'd like to tighten that up if it's possible.

pkg/roll/run_update.go
mpuncel: It's not the only usage of unlocker, but i'm confident it's the only place where it would be nil

mpuncel:
@rudle:

the `update` struct tracks the unlockers it has, using these fields: https://github.com/square/p2/blob/master/pkg/roll/run_update.go#L35

`unlockRCs` can be called in a variety of states, due to https://github.com/square/p2/blob/master/pkg/roll/run_update.go#L103

Therefore, we have to account for the possibility that those struct fields weren't set yet. Maybe there's a cleaner way to organize this?

PR #405, opened 2016-03-05
Allow RU farms to limit which RCs they operate on.

anthonybishopric:
Roll farms may now take an RC selector to decide whether it should
pick up a particular RU request. This can be used to assist in RU
partitioning of work or to create test environments.

Note that this is _not_ required for RU farms to cooperatively schedule work.

### DIFF
--- a/pkg/roll/farm.go
+++ b/pkg/roll/farm.go
@@ -16,6 +16,8 @@ import (
 	"github.com/square/p2/pkg/rc/fields"
 	roll_fields "github.com/square/p2/pkg/roll/fields"
 	"github.com/square/p2/pkg/util"
+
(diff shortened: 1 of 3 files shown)


pkg/roll/farm.go
```diff
@@ -195,6 +230,18 @@ START_LOOP:
 	}
 }
 
+// test if the farm should work on the given replication controller ID
+func (rlf *Farm) shouldWorkOn(rcID fields.ID) (bool, error) {
+	if rlf.rcSelector.Empty() {
```
mpuncel: empty doesn't just match on its own?

mpuncel:
awesome idea :+1:

pkg/roll/farm.go
anthonybishopric: Yep, but I wanted to eliminate the `GetLabels` query if possible

PR #407, opened 2016-03-09
Add functions to rcstore for various rc locks

mpuncel:
There are 3 kinds of locks that are applied to a replication controller:

1) an "ownership" lock, which should be held by the goroutine that
carries out the intent of the replication controller. These goroutines
are spawned by an rc farm.
2) a "mutation" lock. This should be held by any tool that intends to
modify a replication controller. This could be a commandline tool, or an
rolling update goroutine spawned by the roll farm, for instance.
3) an "update creation" lock. This is not currently used, but is useful
for guaranteeing that no conflicting rolling updates are admitted to the
store. A process trying to schedule a rolling update should first grab
the update creation lock for each replication controller it will operate
on, then confirm there are no rolling updates that refer to either of
the replication controllers, and then create the update and release the
locks.

### DIFF
--- a/pkg/kp/rcstore/consul_store.go
+++ b/pkg/kp/rcstore/consul_store.go
@@ -3,20 +3,25 @@ package rcstore
 import (
 	"encoding/json"
 	"fmt"
+	"path"
 
 	"github.com/square/p2/Godeps/_workspace/src/github.com/hashicorp/consul/api"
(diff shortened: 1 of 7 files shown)


rudle:
This LGTM

anthonybishopric:
:shipit:

PR #408, opened 2016-03-10
roll: conservatively treat unknown-health nodes as healthy; support capacity increase

petertseng:
We wish to be conservative because we don't want to mistakenly take down
a node of unknown health only to find out that it is healthy. The other
advantage of this is that it supports easily migrating apps from other
deployment systems which may not report health.

### DIFF
--- a/pkg/roll/run_update.go
+++ b/pkg/roll/run_update.go
@@ -429,7 +429,7 @@ func (u *update) shouldRollAfterDelay(newFields rcf.RC) (int, int, error) {
 	return afterDelayRemove, afterDelayAdd, nil
 }
 
-func (u *update) rollAlgorithmParams(oldHealth, newHealth rcNodeCounts) (oldHealthy, newHealthy, desired, minHealthy int) {
+func (u *update) rollAlgorithmParams(oldHealth, newHealth rcNodeCounts) (oldHealthy, newHealthy, currentDesired, targetDesired, minHealthy int) {
 	// We conservatively treat Unknown nodes as healthy on the old side.
 	oldHealthy = oldHealth.Healthy + oldHealth.Unknown
 	if oldHealth.Desired < oldHealthy {
@@ -444,17 +444,19 @@ func (u *update) rollAlgorithmParams(oldHealth, newHealth rcNodeCounts) (oldHeal
 		oldHealthy = oldHealth.Desired
 	}
 	newHealthy = newHealth.Healthy
-	desired = u.DesiredReplicas
+	currentDesired = oldHealth.Desired + newHealth.Desired
+	targetDesired = u.DesiredReplicas
 	minHealthy = u.MinimumReplicas
 	return
 }
 
-// the roll algorithm defines how to mutate RCs over time. it takes four args:
+// the roll algorithm defines how to mutate RCs over time. it takes five args:
 // - old: the number of healthy nodes on the old RC
 // - new: the number of healthy nodes on the new RC
-// - desired: the number of nodes desired on the new RC (ie the target)
+// - currentDesired: the number of nodes currently desired in total between the old and new RCs
+// - targetDesired: the number of nodes desired on the new RC (ie the target) in the final state
 // - minHealthy: the number of nodes that must always be up (ie the minimum)
-// given these four arguments, rollAlgorithm returns the number of nodes to add
+// given these five arguments, rollAlgorithm returns the number of nodes to add
 // to the new RC and to delete from the old
 //
 // Returns two values, both of which are a positive number or zero:
@@ -464,17 +466,35 @@ func (u *update) rollAlgorithmParams(oldHealth, newHealth rcNodeCounts) (oldHeal
 // Under the following circumstances, both return values will be zero:
 // - new >= desired (the update is done)
 // - old+new <= minHealthy (at or below the minimum, update has to block)
-func rollAlgorithm(old, new, desired, minHealthy int) (nodesToRemove, nodesToAdd int) {
+func rollAlgorithm(old, new, currentDesired, targetDesired, minHealthy int) (nodesToRemove, nodesToAdd int) {
 	// how much "headroom" do we have between the number of nodes that are
 	// currently healthy, and the number that must be healthy?
 	// if we schedule more than this, we'll go below the minimum
+	// Note that this can go negative (if old + new don't satisfy minHealthy).
 	headroom := old + new - minHealthy
+
+	capacityIncrease := targetDesired - currentDesired
+	if capacityIncrease > 0 {
+		// If we intend to schedule new nodes (nodes not currently managed by either RC),
+		// then we increase headroom by the capacity difference.
+		// Note that headroom could previously have been negative.
+		// This means we will schedule some number between 0 and capacityIncrease.
+		// For example, with old = new = 0, minHealthy = 2, capacityIncrease = 3,
+		// we'll schedule 0 - 2 + 3 = 1 node.
+		// This conservatively respects minimum health,
+		// on the assumption that there may be healthy nodes we don't know about.
+		// This is particularly useful when migrating from other deployment systems.
+		headroom += capacityIncrease
+	} else {
+		// We're not yet sure what the effects of a capacity decrease in a rolling update will be.
+		capacityIncrease = 0
+	}
 
 	// how many nodes do we have left to go? we can't schedule more than this
 	// or we'll go over the target
 	// note that remaining < headroom is possible depending on how many
 	// old nodes are still alive
-	remaining := desired - new
+	remaining := targetDesired - new
 
 	// heuristic time:
 	if remaining <= 0 {
@@ -484,7 +504,7 @@ func rollAlgorithm(old, new, desired, minHealthy int) (nodesToRemove, nodesToAdd
 	if new >= minHealthy {
 		// minimum is satisfied by new nodes, doesn't matter how many old ones
 		// we kill. this includes the edge case where minHealthy==0
-		return remaining, remaining
+		return clampToZero(remaining - capacityIncrease), remaining
 	}
 	if headroom <= 0 {
 		// the case of minHealthy==0 was caught above. in this case, minHealthy>0 and
@@ -494,9 +514,21 @@ func rollAlgorithm(old, new, desired, minHealthy int) (nodesToRemove, nodesToAdd
 		return 0, 0
 	}
 	// remaining>0 and headroom>0. each one imposes an upper bound on how
-	// many we can schedule at once, so return the minimum of the two
+	// many nodes we can add at once, so nodesToAdd is their minimum.
+	// nodesToRemove is the same number minus any capacity increase
+	// (so the net number of nodes added equals the capacity increase)
 	if remaining < headroom {
-		return remaining, remaining
+		nodesToAdd = remaining
+	} else {
+		nodesToAdd = headroom
 	}
-	return headroom, headroom
+	nodesToRemove = clampToZero(nodesToAdd - capacityIncrease)
+	return
+}
+
+func clampToZero(n int) int {
+	if n < 0 {
+		return 0
+	}
+	return n
 }
--- a/pkg/roll/run_update.go
+++ b/pkg/roll/run_update.go
@@ -425,7 +425,8 @@ func (u *update) shouldRollAfterDelay(newFields rcf.RC) (int, error) {
 }
 
 func (u *update) rollAlgorithmParams(oldHealth, newHealth rcNodeCounts) (oldHealthy, newHealthy, desired, minHealthy int) {
-	oldHealthy = oldHealth.Healthy
+	// We conservatively treat Unknown nodes as healthy on the old side.
+	oldHealthy = oldHealth.Healthy + oldHealth.Unknown
 	if oldHealth.Desired < oldHealthy {
 		// Because of the non-atomicity of our KV stores,
 		// we may run into this situation while decrementing old RC's count:
--- a/pkg/roll/run_update.go
+++ b/pkg/roll/run_update.go
@@ -175,8 +175,8 @@ ROLL_LOOP:
 				break
 			}
 
-			next := rollAlgorithm(u.rollAlgorithmParams(oldNodes, newNodes))
-			if next > 0 {
+			nextRemove, nextAdd := rollAlgorithm(u.rollAlgorithmParams(oldNodes, newNodes))
+			if nextRemove > 0 || nextAdd > 0 {
 				// apply the delay only if we've already added to the new RC, since there's
 				// no value in sitting around doing nothing before anything has happened.
 				if newNodes.Desired > 0 && u.RollDelay > time.Duration(0) {
@@ -190,7 +190,7 @@ ROLL_LOOP:
 
 					// determine the new value of `next`, which may have changed
 					// following the delay.
-					next, err = u.shouldRollAfterDelay(newFields)
+					nextRemove, nextAdd, err = u.shouldRollAfterDelay(newFields)
 
 					if err != nil {
 						u.logger.NoFields().Errorln(err)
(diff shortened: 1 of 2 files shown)


anthonybishopric:
Looks good! May want to write a test for `shouldUpdateAfterRoll`, and possibly the main loop if possible.

anthonybishopric:
*`shouldRollAfterDelay`

pkg/roll/run_update.go
```diff
@@ -360,10 +363,18 @@ func (u *update) countHealthy(id rcf.ID, checks map[string]health.Result) (rcNod
...
+			if hres.Status == health.Passing {
+				ret.Healthy++
+			} else if hres.Status != health.Unknown {
+				ret.Unhealthy++
+			}
 		}
 	}
+
+	// We set unknown here to deal with the case of nodes not being in CurrentPods()
+	ret.Unknown = ret.Desired - ret.Healthy - ret.Unhealthy
```
petertseng: check whether this should be `ret.Real - ret.Healthy - ret.Unhealthy` instead. See discussion in comment on test

pkg/roll/update_test.go
```diff
@@ -14,6 +21,138 @@ import (
...
+		"node3": {Status: health.Critical},
+	}
+	counts, err := upd.countHealthy(id, checks)
+	Assert(t).IsNil(err, "expected no error counting health")
+	expected := rcNodeCounts{
+		Desired: 3,
+		Current: 3,
+		Real:    2,
+		Healthy: 2,
+		Unknown: 1,
```
petertseng: in this state, should `Unknown` be 0 or 1? Need to think about this for a bit.

pkg/roll/run_update.go
petertseng: upon further deliberation, I think that would be incorrect. I believe this line should stand as written. A new test case will be added that will need this line to subtract from `Desired`, not `Real`

petertseng:
Need to add a test for the case where number of nodes belonging to the RC is less than the desired number (like the case when converting from a different deployment system)

petertseng:
Along with the extra test to `countHealthy`, tests for `shouldRollAfterUpdate` have been added. It does not appear to be feasible to test the `ROLL_LOOP` at this time (unless an extraction is performed) since it requires significant setup and various channels., but since fundamentally it is using the same `rollAlgorithm(oldNodes.Healthy+oldNodes.Unknown, newNodes.Healthy, u.DesiredReplicas, u.MinimumReplicas)` line as in `shouldRollAfterUpdate`, that means that the behavior I want to test has already been covered and any other tests would be outside the scope of this PR

petertseng:
For your consideration are new commits added to this PR that supports the case of increasing capacity in conjunction with a rolling update.

The use case we have is that when we are converting from a different deployment system, we create two RCs, both with desired replicas 0. Then we increase capacity with a rolling update. Details in the individual commit messages.

pkg/roll/update_test.go
```diff
@@ -6,7 +6,15 @@ import (
 	"testing"
```
mpuncel: :bow: thanks for being rigorous about this

pkg/roll/update_test.go
```diff
@@ -14,15 +22,27 @@ import (
...
+	Assert(t).AreEqual(uniformRollAlgorithm(t, 2, 2, 6, 3), 1, "should schedule difference if minimum must be maintained")
+	Assert(t).AreEqual(uniformRollAlgorithm(t, 1, 3, 6, 3), 3, "should schedule remaining if minimum is satisfied by new")
+	Assert(t).AreEqual(uniformRollAlgorithm(t, 3, 0, 6, 0), 6, "should schedule remaining if no minimum")
+}
 
-	Assert(t).AreEqual(rollAlgorithm(2, 2, 6, 3), 1, "should schedule difference if minimum must be maintained")
-	Assert(t).AreEqual(rollAlgorithm(1, 3, 6, 3), 3, "should schedule remaining if minimum is satisfied by new")
-	Assert(t).AreEqual(rollAlgorithm(3, 0, 6, 0), 6, "should schedule remaining if no minimum")
+func TestRollAlgorithmIncreases(t *testing.T) {
+	//Assert(t).AreEqual(rollAlgorithm(0, 0, 0, 3, 2), 1, "should schedule difference if increasing capacity from zero")
```
mpuncel: are you intentionally leaving these around or should they be uncommented?

petertseng: oops. those should be uncommented. just a moment.

pkg/roll/update_test.go
```diff
@@ -138,19 +158,389 @@ func SimulateRollingUpgradeDisable(t *testing.T, full, nonew bool) {
...
 		// choose nodes from the eligible list, randomly, and put the new pod
 		// on them
-		for _, index := range rand.Perm(len(eligible))[:nextUpdate] {
+		for _, index := range rand.Perm(len(eligible))[:nextAdd] {
 			nodes[eligible[index]] = 1
 		}
 	}
 }
+
+type fakeHealthChecker struct {
```
mpuncel: can you put this in a package like `pkg/health/checker/checkertest` or something like that? looks like it's generic enough to be useful (and people can implement what they need as they need it)

pkg/roll/update_test.go
```diff
@@ -14,15 +22,27 @@ import (
...
 
+func uniformRollAlgorithm(t *testing.T, old, new, want, need int) int {
+	remove, add := rollAlgorithm(old, new, want, want, need)
+	Assert(t).AreEqual(remove, add, "expected nodes removed and nodes added to be equal")
+	return add
+}
+
 func TestWouldBlock(t *testing.T) {
 	// in the following cases, rollAlgorithm should return 0
-	Assert(t).AreEqual(rollAlgorithm(1, 1, 3, 4), 0, "should do nothing if below minimum")
```
petertseng: this test was accidentally removed and was not readded. should fix that.

pkg/roll/update_test.go
petertseng: to my slight shame, there appears to already be one. I may simply add mine to this

petertseng: oh wait, no, that was `checker_test.go` instead of `checkertest` package.

pkg/roll/update_test.go
petertseng: changed to real tests

pkg/roll/update_test.go
petertseng: readded.

pkg/roll/update_test.go
petertseng: added in 3c5192ce10c0ad0e93fed3b2a7515c48c0aa3a53 though I left a TODO that replication tests should probably put some stuff in there. maybe I'll just do that now instead of leaving the TODO

petertseng:
splitting c376de4094ca0775f82b0ddd060217ad823a30dd into a commit for tests added merely to test existing behavior to make them separate from the tests that test new behavior. no code changes planned

pkg/roll/run_update.go
anthonybishopric: Why set `ret.Unknown` to this value instead of the sum of all healthchecks that `== health.Unknown`? Are we guarding against unknown unknowns?

pkg/roll/run_update.go
```diff
@@ -360,84 +372,119 @@ func (u *update) countHealthy(id rcf.ID, checks map[string]health.Result) (rcNod
...
 	if err != nil {
-		return 0, util.Errorf("Could not determine old service health: %v", err)
+		return 0, 0, util.Errorf("Could not determine old service health: %v", err)
 	}
 
-	afterDelayNext := rollAlgorithm(afterDelayOld.Healthy, afterDelayNew.Healthy, u.DesiredReplicas, u.MinimumReplicas)
+	// We conservatively treat Unknown nodes as healthy on the old side.
+	afterDelayRemove, afterDelayNext := rollAlgorithm(
+		afterDelayOld.Healthy+afterDelayOld.Unknown, afterDelayNew.Healthy,
+		afterDelayOld.Desired+afterDelayNew.Desired, u.DesiredReplicas,
```
anthonybishopric: Should these summary objects be parameters to `rollAlgorithm`? I ask because the decision to sum the health and unknown is copied to multiple locations in this file, and maybe we want tests to verify that it does the right thing.

pkg/roll/run_update.go
petertseng: We are. Should I take your question as an indication that the comment on the line above needs to be made clearer?

anthonybishopric: More that I'd like to know why we have unknown unknowns. We should be explicit if we're guarding against something. Improving the comment to state exactly what cases we are trying to defend against would be a good first step.

pkg/roll/run_update.go
```diff
@@ -360,84 +372,119 @@ func (u *update) countHealthy(id rcf.ID, checks map[string]health.Result) (rcNod
...
+//
+// Under the following circumstances, both return values will be zero:
 // - new >= desired (the update is done)
 // - old+new <= minHealthy (at or below the minimum, update has to block)
-func rollAlgorithm(old, new, desired, minHealthy int) int {
+func rollAlgorithm(old, new, currentDesired, targetDesired, minHealthy int) (int, int) {
 	// how much "headroom" do we have between the number of nodes that are
 	// currently healthy, and the number that must be healthy?
 	// if we schedule more than this, we'll go below the minimum
 	headroom := old + new - minHealthy
```
petertseng: add comment that this can go negative! and then `headroom += capacityIncrease` may increase that to just above positive, not the full value of `capacityIncrease`

petertseng: A comment has been added in https://github.com/square/p2/pull/408/commits/01913a9c18a1733e765f336c565ad659e4fee426

petertseng: added in aeb5b1fc8ab45241a64980efeaf77bfe5b9673fe

pkg/roll/run_update.go
petertseng: b627695084323a14eaccd60ae403b4fddf225ba7 consolidates the logic in one place

pkg/roll/run_update.go
petertseng: 0d88762ba6967c3e87c8e211404b28541a598233 now makes explicit the various cases where `Unknown` would get incremented. Take a look.

petertseng:
I believe that addresses the current comments (responses inline on the respective comments)

pkg/roll/run_update.go
```diff
@@ -332,6 +334,13 @@ func (u *update) countHealthy(id rcf.ID, checks map[string]health.Result) (rcNod
 	}
 	ret.Current = len(currentPods)
 
+	if ret.Desired > ret.Current {
+		// This implies that the RC hasn't yet scheuled pods that it desires to have.
```
petertseng: s/scheuled/scheduled

pkg/health/checker/test/fake_checker.go
```diff
@@ -0,0 +1,39 @@
+package test
```
mpuncel: thanks for doing this

pkg/roll/run_update.go
```diff
@@ -398,6 +398,10 @@ func (u *update) shouldRollAfterDelay(newFields rcf.RC) (int, error) {
 	return afterDelayNext, nil
 }
 
+func (u *update) rollAlgorithmParams(oldHealth, newHealth rcNodeCounts) (int, int, int, int) {
```
mpuncel: crafty

pkg/roll/run_update.go
```diff
@@ -399,7 +399,8 @@ func (u *update) shouldRollAfterDelay(newFields rcf.RC) (int, error) {
 }
 
 func (u *update) rollAlgorithmParams(oldHealth, newHealth rcNodeCounts) (int, int, int, int) {
```
mpuncel: nit: name the return parameters so it's more obvious what is happening now that there is some logic here

pkg/roll/run_update.go
```diff
@@ -423,17 +426,35 @@ func (u *update) rollAlgorithmParams(oldHealth, newHealth rcNodeCounts) (int, in
 // Under the following circumstances, both return values will be zero:
 // - new >= desired (the update is done)
 // - old+new <= minHealthy (at or below the minimum, update has to block)
-func rollAlgorithm(old, new, desired, minHealthy int) (int, int) {
+func rollAlgorithm(old, new, currentDesired, targetDesired, minHealthy int) (int, int) {
```
mpuncel: nit: name the return params here too

pkg/roll/run_update.go
```diff
@@ -455,7 +476,14 @@ func rollAlgorithm(old, new, desired, minHealthy int) (int, int) {
 	// remaining>0 and headroom>0. each one imposes an upper bound on how
 	// many we can schedule at once, so return the minimum of the two
```
mpuncel: this comment seems irrelevant now

mpuncel:
looks great, just some nits.

Thanks for writing comprehensive tests!!

pkg/roll/run_update.go
petertseng: Hmm, I think it is still relevant, though it's only accurate for nodes _added_. I'll see what I can do to bring it up to date.

mpuncel: err yeah irrelevant was a strong word but it seems to suggest there is only one return value

petertseng:
Named return variables and clarified the comment.

```diff
diff --git a/pkg/roll/run_update.go b/pkg/roll/run_update.go
index 7071aad..c6ef237 100644
--- a/pkg/roll/run_update.go
+++ b/pkg/roll/run_update.go
@@ -403,11 +403,13 @@ func (u *update) shouldRollAfterDelay(newFields rcf.RC) (int, int, error) {
        return afterDelayRemove, afterDelayAdd, nil
 }

-func (u *update) rollAlgorithmParams(oldHealth, newHealth rcNodeCounts) (int, int, int, int, int) {
-       // We conservatively treat Unknown nodes as healthy on the old side.
-       return oldHealth.Healthy + oldHealth.Unknown, newHealth.Healthy,
-               oldHealth.Desired + newHealth.Desired, u.DesiredReplicas,
-               u.MinimumReplicas
+func (u *update) rollAlgorithmParams(oldHealth, newHealth rcNodeCounts) (oldHealthy, newHealthy, currentDesired, targetDesired, minHealthy int) {
+       oldHealthy = oldHealth.Healthy + oldHealth.Unknown
+       newHealthy = newHealth.Healthy
+       currentDesired = oldHealth.Desired + newHealth.Desired
+       targetDesired = u.DesiredReplicas
+       minHealthy = u.MinimumReplicas
+       return
 }

 // the roll algorithm defines how to mutate RCs over time. it takes five args:
@@ -426,7 +428,7 @@ func (u *update) rollAlgorithmParams(oldHealth, newHealth rcNodeCounts) (int, in
 // Under the following circumstances, both return values will be zero:
 // - new >= desired (the update is done)
 // - old+new <= minHealthy (at or below the minimum, update has to block)
-func rollAlgorithm(old, new, currentDesired, targetDesired, minHealthy int) (int, int) {
+func rollAlgorithm(old, new, currentDesired, targetDesired, minHealthy int) (nodesToRemove, nodesToAdd int) {
        // how much "headroom" do we have between the number of nodes that are
        // currently healthy, and the number that must be healthy?
        // if we schedule more than this, we'll go below the minimum
@@ -474,11 +476,16 @@ func rollAlgorithm(old, new, currentDesired, targetDesired, minHealthy int) (int
                return 0, 0
        }
        // remaining>0 and headroom>0. each one imposes an upper bound on how
-       // many we can schedule at once, so return the minimum of the two
+       // many nodes we can add at once, so nodesToAdd is their minimum.
+       // nodesToRemove is the same number minus any capacity increase
+       // (so the net number of nodes added equals the capacity increase)
        if remaining < headroom {
-               return clampToZero(remaining - capacityIncrease), remaining
+               nodesToAdd = remaining
+       } else {
+               nodesToAdd = headroom
        }
-       return clampToZero(headroom - capacityIncrease), headroom
+       nodesToRemove = clampToZero(nodesToAdd - capacityIncrease)
+       return
 }

 func clampToZero(n int) int {
```

Where I deemed it appropriate to do so, I assigned variables to the named return values rather than using `return a, b, c`. The tests would have had my back if I had gotten it wrong

petertseng:
I lost the comment. Putting it back.

petertseng:
Rebased on master so as to make an upcoming test actually indicative of the state of the world

petertseng:
tested in staging, correctly brings up a new service from capacity 0 -> 2

mpuncel:
ship it!

petertseng:
scheduled for ship after all problems revealed by https://github.com/square/p2/pull/414 have been fixed simply because I would much rather only try to reason about one change at once.

I may pull off some changes from this PR and use them in an upcoming bugfix.

petertseng:
Rebase on master was necessary because the first four commits of this have been stolen by #426. I'll test in staging again for this reason, but I consider the previous approval to still be sufficient simply because the remaining three commits on this PR have not changed (except as necessary to deal with #426 's new behavior)

petertseng:
If there is one place where more attention should be given, it's now that #426 interacts with 0ffe7f9987a432451a241fb7a394762c22848a3a in the following manner:

old healthy is now the minimum of (old desired, old healthy + old unknown).

I believe this is reasonable behavior.

mpuncel:
still looks good :+1:

anthonybishopric:
:shipit:

PR #409, opened 2016-03-10
When an RC schedules nodes, use a sorted list

mpuncel:
This solves a use case in which a deployer wants to know which node will
be scheduled first by a new RC.

### DIFF
--- a/pkg/rc/replication_controller.go
+++ b/pkg/rc/replication_controller.go
@@ -201,18 +201,22 @@ func (rc *replicationController) addPods(current PodLocations) error {
 	// TODO: With Docker or runc we would not be constrained to running only once per node.
 	// So it may be the case that we need to make the Scheduler interface smarter and use it here.
 	possible := sets.NewString(eligible...).Difference(sets.NewString(currentNodes...))
+
+	// Users want deterministic ordering of nodes being populated to a new
+	// RC. Move nodes in sorted order by hostname to achieve this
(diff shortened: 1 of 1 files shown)


pkg/rc/replication_controller.go
```diff
@@ -201,18 +201,22 @@ func (rc *replicationController) addPods(current PodLocations) error {
 	// TODO: With Docker or runc we would not be constrained to running only once per node.
 	// So it may be the case that we need to make the Scheduler interface smarter and use it here.
 	possible := sets.NewString(eligible...).Difference(sets.NewString(currentNodes...))
+
+	// Users want deterministic ordering of nodes being populated to a new
+	// RC. Move nodes in sorted order by hostname to achieve this
+	possibleSorted := possible.List()
```
anthonybishopric: How are you guaranteed this is sorted?

mpuncel: Read the source https://github.com/kubernetes/kubernetes/blob/master/pkg/util/sets/string.go#L168.
Should I sort again on top of that in case that code changes?

mpuncel: to be clear that was "[i] read the source", not me barking an order at you :)

anthonybishopric: Oh ha, no worries :) I just didn't know it was guaranteed to be sorted. :+1:

anthonybishopric:
:ship:

PR #410, opened 2016-03-11
Mpuncel/simpler cas ru put

mpuncel:
I still need to add tests for `CreateRollingUpdateFromOneMaybeExistingWithLabelSelector`


mpuncel:
also this will likely be much easier to review commit-by-commit

pkg/kp/rollstore/consul_store.go
```diff
@@ -0,0 +1,556 @@
...
+		"RU create lock for %s",
+		hostname,
+	)
+
+	return s.store.NewSession(lockName, nil)
+}
+
+// Obtains a lock for each RC in the list, or errors. RCs are locked in
+// lexicographical order by RC id to avoid deadlocks. Returns a function that
+// should be called to release the locks
```
anthonybishopric: I don't think it returns a function

pkg/kp/rollstore/consul_store.go
```diff
@@ -0,0 +1,555 @@
...
+		for _, id := range rcIDs {
+			if ru.NewRC == id || ru.OldRC == id {
+				conflictingID = ru.ID()
+				break
+			}
+		}
+	}
+
+	if conflictingID != "" {
+		return util.Errorf("RC %s conflicts with the request", conflictingID)
```
anthonybishopric: This might be an error worth structuring more than just an error message, particularly since it will be shown to developers when deploy requests fail. It'd be nice to indicate which RU is causing the conflict, for instance.

pkg/kp/rollstore/consul_store.go
```diff
@@ -0,0 +1,555 @@
...
+			"Can't create update: old RC selector %s was ambiguous, provided %d matches",
+			oldRCSelector.String(),
+			len(matches),
+		)
+	}
+
+	var oldRCID rc_fields.ID
+	if len(matches) == 1 {
+		oldRCID = rc_fields.ID(matches[0].ID)
+	} else {
```
anthonybishopric: Should handle the ` > 1` case as an error probably.

pkg/kp/rollstore/consul_store.go
```diff
@@ -0,0 +1,555 @@
...
+		if leaveOld {
```
anthonybishopric: Is it ever possible that someone would want to set `leaveOld = true` if an old RC is found?

pkg/kp/rollstore/consul_store.go
```diff
@@ -0,0 +1,555 @@
...
+	}
+
+	// Check for updates that exist that operate on the old RC
+	err = s.checkForConflictingUpdates(rc_fields.IDs{oldRCID})
+	if err != nil {
+		return roll_fields.Update{}, err
+	}
+
+	// Create the new RC
+	rc, err := s.rcstore.Create(newRCManifest, newRCNodeSelector, newRCPodLabels)
```
anthonybishopric: In the previous function you attempted to drain the `sessionRenewalErr` channel before doing this, why not here?

pkg/kp/rollstore/consul_store.go
```diff
@@ -0,0 +1,555 @@
...
+		if err == nil {
+			err = util.Errorf("Cannot create ru because session was destroyed")
+		}
+		return u, err
+	default:
+		success, _, err := s.kv.CAS(&api.KVPair{
+			Key:   key,
+			Value: b,
+			// it must not already exist
+			ModifyIndex: 0,
```
anthonybishopric: Won't it be zero by default?

pkg/kp/rollstore/consul_store_test.go
```diff
@@ -0,0 +1,720 @@
...
+func testManifest() pods.Manifest {
+	builder := pods.NewManifestBuilder()
+	builder.SetID("slug")
+	return builder.GetManifest()
+}
+
+func testNodeSelector() klabels.Selector {
+	return klabels.Everything().
+		Add("some_key", klabels.EqualsOperator, []string{"some_value"})
+}
```
anthonybishopric: :star: for this whole file

bin/p2-rctl/main.go
```diff
@@ -335,7 +335,7 @@ LOOP:
 }
 
 func (r RCtl) ScheduleUpdate(oldID, newID string, want, need int) {
-	err := r.rls.Put(roll_fields.Update{
+	_, err := r.rls.CreateRollingUpdateFromExistingRCs(roll_fields.Update{
```
anthonybishopric: Future request: it'd be great to get all three creation methods into `p2-rctl` as part of the CLI.

pkg/kp/rollstore/consul_store.go
mpuncel: it's above up above, but probably putting the whole chain together would make that clearer

pkg/kp/rollstore/consul_store.go
mpuncel: I can't think of why you would. In our use cases, when doing a first time RC deploy, the old RC has the same pod manifest as the new one, so I can't imagine you'd want to leave that around.

pkg/kp/rollstore/consul_store.go
mpuncel: I can remove that, but then it might be less clear why it's a `CAS()`

mpuncel: although it probably doesn't even need to be a `CAS()` now that i'm checking the rest of it now that I think about it

anthonybishopric:
Nice work thinking through these cases. So far I can't think of a scenario in which you have the new RC's ID and need to create the old one via the selector, but plenty of scenarios where I'll use the three methods you've added here.

Definitely interested in how this will affect our server code :)

pkg/kp/rollstore/consul_store.go
mpuncel: good call

pkg/kp/rollstore/consul_store.go
mpuncel: I guess the caller might lazily set `leaveOld` without checking if there even is an old, although I'm not sure what the desired outcome would be

pkg/kp/rollstore/consul_store.go
mpuncel: actually going to leave it if for no reason other than that's what we do in other stores (like rcstore) and there's no significant additional cost do doing a CAS operation. I'll remove the modifyindex explicit setting though

anthonybishopric: By virtue of being CAS, it should be clear that something else can't have created it already (hopefully that's clear to others at least)

mpuncel:
addressed comments so far

anthonybishopric:
:ship:

PR #411, opened 2016-03-14
Recover from panics in a goroutine spawned by a farm

mpuncel:
A farm will spawn many goroutines that are activated by changes in
consul data, and it is common to run multiple farms on a server (e.g.
rc farm and roll farm). Since consul data is not transactional, there
have been repeated issues of null pointer dereferences in farm
goroutines crashing the entire process.

While pointers are now pessimistically dereferenced to avoid these
issues, it is probably good to have an additional level of protection to
stop bad data in consul from grinding the farms to a halt if a nil
pointer dereference is missed.


mpuncel:
gonna reopen this pull request because i accidentally created a phantom travis build :(

PR #413, opened 2016-03-14
Recover from panics in a goroutine spawned by a farm

mpuncel:
A farm will spawn many goroutines that are activated by changes in
consul data, and it is common to run multiple farms on a server (e.g.
rc farm and roll farm). Since consul data is not transactional, there
have been repeated issues of null pointer dereferences in farm
goroutines crashing the entire process.

While pointers are now pessimistically dereferenced to avoid these
issues, it is probably good to have an additional level of protection to
stop bad data in consul from grinding the farms to a halt if a nil
pointer dereference is missed.

### DIFF
--- a/pkg/rc/farm.go
+++ b/pkg/rc/farm.go
@@ -1,6 +1,8 @@
 package rc
 
 import (
+	"sync"
+
 	"github.com/square/p2/Godeps/_workspace/src/github.com/Sirupsen/logrus"
 
 	"github.com/square/p2/pkg/kp"
@@ -9,6 +11,7 @@ import (
 	"github.com/square/p2/pkg/labels"
 	"github.com/square/p2/pkg/logging"
 	"github.com/square/p2/pkg/rc/fields"
+	"github.com/square/p2/pkg/util"
 )
 
 // The Farm is responsible for spawning and reaping replication controllers
@@ -27,6 +30,7 @@ type Farm struct {
 	sessions <-chan string
 
 	children map[fields.ID]childRC
+	childMu  sync.Mutex
 	session  kp.Session
 
 	logger logging.Logger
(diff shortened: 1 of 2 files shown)


pkg/rc/farm.go
```diff
@@ -138,24 +142,42 @@ START_LOOP:
...
 					}
 
-					// NOTE: if WatchDesires experiences an unrecoverable error, we don't release the replication controller.
-					// However, it is unlikely that another farm instance would fare any better so that's okay
-				}()
+					// Release the child so that another farm can reattempt
+					rcf.childMu.Lock()
+					defer rcf.childMu.Unlock()
+					if _, ok := foundChildren[id]; !ok {
+						rcf.releaseChild(id)
```
anthonybishopric: call `releaseDeletedChildren`?

mpuncel: i'm releasing a child that was NOT deleted from consul, due to a recovered panic

anthonybishopric: Sure, just noting that the only code difference is that there's a logging line and that it takes a map

anthonybishopric:
Seems good - I do wonder if there was some reason (code or otherwise) for not releasing child locks in a panic situation. I assume it's also possible to panic when releasing locks.

mpuncel:
I think it was probably just thought not to matter, like why release a lock on a corrupt piece of data?

Also panicking would implicitly release any locks due to invalidating the session.

I think the risk of panicking when releasing a lock is low if not zero, the main worry would be nil pointer exceptions

pkg/rc/farm.go
mpuncel: i'm realizing this is a mistake, I should be checking `rcf.children` for it not `foundChildren`. Of course it's in the latter, because we found it and attempted to operate on it.

mpuncel:
rebased, which involved changing a bunch of `rc_fields.ID` types to `roll_fields.ID`

anthonybishopric:
:ship:

PR #414, opened 2016-03-14
roll: Finish update when new healthy > minimum requested

petertseng:
This removes the old block for healthy nodes that was effectively a
reimplementation of slice divisors. This check is unnecessary because
`rollAlgorithm` already ensures safe operation. Removing it has the
benefit of allowing deploys to proceed with more parallelism - if one
node is slow to become healthy but there is health headroom, the deploy
can proceed to other nodes in the meantime. This was not possible with
the block in place.

The other benefit is that in the case of a deploy where minimum health
requested is zero, the deploy terminates as soon as all nodes have
converged, without waiting for nodes to become healthy. This is useful
in the case where we are knowingly deploying a bad sha, for which nodes
will never become healthy.

### DIFF
--- a/pkg/roll/run_update.go
+++ b/pkg/roll/run_update.go
@@ -152,24 +152,19 @@ ROLL_LOOP:
 				break
 			}
 
-			if newNodes.Desired > newNodes.Healthy {
-				// we assume replication controllers do in fact "work", ie healthy
-				// and current converge towards desired
-				// in this case, they have not yet met, so we block
-				u.logger.WithFields(logrus.Fields{
-					"old": oldNodes,
-					"new": newNodes,
-				}).Debugln("Blocking for more healthy new nodes")
-				break
-			}
-			if newNodes.Desired >= u.DesiredReplicas {
-				// note that we only exit the loop AFTER the desired nodes
-				// become healthy - this ensures that, when the old RC is
-				// cleaned up, we do not accidentally remove the nodes we just
-				// assigned to the new RC
-				// normal execution of this rollAlgorithm will never cause
-				// newDesired > u.Desired, but if Run is resuming from an unusual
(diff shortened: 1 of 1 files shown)


pkg/roll/run_update.go
```diff
@@ -152,17 +152,7 @@ ROLL_LOOP:
...
-				// in this case, they have not yet met, so we block
-				u.logger.WithFields(logrus.Fields{
-					"old": oldNodes,
-					"new": newNodes,
-				}).Debugln("Blocking for more healthy new nodes")
-				break
-			}
-			if newNodes.Desired >= u.DesiredReplicas {
+			if newNodes.Desired >= u.DesiredReplicas && newNodes.Healthy >= u.MinimumReplicas {
 				// note that we only exit the loop AFTER the desired nodes
```
petertseng: slight trepidation - I do not fully understand this comment and whether it's still applicable. Stand by while I read and decide.

mpuncel: i also don't understand it

petertseng: I think my trouble is I don't understand when this thing would happen. Under what circumstances would cleaning up an old RC remove nodes from the new RC? Why are they related at all?

petertseng: there's a second sentence beginning with "normal execution of this rollAlgorithm..." which is just explaining why the condition is `newNodes.Desired >= u.DesiredReplicas` rather than `>`, that part I understand.

petertseng: I'm going to update the comment to reflect my current understanding of the situation.

petertseng:
Comment is updated.

mpuncel:
:+1: thanks for updating the comment too

PR #415, opened 2016-03-15
Implement basic finish functionality

rudle:
Here's a first cut at configurable finish executables for our runit integration. Pay special attention to the changes to to servicebuilder.go wrt the process _shell_ for the call to finish (keeping in mind that the app's env will also be present at this time).

This PR assumes we will ship one version of finish for all pods. I could imagine a world in which pod's could override finish but I can't imagine the use case just yet.

I hope to write some more tests for this - though I'm hoping to get feedback on the direction first. I've verified that this works in a testing environment.

### DIFF
--- a/pkg/p2exec/p2_exec.go
+++ b/pkg/p2exec/p2_exec.go
@@ -38,5 +38,9 @@ func (args P2ExecArgs) CommandLine() []string {
 		cmd = append(cmd, "-w", args.WorkDir)
 	}
 
+	if len(cmd) > 0 {
+		cmd = append(cmd, "--")
+	}
+
 	return append(cmd, args.Command...)
 }
--- a/pkg/pods/pod.go
+++ b/pkg/pods/pod.go
@@ -15,6 +15,7 @@ import (
 	"github.com/square/p2/pkg/launch"
 	"github.com/square/p2/pkg/logging"
 	"github.com/square/p2/pkg/opencontainer"
+	"github.com/square/p2/pkg/p2exec"
 	"github.com/square/p2/pkg/runit"
 	"github.com/square/p2/pkg/types"
 	"github.com/square/p2/pkg/uri"
@@ -35,9 +36,10 @@ var (
 )
 
 const DEFAULT_PATH = "/data/pods"
+const DefaultP2Exec = "/usr/local/bin/p2-exec"
 
-var DefaultP2Exec = "/usr/local/bin/p2-exec"
-var defaultLogExec = []string{"chpst", "-unobody", "svlogd", "-tt", "./main"}
+var DefaultFinishExec = []string{"/bin/true"} // type must match preparerconfig
+var DefaultLogExec = []string{"svlogd", "-tt", "./main"}
 
 func init() {
 	Log = logging.NewLogger(logrus.Fields{})
@@ -55,7 +57,8 @@ type Pod struct {
 	ServiceBuilder *runit.ServiceBuilder
 	P2Exec         string
 	DefaultTimeout time.Duration // this is the default timeout for stopping and restarting services in this pod
-	LogExec        runit.LogExec // this is exposed to support a gradual rollout of p2-log-bridge
+	LogExec        runit.Exec
+	FinishExec     runit.Exec
 }
 
 func NewPod(id types.PodID, path string) *Pod {
@@ -67,7 +70,8 @@ func NewPod(id types.PodID, path string) *Pod {
 		ServiceBuilder: runit.DefaultBuilder,
 		P2Exec:         DefaultP2Exec,
 		DefaultTimeout: 60 * time.Second,
-		LogExec:        defaultLogExec,
+		LogExec:        DefaultLogExec,
+		FinishExec:     DefaultFinishExec,
 	}
 }
 
@@ -248,8 +252,9 @@ func (pod *Pod) buildRunitServices(launchables []launch.Launchable, newManifest 
 				return util.Errorf("Duplicate executable %q for launchable %q", executable.Service.Name, launchable.ServiceID())
 			}
 			sbTemplate[executable.Service.Name] = runit.ServiceTemplate{
-				Log: pod.LogExec,
-				Run: executable.Exec,
+				Log:    pod.LogExec,
+				Run:    executable.Exec,
+				Finish: pod.FinishExecForLaunchable(launchable),
 			}
 		}
 	}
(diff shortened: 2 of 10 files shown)


pkg/p2exec/p2_exec.go
```diff
@@ -38,5 +38,9 @@ func (args P2ExecArgs) CommandLine() []string {
 		cmd = append(cmd, "-w", args.WorkDir)
 	}
 
+	if len(cmd) > 0 {
+		cmd = append(cmd, "--")
```
anthonybishopric: :star:

pkg/pods/pod.go
```diff
@@ -67,7 +70,6 @@ func NewPod(id types.PodID, path string) *Pod {
 		ServiceBuilder: runit.DefaultBuilder,
 		P2Exec:         DefaultP2Exec,
 		DefaultTimeout: 60 * time.Second,
-		LogExec:        defaultLogExec,
```
anthonybishopric: Why not pass defaults anymore?

pkg/pods/pod.go
```diff
@@ -593,6 +596,34 @@ func (pod *Pod) Launchables(manifest Manifest) ([]launch.Launchable, error) {
...
+		runAs = string(pod.Id)
+	}
+
+	p2ExecArgs := p2exec.P2ExecArgs{
+		Command: finishExec,
+		User:    runAs,
+		EnvDirs: []string{pod.EnvDir()},
+	}
+
+	pod.FinishExec = append([]string{pod.P2Exec}, p2ExecArgs.CommandLine()...)
```
anthonybishopric: We'll need to write a `finish` for each launchable, it may be sensible to let the launchable's environment also be part of each one's respective exec

pkg/pods/pod_test.go
```diff
@@ -19,6 +19,8 @@ import (
 	"github.com/square/p2/pkg/types"
 	"github.com/square/p2/pkg/util"
 
+	"reflect"
```
anthonybishopric: That instantaneous feeling of dread when I get to a file with a `+     "reflect"`

pkg/pods/pod.go
anthonybishopric: I think we'll always want to provide defaults here. Not all `Pod`s are constructed by the preparer / in `orchestrate.go`. As it stands, you can't just build a default `Pod` and safely launch it, whereas previously you could.

pkg/runit/servicebuilder.go
```diff
@@ -83,13 +85,15 @@ __END__
...
-func (s ServiceTemplate) LogScript() ([]byte, error) {
+func (s ServiceTemplate) logScript() ([]byte, error) {
 	sleep := 2
 	if s.LogSleep != nil && *s.LogSleep >= 0 {
 		sleep = *s.LogSleep
 	}
 
 	log := s.Log
+	// TODO this leaks the type information :<
+	// should we implement empty? on this type
```
anthonybishopric: Seems like a decent idea

pkg/runit/servicebuilder.go
```diff
@@ -113,6 +117,19 @@ __END__
 	return []byte(ret), nil
 }
 
+func (s ServiceTemplate) finishScript() ([]byte, error) {
+	finish_exec := s.Finish
+	if len(finish_exec) == 0 {
+		finish_exec = []string{"/bin/true", "# finish not implemented"}
```
anthonybishopric: Does the bash comment work with `os.Exec`?

anthonybishopric: I'm guessing not, but nevermind because it's just interpolated into a real bash script below

rudle: The template is evaluated into a file on the disk that will be invoked by runit. I don't see the connection to the `os` package.

anthonybishopric:
Great - just a few suggestions, otherwise I'm excited for this.

pkg/pods/pod.go
```diff
@@ -593,6 +596,34 @@ func (pod *Pod) Launchables(manifest Manifest) ([]launch.Launchable, error) {
 	return launchables, nil
 }
 
+func (pod *Pod) SetFinishExec(finishExec []string) {
+	var runAs string
+
+	if manifest, err := pod.CurrentManifest(); err == nil {
```
mpuncel: what does it mean to have an `err` here? does that mean the pod is being deployed for the first time? seems weird that the user would be different on a first time deploy than a second time deploy.

pkg/pods/pod.go
```diff
@@ -593,6 +596,34 @@ func (pod *Pod) Launchables(manifest Manifest) ([]launch.Launchable, error) {
...
+		runAs = manifest.RunAsUser()
```
mpuncel: I wonder if we really want finish scripts running as root?

pkg/pods/pod.go
anthonybishopric: Good catch. This is definitely strange - if the `run_as` were to change between deploys, we don't want to be caught with the _last_ deploy's user. Most functions to `Pod` also pass the manifest we're operating on, why not pass it here too?

pkg/pods/pod.go
anthonybishopric: If the pod is allowed to, why wouldn't we?

mpuncel: sure, i guess it would have to be the pod's responsibility to drop privileges in the finish script just like the other launchables need to

anthonybishopric: For now, and for the indefinite future, the `finish` program is only configurable by preparer config. That makes it safer than arbitrary user code - but yes, if they were already running as root, this problem exists no matter what.

pkg/runit/servicebuilder.go
rudle: The line below is a bit more disconcerting wrt type abstraction. I'm not sure this TODO is worth addressing - we're getting a lot of benefits from the type checks at function boundaries, adding any more feels like contortionism.

rudle:
One open question: why not run the finish script as `nobody`? I don't think there's any reason that the finish script needs to access the apps secrets and using the Pod + Launchable environments should provide us enough context to generate some useful metrics.

Bonus is that the code gets a bit simpler if we use the `nobody` user.

anthonybishopric:
Well, in our case, secrets for the app won't be accessible to `nobody`. But you will get plenty of information from the environment variables. That may be sufficient for what we plan on doing.

rudle:
I'll do that then. If it turns out we need access to secrets I can implement something similar to what we have here now.

rudle:
I think this is ready for another (final?) review.

mpuncel:
ship it!

PR #417, opened 2016-03-15
Make update.LockRCs have pointer receiver in pkg/roll/run_update.go

mpuncel:
If it's a value receiver instead of pointer receiver, the struct fields
for the kp.Unlockers don't get updated, and we never end up unlocking
the held locks.


petertseng:
:+1:

petertseng:
thanks for adding test! Also :+1:

PR #418, opened 2016-03-16
Mpuncel/label rc ru creation

mpuncel:


pkg/kp/rollstore/consul_store.go
```diff
@@ -204,6 +218,15 @@ func (s consulStore) CreateRollingUpdateFromExistingRCs(u roll_fields.Update) (r
 		return roll_fields.Update{}, err
 	}
 
+	err = s.labelNewRC(u.NewRC, newRCLabels)
+	if err != nil {
+		// This is potentially bad, because the labels that were there
+		// before have been overwritten. However its unreasonable to
+		// expect that a rollback will succeed if the roll-forward
+		// didn't succeed, so just report the error
```
mpuncel: In practice this doesn't pose a problem in Square's use of this, because no ongoing action is taken as a result of these lingering labels. Informational endpoints may have incorrect data, however

pkg/kp/rollstore/store.go
```diff
@@ -25,10 +25,12 @@ type Store interface {
 	// Creates a rolling update from two existing RCs. Will check that the
 	// RCs actually exist before applying the update, and acquire locks on
 	// them in a deterministic order to guarantee that no two RUs will
-	// operate on the same RC and will avoid deadlock scenarios
-	CreateRollingUpdateFromExistingRCs(update fields.Update) (fields.Update, error)
+	// operate on the same RC and will avoid deadlock scenarios. Before
+	// the RU is created, newRCLabels will be
```
petertseng: will be what?

pkg/kp/rollstore/consul_store.go
anthonybishopric: It might mess with the farm's RC selector though?

pkg/kp/rollstore/consul_store.go
```diff
@@ -171,6 +171,20 @@ func (s consulStore) checkForConflictingUpdates(rcIDs rc_fields.IDs) error {
 	return nil
 }
 
+// Best-effort applies labels to the specified RC. It's the callers
+// responsibility to put things back the way they were in an error scenario due
+// to the variety of situations this is called in.
+func (s consulStore) labelNewRC(rcID rc_fields.ID, rcLabels klabels.Set) error {
+	for k, v := range rcLabels {
+		err := s.labeler.SetLabel(labels.RC, rcID.String(), k, v)
```
anthonybishopric: I'm worried about this pattern - we need `labels.Applicator` to have a `SetLabels` method that can apply them atomically. In the near future, where we have active watchers for labels, we will want to guarantee atomicity when writing sets of related labels. Doing this would also reduce round-trips to Consul.

pkg/kp/rollstore/consul_store.go
mpuncel: that label (pod_id) is written as part of `rcstore.New()`

pkg/kp/rollstore/consul_store.go
mpuncel: agreed

mpuncel: although it'll be weird because you don't want to overwrite any other labels, it'd more be like batch apply and batch delete

pkg/kp/rollstore/consul_store.go
anthonybishopric: That's our internal label selector, but any RC selector may be used for the RU farm. That said, it may not have that much of an impact since the RU farm would hopefully never see an RC that has lingering labels like this.

anthonybishopric:
:ship:

pkg/labels/http_applicator.go
```diff
@@ -43,6 +43,10 @@ func (h *httpApplicator) SetLabel(labelType Type, id, name, value string) error
 	return util.Errorf("SetLabel not implemented for HttpApplicator (type %s, id %s, name %s, value %s)", labelType, id, name, value)
 }
 
+func (h *httpApplicator) SetLabels(labelType Type, id string, labels map[string]string) error {
+	return util.Errorf("SetLabel not implemented for HttpApplicator (type %s, id %s, labels %s)", labelType, id, labels)
```
petertseng: would have been best if that was `SetLabels`, but not too important I suppose.

I'm not sure you can `%s` a map though.

mpuncel: works well enough: http://play.golang.org/p/QwfOT9-GQc

PR #419, opened 2016-03-16
Replace rollstore.Put() function

mpuncel:
This was the only operation provided in the past for creating rolling
updates. Its use is discouraged because it does not provide any safety
guarantees around consistency of data (e.g. a RU being created with
references to RCs which don't exist).

It is being resurrected to ease clients of this library onto the new
functions without forcing.


petertseng:
I'm :+1:  on this

PR #420, opened 2016-03-16
P2 rm for managed pods

rudle:


bin/p2-rm/main.go
```diff
@@ -2,34 +2,65 @@
...
+	Labeler     *labels.ConsulApplicator
+
+	LabelID  string
+	NodeName string
+	PodName  string
+}
+
+// NewP2RM is a constructor for the P2RM type. It will generate the necessary
+// storage types based on its api.Client argument
+func NewP2RM(client *api.Client, podName, nodeName string) *P2RM {
```
anthonybishopric: Migrate this pretty useful thing to `pkg/` somewhere?

bin/p2-rm/main.go
```diff
@@ -2,34 +2,65 @@
...
-	nodeName = kingpin.Flag("node", "The node to do the scheduling on. Uses the hostname by default.").String()
+	podName      = kingpin.Arg("pod", "The names of the pod to be removed").Required().String()
+	nodeName     = kingpin.Flag("node", "The node to do the scheduling on. Uses the hostname by default.").String()
+	deallocation = kingpin.Flag("deallocate", "Specifies that we are deallocating this pod on this node. Using this switch will mutate the desired_replicas value on a managing RC, if one exists.").Bool()
 )
 
+// P2RM contains the state necessary to safely remove a pod from a node
+type P2RM struct {
+	Store       kp.Store
+	ConsulStore rcstore.Store
```
anthonybishopric: The store above this one is also a Consul store, technically. what about `RCStore`?

bin/p2-rm/main.go
```diff
@@ -40,37 +71,154 @@ func main() {
...
+	}
+
+	err = rm.deletePod()
+	if err != nil {
+		fmt.Fprintf(os.Stderr, "Unable to delete pod. Please re-run this command to clean up. \n%v", err)
 		os.Exit(1)
 	}
-	err = labeler.RemoveAllLabels(labels.POD, labelID)
+
+	err = rm.ConsulStore.Enable(id)
```
anthonybishopric: If you enable for decrementing replicas, there is a possible race where it tries to reinstall this thing before you decrement.

anthonybishopric: Also, why re-enable at all?

rudle: I agree that the order is wrong.

bin/p2-rm/main.go
```diff
@@ -40,37 +71,154 @@ func main() {
...
+	if err == nil {
+		username = currentUser.Username
+	}
+
+	return fmt.Sprintf("p2-rm:user:%s:rcID:%s", username, rcID)
+}
+
+func (rm *P2RM) decrementDesiredCount(id fields.ID) error {
+	sessionMutex := &sync.Mutex{}
+	rm.Client.LockKey(rm.LabelID)
```
mpuncel: don't wanna blow up the scope of this PR but i've been trying to get away with random code knowing the structure of consul keys

bin/p2-rm/main.go
```diff
@@ -40,37 +71,154 @@ func main() {
...
+				err)
+		}
+	}
+
+	fmt.Printf("%s: successfully removed %s\n", rm.NodeName, rm.PodName)
+}
+
+func sessionName(rcID fields.ID) string {
+	currentUser, err := user.Current()
+	username == "unknown"
```
rudle: what is a programming :dizzy_face:

bin/p2-rm/main.go
```diff
@@ -40,37 +71,154 @@ func main() {
...
+}
+
+func (rm *P2RM) decrementDesiredCount(id fields.ID) error {
+	sessionMutex := &sync.Mutex{}
+	rm.Client.LockKey(rm.LabelID)
+
+	sessions := make(chan string)
+	quit := make(chan struct{})
+
+	go consulutil.SessionManager(api.SessionEntry{
```
mpuncel: for such a quick operation, this is probably heavyweight. You can probably just use `(kp.Store) NewSession()` and check for renewal errors before you do any work

bin/p2-rm/main.go
```diff
@@ -40,37 +71,154 @@ func main() {
...
+
+				sessionMutex.Lock()
+				session = rm.Store.NewUnmanagedSession(sessionID, "")
+				sessionMutex.Unlock()
+			default:
+			}
+		}
+	}(sessionMutex)
+
+	rcLock, err := rm.ConsulStore.LockForMutation(id, session)
```
mpuncel: :+1:

bin/p2-rm/main.go
rudle: TIL: I'll simplify this

bin/p2-rm/main.go
mpuncel: does disabling/enabling guarantee that the RC will notice that our node is down before it attempts to bring one down? any race conditions there?

mpuncel: also should the enable happen after `AddDesiredReplicas(-1)`

bin/p2-rm/main.go
mpuncel: that said, it's fine since you already wrote it

bin/p2-rm/main.go
```diff
@@ -2,34 +2,65 @@
...
+	nodeName     = kingpin.Flag("node", "The node to do the scheduling on. Uses the hostname by default.").String()
+	deallocation = kingpin.Flag("deallocate", "Specifies that we are deallocating this pod on this node. Using this switch will mutate the desired_replicas value on a managing RC, if one exists.").Bool()
 )
 
+// P2RM contains the state necessary to safely remove a pod from a node
+type P2RM struct {
+	Store       kp.Store
+	ConsulStore rcstore.Store
+	Client      *api.Client
+	Labeler     *labels.ConsulApplicator
```
mpuncel: why can't this just be a `labels.Applicator` and then you don't have to export `labels.consulApplicator`

bin/p2-rm/main.go
rudle: Do you have a handle on what would reuse this? I can't think of anything offhand.

bin/p2-rm/main.go
rudle: right

bin/p2-rm/main.go
anthonybishopric: Maybe not anything, but it does seem like a generally useful thing - safely remove one thing from a replication controller.

bin/p2-rm/main.go
```diff
@@ -40,37 +67,123 @@ func main() {
...
 	if err != nil {
-		fmt.Fprintf(os.Stderr, "unable to check node for labels: %v\n", err)
-		os.Exit(1)
+		os.Exit(2)
 	}
-	if podLabels.Labels.Has(rc.RCIDLabel) {
+
+	if !podIsManagedByRC {
+		err = rm.deletePod()
+		if err != nil {
```
anthonybishopric: Probably worth printing this error

anthonybishopric: Nevermind, I see that it's printed below

bin/p2-rm/main.go
```diff
@@ -40,37 +67,123 @@ func main() {
...
-			labelID,
-			podLabels.Labels.Get(rc.RCIDLabel),
-		)
-		input, err := bufio.NewReader(os.Stdin).ReadString('\n')
-		answer := (strings.ToLower(strings.TrimSpace(input)) + "n")[0]
-		if err != nil || answer != 'y' {
-			fmt.Fprintf(os.Stderr, "aborting\n")
+			"error: %s is managed by replication controller: %v\n"+
+				"It's possible you meant you deallocate this pod on this node. If so, please confirm your intention with --deallocate\n", rcID)
+		os.Exit(2)
```
anthonybishopric: In case you're interested, the `log` package has a `Fatalf` function that might shave a few lines off.

bin/p2-rm/main.go
```diff
@@ -40,37 +67,123 @@ func main() {
...
+	}
+
+	if podIsManagedByRC && *deallocation {
```
anthonybishopric: What problem are you attempting to prevent by requiring the `deallocation` flag?

anthonybishopric:
I think I generally understand the reasoning for the `--deallocate`. Also Github has eaten all the previous comments, so apologies if you've already discussed it with Mike.

Otherwise, this is going to be very useful and very in demand. :ship:

bin/p2-rm/main.go
mpuncel: is it printed?

bin/p2-rm/main.go
```diff
@@ -40,37 +67,123 @@ func main() {
...
+			fmt.Fprintf(os.Stderr, "Got fatal renewal error!\n%v\n", err)
 			os.Exit(1)
+		default:
 		}
+	}(renewalErrChan)
+
+	rcLock, err := rm.RCStore.LockForMutation(id, session)
+	if err != nil {
+		fmt.Fprintf(os.Stderr, "Unable to lock RC for mutation. Halting.\n %v", err)
+		os.Exit(1)
```
mpuncel: I'm concerned about this exit happening between `Disable()` and `Enable()` which would leave things in a bad state. It's probably better to just drain the channel right before `Disable()` and then just do all of the operations without checking again

anthonybishopric: Agreed that, at the very least, it could cause surprising raciness in a pathological case.

mpuncel:
looks good, just the one comment about exiting between operations

bin/p2-rm/main.go
rudle: All error paths in deletePod() have logging.

bin/p2-rm/main.go
rudle: Good to know, but I'm not keen to add a dependency on the log package here. Right now we're doing human friendly printing to stderr

bin/p2-rm/main.go
rudle: I'm happy to follow the convention for session renewal, but I'm not clear on what it is.

In my opinion, the safest thing to do here is to check the lock between each operation and print some helpful diagnostics in that case. Is that overkill? I'm fairly confident that the critical section guarded by the consul lock is going to complete without requiring a renewal.

bin/p2-rm/main.go
rudle: I feel safer having it there while we're in a mixed world wrt p2-rc and p2-master. It (hopefully) reminds the operator of the action they're about to perform and what the consequences of an incomplete run are.

bin/p2-rm/main.go
mpuncel: I sort of think it'd be better to do all of the actions atomically and risk doing so without the lock held than do something we don't expect (like leave a RC disabled).

But both cases aren't good, and you have to pick one, so i'll leave it up to you as the author.

mpuncel:
:+1:

bin/p2-rm/main.go
mpuncel: and it would be very improbable 1) for this operation to last longer than a lock session (15 seconds i believe by default and 2) for the renewal to fail even if it did take that long

so in practice this should all be moot :)

rudle: I removed the bit that crashed the program on failure to renew a lock. I think that it's the lesser of two evils.

PR #421, opened 2016-03-16
roll: If update is done but just waiting on healthy, don't schedule

petertseng:
We ran into a situation today where a deploy was in a state where one
canary node was desired, but two nodes had appeared on the new side.
Examination of the logs produced the following

    {
      "msg": "Adding 1 new nodes",
      "new": {
        "Desired": 0,
        "Current": 0,
        "Real": 0,
        "Healthy": 0
      },
      "next": 1,
      "old": {
        "Desired": 19,
        "Current": 19,
        "Real": 19,
        "Healthy": 18
      },
      "pod": "censored",
    },

This is expected.

    {
      "msg": "Adding 1 new nodes",
      "new": {
        "Desired": 1,
        "Current": 1,
        "Real": 1,
        "Healthy": 0
      },
      "next": 1,
      "old": {
        "Desired": 18,
        "Current": 19,
        "Real": 19,
        "Healthy": 18
      },
      "pod": "censored",
    }

This was unexpected, but (assuming minimum healthy was < 18)
understandable because in `rollAlgorithm`, `headroom` would be > 1 and
`remaining` would be 1, so we would schedule one node.

This is a direct consequence of 0768d697962d100bb4f2366d5e6e52a93ff24b8b
which removed the block for healthy nodes. It turns out that this block
is a good thing to have... when the deploy is almost complete ("almost
complete" meaning the desired number of nodes have been scheduled on the
new RC).

If the deploy is almost complete by this definition, we need to check
whether the minimum health requirement is met. If so, we declare the
deploy complete (old behavior unchanged by this commit). If not, we wait
for minimum health (behavior changed by this commit - behavior before
this commit was to schedule new nodes).

### DIFF
--- a/pkg/roll/run_update.go
+++ b/pkg/roll/run_update.go
@@ -152,24 +152,32 @@ ROLL_LOOP:
 				break
 			}
 
-			if newNodes.Desired >= u.DesiredReplicas && newNodes.Healthy >= u.MinimumReplicas {
-				// We only ask for u.MinimumReplicas nodes to be healthy
-				// before declaring an upgrade to be complete.
-				// This is so that if a deployer intentionally deploys a known-bad SHA
-				// for which no nodes become healthy (specifying minimum == 0),
-				// we don't leave an RU and old RC lying around.
-				// The RU and RC would have required manual intervention to clean up.
-				//
+			if newNodes.Desired >= u.DesiredReplicas {
 				// The above condition contains a >= instead of a ==.
 				// This is despite the fact that normally rollAlgorithm will never cause
 				// newDesired > u.Desired.
 				// The benefit of using >= is that it allows termination even if Run
 				// resumes from an unusual state.
-				u.logger.WithFields(logrus.Fields{
-					"old": oldNodes,
-					"new": newNodes,
-				}).Debugln("Upgrade complete")
-				break ROLL_LOOP
+
+				if newNodes.Healthy >= u.MinimumReplicas {
+					// We only ask for u.MinimumReplicas nodes to be healthy
+					// before declaring an upgrade to be complete.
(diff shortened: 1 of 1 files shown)


pkg/roll/run_update.go
```diff
@@ -152,24 +152,32 @@ ROLL_LOOP:
...
+					// This is so that if a deployer intentionally deploys a known-bad SHA
+					// for which no nodes become healthy (specifying minimum == 0),
+					// we don't leave an RU and old RC lying around.
+					// The RU and RC would have required manual intervention to clean up.
+					u.logger.WithFields(logrus.Fields{
+						"old": oldNodes,
+						"new": newNodes,
+					}).Debugln("Upgrade complete")
+					break ROLL_LOOP
+				} else {
```
mpuncel: while this fixes the bug, I think it also introduces a regression in behavior where a node in a particular "batch" might never come up (for example if it was shut down), and the deploy will halt as a result because we block forever here.

That said I'm okay with this for now until we figure out something better.

mpuncel: also, this is strictly better than having the check up above like before

petertseng: My understanding of the issue you describe is if I have nodes A B C D E, and ask for a canary of 1 node, and A is chosen but never becomes healthy, the deploy blocks forever.

I think it has to. The alternative is to keep trying to bring node B in. What if B becomes unhealthy too? We wouldn't want to bring C, D, E in to have them all become unhealthy (granted, because of minimum health guarantees this should only happen if the deployer set minimum health to 0, which we hope doesn't happen)

mpuncel: I was confused, this looks good :+1:

PR #423, opened 2016-03-17
roll: terminate when OLD + new healthy exceed minimum health

petertseng:
This helps when we are in a canary deploy. Imagine a canary deploy with
2 minimum health but only 1 canary desired.

The RU will move 1 node to the new side, then wait for 2 nodes to be
healthy on the new side, which is impossible.

The old side is allowed to contribute toward health too. This will allow
the above-mentioned canary deploy to terminate.

### DIFF
--- a/pkg/roll/run_update.go
+++ b/pkg/roll/run_update.go
@@ -159,7 +159,7 @@ ROLL_LOOP:
 				// The benefit of using >= is that it allows termination even if Run
 				// resumes from an unusual state.
 
-				if newNodes.Healthy >= u.MinimumReplicas {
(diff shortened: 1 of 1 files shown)


anthonybishopric:
:ship: we need some stronger regression tests in this area. Lets figure out a plan to make it more testable

petertseng:
I am really wanting to extract some of this code into a small testable function. Just not sure what.

`deployShouldTerminate(oldNodes, newNodes rcNodeCounts) (shouldTerminate, shouldBlock bool)` ???

petertseng:
all right, I'll merge this for now, we can see how to perform the extraction desired

petertseng:
that extraction is performed in #424

PR #424, opened 2016-03-17
roll: extract shouldStop function and test it

petertseng:


petertseng:
foot note is that I'm not 100% sure this makes our lives better, but I made enough of a mistake in this logic the first time that I think I'm motivated to try to get it right.

Obviously a test for `Roll` would still be great. Still need to investigate how to make that happen.

pkg/roll/run_update.go
```diff
@@ -73,6 +73,15 @@ type Update interface {
 	Run(quit <-chan struct{}) bool
 }
 
+// returned by shouldStop
+type deployAction int
```
anthonybishopric: nit `ruStep` maybe?

petertseng: I'll take it! commit has now been amended.

petertseng:
limited scope of `nextAction` to the `if` block

mpuncel:
I like this

mpuncel:
:+1:

PR #425, opened 2016-03-18
fixes a typo in an error message

mpuncel:


petertseng:
Sure.

(Does `%s` in fact work for `map` after all?)

mpuncel:
yeah:  http://play.golang.org/p/QwfOT9-GQc

PR #426, opened 2016-03-18
roll: Safe behavior if old desired < old healthy

petertseng:
Note for reviewers that the first four commits in this PR are the same as #414 (Except rebased off of master and I switched around the order of values in the `rollAlgorithmsParams` test). This is because for various reasons I don't want to merge #414 yet, but want these commits because they're necessary to write the tests I want for this bugfix.

So the only commit that is new is the last commit in the sequence: https://github.com/square/p2/commit/d72fedd0b7e114f8278cbac82c2a714a3f74d8e4

### DIFF
--- a/pkg/roll/run_update.go
+++ b/pkg/roll/run_update.go
@@ -426,6 +426,17 @@ func (u *update) shouldRollAfterDelay(newFields rcf.RC) (int, error) {
 
 func (u *update) rollAlgorithmParams(oldHealth, newHealth rcNodeCounts) (oldHealthy, newHealthy, desired, minHealthy int) {
 	oldHealthy = oldHealth.Healthy
+	if oldHealth.Desired < oldHealthy {
+		// Because of the non-atomicity of our KV stores,
+		// we may run into this situation while decrementing old RC's count:
+		// old RC has decremented desire, but not yet removed RC label from the pod.
+		// We will see this as {Desired: 1, Healthy: 2}, or similar.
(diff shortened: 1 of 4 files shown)


pkg/roll/run_update.go
```diff
@@ -426,6 +426,17 @@ func (u *update) shouldRollAfterDelay(newFields rcf.RC) (int, error) {
...
 	oldHealthy = oldHealth.Healthy
+	if oldHealth.Desired < oldHealthy {
+		// Because of the non-atomicity of our KV stores,
+		// we may run into this situation while decrementing old RC's count:
+		// old RC has decremented desire, but not yet removed RC label from the pod.
+		// We will see this as {Desired: 1, Healthy: 2}, or similar.
+		// We assume that the old RC will eventually cause Healthy to go to 1 here.
+		// So, we should assume that oldHealthy is is the lesser of the two.
+		// Otherwise, we may over-eagerly remove more nodes from old RC.
+		// That would cause a violation of minimum health.
```
mpuncel: great comment

mpuncel:
:+1: but I didn't look at the first 4 commits since you said they're the same

petertseng:
`git diff 0e5cec1f61b896324698909b57d202f3c5b04055 ac57a5364b923257bb0a2ae47c37cff27fee0fb0` (that'd be the first four commits of this PR and the first four commits of #414) has changes I fully expect to be there, so I feel comfortable with this.

PR #428, opened 2016-03-22
roll: Terminate when current >= desired

petertseng:
The health check would have always been true assuming rollAlgorithm had
correct operation. But we should block until the RC has labeled enough
pods so that post-update operations can examine the nodes just-deployed
to by calling newRC.CurrentPods()

### DIFF
--- a/pkg/roll/run_update.go
+++ b/pkg/roll/run_update.go
@@ -246,17 +246,21 @@ func (u *update) shouldStop(oldNodes, newNodes rcNodeCounts) ruStep {
 
 	// Enough nodes are scheduled on the new side.
 
-	if oldNodes.Healthy+newNodes.Healthy >= u.MinimumReplicas {
-		// We only ask for u.MinimumReplicas nodes to be healthy
+	if newNodes.Current >= u.DesiredReplicas {
+		// We only ask for the new RC to have labeled the desired number of nodes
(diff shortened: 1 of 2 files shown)


pkg/roll/run_update.go
```diff
@@ -246,17 +246,21 @@ func (u *update) shouldStop(oldNodes, newNodes rcNodeCounts) ruStep {
...
-		// The RU and RC would have required manual intervention to clean up.
+		// we terminate and don't leave an RU and old RC lying around.
+		// 2) rollAlgorithm is responsible for maintaining the min health guarantee,
+		// not this code. If we reach this point, rollAlgorithm is finished anyway.
+		// The only decision to make is whether to terminate or block.
 		return ruShouldTerminate
 	}
 
-	// Enough nodes are scheduled, but not enough of them are healthy.
+	// Enough nodes are desired, but not enough have been labeled by the RC.
```
anthonybishopric: "Enough" is not really a sensible way to describe "desired". I'd rewrite this sentence:

> Fewer nodes are owned by the new RC than are desired.

anthonybishopric:
:+1:

pkg/roll/run_update.go
petertseng: took it word for word

mpuncel:
:ship:

PR #431, opened 2016-03-23
Extract roll loop into testable function, then test it

petertseng:


petertseng:
Currently only tests for a single case (a deploy that proceeds along swimmingly), let's talk about other tests to potentially add.

pkg/roll/update_test.go
```diff
@@ -688,3 +689,83 @@ func TestShouldRollMidwayHealthyMigrationFromZeroWhenNewSatisfies(t *testing.T)
...
+	podID := manifest.ID().String()
+	return upd.labeler.SetLabel(labels.POD, node+"/"+podID, rc.RCIDLabel, string(upd.NewRC))
+}
+
+func TestRollLoopTypicalCase(t *testing.T) {
+	upd, _, manifest := updateWithHealth(t, 3, 0, nil, nil, nil)
+	upd.DesiredReplicas = 3
+	upd.MinimumReplicas = 2
+
+	updates := make(chan map[string]health.Result)
```
petertseng: this name is... slightly misleading, since there's an `update` type, but this channel is not for values of that `update` type. I meant for it to be "health updates". I'll probably change the name. `health` doesn't work since it's the package name and I use `health.Passing` later on. so probably `healthChan` or something.

pkg/roll/update_test.go
```diff
@@ -688,3 +689,83 @@ func TestShouldRollMidwayHealthyMigrationFromZeroWhenNewSatisfies(t *testing.T)
...
+	errors := make(chan error)
+	quit := make(chan struct{})
```
petertseng: these quit channels aren't being used (nor `oldRCQuit`, `newRCQuit`), so I might as well not use them in this test.

pkg/roll/update_test.go
```diff
@@ -683,8 +684,88 @@ func TestShouldRollMidwayHealthyMigrationFromZeroWhenNewSatisfies(t *testing.T)
...
+	<-oldRCUpdated
+	Assert(t).AreEqual(oldRC.ReplicasDesired, 0, "expected old RC to decrement")
+	<-newRCUpdated
+	Assert(t).AreEqual(newRC.ReplicasDesired, 3, "expected new RC to increment")
+
+	transferNode("node3", manifest, upd)
+	updates <- checks
+
+	result := <-rollLoopResult
+	Assert(t).IsTrue(result, "expected roll loop to terminate with true (successful)")
```
petertseng: check that the old RC got deleted too?

pkg/roll/update_test.go
```diff
@@ -683,8 +684,88 @@ func TestShouldRollMidwayHealthyMigrationFromZeroWhenNewSatisfies(t *testing.T)
...
+	remove, add, err := upd.shouldRollAfterDelay(manifest.ID())
 	Assert(t).IsNil(err, "expected no error determining nodes to roll")
 	Assert(t).AreEqual(remove, 0, "expected to remove no nodes")
 	Assert(t).AreEqual(add, 1, "expected to add one node")
 }
+
+func panicOnError(desc string, errs <-chan error) {
+	err, ok := <-errs
+	if ok {
+		panic("Error received on " + desc + ":" + err.Error())
```
mpuncel: does `t.Fatal` not work in a goroutine? probably would be nicer if so

petertseng: it maybe would; I'd just need to pass the `testing.T`. Will need to figure out how to test since I don't have anything that exercises it right now

pkg/roll/update_test.go
petertseng: oops, can't, that happens outside of `rollLoop`. So scratch that idea.

pkg/roll/update_test.go
```diff
@@ -683,8 +684,88 @@ func TestShouldRollMidwayHealthyMigrationFromZeroWhenNewSatisfies(t *testing.T)
...
+	}
+}
+
+// Transfers the named node from the old RC to the new RC
+func transferNode(node string, manifest pods.Manifest, upd update) error {
+	if _, err := upd.kps.SetPod(kp.REALITY_TREE, node, manifest); err != nil {
+		return err
+	}
+	podID := manifest.ID().String()
+	return upd.labeler.SetLabel(labels.POD, node+"/"+podID, rc.RCIDLabel, string(upd.NewRC))
```
mpuncel: for another PR, we should have a helper for this `node+"/"+podId` thing in the `kp` or `labels` package

pkg/roll/update_test.go
```diff
@@ -683,8 +684,88 @@ func TestShouldRollMidwayHealthyMigrationFromZeroWhenNewSatisfies(t *testing.T)
...
+	go panicOnError("new RC", newRCErrors)
+
+	rollLoopResult := make(chan bool)
+
+	go func() {
+		rollLoopResult <- upd.rollLoop(manifest.ID(), updates, errors, quit)
+		close(rollLoopResult)
+	}()
+
+	go panicOnError("roll loop", errors)
```
petertseng: this call incorrectly assumes that `rollLoop` *sends* errors on the error channel. Nope, it doesn't. `rollLoop` expects to *receive* errors on this channel, so again we can just pass nil.

pkg/roll/update_test.go
```diff
@@ -683,8 +684,81 @@ func TestShouldRollMidwayHealthyMigrationFromZeroWhenNewSatisfies(t *testing.T)
...
+
+	checks := map[string]health.Result{
+		"node1": {Status: health.Passing},
+		"node2": {Status: health.Passing},
+		"node3": {Status: health.Passing},
+	}
+
+	healths <- checks
+
+	<-oldRCUpdated
```
petertseng: helper function to wait on both `oldRCUpdated` and `newRCUpdated` and assert on their values? This code got a bit repetitive.

petertseng:
rebased on master, fixed typical case test's accidental reliance on "unknown is healthy", added a test for migration from zero

petertseng:
Added a test for a rolling update where a node goes unhealthy

pkg/roll/update_test.go
petertseng: please remind me to use `MakePodLabelKey` which was added in https://github.com/square/p2/pull/452/files#diff-b534dcec1a1a54cde568a086e94fe0d3

mpuncel: reminder!

mpuncel:
this is great 👍

petertseng:
I knew that reminder request comment would come in handy!


```diff
$ git diff 7989d99 f619a4e
diff --git a/pkg/roll/update_test.go b/pkg/roll/update_test.go
index 6bd2ca8..5c99a9c 100644
--- a/pkg/roll/update_test.go
+++ b/pkg/roll/update_test.go
@@ -788,8 +788,7 @@ func transferNode(node string, manifest pods.Manifest, upd update) error {
        if _, err := upd.kps.SetPod(kp.REALITY_TREE, node, manifest); err != nil {
                return err
        }
-       podID := manifest.ID().String()
-       return upd.labeler.SetLabel(labels.POD, node+"/"+podID, rc.RCIDLabel, string(upd.NewRC))
+       return upd.labeler.SetLabel(labels.POD, rc.MakePodLabelKey(node, manifest.ID()), rc.RCIDLabel, string(upd.NewRC))
 }

 func assertRCUpdates(t *testing.T, rc *rc_fields.RC, upd <-chan struct{}, expect int, desc string) {
```

merging soon if it works

mpuncel:
👍

PR #432, opened 2016-03-24
Revert "roll: conservatively treat old unknown-health nodes as healthy"

petertseng:
This reverts commit 0ffe7f9987a432451a241fb7a394762c22848a3a.

The reason we thought we wanted that commit was because enables the case
of a migration from a different deployment tool onto p2. For example, if
we have three unknown nodes on the old side and minimum health is 2, we
want to roll one node at a time.

However, this code is inapplicable in a migration from a different
deployment tool, because the nodes simply will not exist on the old RC.
The old RC's desired node count will be 0, since that is the default
number for an old RC. So the unknown health code path never gets
triggered.

The proper way to support migrations from another deployment tool is
d045410ce859ab06b5c1ad46904e108de5912754.

We ran into a case today where we mistakenly rolled a node under this
condition:

```json
{
  "msg": "Adding 1 new nodes and removing 1 old nodes",
    "new": {
    "Desired": 0,
    "Current": 0,
    "Real": 0,
    "Healthy": 0,
    "Unhealthy": 0,
    "Unknown": 0

    },
  "nextAdd": 1,
  "nextRemove": 1,
  "old": {
    "Desired": 2,
    "Current": 2,
    "Real": 2,
    "Healthy": 1,
    "Unhealthy": 0,
    "Unknown": 1

  },
  "pod": "censored"
}
```

In fact, the old unknown-health node was unhealthy.

It is ALWAYS safer to NEVER include the unknown health node in the
healthy count, because increasing oldHealthy always increases headroom
whereas that headroom might not exist if the unknown health node is
actually unhealthy instead of healthy.

In an ideal world, we would be able to have selective rolling, where we
are able to say "schedule N nodes but only on unhealthy/unknown nodes",
but until we have that capability it is best to be safe.

### DIFF
--- a/pkg/roll/run_update.go
+++ b/pkg/roll/run_update.go
@@ -434,8 +434,7 @@ func (u *update) shouldRollAfterDelay(newFields rcf.RC) (int, int, error) {
 }
 
 func (u *update) rollAlgorithmParams(oldHealth, newHealth rcNodeCounts) (oldHealthy, newHealthy, currentDesired, targetDesired, minHealthy int) {
-	// We conservatively treat Unknown nodes as healthy on the old side.
-	oldHealthy = oldHealth.Healthy + oldHealth.Unknown
+	oldHealthy = oldHealth.Healthy
 	if oldHealth.Desired < oldHealthy {
 		// Because of the non-atomicity of our KV stores,
 		// we may run into this situation while decrementing old RC's count:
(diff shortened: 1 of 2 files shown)


anthonybishopric:
:+1:

mpuncel:
:+1:

petertseng:
just so I have it in records: Reverts one commit in #408

PR #433, opened 2016-03-25
Support rcFarmSelector in RC farm

mpuncel:
Previously only roll farms supported a selector to use to partition work
into a test environment. It's also useful to have this capability for RC
farms.

### DIFF
--- a/bin/p2-rctl-server/main.go
+++ b/bin/p2-rctl-server/main.go
@@ -119,7 +119,7 @@ func main() {
 	}
 
 	// Run the farms!
-	go rc.NewFarm(kpStore, rcStore, scheduler, labeler, rcSub.Chan(), logger, alerter).Start(nil)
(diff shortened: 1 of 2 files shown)


rudle:
LGTM

mpuncel:
resolved conflicts with `Alerter` change

bin/p2-rctl-server/main.go
```diff
@@ -119,7 +119,7 @@ func main() {
 	}
 
 	// Run the farms!
-	go rc.NewFarm(kpStore, rcStore, scheduler, labeler, rcSub.Chan(), logger, alerter).Start(nil)
+	go rc.NewFarm(kpStore, rcStore, scheduler, labeler, rcSub.Chan(), logger, klabels.Everything(), alerter).Start(nil)
```
mpuncel: put it before `alerter` to be consistent with `roll.NewFarm()`

PR #434, opened 2016-03-25
Only write the latest RC list to consumer of rcstore.WatchNew

mpuncel:
Previously, if the consumer of the RC watch was very slow, we might
receive new updates from consul while we're still blocking on writing
the previous update. Now, if we receive a new update while still
attempting to write the old one, we just move on to the most recent
update.

### DIFF
--- a/pkg/kp/rcstore/consul_store.go
+++ b/pkg/kp/rcstore/consul_store.go
@@ -136,7 +136,7 @@ func (s *consulStore) Get(id fields.ID) (fields.RC, error) {
 		// ID didn't exist
 		return fields.RC{}, NoReplicationController
 	}
-	return s.kvpToRC(kvp)
+	return kvpToRC(kvp)
 }
 
 func (s *consulStore) List() ([]fields.RC, error) {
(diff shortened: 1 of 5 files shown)


pkg/kp/rcstore/consul_store.go
```diff
@@ -158,17 +158,26 @@ func (s *consulStore) WatchNew(quit <-chan struct{}) (<-chan []fields.RC, <-chan
...
+		for {
 			out, err := s.kvpsToRCs(listed)
 			if err != nil {
 				select {
 				case errCh <- err:
 				case <-quit:
+					break
 				}
 			} else {
 				select {
```
anthonybishopric: To avoid rewriting the same result on `outCh` repeatedly, perhaps check for an update with a second select right above this?
```
select {
   case listed = <- inCh:
   default:
}
``` 
?

pkg/kp/rcstore/consul_store.go
```diff
@@ -144,31 +144,87 @@ func (s *consulStore) List() ([]fields.RC, error) {
...
 	go consulutil.WatchPrefix(rcTree+"/", s.kv, inCh, quit, errCh)
 
+	return outCh, errCh
+}
+
+// Pulled out of WatchNew for testing purposes (faking watches on a mock KV is
+// hard). It converts api.KVPairs values received on inCh to []fields.RC, and
+// attempts to pass them on outCh. It will discard updates that are not read
+// quickly enough (see WatchNew)
+func publishLatestRCs(inCh <-chan api.KVPairs, quit <-chan struct{}) (<-chan []fields.RC, chan error) {
```
petertseng: Is the `chan error` intentionally bidirectional?

pkg/kp/rollstore/consul_store.go
```diff
@@ -521,35 +521,75 @@ func (s consulStore) Lock(id roll_fields.ID, session string) (bool, error) {
...
-	errCh := make(chan error)
 	inCh := make(chan api.KVPairs)
 
+	outCh, errCh := publishLatestRolls(inCh, quit)
 	go consulutil.WatchPrefix(rollTree+"/", s.kv, inCh, quit, errCh)
 
+	return outCh, errCh
+}
+
+func publishLatestRolls(inCh <-chan api.KVPairs, quit <-chan struct{}) (<-chan []roll_fields.Update, chan error) {
```
petertseng: Is the `chan error` intentionally bidirectional?

mpuncel: yeah because it also gets passed to the roll watch (two publishers). Weird, I know, but not something i wanted to tackle in this PR

pkg/kp/rcstore/consul_store.go
mpuncel: yeah, see my other comment

pkg/kp/rollstore/consul_store_test.go
```diff
@@ -787,6 +788,99 @@ func TestLeaveOldInvalidIfNoOldRC(t *testing.T) {
...
+	inCh <- rollsWithIDs(t, "e", 6)
+
+	select {
+	case val = <-outCh:
+	case <-time.After(1 * time.Second):
+		t.Fatalf("timed out reading from channel")
+	}
+
+	if len(val) != 6 {
+		t.Errorf("Expected %d values on outCh, got %d", 2, len(val))
```
petertseng: 6 not 2

pkg/kp/rcstore/consul_store_test.go
```diff
@@ -0,0 +1,109 @@
...
+	inCh <- rcsWithIDs(t, "e", 6)
+
+	select {
+	case val = <-outCh:
+	case <-time.After(1 * time.Second):
+		t.Fatalf("timed out reading from channel")
+	}
+
+	if len(val) != 6 {
+		t.Errorf("Expected %d values on outCh, got %d", 2, len(val))
```
petertseng: 6 not 2

pkg/kp/rollstore/consul_store.go
```diff
@@ -521,35 +521,75 @@ func (s consulStore) Lock(id roll_fields.ID, session string) (bool, error) {
...
+			// the channel was closed
+			return
+		}
+
+		needToWrite := true
+
+		for {
+			if !needToWrite {
+				select {
+				case listed, ok = <-inCh:
```
petertseng: should this set `needToWrite` true?

pkg/kp/rollstore/consul_store.go
```diff
@@ -521,35 +521,75 @@ func (s consulStore) Lock(id roll_fields.ID, session string) (bool, error) {
...
+				select {
+				case outCh <- out:
+				case listed, ok = <-inCh:
+					if !ok {
+						// channel closed
+						return
+					}
+					// We got a new value before we write the old one,
+					// drop the old one
+					needToWrite = true
```
petertseng: At this point, `needToWrite` is bound to be true anyway due to line 579 though?

mpuncel: yeah, but honestly I like it for clarity. Although i'm realizing i missed setting it to `false` in the case above

pkg/kp/rollstore/consul_store.go
petertseng: OK you got it now

pkg/kp/rcstore/consul_store_test.go
petertseng: you missed this one!

petertseng:
Seems reasonable, tests are good. Is it possible to add a test for the case when `kvToRCs` and the `json.Unmarshal` fail? Would make the tests even more ironclad I think

mpuncel:
yeah i think that's a worth goal given that the review caught some bugs that the tests didn't. Will revisit this later today or tomorrow.

petertseng:
I'm :+1:  with these new tests. Don't forget squash the thing you want to

mpuncel:
you know me too well @petertseng

petertseng:
yup :+1: 

By the way any thought on deduplicating the "expect this channel to be closed" code used a few times in tests?

mpuncel:
@petertseng that'd be a little hard since `outCh` and `errCh` have different types. I also don't really feel that test code needs to be deduped super strongly except for test setup. If a test fails it's nice to be able to quickly see what's going wrong imo

PR #435, opened 2016-03-28
Mpuncel/alerting

mpuncel:


pkg/roll/run_update.go
```diff
@@ -363,7 +363,7 @@ func (u *update) countHealthy(id rcf.ID, checks map[string]health.Result) (rcNod
 
 	ret.Desired = rcFields.ReplicasDesired
 
-	currentPods, err := rc.New(rcFields, u.kps, u.rcs, u.sched, u.labeler, u.logger).CurrentPods()
+	currentPods, err := rc.New(rcFields, u.kps, u.rcs, u.sched, u.labeler, u.logger, nil).CurrentPods()
```
anthonybishopric: Should we inject an `Alerter` into the roll farm in this PR? Even if we don't use it except to pass through to this function, it will inevitably appear here later I'd guess.

anthonybishopric:
:+1: mod one question

pkg/roll/run_update.go
mpuncel: sure, although I don't anticipate `CurrentPods()` alerting ever. the roll farm may alert in other places though so I'll do it.

mpuncel:
@anthonybishopric I added it to roll. The only thing I don't like is the prolific nil-checking of alerter, but I suppose that's not really a huge issue.

anthonybishopric:
You've probably mitigated most of the risk by making the no-op version easy to use.

petertseng:
Thinking about the nil checks, I kind of liked what happened in https://github.com/golang/crypto/blob/master/openpgp/packet/config.go - that the methods of `*Config` check whether the `*Config` is nil. However that works in that case since `Config` is a struct, whereas `Alerter` is an interface, so I believe we can't do the same here. So I couldn't come up with a better solution. But maybe someone else will be able to run somewhere with this idea.

PR #436, opened 2016-03-29
if there is an error sending an alert, log a warning

mpuncel:


anthonybishopric:
:shipit:  Could argue over `Warn` vs `Error` but I'll probably be grepping by the message, not the level anyway.

mpuncel:
happy to change to Error

PR #437, opened 2016-03-29
Fix rollstore by populating all member fields in constructor

mpuncel:
Some rollstore operations depend on member fields that were not being
set in the rollstore constructor. This commit fixes that.


anthonybishopric:
:+1: although maybe we should be more diligent about testing constructor functions when we use kwarg initialization instead of positional init? If we didn't use the kwarg init approach it would have been caught

mpuncel:
agreed, generally I prefer kwarg init because it makes it hard to switch two fields of the same type. I should write a test for this.

PR #439, opened 2016-03-30
Failsafes for rc farms

anthonybishopric:
This PR adds two failsafes to the RC farms to prevent
catastrophic unscheduling of pods across the cluster. This
situation is generally hypothetical, but I'd rather be safe
than sorry.

The first situation is - the farm will panic if zero RCs have
been scheduled in Consul. This currently doesn't prevent any
specific behavior other than the farm abandoning working on RCs,
but is worth signaling that it has occurred.

The second situation would be genuinely catastrophic. If replica
counts were all 0, the RC farm would go on a killing spree across
the cluster, unscheduling all services. In practice this should
never happen.

### DIFF
--- a/pkg/rc/farm.go
+++ b/pkg/rc/farm.go
@@ -112,6 +112,8 @@ START_LOOP:
 		case rcFields := <-rcWatch:
 			rcf.logger.WithField("n", len(rcFields)).Debugln("Received replication controller update")
 
+			rcf.failsafe(rcFields)
+
 			// track which children were found in the returned set
 			foundChildren := make(map[fields.ID]struct{})
 			for _, rcField := range rcFields {
@@ -197,6 +199,31 @@ START_LOOP:
 			// now remove any children that were not found in the result set
 			rcf.releaseDeletedChildren(foundChildren)
 		}
+	}
+}
+
+func (rcf *Farm) failsafe(rcFields []fields.RC) {
+	// FAILSAFES. If no RCs are scheduled, or there are zero replicas of anything scheduled, panic
+	if len(rcFields) == 0 {
+		rcf.alerter.Alert(alerting.AlertInfo{
+			Description: "No RCs have been scheduled",
+			IncidentKey: "no_rcs_found",
+		})
+		panic("No RCs are scheduled at all. Create one RC to enable the farm. Panicking to escape a potentially bad situation.")
+	}
+	globalReplicaCount := 0
+	for _, rc := range rcFields {
+		if rc.ReplicasDesired > 0 {
+			return
+		}
+		globalReplicaCount += rc.ReplicasDesired
+	}
+	if globalReplicaCount == 0 {
+		rcf.alerter.Alert(alerting.AlertInfo{
+			Description: "All RCs have zero replicas requested",
+			IncidentKey: "zero_replicas_found",
+		})
+		panic("The sum of all replicas is 0. Panicking to escape a potentially bad situation")
 	}
 }
 
--- a/pkg/rc/farm_test.go
+++ b/pkg/rc/farm_test.go
@@ -1,0 +1,76 @@
+package rc
+
+import (
+	"testing"
+
+	"github.com/square/p2/pkg/alerting"
+	"github.com/square/p2/pkg/rc/fields"
+
+	. "github.com/square/p2/Godeps/_workspace/src/github.com/anthonybishopric/gotcha"
+)
+
+type failsafeAlerter struct {
+	savedInfo alerting.AlertInfo
+}
+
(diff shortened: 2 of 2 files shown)


pkg/rc/farm.go
```diff
@@ -200,6 +202,28 @@ START_LOOP:
...
 }
 
+func (rcf *Farm) failsafe(rcFields []fields.RC) {
+	// FAILSAFES. If no RCs are scheduled, or there are zero replicas of anything scheduled, panic
+	if len(rcFields) == 0 {
+		rcf.alerter.Alert(alerting.AlertInfo{
+			Description: "No RCs have been scheduled",
+			IncidentKey: "no_rcs_found",
+		})
+		panic("No RCs are scheduled at all. Create one RC to enable the farm. Panicking to escape a potentially bad situation.")
```
mpuncel: this might make it hard for someone to spin up a p2 cluster

pkg/rc/farm.go
```diff
@@ -200,6 +202,28 @@ START_LOOP:
...
+	}
+	globalReplicaCount := 0
+	for _, rc := range rcFields {
+		globalReplicaCount += rc.ReplicasDesired
```
mpuncel: you can early exit this if `rc.ReplicasDesired > 0` right? shouldn't need to account for negative numbers

pkg/rc/farm.go
anthonybishopric: When we get around to documenting it we can say to use `p2-rctl` to schedule an RC before starting the farms.

mpuncel:
:+1:

pkg/rc/farm.go
anthonybishopric: If this really becomes a PITA for testing though we can make it configurable.

pkg/rc/farm.go
anthonybishopric: Good point

pkg/rc/farm.go
```diff
@@ -200,6 +202,31 @@ START_LOOP:
...
+		panic("No RCs are scheduled at all. Create one RC to enable the farm. Panicking to escape a potentially bad situation.")
+	}
+	globalReplicaCount := 0
+	for _, rc := range rcFields {
+		if rc.ReplicasDesired > 0 {
+			return
+		}
+		globalReplicaCount += rc.ReplicasDesired
+	}
+	if globalReplicaCount == 0 {
```
mpuncel: should you remove this condition? The only way this would be untrue is if the total is negative which should probably be alerted for as well

mpuncel: I guess that makes alert message more difficult and it's an unlikely scenario, so i'm fine with you keeping it

petertseng:
Code change seems good.

You already noticed that the first check (no RCs at all) doesn't prevent any undesired behavior but given that 0 RCs will mean 0 total replica count it's good to distinguish the first case from the second.

The `failsafe` function seems like it would be super easy to write tests for, and I would appreciate those.

anthonybishopric:
@petertseng added test, was quite easy indeed

pkg/rc/farm_test.go
```diff
@@ -0,0 +1,76 @@
...
+	savedInfo alerting.AlertInfo
+}
+
+func (f *failsafeAlerter) Alert(info alerting.AlertInfo) error {
+	f.savedInfo = info
+	return nil
+}
+
+func TestNoRCsWillCausePanic(t *testing.T) {
+	alerter := &failsafeAlerter{}
```
mpuncel: you could use `alertingtest.NewRecorder()`

pkg/rc/farm_test.go
```diff
@@ -0,0 +1,76 @@
...
+		fields.RC{
+			ReplicasDesired: 0,
+		},
+	}
+
+	defer func() {
+		if p := recover(); p == nil {
+			t.Fatal("Should have panicked due to no replicas")
+		}
+		Assert(t).AreEqual("zero_replicas_found", alerter.savedInfo.IncidentKey, "should not have had a fired alert")
```
petertseng: message should say "should have had a fired alert"

anthonybishopric: good catch

pkg/rc/farm_test.go
anthonybishopric: gonna skip the import, it's only 5 lines of code if that's okay with you

mpuncel: just making sure you knew the options, fine with me.

petertseng:
I say :+1: , don't really have a strong opinion on the alerter imports

PR #442, opened 2016-03-31
roll: set remaining as target desire - new desire, not new healthy

petertseng:
We experienced the following with a deploy with desired 3 nodes and
minimum 1 node.

```json
{
  "msg": "Adding 2 new nodes and removing 2 old nodes",
  "new": {
    "Desired": 0,
    "Current": 0,
    "Real": 0,
    "Healthy": 0,
    "Unhealthy": 0,
    "Unknown": 0
  },
  "nextAdd": 2,
  "nextRemove": 2,
  "old": {
    "Desired": 3,
    "Current": 3,
    "Real": 3,
    "Healthy": 3,
    "Unhealthy": 0,
    "Unknown": 0
  },
  "pod": "censored",
},
{
  "msg": "Adding 2 new nodes and removing 2 old nodes",
  "new": {
    "Desired": 2,
    "Current": 2,
    "Real": 2,
    "Healthy": 1,
    "Unhealthy": 1,
    "Unknown": 0
  },
  "nextAdd": 2,
  "nextRemove": 2,
  "old": {
    "Desired": 1,
    "Current": 1,
    "Real": 1,
    "Healthy": 1,
    "Unhealthy": 0,
    "Unknown": 0
    },
  "pod": "censored",
}
```

The first is OK, but the second represents a serious error since now
there are four desired nodes which is greater than that requested. This
is because the minimum remaining count was being calculated as
`targetDesired - newHealthy`, which is incorrect if newHealthy hasn't
caught up with newDesired yet. Instead the correct calculation is
`targetDesired - newDesired`.

This was not visible before #414 and got merged, because a conditional
removed in that PR caused it to always be the case that newHealthy
would have caught up with newDesired. The conditional was removed in
11d7b03.

### DIFF
--- a/pkg/roll/update_test.go
+++ b/pkg/roll/update_test.go
@@ -51,6 +51,16 @@ func assertRollAlgorithmResults(t *testing.T, old, new, final, need, remove, add
 	Assert(t).AreEqual(gotAdd, add, "added nodes incorrect: "+message)
 }
 
+func TestRollAlgorithmDoesNotExceed(t *testing.T) {
+	// newHealthy < newDesired, and newHealthy >= minHealthy.
+	// In this case, we schedule the remaining nodes.
+	// We want to ensure that remaining == targetDesired - newDesired
+	// instead of targetDesired - newHealthy
+	gotRemove, gotAdd := rollAlgorithm(1, 1, 1, 2, 3, 1)
+	Assert(t).AreEqual(gotRemove, 1, "expected only one node to be removed")
+	Assert(t).AreEqual(gotAdd, 1, "expected only one node to be added")
+}
+
 func TestRollAlgorithmIncreases(t *testing.T) {
 	assertRollAlgorithmResults(t, 0, 0, 3, 2, 0, 1, "should schedule difference if increasing capacity from zero")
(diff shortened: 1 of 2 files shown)


pkg/roll/update_test.go
```diff
@@ -281,7 +297,7 @@ func SimulateRollingUpgradeDisable(t *testing.T, full, nonew bool) {
 		}
 
 		// calculate the next update
-		nextRemove, nextAdd := rollAlgorithm(len(old), len(new), target, target, minimum)
+		nextRemove, nextAdd := rollAlgorithm(len(old), len(new), target-len(new), len(new), target, minimum)
```
petertseng: the fuzz test is doing weird things to `oldDesired` because it may randomly choose to schedule on an old node or an empty node, which means `oldDesired != len(old)` sometimes... and also means that `nextRemove` is pretty much not being respected - it's not always the case that the number of nodes being removed will equal `nextRemove`

petertseng: I took the path of least resistance here, but I may need to reconsider either this test, or what `nextRemove` should mean in a mutable world

mpuncel:
Ship it!

PR #443, opened 2016-03-31
roll: Add fuzz test for strict removals

petertseng:
While making #442 we discovered that #408 kind of made the fuzz test do
odd things: It wasn't necessarily removing the number of nodes it said
it would be removing.

We'd like to keep the current fuzz test, as it seems to be useful in the
mutable world (where the old RC gets disabled thus doesn't necessarily
remove any nodes).

Now we'd like to add a new test that shows how things will progress in a
world of immutable deployments, where the old RC will indeed remove
exactly the number of nodes specified.

### DIFF
--- a/pkg/roll/update_test.go
+++ b/pkg/roll/update_test.go
@@ -205,15 +205,21 @@ func TestLockRCs(t *testing.T) {
 	Assert(t).IsNotNil(update.oldRCUnlocker, "should have kp.Unlocker for unlocking old rc")
 }
 
-func TestSimulateRollingUpgradeDisable(t *testing.T) {
+func TestSimulateRollingUpgrade(t *testing.T) {
 	rand.Seed(time.Now().UTC().UnixNano())
 	for i := 0; i < 20000; i++ {
-		SimulateRollingUpgradeDisable(t, false, false)
+		SimulateRollingUpgrade(t, false, false, false)
 	}
 }
 
-// this fuzzer tests the rolling upgrade rollAlgorithm in an environment where new
-// pods replace old pods (eg hoist artifacts). it creates an imaginary list of
+func TestSimulateRollingUpgradeStrictRemove(t *testing.T) {
+	rand.Seed(time.Now().UTC().UnixNano())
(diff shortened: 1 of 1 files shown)


petertseng:
going to change its name from `RollingUpgradeDisable` to just `RollingUpgrade`

pkg/roll/update_test.go
```diff
@@ -297,10 +319,29 @@ func SimulateRollingUpgradeDisable(t *testing.T, full, nonew bool) {
...
+		if !strictRemove {
+			Assert(t).AreEqual(nextRemove, nextAdd, "got asymmetric update, not expected for this fuzz test")
+		}
+
 		t.Logf("Scheduling %d new out of %v eligible\n", nextAdd, eligible)
 		Assert(t).AreNotEqual(nextAdd, 0, "got noop update, would never terminate")
+
+		if strictRemove {
+			// choose nodes from the old list, randomly, and remove the pod from them.
+			t.Logf("Unscheduling %d new out of %v old\n", nextRemove, old)
```
petertseng: old not new

anthonybishopric:
:+1: nice alteration, makes the test a bit clearer

petertseng:
Text change only, will merge when Travis succeeds

```diff
$ git diff -u 2f63e30 b686ae4
diff --git a/pkg/roll/update_test.go b/pkg/roll/update_test.go
index 35c389d..b351017 100644
--- a/pkg/roll/update_test.go
+++ b/pkg/roll/update_test.go
@@ -334,7 +334,7 @@ func SimulateRollingUpgrade(t *testing.T, full, nonew, strictRemove bool) {

                if strictRemove {
                        // choose nodes from the old list, randomly, and remove the pod from them.
-                       t.Logf("Unscheduling %d new out of %v old\n", nextRemove, old)
+                       t.Logf("Unscheduling %d out of %v old\n", nextRemove, old)
                        for _, index := range rand.Perm(len(old))[:nextRemove] {
                                nodes[old[index]] = 0
                                eligible = append(eligible, old[index])
```

PR #445, opened 2016-04-05
Allow updates to other clients to proceed

anthonybishopric:
This change makes it possible for many label watchers to be subscribed to the same aggregator without incurring penalties due to individual watches being slow.

### DIFF
--- a/pkg/labels/consul_aggregator.go
+++ b/pkg/labels/consul_aggregator.go
@@ -13,13 +13,12 @@ import (
 )
 
 // minimum required time between retrievals of label subtrees.
-var AggregationRateCap = 10 * time.Second
+var DefaultAggregationRate = 10 * time.Second
 
 // linked list of watches with control channels
 type selectorWatch struct {
 	selector labels.Selector
 	resultCh chan *[]Labeled
-	canceled chan struct{}
 	next     *selectorWatch
 }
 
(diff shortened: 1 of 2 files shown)


pkg/labels/consul_aggregator.go
```diff
@@ -13,13 +13,12 @@ import (
...
 
 // minimum required time between retrievals of label subtrees.
-var AggregationRateCap = 10 * time.Second
+var DefaultAggregationRate = 10 * time.Second
 
 // linked list of watches with control channels
 type selectorWatch struct {
 	selector labels.Selector
 	resultCh chan *[]Labeled
-	canceled chan struct{}
```
anthonybishopric: `canceled` is no longer necessary since failed sends on the channel due to closure will be handled gracefully.

pkg/labels/consul_aggregator.go
```diff
@@ -156,9 +155,26 @@ func (c *consulAggregator) Aggregate() {
...
+				// watchers. We then subsequently put the newer value into the channel.
+				//
+				// This approach has the effect of a slower watcher potentially missing updates,
+				// but also means one watcher can't cause a DoS of other watchers as the main
+				// aggregation goroutine waits.
+
+				// first drain the previous (stale) cached value if present...
+				select {
+				case <-watcher.resultCh:
+				default:
```
mpuncel: this might have been a better approach for me to use in the RC/RU update PR

mpuncel:
:+1:

PR #446, opened 2016-04-05
Label watch optimizations

anthonybishopric:
These are two commits that improve the performance of label watches, see commit messages for details.

### DIFF
--- a/pkg/labels/applicator.go
+++ b/pkg/labels/applicator.go
@@ -54,11 +54,6 @@ func (l Labeled) SameAs(o Labeled) bool {
 	return l.ID == o.ID && l.LabelType == o.LabelType
 }
 
-type WatchResult struct {
-	Labeled []Labeled
-	Valid   bool // Will be false if the underlying watcher has terminated
-}
-
 // General purpose backing store for labels and assignment
 // to P2 objects
 type Applicator interface {
@@ -86,7 +81,7 @@ type Applicator interface {
 	//
 	// The watch may be terminated by the underlying implementation without signaling on
 	// the quit channel - this will be indicated by the closing of the result channel. For
-	// this reason, the resulting WatchResult have a Valid flag to help determine if
-	// the channel has been closed.
-	WatchMatches(selector labels.Selector, labelType Type, quitCh chan struct{}) chan WatchResult
+	// this reason, callers should **always** verify that the channel is closed by checking
+	// the "ok" boolean or using `range`.
+	WatchMatches(selector labels.Selector, labelType Type, quitCh chan struct{}) chan []Labeled
 }
--- a/pkg/labels/applicator.go
+++ b/pkg/labels/applicator.go
@@ -54,6 +54,11 @@ func (l Labeled) SameAs(o Labeled) bool {
 	return l.ID == o.ID && l.LabelType == o.LabelType
 }
 
+type WatchResult struct {
+	Labeled []Labeled
+	Valid   bool // Will be false if the underlying watcher has terminated
+}
+
 // General purpose backing store for labels and assignment
 // to P2 objects
 type Applicator interface {
@@ -81,7 +86,7 @@ type Applicator interface {
 	//
 	// The watch may be terminated by the underlying implementation without signaling on
 	// the quit channel - this will be indicated by the closing of the result channel. For
-	// this reason, the safest way to process the result of this channel is to use
-	// it in a for loop. Alternatively, you may check for nullity of the result []Labeled
-	WatchMatches(selector labels.Selector, labelType Type, quitCh chan struct{}) chan *[]Labeled
+	// this reason, the resulting WatchResult have a Valid flag to help determine if
+	// the channel has been closed.
+	WatchMatches(selector labels.Selector, labelType Type, quitCh chan struct{}) chan WatchResult
 }
--- a/pkg/labels/consul_aggregator.go
+++ b/pkg/labels/consul_aggregator.go
@@ -18,8 +18,7 @@ var DefaultAggregationRate = 10 * time.Second
 // linked list of watches with control channels
 type selectorWatch struct {
 	selector labels.Selector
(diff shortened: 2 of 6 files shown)


pkg/labels/applicator.go
```diff
@@ -81,7 +86,7 @@ type Applicator interface {
 	//
 	// The watch may be terminated by the underlying implementation without signaling on
 	// the quit channel - this will be indicated by the closing of the result channel. For
-	// this reason, the safest way to process the result of this channel is to use
-	// it in a for loop. Alternatively, you may check for nullity of the result []Labeled
-	WatchMatches(selector labels.Selector, labelType Type, quitCh chan struct{}) chan *[]Labeled
+	// this reason, the resulting WatchResult have a Valid flag to help determine if
```
mpuncel: s/have/has

pkg/labels/applicator.go
```diff
@@ -54,6 +54,11 @@ func (l Labeled) SameAs(o Labeled) bool {
 	return l.ID == o.ID && l.LabelType == o.LabelType
 }
 
+type WatchResult struct {
+	Labeled []Labeled
+	Valid   bool // Will be false if the underlying watcher has terminated
```
mpuncel: is this necessary? can callers not use `range` or check `ok` when pulling a value off the channel?

anthonybishopric: `range` and `ok` would both be ways to check that the value is valid, but I assume most clients of this code will be using it in a `select` statement, given that it's a watch that can be interrupted. 

I know it's maybe a bit paranoid, but I really don't want clients accidentally treating an empty result the same as a nil result set.

mpuncel: aren't clients going to have to do one or the other anyway to exit their loops?

That said i'm okay with this as a safety mechanism

pkg/labels/consul_aggregator.go
```diff
@@ -127,55 +131,21 @@ func (c *consulAggregator) Aggregate() {
...
-				if err != nil {
-					c.logger.WithErrorAndFields(err, logrus.Fields{
-						"key":   kvp.Key,
-						"value": string(kvp.Value),
-					}).Errorln("Invalid key encountered, skipping this value")
-					continue
-				}
-				c.labeledCache[i] = val
-			}
+			c.watcherLock.Lock()
```
mpuncel: might be worth extracting this to a function, that would let you `defer` the unlock here. It's correct as written but if any error scenarios are introduced it would be easy to make a mistake

mpuncel: actually this is already very concise i was just thrown off by the sea of red. don't worry about it

pkg/labels/consul_aggregator.go
```diff
@@ -184,7 +154,56 @@ func (c *consulAggregator) Aggregate() {
...
+			// we purposely don't case outErrors here, since loopTime lets us
+			// back off of Consul watches. If an error repeatedly were occurring,
+			// we could end up in a nasty busy loop.
+		}
+	}
+}
+
+func (c *consulAggregator) fillCache(pairs api.KVPairs) {
+	cache := make([]Labeled, len(pairs))
+	c.labeledCache = cache
```
mpuncel: it seems weird to set this before we've populated any values. I think the behavior is correct because this is done while holding the locks but maybe do this after?

anthonybishopric: that seems like a sensible change

pkg/labels/consul_aggregator.go
```diff
@@ -184,7 +154,56 @@ func (c *consulAggregator) Aggregate() {
...
+				"value": string(kvp.Value),
+			}).Errorln("Invalid key encountered, skipping this value")
+			continue
 		}
+		c.labeledCache[i] = val
+	}
+}
 
+// this must be called within the watcherLock mutex.
+func (c *consulAggregator) doMatch(watcher *selectorWatch) {
```
mpuncel: nit: i'd maybe call this `sendMatches` or something because it does the actual sending on the channel, wasn't obvious from the name.

pkg/labels/consul_applicator.go
```diff
@@ -229,7 +231,9 @@ func convertLabeledToKVP(l Labeled) (*api.KVPair, error) {
 // to the cost of querying for this subtree on any sizeable fleet of machines. Instead, preparers should
 // use the httpApplicator from a server that exposes the results of this (or another)
 // implementation's watch.
-func (c *consulApplicator) WatchMatches(selector labels.Selector, labelType Type, quitCh chan struct{}) chan *[]Labeled {
+func (c *consulApplicator) WatchMatches(selector labels.Selector, labelType Type, quitCh chan struct{}) chan WatchResult {
+	c.aggregatorMux.Lock()
```
mpuncel: good call

mpuncel:
looks great! just some nits

anthonybishopric:
i like your nits :)

pkg/labels/consul_aggregator.go
anthonybishopric: :+1:

pkg/labels/applicator.go
anthonybishopric: Changed back to `[]Labeled`, realized that this safety guarantee can be provided in `select` statements by using `case res, ok := <- resCh`. Adding that to the documentation.

pkg/labels/consul_aggregator_test.go
```diff
@@ -107,8 +105,9 @@ func TestQuitAggregateBeforeResults(t *testing.T) {
 	aggreg.Quit()
 
 	select {
-	case labeled := <-res:
-		Assert(t).IsNotNil(labeled, "Should not have received any results")
+	case labeled, ok := <-res:
+		Assert(t).IsFalse(ok, "should have been okay")
```
petertseng: not ok?

PR #447, opened 2016-04-05
Return a specific error type if oldRCSelector is ambiguous

mpuncel:
One way to specify the old RC to use when creating an RU is by using a
label selector. If the selector matches more than one RC, an error is
returned.

This commit makes the roll store return a special error type in this
case so that callers can detect this error type and generate more useful
error messages. In Square's application of this function, an error of
this type means that a deploy is already in progress and thus there are
two RCs that match the selector.

### DIFF
--- a/pkg/kp/rollstore/consul_store.go
+++ b/pkg/kp/rollstore/consul_store.go
@@ -392,11 +392,7 @@ func (s consulStore) CreateRollingUpdateFromOneMaybeExistingWithLabelSelector(
 
 	var oldRCID rc_fields.ID
 	if len(matches) > 1 {
-		return roll_fields.Update{}, util.Errorf(
-			"Can't create update: old RC selector %s was ambiguous, provided %d matches",
-			oldRCSelector.String(),
-			len(matches),
-		)
+		return roll_fields.Update{}, AmbiguousRCSelector
 	} else if len(matches) == 1 {
 		oldRCID = rc_fields.ID(matches[0].ID)
 	} else {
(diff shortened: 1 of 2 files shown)


pkg/kp/rollstore/consul_store.go
```diff
@@ -392,11 +392,7 @@ func (s consulStore) CreateRollingUpdateFromOneMaybeExistingWithLabelSelector(
 
 	var oldRCID rc_fields.ID
 	if len(matches) > 1 {
-		return roll_fields.Update{}, util.Errorf(
-			"Can't create update: old RC selector %s was ambiguous, provided %d matches",
-			oldRCSelector.String(),
-			len(matches),
-		)
+		return roll_fields.Update{}, AmbiguousRCSelector
```
mpuncel: it's somewhat unfortunate that we lose the number of matches here, but the caller will know the selector which we can easily use to debug with p2-rctl

mpuncel: I could implement an error type and return it, but that opens a different can of worms re: nil checking but maybe worth it?

rudle: > which we can easily use to debug with p2-rctl
agree

It seems unlikely that this number will be anything other than 1 or 2. Is there a case that I'm missing?

rudle:
:ship:

pkg/kp/rollstore/consul_store.go
mpuncel: yeah i agree

PR #448, opened 2016-04-05
Log the amount of time it takes to process an RC update in the farm

mpuncel:


mpuncel:
approved in person by @anthonybishopric

PR #449, opened 2016-04-05
log the time spent processing RC updates as a string instead of nanos…

mpuncel:
…econd integer


rudle:
:+1: looks like it works to me. How did you test this?

mpuncel:
just by running an equivalent on play: http://play.golang.org/p/5odM22XwUA

mpuncel:
although maybe integers are easier to do math on

rudle:
I think it's fine as is. We definitely want some units in the output and the quantities should be small enough to allow mental arithmetic.

PR #451, opened 2016-04-06
Adds the ability to collect metrics in RC farms

mpuncel:
The only metric currently collected is "rc_processing_time" which
represents the amount of time it takes for the farm to respond to an
update from the rc watch.


pkg/rc/farm.go
```diff
@@ -4,6 +4,7 @@ import (
 	"sync"
 	"time"
 
+	"github.com/rcrowley/go-metrics"
```
anthonybishopric: godep me

anthonybishopric:
Neat! Perhaps we should move the logic for this into a RC-metrics object though - I could see a number of different things I'd want to track over time and code that manages the RC farm should probably be separate from the metrics publication sausage.

Something like

```
rcf.rcMetrics.setRCProcessingTime(val)
```

What do you think?

pkg/rc/farm.go
```diff
@@ -44,6 +46,9 @@ type Farm struct {
 	logger     logging.Logger
 	alerter    alerting.Alerter
 	rcSelector klabels.Selector
+
+	metrics         *rcmetrics.Metrics
+	metricsRegistry metrics.Registry
```
anthonybishopric: This field is unused now right?

anthonybishopric:
:ship:

pkg/rc/rcmetrics/metrics.go
```diff
@@ -0,0 +1,41 @@
...
+)
+
+type Metrics struct {
+	Registry metrics.Registry
+	Logger   logging.Logger
+}
+
+func (m *Metrics) SetRegistry(registry metrics.Registry) error {
+	m.Registry = registry
+	return m.Registry.Register(RCProcessingTimeMetric, metrics.NewHistogram(metrics.NewExpDecaySample(1028, 0.015)))
```
petertseng: how these numbers determined?

mpuncel: This went away in a future PR, but it seems like those are just the standard values for a histogram using an exponential decay sample. This is based on docs and also our internal metrics system has these hardcoded

PR #452, opened 2016-04-07
Add p2-rctl-server to integration test

anthonybishopric:
This change expands the role of the integration test to run
a p2-rctl-server. Instead of scheduling a `hello` pod directly
on the host, we instead create a `hello` replication controller
and verify that hello gets correctly scheduled on the host
running the integration test.

One of the new conditions verified is that the `hello` pod gets
labeled with the RC that scheduled it in the label tree. This
exercises the relatively new label aggregation and watch behavior.

This change was relatively painless. The next step is to actually
attempt a rolling update through the same p2-rctl-server.

### DIFF
--- a/integration/single-node-slug-deploy/check.go
+++ b/integration/single-node-slug-deploy/check.go
@@ -14,13 +14,19 @@ import (
 	"path"
 	"path/filepath"
 	"runtime"
+	"strings"
 	"time"
 
 	"github.com/square/p2/pkg/health"
 	"github.com/square/p2/pkg/kp"
+	"github.com/square/p2/pkg/labels"
 	"github.com/square/p2/pkg/pods"
 	"github.com/square/p2/pkg/preparer"
+	"github.com/square/p2/pkg/rc"
+	"github.com/square/p2/pkg/rc/fields"
 	"github.com/square/p2/pkg/util"
+
+	klabels "github.com/square/p2/Godeps/_workspace/src/k8s.io/kubernetes/pkg/labels"
 )
 
 const preparerStatusPort = 32170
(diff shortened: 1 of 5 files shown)


integration/single-node-slug-deploy/check.go
```diff
@@ -78,9 +87,19 @@ func main() {
...
+	rcID, err := createHelloReplicationController(tempdir)
+	if err != nil {
+		log.Fatalf("Could not create hello pod / rc: %s\n", err)
+	}
+	log.Printf("Created RC #%s for hello\n", rcID)
+
+	err = waitForPodLabeledWithRC(klabels.Everything().Add(rc.RCIDLabel, klabels.EqualsOperator, []string{rcID.String()}), rcID)
 	if err != nil {
-		log.Fatalf("Could not generate hello pod: %s\n", err)
+		log.Fatalf("Failed waiting for pods labeled with the given RC", err)
```
mpuncel: you probably want `Fatalln` or put a format placeholder

integration/single-node-slug-deploy/check.go
```diff
@@ -315,7 +334,28 @@ func executeBootstrap(preparerManifest, consulManifest string) error {
...
+func scheduleRCTLServer(dir string) error {
+	p2RCTLServerPath, err := exec.Command("which", "p2-rctl-server").CombinedOutput()
+	if err != nil {
+		return fmt.Errorf("Could not find p2-rctl-server on PATH")
+	}
+	chomped := strings.TrimSpace(string(p2RCTLServerPath))
+	if _, err = os.Stat(chomped); os.IsNotExist(err) {
+		return fmt.Errorf("%v does not exist", chomped)
+	}
+	cmd := exec.Command("p2-bin2pod", "--work-dir", dir, chomped)
```
mpuncel: forgot about this guy

mpuncel:
👍  this is awesome

integration/single-node-slug-deploy/check.go
```diff
@@ -333,21 +373,81 @@ func postHelloManifest(dir string) error {
...
+	cmd.Stderr = &out
+	err = cmd.Run()
+	if err != nil {
+		return fields.ID(""), fmt.Errorf("Couldn't create replication controller for hello: %s %s", out.String(), err)
+	}
+	var rctlOut struct {
+		ID string `json:"id"`
+	}
+
+	err = json.Unmarshal(out.Bytes(), &rctlOut)
```
petertseng: you did it better than I did (see my code that just parses the UUID out of the message with a regex...)

PR #453, opened 2016-04-07
simplify the registry interface used by the farm

mpuncel:


pkg/rc/farm.go
```diff
@@ -89,7 +88,7 @@ func NewFarm(
 // Set the metrics registry to use (see
 // https://godoc.org/github.com/rcrowley/go-metrics#Registry). If set,
 // performance metrics will be recorded to the registry.
-func (rcf *Farm) SetMetricsRegistry(registry metrics.Registry) error {
+func (rcf *Farm) SetMetricsRegistry(registry rcmetrics.MetricsRegistry) error {
```
anthonybishopric: Out of curiosity, why isn't this method `SetMetrics(metrics rcmetrics.Metrics)`?

mpuncel: what's the benefit of that? That would also work but i guess i like the logging in the implementation of `RecordRCProcessingTime`, and we would probably mutate the `rcmetrics.Metrics` instance passed in to layer on metrics

mpuncel: Is it that you think people might want to do things with rc_processing_time other than put them in a Histogram?

anthonybishopric: Maybe this is a broader question - why does anything in this file need to be aware of the underlying `rcmetrics.MetricsRegistry`? As far as I can tell, it's an implementation detail of `Metrics`, not of the RC farm. The logger appears to be passed in directly without alteration from the init of the farm to the init of `&rcmetrics.Metrics{}` on line 82.

mpuncel: that actually might be easier for my use case though now that I think about it.

anthonybishopric:
:+1:

PR #454, opened 2016-04-08
Remove a common log line in rc farm

mpuncel:
In steady state, all RCs are locked and the RC farm was emitting a log
message for each RC it encounters that is already locked. This
contributes to a lot of log lines that are not very informative. This
commit removes that


anthonybishopric:
:ship:

PR #455, opened 2016-04-08
rollstore: If created an old RC, apply new RC's labels

petertseng:
Since we expect the old RC to be a copy of the new one, then we should
copy the labels as well, except that was not done.

Not copying the labels results in some schemes that assume the label
presence to not be able to find the old RC (since it has no labels).

### DIFF
--- a/pkg/kp/rollstore/consul_store.go
+++ b/pkg/kp/rollstore/consul_store.go
@@ -419,6 +419,12 @@ func (s consulStore) CreateRollingUpdateFromOneMaybeExistingWithLabelSelector(
 
 			// Any labels we wrote will be deleted by rcstore.Delete()
 		}
+
+		// Copy the new RC labels to the old RC as well
+		err = s.labeler.SetLabels(labels.RC, oldRCID.String(), newRCLabels)
+		if err != nil {
+			return roll_fields.Update{}, err
+		}
 	}
 
 	// Lock the old RC to guarantee that no new updates can use it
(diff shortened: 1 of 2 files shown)


petertseng:
going to add a line to the test.

petertseng:
added

pkg/kp/rollstore/consul_store.go
```diff
@@ -476,6 +480,13 @@ func (s consulStore) CreateRollingUpdateFromOneMaybeExistingWithLabelSelector(
 	if err != nil {
 		return roll_fields.Update{}, err
 	}
+	if createdOld {
```
anthonybishopric: Why create the labels down here, not up after the actual creation statement?

petertseng: The reason for delaying label creating for the new RC is "Now that we know there are no RUs in progress, and we have the update creation locks, we can safely apply labels.". Does the same reasoning apply to the old RC as well? If it does not, the labels can be created earlier. I will attempt to reason about this question.

mpuncel: I think you can just create the labels after creating the old RC.

petertseng:
labeling immediately after old RC creation (much smaller change now! this is good!)

pkg/kp/rollstore/consul_store.go
```diff
@@ -410,6 +410,12 @@ func (s consulStore) CreateRollingUpdateFromOneMaybeExistingWithLabelSelector(
 			return roll_fields.Update{}, err
 		}
 
+		// Copy the new RC labels to the old RC as well
```
mpuncel: you should do this after `cleanupOldRC` func is set so that the label set failure cleans up the RC

petertseng: clever!

it is moved to be after `cleanupOldRC` set

mpuncel:
👍

PR #456, opened 2016-04-08
Use the metrics.Registry interface for rc farm metrics

mpuncel:
This means less work for callers using the RC Farm library for setting
up metrics, and it means that the RC farm can add metrics at any time
without callers needing to write any code.


anthonybishopric:
:+1:

PR #457, opened 2016-04-11
Add basic pcstore.Store interface and PodCluster struct

mpuncel:


mpuncel:
only did `Get()` and `Create()` for now but undoubtedly we'll want things like `List()` and `Watch()` soon

pkg/kp/pcstore/store.go
```diff
@@ -0,0 +1,13 @@
...
+
+import (
+	"github.com/square/p2/pkg/pc/fields"
+	"github.com/square/p2/pkg/types"
+)
+
+const podClusterTree string = "pod_clusters"
+
+type Store interface {
+	Create(podId types.PodID, availabilityZone fields.AvailabilityZone, clusterName fields.ClusterName) (fields.PodCluster, error)
```
mpuncel: I need Annotations here

I deliberately didn't put pod selector here because it will be constructed from the other types. Maybe someday we'll want more flexbility

anthonybishopric:
:+1:

pkg/pc/fields/fields.go
```diff
@@ -0,0 +1,48 @@
...
+	// The ID of the pods that the cluster contains
+	PodID types.PodID `json:"pod_id"`
+
+	// Represents a region the pod cluster inhabits. P2 doesn't use this
+	// value but it is useful for implementations that care about
+	// geographical location of pod clusters
+	AvailabilityZone AvailabilityZone `json:"availability_zone"`
+
+	// Human-readable name for the pod cluster. Must be unique within a
+	// (PodID, AvaialibilityZone) space
```
petertseng: there are too many a's or i's here, or something

PR #459, opened 2016-04-11
Implement the pcstore.Store interface backed by Consul

mpuncel:


pkg/kp/pcstore/consul_store.go
```diff
@@ -0,0 +1,109 @@
...
+	jsonPC, err := json.Marshal(pc)
+	if err != nil {
+		// Probably the annotations don't marshal to JSON
+		return fields.PodCluster{}, util.Errorf("Unable to marshal pod cluster as JSON: %s", err)
+	}
+
+	success, _, err := s.kv.CAS(&api.KVPair{
+		Key:   key,
+		Value: jsonPC,
+		// the chance of the UUID already existing is vanishingly small, but
```
rudle: I like this comment, but I think it belongs above line 57.

pkg/pc/fields/fields.go
```diff
@@ -1,16 +1,26 @@
...
 type ID string
 type AvailabilityZone string
 type ClusterName string
 type Annotations map[string]interface{}
 
+// label keys used by pod selector
+const (
+	AvailabilityZoneLabel = "availability_zone"
+	ClusterNameLabel      = "cluster_name"
+	// Pod id label name is imported from the rcstore package
```
rudle: What does this mean? I don't see `rcstore` imported by this file. Would it be a bad idea to define a more local constant in this file that's equal to the PodLabelName in `rstore`?

mpuncel: that's a great idea

rudle:
A couple of small comments. This looks mostly sound to me.

PR #460, opened 2016-04-12
provide test implementation of pcstore.Store interface

mpuncel:


petertseng:
I say 👍 , seems super easy.

PR #462, opened 2016-04-13
Add metrics to the Consul label aggregator

anthonybishopric:
Adds three metrics per aggregator; the number of cached `labels.Labeled`, the number of outstanding watchers, and the number of watchers that missed a selector update.

### DIFF
--- a/pkg/labels/consul_aggregator.go
+++ b/pkg/labels/consul_aggregator.go
@@ -1,6 +1,7 @@
 package labels
 
 import (
+	"fmt"
 	"sync"
 	"time"
 
@@ -10,10 +11,16 @@ import (
 	"github.com/square/p2/Godeps/_workspace/src/github.com/Sirupsen/logrus"
 	"github.com/square/p2/Godeps/_workspace/src/github.com/hashicorp/consul/api"
 	"github.com/square/p2/Godeps/_workspace/src/k8s.io/kubernetes/pkg/labels"
+
+	"github.com/square/p2/Godeps/_workspace/src/github.com/rcrowley/go-metrics"
 )
 
 // minimum required time between retrievals of label subtrees.
 var DefaultAggregationRate = 10 * time.Second
+
+type MetricsRegistry interface {
+	Register(metricName string, metric interface{}) error
+}
 
 // linked list of watches with control channels
 type selectorWatch struct {
@@ -44,6 +51,16 @@ func (s *selectorWatch) delete(w *selectorWatch) {
 	}
 }
 
+func (s *selectorWatch) len() int {
+	watch := s
+	ret := 0
+	for watch != nil {
+		ret++
+		watch = watch.next
+	}
+	return ret
+}
+
 type consulAggregator struct {
 	logger          logging.Logger
 	labelType       Type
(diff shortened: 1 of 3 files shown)


pkg/labels/consul_aggregator.go
```diff
@@ -174,10 +223,11 @@ func (c *consulAggregator) fillCache(pairs api.KVPairs) {
 		cache[i] = labeled
 	}
 	c.labeledCache = cache
+	c.metCacheSize.Update(int64(len(cache)))
 }
 
 // this must be called within the watcherLock mutex.
-func (c *consulAggregator) sendMatches(watcher *selectorWatch) {
+func (c *consulAggregator) sendMatches(watcher *selectorWatch) bool {
```
rudle: Do you think it's more gopher-like to use an error here? the bool is a bit C

pkg/labels/consul_aggregator.go
```diff
@@ -54,16 +71,38 @@ type consulAggregator struct {
...
+func NewConsulAggregator(labelType Type, kv consulutil.ConsulLister, logger logging.Logger, metReg MetricsRegistry) *consulAggregator {
+	if metReg == nil {
+		metReg = metrics.NewRegistry()
+	}
+	watchCount := metrics.NewGauge()
+	watchSendMiss := metrics.NewGauge()
+	cacheSize := metrics.NewGauge()
+	metReg.Register(fmt.Sprintf("%v_aggregate_watches", labelType.String()), watchCount)
+	metReg.Register(fmt.Sprintf("%v_aggregate_send_miss", labelType.String()), watchSendMiss)
+	metReg.Register(fmt.Sprintf("%v_aggregate_cache_size", labelType.String()), cacheSize)
```
mpuncel: intentionally not checking `err` on these? the only err i think there could be is a duplicate metric which should be fine

mpuncel:
👍

pkg/labels/consul_aggregator.go
```diff
@@ -111,6 +151,11 @@ func (c *consulAggregator) removeWatch(watch *selectorWatch) {
 	} else {
 		c.watchers.delete(watch)
 	}
+	if c.watchers != nil {
+		c.metWatchCount.Update(int64(c.watchers.len()))
+	} else {
+		c.metWatchCount.Update(0)
```
petertseng: they way you've written `len` it may actually have been safe to call `len` on nil, but I'll not complain about this since this way is sure to be safe

petertseng:
👍

rudle:
:boat:

pkg/labels/consul_aggregator.go
petertseng: Because you asked me about it - yes it would have been safe http://play.golang.org/p/6uR_1sgNRT

My understanding of the situation is even though it is nil, the nil is a typed nil, therefore you can call methods on it, so long as the methods do not attempt to deref the nil (which `len` does not).

The case I usually use an example for this is https://github.com/golang/crypto/blob/master/openpgp/packet/config.go where the config checks if it's nil so that calling these methods on a nil config gives you the defaults.

Interesting but I'm not a strong advocate for changing the code since as written it's more obvious.

PR #464, opened 2016-04-15
Mpuncel/pod cluster labels

mpuncel:


pkg/kp/pcstore/consul_store.go
```diff
@@ -75,7 +81,18 @@ func (s *consulStore) Create(
 	if !success {
 		return fields.PodCluster{}, util.Errorf("Could not set pod cluster at path '%s'", key)
 	}
-	return pc, nil
+
+	// Should we delete the PC if the labels fail?
```
anthonybishopric: Yes, lets

pkg/kp/rcstore/consul_store.go
```diff
@@ -85,6 +85,8 @@ func NewConsul(client *api.Client, retries int) *consulStore {
 
 func (s *consulStore) Create(manifest pods.Manifest, nodeSelector klabels.Selector, podLabels klabels.Set) (fields.RC, error) {
 	rc, err := s.innerCreate(manifest, nodeSelector, podLabels)
+
+	// TODO: measure whether retries are is important in practice
```
anthonybishopric: Good candidate for a metric

pkg/kp/pcstore/consul_store.go
```diff
@@ -75,9 +82,25 @@ func (s *consulStore) Create(
 	if !success {
 		return fields.PodCluster{}, util.Errorf("Could not set pod cluster at path '%s'", key)
 	}
+
+	// Should we delete the PC if the labels fail?
+	err = s.setLabelsForPC(pc)
+	if err != nil {
+		return fields.PodCluster{}, s.Delete(pc.ID)
```
anthonybishopric: You should still return `err` from the label set failure.

mpuncel: yeah, although now i don't know what to do if delete fails. I guess i'll add a logger and log it. we should probably throw exception reporting integration on these stores as well

anthonybishopric:
:+1:

pkg/kp/pcstore/consul_store.go
```diff
@@ -75,9 +82,27 @@ func (s *consulStore) Create(
 	if !success {
 		return fields.PodCluster{}, util.Errorf("Could not set pod cluster at path '%s'", key)
 	}
+
+	// Should we delete the PC if the labels fail?
```
petertseng: probably rm comment

PR #465, opened 2016-04-15
Refactor: move Session abstraction to consulutil

anthonybishopric:
We'll leave aside for the moment that the package is called `consulutil`. The `kp` package currently owns the responsibility of basic Consul interaction as well as persistence logic for Pods. Several functions related to locking and health management also go on `kp.Store`, which do not belong.

This is the first step towards moving lock logic out of `kp` and into a generic Consul management package that can be shared between `kp` and the other stores.


pkg/kp/consulutil/session_controller.go
```diff
@@ -0,0 +1,161 @@
+package consulutil
+
```
anthonybishopric: This file is not new content, it was moved from `kp/consulutil/session.go`

pkg/kp/consulutil/session.go
```diff
@@ -1,161 +1,175 @@
...
-	output chan<- string,
-	done chan struct{},
-	logger logging.Logger,
-) {
-	logger.NoFields().Info("session manager: starting up")
+// attempts to acquire the lock on the targeted key. keys used for
+// locking/synchronization should be ephemeral (ie their value does not matter
+// and you don't care if they're deleted)
+func (s Session) Lock(key string) (Unlocker, error) {
+	success, _, err := s.client.KV().Acquire(&api.KVPair{
```
anthonybishopric: This content was moved from `kp/session.go`

pkg/kp/kv.go
```diff
@@ -25,6 +25,10 @@ type ManifestResult struct {
 	Path     string
 }
 
+type Lockable interface {
```
mpuncel: is this used anywhere?

anthonybishopric: No, will remove

mpuncel:
Looks like a great reorganization! 👍

PR #467, opened 2016-04-15
Check for existing Pod Clusters on creation

anthonybishopric:
This adds a required Session argument to pod cluster creation, ensuring that the pod cluster for a given 3-ple of pod ID, availability zone, and cluster name does not already exist.

### DIFF
--- a/pkg/kp/pcstore/consul_store.go
+++ b/pkg/kp/pcstore/consul_store.go
@@ -49,7 +49,7 @@ func (s *consulStore) Create(
 ) (fields.PodCluster, error) {
 	id := fields.ID(uuid.New())
 
-	unlocker, err := s.lockForUpdateCreation(podID, availabilityZone, clusterName, session)
+	unlocker, err := s.lockForCreation(podID, availabilityZone, clusterName, session)
 	if err != nil {
 		return fields.PodCluster{}, err
 	}
@@ -159,7 +159,7 @@ func (s *consulStore) pcPath(pcID fields.ID) (string, error) {
 	return path.Join(podClusterTree, pcID.String()), nil
 }
 
-func (s *consulStore) lockForUpdateCreation(podID types.PodID,
+func (s *consulStore) lockForCreation(podID types.PodID,
 	availabilityZone fields.AvailabilityZone,
 	clusterName fields.ClusterName,
 	session Session) (consulutil.Unlocker, error) {
--- a/pkg/kp/pcstore/consul_store.go
+++ b/pkg/kp/pcstore/consul_store.go
@@ -4,6 +4,7 @@ import (
 	"encoding/json"
 	"path"
 
+	"github.com/square/p2/pkg/kp"
 	"github.com/square/p2/pkg/kp/consulutil"
 	"github.com/square/p2/pkg/labels"
 	"github.com/square/p2/pkg/pc/fields"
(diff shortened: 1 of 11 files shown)


pkg/kp/pcstore/consul_store.go
```diff
@@ -44,9 +45,24 @@ func (s *consulStore) Create(
 	clusterName fields.ClusterName,
 	podSelector klabels.Selector,
 	annotations fields.Annotations,
+	session Session,
 ) (fields.PodCluster, error) {
 	id := fields.ID(uuid.New())
 
+	unlocker, err := s.lockForUpdateCreation(podID, availabilityZone, clusterName, session)
```
anthonybishopric: I just realized this was named this way after "rolling updates" not "update or creation"

pkg/kp/pcstore/consul_store.go
```diff
@@ -143,6 +159,42 @@ func (s *consulStore) pcPath(pcID fields.ID) (string, error) {
 	return path.Join(podClusterTree, pcID.String()), nil
 }
 
+func (s *consulStore) lockForUpdateCreation(podID types.PodID,
```
mpuncel: probably should just be `lockForCreation`

pkg/kp/pcstore/consul_store.go
```diff
@@ -44,9 +45,24 @@ func (s *consulStore) Create(
...
+		return fields.PodCluster{}, err
+	}
+	defer unlocker.Unlock()
+
+	existing, err := s.FindWhereLabeled(podID, availabilityZone, clusterName)
+	if err != nil {
+		return fields.PodCluster{}, util.Errorf("Couldn't determine if pod cluster exists already: %v", err)
+	}
+	if len(existing) > 0 {
+		return existing[0], util.Errorf("Pod cluster already exists for %v", pcCreateLockPath(podID, availabilityZone, clusterName))
```
mpuncel: maybe give the ID of the pod cluster too?

mpuncel: i guess that forces you to deal with `len(existing)` which is annoying so feel free to ignore

pkg/kp/pcstore/consul_store.go
```diff
@@ -143,6 +159,42 @@ func (s *consulStore) pcPath(pcID fields.ID) (string, error) {
...
+	availabilityZone fields.AvailabilityZone,
+	clusterName fields.ClusterName) string {
+	return path.Join(kp.LOCK_TREE, podID.String(), availabilityZone.String(), clusterName.String())
+}
+
+func (s *consulStore) FindWhereLabeled(podID types.PodID,
+	availabilityZone fields.AvailabilityZone,
+	clusterName fields.ClusterName) ([]fields.PodCluster, error) {
+
+	sel := klabels.Everything().
```
mpuncel: we should probably put the label set in an accessible place and then use `labelSet.AsSelector()`

pkg/kp/pcstore/consul_store_test.go
```diff
@@ -217,3 +236,22 @@ func consulStoreWithFakeKV() *consulStore {
 		applicator: labels.NewFakeApplicator(),
 	}
 }
+
+type fakeSession struct {
```
mpuncel: any chance of reusing `fakeSession` from `pkg/kp/kptest`?

mpuncel: i dont' feel strongly

pkg/kp/pcstore/consul_store.go
anthonybishopric: Easy to add the first one at least

mpuncel:
looks great just nits 👍

pkg/kp/pcstore/consul_store_test.go
anthonybishopric: Didn't realize it existed, will remove this one

pkg/kp/pcstore/pcstoretest/fake_pcstore.go
```diff
@@ -56,3 +57,19 @@ func (p *FakePCStore) Delete(id fields.ID) error {
 	delete(p.podClusters, id)
 	return nil
 }
+
+func (p *FakePCStore) FindWhereLabeled(
```
mpuncel: might be worth putting a `labels.NewFakeApplicator()` onto the fake store and actually write/delete/query labels so if the label set ever diverges from the fields the behavior will stay correct, but not necessary today.

mpuncel: or maybe this fake store could just be the real store but with a fake consul api client put in? that'd be cool but can be a different PR

anthonybishopric: I like the idea of using a fake applicator, since there will be needs for a fake pcstore. Will save for a future change though.

mpuncel:
👍

PR #468, opened 2016-04-15
use a metric to track retries in label applicator

mpuncel:


anthonybishopric:
👍 mod the test failure

PR #469, opened 2016-04-15
pcstore implements Watch()

rudle:
This is an early pass at implement Watch() for pcstore. It is explicitly missing a filter on the ID parameter inside of the watch function. I'm likely to do the naive thing unless someone suggests something more clever.

### DIFF
--- a/pkg/kp/consulutil/watch.go
+++ b/pkg/kp/consulutil/watch.go
@@ -22,6 +22,7 @@ func WatchPrefix(
 	done <-chan struct{},
 	outErrors chan<- error,
 ) {
+	// TODO return this for closing on the client side
 	defer close(outPairs)
 	var currentIndex uint64
 	timer := time.NewTimer(time.Duration(0))
--- a/pkg/kp/pcstore/consul_store.go
+++ b/pkg/kp/pcstore/consul_store.go
@@ -4,37 +4,43 @@ import (
 	"encoding/json"
 	"path"
 
-	"github.com/square/p2/pkg/kp/consulutil"
-	"github.com/square/p2/pkg/labels"
-	"github.com/square/p2/pkg/pc/fields"
-	"github.com/square/p2/pkg/types"
-	"github.com/square/p2/pkg/util"
-
 	"github.com/square/p2/Godeps/_workspace/src/github.com/hashicorp/consul/api"
 	"github.com/square/p2/Godeps/_workspace/src/github.com/pborman/uuid"
 	klabels "github.com/square/p2/Godeps/_workspace/src/k8s.io/kubernetes/pkg/labels"
+	"github.com/square/p2/pkg/kp/consulutil"
+	"github.com/square/p2/pkg/labels"
+	"github.com/square/p2/pkg/logging"
+	"github.com/square/p2/pkg/pc/fields"
+	"github.com/square/p2/pkg/types"
+	"github.com/square/p2/pkg/util"
 )
 
 type consulKV interface {
 	Get(key string, opts *api.QueryOptions) (*api.KVPair, *api.QueryMeta, error)
 	CAS(pair *api.KVPair, opts *api.WriteOptions) (bool, *api.WriteMeta, error)
 	Delete(key string, w *api.WriteOptions) (*api.WriteMeta, error)
+	List(prefix string, opts *api.QueryOptions) (api.KVPairs, *api.QueryMeta, error)
 }
 
 type consulStore struct {
-	applicator labels.Applicator
 	kv         consulKV
+	applicator labels.Applicator
+
+	logger logging.Logger
 }
+
+var ErrNoPodCluster error
 
 var _ Store = &consulStore{}
 
 // NOTE: The "retries" concept is mimicking what is built in rcstore.
 // TODO: explore transactionality of operations and returning errors instead of
 // using retries
-func NewConsul(client api.Client, retries int) Store {
+func NewConsul(client api.Client, retries int, logger *logging.Logger) Store {
 	return &consulStore{
-		kv:         client.KV(),
 		applicator: labels.NewConsulApplicator(&client, retries),
+		kv:         client.KV(),
+		logger:     *logger,
 	}
 }
 
@@ -189,6 +195,123 @@ func (s *consulStore) FindWhereLabeled(podID types.PodID,
 		if err != nil {
 			return nil, err
 		}
+	}
+	return ret, nil
+}
+
+// Watch watches the entire podClusterTree for changes.
+// It will return a blocking channel on which the client can read
+// WatchedPodCluster objects. The goroutine maintaining the watch will block on
+// writing to this channel so it's up to the caller to read it with haste.
+func (s *consulStore) Watch(quit <-chan struct{}) <-chan WatchedPodCluster {
+	inCh := make(chan api.KVPairs)
+	outCh := make(chan WatchedPodCluster)
+	errChan := make(chan error, 1)
+
+	go consulutil.WatchPrefix(podClusterTree, s.kv, inCh, quit, errChan)
+
+	go func() {
+		var kvp api.KVPairs
+		for {
+			select {
+			case <-quit:
+				return
+			case err := <-errChan:
+				s.logger.WithError(err).Errorf("WatchPrefix returned error, recovered.")
+			case kvp = <-inCh:
+				if kvp == nil {
+					// nothing to do
+					continue
+				}
+			}
+
+			pcs, err := kvpsToPC(kvp)
+			if err != nil {
+				select {
+				case <-quit:
+					return
+				case outCh <- WatchedPodCluster{PodCluster: nil, Err: err}:
+					continue
+				}
+			}
+
+			for _, pc := range pcs {
+				select {
+				case outCh <- WatchedPodCluster{PodCluster: &pc, Err: nil}:
+				case <-quit:
+					return
+				}
+			}
+		}
+	}()
+
+	return outCh
+}
+
+// WatchPodCluster implements a watch for the Pod cluster at _id_
+// It will return a blocking channel on which the client can read
+// WatchedPodCluster objects. The goroutine maintaining the watch will block on
+// writing to this channel so it's up to the caller to read it with haste.
+// This function will return ErrNoPodCluster if the podCluster goes away. In
+// this case, the caller should close the quit chan.
+// The caller may shutdown this watch by sending a sentinel on the quitChan.
+func (s *consulStore) WatchPodCluster(id fields.ID, quit <-chan struct{}) <-chan WatchedPodCluster {
+	inCh := make(chan *api.KVPair)
+	outCh := make(chan WatchedPodCluster)
+	errChan := make(chan error, 1)
+	quitWatch := make(chan struct{})
+	key := path.Join(podClusterTree, string(id))
+
+	go consulutil.WatchSingle(key, s.kv, inCh, quitWatch, errChan)
+
+	go func() {
+		var kvp *api.KVPair
+		for {
+			select {
+			case <-quit:
+				return
+			case err := <-errChan:
+				s.logger.WithError(err).Errorf("WatchSingle returned error, recovered.")
+			case kvp = <-inCh:
+				if kvp == nil { // PodCluster at _id_ has been deleted
+					select {
+					case <-quit:
+						return
+					case outCh <- WatchedPodCluster{PodCluster: nil, Err: ErrNoPodCluster}:
+					}
+
+					return
+				}
+			}
+
+			pc, err := kvpToPC(kvp)
+			var wpc WatchedPodCluster
+			if err != nil {
+				wpc.Err = err
+			} else {
+				wpc.PodCluster = &pc
+			}
+
+			select {
+			case <-quit:
+				return
+			case outCh <- wpc:
+			}
+		}
+	}()
+
+	return outCh
+}
+
+func kvpsToPC(pairs api.KVPairs) ([]fields.PodCluster, error) {
+	ret := make([]fields.PodCluster, 0, len(pairs))
+	for _, kvp := range pairs {
+		var pc fields.PodCluster
+		var err error
+		if pc, err = kvpToPC(kvp); err != nil {
+			return nil, err
+		}
+		ret = append(ret, pc)
 	}
 	return ret, nil
 }
--- a/pkg/kp/pcstore/consul_store_test.go
+++ b/pkg/kp/pcstore/consul_store_test.go
@@ -2,6 +2,7 @@ package pcstore
 
 import (
 	"testing"
+	"time"
 
 	"github.com/square/p2/pkg/kp/kptest"
 	"github.com/square/p2/pkg/labels"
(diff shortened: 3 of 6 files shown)


pkg/kp/pcstore/consul_store.go
```diff
@@ -113,3 +127,72 @@ func kvpToPC(pair *api.KVPair) (fields.PodCluster, error) {
...
+
+	go consulutil.WatchPrefix(podClusterTree, s, inCh, quit, errChan)
+
+	go func() {
+		for {
+			select {
+			case <-quit:
+				return
+			case err := <-errChan:
+				s.logger.Errorf("WatchPrefix returned error, recovered.%v\n", err)
```
mpuncel: nit: `s.logger.WithError(err).Errorf()`

pkg/kp/pcstore/consul_store.go
```diff
@@ -151,3 +162,72 @@ func kvpToPC(pair *api.KVPair) (fields.PodCluster, error) {
...
+					return
+				}
+			}
+		}
+	}()
+
+	return outCh
+}
+
+func (s *consulStore) List(prefix string, opts *api.QueryOptions) (api.KVPairs, *api.QueryMeta, error) {
```
mpuncel: is this used by anything? i'd expect it to take no arguments and return `[]PodCluster`

pkg/kp/pcstore/consul_store_test.go
```diff
@@ -211,6 +212,49 @@ func TestDelete(t *testing.T) {
...
+		Add(fields.ClusterNameLabel, klabels.EqualsOperator, []string{clusterName.String()})
+
+	annotations := fields.Annotations(map[string]interface{}{
+		"foo": "bar",
+	})
+
+	quit := make(chan struct{})
+	watch := store.Watch(pod, quit)
+
+	var watched []WatchedPodCluster
```
mpuncel: looks like you only attempt to put one value in here, should this be a pointer instead?

pkg/kp/pcstore/consul_store_test.go
```diff
@@ -211,6 +212,49 @@ func TestDelete(t *testing.T) {
...
+
+	select {
+	case watchedPC := <-watch:
+		watched = append(watched, watchedPC)
+	case <-time.After(5 * time.Second):
+		quit <- struct{}{}
+		t.Fatal("nothing on the channel")
+	}
+
+	if len(watched) != 1 {
```
mpuncel: maybe test the `watched[0].PodCluster.ID` against `pc.ID` too?

pkg/kp/pcstore/consul_store.go
```diff
@@ -151,3 +162,72 @@ func kvpToPC(pair *api.KVPair) (fields.PodCluster, error) {
 
 	return pc, nil
 }
+
+// Watch implements a watch for the Pod cluster at _id_
+// It will return a blocking channel on which the client can read
+// WatchedPodCluster objects. The goroutine maintaining the watch will block on
+// writing to this channel so it's up to the caller to read it with haste.
+// One blocking watch will not affect any of the other watches.
+func (s *consulStore) Watch(id fields.ID, quit <-chan struct{}) <-chan WatchedPodCluster {
```
mpuncel: is `id` used?

pkg/kp/pcstore/pcstoretest/fake_pcstore.go
```diff
@@ -52,7 +54,18 @@ func (p *FakePCStore) Get(id fields.ID) (fields.PodCluster, error) {
...
 func (p *FakePCStore) Delete(id fields.ID) error {
 	delete(p.podClusters, id)
 	return nil
 }
+
+func (p *FakePCStore) Watch(id fields.ID, quit <-chan struct{}) <-chan pcstore.WatchedPodCluster {
+	pcChan, ok := p.watchers[id]
+	if !ok {
+		ret := make(chan pcstore.WatchedPodCluster)
+		return ret // reading from a closed channel returns the 0 value of the type immediately
```
mpuncel: maybe just `close(pcChan)`

mpuncel: actually this looks very different if `id` doesn't exist

pkg/kp/consulutil/watch.go
```diff
@@ -22,6 +22,7 @@ func WatchPrefix(
 	done <-chan struct{},
 	outErrors chan<- error,
 ) {
+	// TODO return this for closing on the client side
 	defer close(outPairs)
```
anthonybishopric: if it's being passed in, why does it need to be returned?

pkg/kp/pcstore/consul_store.go
```diff
@@ -151,3 +162,72 @@ func kvpToPC(pair *api.KVPair) (fields.PodCluster, error) {
...
+			case <-quit:
+				return
+			case err := <-errChan:
+				s.logger.WithError(err).Errorf("WatchPrefix returned error, recovered.")
+			}
+		}
+	}()
+
+	go func() {
+		for kvp := range inCh {
```
anthonybishopric: You probably want a `for { select { ` here so you can case the `quit` channel.

pkg/kp/pcstore/consul_store.go
anthonybishopric: +1, as written it just looks like a `kv` passthrough

pkg/kp/pcstore/consul_store.go
```diff
@@ -151,3 +162,72 @@ func kvpToPC(pair *api.KVPair) (fields.PodCluster, error) {
...
+// It will return a blocking channel on which the client can read
+// WatchedPodCluster objects. The goroutine maintaining the watch will block on
+// writing to this channel so it's up to the caller to read it with haste.
+// One blocking watch will not affect any of the other watches.
+func (s *consulStore) Watch(id fields.ID, quit <-chan struct{}) <-chan WatchedPodCluster {
+	inCh := make(chan api.KVPairs)
+	outCh := make(chan WatchedPodCluster)
+	errChan := make(chan error, 1)
+
+	go consulutil.WatchPrefix(podClusterTree, s, inCh, quit, errChan)
```
anthonybishopric: Probably should be `path.join(podClusterTree, id)`?

pkg/kp/pcstore/consul_store_test.go
```diff
@@ -211,6 +212,49 @@ func TestDelete(t *testing.T) {
...
+	quit := make(chan struct{})
+	watch := store.Watch(pod, quit)
+
+	var watched []WatchedPodCluster
+	pc, err := store.Create(podID, az, clusterName, selector, annotations)
+	if err != nil {
+		t.Fatalf("Unable to create pod cluster: %s", err)
+	}
+
+	pc, err = store.Get(pc.ID)
```
anthonybishopric: Is this a sanity check?

pkg/kp/consulutil/watch.go
mpuncel: my understanding is that it's not idiomatic to close a channel you were passed as an argument, it would make more sense for this function to do its work in a goroutine and return a channel it created

pkg/kp/pcstore/consul_store.go
mpuncel: `inCh` will be closed by `WatchPrefix` but defense in depth maybe

pkg/kp/pcstore/consul_store.go
rudle: This is necessary to implement the `ConsulLister` interface as required by WatchPrefix. I could instead pass `s.kv` into `WatchPrefix` which might work.

pkg/kp/pcstore/consul_store.go
rudle: We'll want to use it to filter the returned pod clusters, so the answer to your question is "no, not yet".

pkg/kp/pcstore/consul_store_test.go
rudle: It successfully ensured my sanity but is no longer necessary. Good catch.

rudle:
rebase across a conflict in `fake_pcstore.go`

rudle:
PTAL, I've implemented both types of watches.

pkg/kp/pcstore/consul_store.go
```diff
@@ -204,22 +198,11 @@ func (s *consulStore) FindWhereLabeled(podID types.PodID,
...
-	err := json.Unmarshal(pair.Value, &pc)
-	if err != nil {
-		return pc, util.Errorf("Could not unmarshal pod cluster ('%s') as json: %s", string(pair.Value), err)
-	}
-
-	return pc, nil
-}
-
-// Watch implements a watch for the Pod cluster at _id_
+// WatchPodCluster watches the podClusterTree for changes.
```
petertseng: I believe this one's just Watch.

pkg/kp/pcstore/consul_store.go
```diff
@@ -272,6 +255,62 @@ func (s *consulStore) Watch(id fields.ID, quit <-chan struct{}) <-chan WatchedPo
...
+
+	go consulutil.WatchSingle(key, s.kv, inCh, quit, errChan)
+
+	go func() {
+		var kvp *api.KVPair
+		for {
+			select {
+			case kvp = <-inCh:
+				if kvp == nil {
+					// PodCluster at id has been deleted
```
petertseng: shall we do something to let the caller know? for example return `WatchedPodCluster{Err: NoPodCluster}`?

Then again, `rcstore` does not do that either, it just does `continue`... which is different from `return` in that I believe it will wait for the pod cluster to reappear, though I'm not sure how probable that is either. Will give it a little thought.

pkg/kp/pcstore/consul_store_test.go
```diff
@@ -244,10 +244,54 @@ func TestWatch(t *testing.T) {
...
+		Add(fields.PodIDLabel, klabels.EqualsOperator, []string{pod.String()}).
+		Add(fields.AvailabilityZoneLabel, klabels.EqualsOperator, []string{az.String()}).
+		Add(fields.ClusterNameLabel, klabels.EqualsOperator, []string{clusterName.String()})
+
+	annotations := fields.Annotations(map[string]interface{}{
+		"foo": "bar",
+	})
+
+	quit := make(chan struct{})
+	watch := store.WatchPodCluster(pod, quit)
```
petertseng: I was considering suggesting a test that watched an existing PC ID and seeing if the watcher perceived changes to that PC. But that's maybe more of a test of consul than of the PC store...

I thought about that suggestion because in our usage it would be unusual for someone to know the name of a pod cluster before it got created (since we use UUIDs)

petertseng: actually now I am slightly confused. I would have thought that instead of `pod`, I would need to pass in `pc.ID` here (where `pc` is the result form the `Create` a few lines below)? May need someone to help shed some light on how this works.

pkg/kp/pcstore/consul_store.go
petertseng: yeah I guess `return` here is fine if we expect the pod cluster at a given ID to never come back after it's deleted. Undecided on whether to let the caller know about it.

pkg/kp/pcstore/consul_store.go
```diff
@@ -193,6 +198,136 @@ func (s *consulStore) FindWhereLabeled(podID types.PodID,
 	return ret, nil
 }
 
+// WatchPodCluster watches the podClusterTree for changes.
```
anthonybishopric: `s/WatchPodCluster/Watch`

pkg/kp/pcstore/consul_store_test.go
petertseng: only now do I realise I commented on the old version; you already did the `pc.ID` thing in the fixup commit

pkg/kp/pcstore/consul_store.go
rudle: I'm envisioning a workflow somewhat like the following:

1) An outer loop calls `Watch` to observe changes to the pod cluster tree, forking go routines to call `WatchPodCluster` on new PodClusters
2) These go routines live as long as the PodCluster exists, shutting themselves down when the PodCluster is deleted.

So, if a pod cluster goes away and comes back it will be observed by the outer loop and get a watching goroutine fairly expediently. I'm not seeing anything undesirable about this, but I'm happy to be wrong.

pkg/kp/pcstore/consul_store.go
```diff
@@ -193,6 +198,136 @@ func (s *consulStore) FindWhereLabeled(podID types.PodID,
...
+	errChan := make(chan error, 1)
+	key := path.Join(podClusterTree, string(id))
+
+	go func() {
+		for {
+			select {
+			case <-quit:
+				return
+			case err := <-errChan:
+				s.logger.WithError(err).Errorf("WatchPrefix returned error, recovered.")
```
petertseng: WatchSingle, perhaps?

pkg/kp/pcstore/consul_store.go
petertseng: Seems reasonable to me. So this `return` makes it so that the goroutine watching `inCh` will stop. Is there something to stop the one watching `errChan` (and should there be?)?

pkg/kp/pcstore/consul_store.go
rudle: Addressed by `WatchPodCluster`

pkg/kp/pcstore/consul_store.go
mpuncel: WatchPrefix is correct

petertseng: oh... I meant elsewhere I guess

mpuncel: this should be WatchSingle

petertseng: actually I think I meant this one - this one should be `WatchSingle` right?

pkg/kp/pcstore/consul_store.go
mpuncel: yeah passing `s.kv` would be more inline with what we do already

mpuncel:
looks pretty good, I have the same questions as peter regarding a pod cluster disappearing. I'd probably favor passing a `NoPodCluster` error and making the caller `close(quit)`

pkg/kp/pcstore/consul_store.go
petertseng: damn github is displaying our comments in the wrong place... we're both commenting on line 274 but our comment is being associated with  218 because they're the same line!!!! whoaaaaaaaaaaaaa

mpuncel: ahahahaha @rudle help us

rudle: sad_cat


I'm not sure I can help.

pkg/kp/pcstore/consul_store.go
rudle: The errchan check has been merged with the other go routine. This doesn't change the halting semantics - it still depends on the caller closing the quitChan.

pkg/kp/pcstore/consul_store.go
```diff
@@ -193,6 +199,122 @@ func (s *consulStore) FindWhereLabeled(podID types.PodID,
...
+					continue
+				}
+			}
+
+			pcs, err := kvpsToPC(kvp)
+			if err != nil {
+				select {
+				case <-quit:
+					return
+				case outCh <- WatchedPodCluster{PodCluster: nil, Err: err}:
```
anthonybishopric: Should there be a `continue` here?

mpuncel: might be nice to be explicit, but if `err != nil` then `pcs` is a 0 length slice so it's fine

rudle: continue added!

anthonybishopric:
:barge:

rudle:
squashed the commits, will merge on a green build

PR #470, opened 2016-04-22
When watching the pod cluster tree, return the full set of pod clusters

anthonybishopric:
the full set of all pod clusters, not just individuals.

This makes the pc store match the same pattern as the
preparer's pod watch, allowing both addition and deletion
to be supported.

### DIFF
--- a/pkg/kp/pcstore/consul_store.go
+++ b/pkg/kp/pcstore/consul_store.go
@@ -203,9 +203,9 @@ func (s *consulStore) FindWhereLabeled(podID types.PodID,
 // It will return a blocking channel on which the client can read
 // WatchedPodCluster objects. The goroutine maintaining the watch will block on
 // writing to this channel so it's up to the caller to read it with haste.
-func (s *consulStore) Watch(quit <-chan struct{}) <-chan WatchedPodCluster {
+func (s *consulStore) Watch(quit <-chan struct{}) <-chan WatchedPodClusters {
 	inCh := make(chan api.KVPairs)
-	outCh := make(chan WatchedPodCluster)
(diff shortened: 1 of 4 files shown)


pkg/kp/pcstore/consul_store.go
```diff
@@ -225,22 +225,29 @@ func (s *consulStore) Watch(quit <-chan struct{}) <-chan WatchedPodCluster {
 				}
 			}
 
+			clusters := WatchedPodClusters{
+				Clusters: []*fields.PodCluster{},
```
mpuncel: nit: don't think you need to initialize `Clusters`, `append` will do the right thing

mpuncel: although maybe you want to initialize in case no append happens

anthonybishopric: Ah of course

mpuncel:
👍

pkg/kp/pcstore/consul_store_test.go
```diff
@@ -255,18 +256,17 @@ func TestWatch(t *testing.T) {
...
-		watched = &watchedPC
+		watched = watchedPC
 	case <-time.After(5 * time.Second):
-		quit <- struct{}{}
 		t.Fatal("nothing on the channel")
 	}
 
-	if watched == nil {
-		t.Fatalf("Expected to get a watched PodCluster, but did not")
+	if len(watched.Clusters) != 1 {
```
petertseng: This should have been a comment from when this file was first added, but I guess I gotta do it retroactively: Check that `watched.Err` is nil?

PR #471, opened 2016-04-26
Fix a potential deadlock in fake pcstore

mpuncel:
The Create() function on the fake pcstore does not check if a watcher
exists before attempting to write to it, which can cause deadlocks in
tests that depend on it.


rudle:
d'oh. Sorry.

rudle:
:+1:

petertseng:
Hmm, interesting that we didn't figure out. OK 👍

PR #472, opened 2016-04-26
Drop privs in the default log exec

rudle:
If we don't do this, we end up running a process as the root user.

### DIFF
--- a/pkg/pods/pod.go
+++ b/pkg/pods/pod.go
@@ -39,7 +39,6 @@ const DEFAULT_PATH = "/data/pods"
 const DefaultP2Exec = "/usr/local/bin/p2-exec"
 
 var DefaultFinishExec = []string{"/bin/true"} // type must match preparerconfig
-var DefaultLogExec = []string{"svlogd", "-tt", "./main"}
 
 func init() {
 	Log = logging.NewLogger(logrus.Fields{})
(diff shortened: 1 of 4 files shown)


pkg/pods/pod.go
```diff
@@ -39,7 +39,7 @@ const DEFAULT_PATH = "/data/pods"
 const DefaultP2Exec = "/usr/local/bin/p2-exec"
 
 var DefaultFinishExec = []string{"/bin/true"} // type must match preparerconfig
-var DefaultLogExec = []string{"svlogd", "-tt", "./main"}
+var DefaultLogExec = []string{"chpst", "-unobody", "svlogd", "-tt", "./main"}
```
anthonybishopric: Can we use `p2-exec` instead? I've been meaning to kill the `chpst` dependency from this project.

petertseng:
I didn't test very thoroughly but it looks like it worked (just ran `echo a | sudo chpst -unobody svlogd -tt ./main` and it seemed to be OK) so 👍

pkg/pods/pod.go
rudle: sure, I'll change it.

pkg/runit/servicebuilder.go
```diff
@@ -48,6 +49,12 @@ const (
 // trailing space must be used.
 const yamlSeparator = "--- "
 
+// This matches the pods package. Copied to avoid import cycles
```
anthonybishopric: Could `pkg/pods` just import `pkg/runit` and use `runit.DefaultLogExec`? I don't see any imports pointing in the other direction right now

rudle: that's a good fix. just pushed a change to do that.

anthonybishopric:
👍

pkg/runit/servicebuilder.go
petertseng: still relevant coment?

PR #474, opened 2016-04-28
if --node specified, don't query for everything

anthonybishopric:
`p2-inspect` could be a lot faster if we don't assume we're always querying for everything


mpuncel:
👍

PR #475, opened 2016-04-28
Implement `WatchAndSync` for pcstore

anthonybishopric:
The code to watch both pod cluster definitions / annotations and their
labeled pods is complex enough that it warrants being factored into a
common library. Clients need only implement the `pcstore.ConcreteSyncer`
interface and pass it to `store.WatchAndSync(syncer, quitCh)` to implement
a concrete implementation of a pod cluster synchronization loop.

### DIFF
--- a/pkg/kp/pcstore/consul_store.go
+++ b/pkg/kp/pcstore/consul_store.go
@@ -361,6 +361,9 @@ func (s *consulStore) WatchAndSync(syncer ConcreteSyncer, quit <-chan struct{}) 
 	return nil
 }
 
+// zipResults takes two sets of watched pod clusters and joins them such that they
+// are paired together in a map of pc ID -> change objects. Each change will be sent
+// to the respective sync channels of each pod cluster later on.
 func (s *consulStore) zipResults(current, previous WatchedPodClusters) map[fields.ID]podClusterChange {
 	allPrevious := make(map[fields.ID]*fields.PodCluster)
 	for _, prev := range previous.Clusters {
@@ -389,9 +392,6 @@ func (s *consulStore) zipResults(current, previous WatchedPodClusters) map[field
 // If a change fails to take, this function will retry that change forever until
 // it works as expected or a newer change appears on the channel. This routine also
 // executes and monitors the label watch for the pod's label selector.
-// TODO: since most of this code is generic to pod cluster syncers, this will likely
-// be extracted to github.com/square/p2. The P2 VIP agent, for f5 syncing, will still
-// implement ConcreteSyncer, whose methods are implemented below.
 func (s *consulStore) handlePCUpdates(concrete ConcreteSyncer, changes chan podClusterChange) {
 	var change podClusterChange
 	podWatch := make(chan []labels.Labeled)
--- a/pkg/kp/pcstore/consul_store.go
+++ b/pkg/kp/pcstore/consul_store.go
@@ -4,6 +4,7 @@ import (
 	"encoding/json"
 	"path"
 
+	"github.com/square/p2/Godeps/_workspace/src/github.com/Sirupsen/logrus"
 	"github.com/square/p2/Godeps/_workspace/src/github.com/hashicorp/consul/api"
 	"github.com/square/p2/Godeps/_workspace/src/github.com/pborman/uuid"
 	klabels "github.com/square/p2/Godeps/_workspace/src/k8s.io/kubernetes/pkg/labels"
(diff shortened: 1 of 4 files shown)


pkg/kp/pcstore/consul_store.go
```diff
@@ -308,6 +309,145 @@ func (s *consulStore) WatchPodCluster(id fields.ID, quit <-chan struct{}) <-chan
...
+			prevResults = curResults
+		case <-quit:
+			return nil
+		}
+	}
+
+	return nil
+}
+
+func (s *consulStore) zipResults(current, previous WatchedPodClusters) map[fields.ID]podClusterChange {
```
mpuncel: maybe throw a comment onto this function explaining expected output?

anthonybishopric: Done

pkg/kp/pcstore/consul_store.go
```diff
@@ -308,6 +309,147 @@ func (s *consulStore) WatchPodCluster(id fields.ID, quit <-chan struct{}) <-chan
...
+		}
+	}
+	return ret
+}
+
+// try forever to match the expectations as defined in the provided change channel.
+// If a change fails to take, this function will retry that change forever until
+// it works as expected or a newer change appears on the channel. This routine also
+// executes and monitors the label watch for the pod's label selector.
+// TODO: since most of this code is generic to pod cluster syncers, this will likely
```
mpuncel: remove TODO

anthonybishopric: Done

pkg/kp/pcstore/consul_store.go
```diff
@@ -308,6 +309,147 @@ func (s *consulStore) WatchPodCluster(id fields.ID, quit <-chan struct{}) <-chan
...
+				return // we're closed for business
+			}
+
+			if change.current == nil && change.previous != nil {
+				// if no current cluster exists, but there is a previous cluster,
+				// it means we need to destroy this concrete cluster
+				s.logger.WithField("pc_id", change.previous.ID).Infof("Calling DeleteCluster with %v", change.previous)
+				err := concrete.DeleteCluster(change.previous)
+				if err != nil {
+					s.logger.Errorf("Deletion of cluster failed! %v", err)
```
mpuncel: i wonder if it makes more sense for the implementations to log this kind of stuff?

pkg/kp/pcstore/consul_store.go
```diff
@@ -308,6 +309,147 @@ func (s *consulStore) WatchPodCluster(id fields.ID, quit <-chan struct{}) <-chan
...
+		select {
+		case labeledPods := <-podWatch:
+			s.logger.Debugf("Calling SyncCluster with %v / %v", change.current, labeledPods)
+			concrete.SyncCluster(change.current, labeledPods)
+		case change, ok = <-changes:
+			if !ok {
+				return // we're closed for business
+			}
+
+			if change.current == nil && change.previous != nil {
```
mpuncel: might be nice to define functions like `change.IsDeletion()` or maybe `change.Type()` that you can `switch` on

pkg/kp/pcstore/consul_store.go
```diff
@@ -308,6 +309,147 @@ func (s *consulStore) WatchPodCluster(id fields.ID, quit <-chan struct{}) <-chan
...
+				if change.current.PodSelector.String() != change.previous.PodSelector.String() {
+					close(podWatchQuit)
+					podWatchQuit = make(chan struct{})
+					s.logger.WithFields(logrus.Fields{
+						"pc_id":        change.current.ID,
+						"old_selector": change.previous.PodSelector.String(),
+						"new_selector": change.current.PodSelector.String(),
+					}).Debugf("Altering pod selector for %v", change.current.ID)
+					podWatch = s.applicator.WatchMatches(change.current.PodSelector, labels.POD, podWatchQuit)
+				}
```
mpuncel: shouldn't you call `concrete.SyncCluster`? for example if annotations changed?

pkg/kp/pcstore/consul_store.go
anthonybishopric: I figured there's no harm in doing it here too, but I'm open to delegating this away

pkg/kp/pcstore/consul_store.go
anthonybishopric: `change` objects never make it to clients, so it's really just logic confined to this function. I can add it if you think it makes the code more readable though

pkg/kp/pcstore/consul_store.go
```diff
@@ -308,6 +309,147 @@ func (s *consulStore) WatchPodCluster(id fields.ID, quit <-chan struct{}) <-chan
...
+			} else {
+				// if there's no previous pod cluster but there is a current, create the concrete cluster
```
mpuncel: i'm not sure what this comment means regarding "create the concrete cluster"

pkg/kp/pcstore/consul_store.go
mpuncel: on second thought I think it's obvious, especially since you're calling `Delete()`

pkg/kp/pcstore/consul_store.go
mpuncel: sure let's leave it for now

rudle: nit: `WithError(err)`

pkg/kp/pcstore/consul_store.go
anthonybishopric: You don't want to call `concrete.SyncCluster` here because you don't have any results from the `podWatch` yet. The new annotations will be provided when `concrete.SyncCluster` is called after the `podWatch` fires.

It does mean that there is a somewhat artificial delay in updating annotations since you have to wait on the label watcher. In practice, the Consul label watcher is guaranteed to fire every 10 seconds at least, so this is probably fine but not great.

I think maybe we could keep the last result from `podWatch` as a variable out of scope of the `for` loop and sync the cluster immediately to get the desired effect. Going to think about whether that's safe to do.

pkg/kp/pcstore/consul_store.go
anthonybishopric: Yeah it doesn't really make sense as is. Will change to just `start a pod watch`

mpuncel:
Looks good to me 👍 , you'll just have to implement `WatchAndSync` on fake pc store

PR #477, opened 2016-04-29
errcheck the repo

rudle:
This list populated by `errcheck -ignoretests github.com/square/p2/pkg/... | grep -v defer`

I ignored defer cases because they are tricky to get right. I also wanted to get this merged ASAP as I caught a few pretty scary ignored errors!

Some discussion on err handling in defer can be found here: https://github.com/kisielk/errcheck/issues/55, https://github.com/kisielk/errcheck/issues/77

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,6 +10,7 @@ before_install:
 install:
   - gem install rake
   - go get github.com/tools/godep
+  - go get github.com/kisielk/errcheck
   - rake install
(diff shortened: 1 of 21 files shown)


rudle:
Still need to reason about how to get this into travis given that the exit code from

anthonybishopric:
👍  yesss

mpuncel:
👍

rudle:
PTAL travis is now wired up and verified working. By working I mean that the build will fail if someone introduces an unchecked error.

mpuncel:
👍

PR #478, opened 2016-05-02
add function to pcstore to allow locking a pod cluster for sync

mpuncel:
This will allow multiple copies of the same implementation of
ConcreteSyncer for pod clusters to run and use locking to ensure that
only one instance is handling an update for a given pod cluster.


anthonybishopric:
👍

mpuncel:
just rebased on master

PR #479, opened 2016-05-02
Fixes bug in pcstore.Watch()

mpuncel:
Watch() gives a list of *pcfields.PodCluster, but was unintentionally
populating the list with a pointer to a loop variable that gets
rewritten so that all values passed would be pointers to the same value
(the last value in the for loop).


rudle:
:ship:

anthonybishopric:
👍

PR #481, opened 2016-05-02
fix a test flake

mpuncel:
The test would flake because we start a watch before creating the
watched objects, and depend on the length of the returned results.
However the watch might return before the creations and result in
flakes.


anthonybishopric:
yep 🚢

PR #482, opened 2016-05-03
Mpuncel/status store

mpuncel:


pkg/kp/statusstore/consul_store.go
```diff
@@ -0,0 +1,161 @@
...
+	return path.Join(resourceTypePath(t), id)
+}
+
+func namespacedResourcePath(t ResourceType, id string, namespace Namespace) string {
+	return path.Join(resourcePath(t, id), namespace.String())
+}
+
+// in consul, the key for a status looks like
+// '/status/pod_clusters/<pod_cluster_id>/<namespace>'.
+func keyParts(key string) (ResourceType, string, Namespace, error) {
```
rudle: I'm wondering why podClusterID is a string instead of a typedef'd string like the other values. It would make a lot of this code clearer, including the `map[string]map[ResourceType]Status` above.

mpuncel: because ResourceType might not be pod cluster. I suppose I could make it be some generic ID but that would probably be weird for callers? e.g. `statusstore.ID(podId)`

rudle: Oh, I see. Its type is dependent on the type of ResourceType. Bummer. What about `statusstore.ResourceID`? That seems generic enough…

rudle:
What's here LGTM! Discussed the ResourceID type offline, I approve pending that change.

mpuncel:
squashed commit

PR #483, opened 2016-05-04
implement GetInitialClusters()

anthonybishopric:
This change requires implementations of `pcstore.ConcreteSyncer` to implement `GetInitialClusters()`. 

The use case for this function is to enable concrete syncers to be notified of potential deletions that may have occurred during period of downtime or lack of connectivity. Assume the following scenario:

1. Pod cluster A is created
2. ConcreteSyncer S synchronizes A
3. S shuts down for maintenance
4. A is deleted
5. S resumes, unaware of A's deletion

The desired behavior above is to synchronize the deletion of A, but this is not possible because the `WatchAndSync` function can only watch the generic pcstore as defined in this project. Syncers track their own status information in their own implementation-specific ways. Soon we'll be adding a `StatusStore`, but that too must allow vendor-specific status data to be written. For comparison, the P2 Preparer is able to fully use Consul to do syncing without fear of missing deletions because there are no vendors external to P2 to synchronize.

### DIFF
--- a/pkg/kp/pcstore/consul_store.go
+++ b/pkg/kp/pcstore/consul_store.go
@@ -341,14 +341,17 @@ func (s *consulStore) WatchAndSync(syncer ConcreteSyncer, quit <-chan struct{}) 
 	var prevResults WatchedPodClusters
 	for _, id := range initial {
 		existing, err := s.Get(id)
-		if err != nil {
+		if err == ErrNoPodCluster {
 			s.logger.WithField("pc_id", id).Warnf("Could not find initial cluster %v, will call DeleteCluster momentarily", id)
 			existing = fields.PodCluster{
 				ID: id,
 			}
+		} else if err != nil {
+			s.logger.WithField("pc_id", id).Errorln("Error retrieving pod cluster from consul")
+			return err
+		} else {
+			prevResults.Clusters = append(prevResults.Clusters, &existing)
 		}
-
-		prevResults.Clusters = append(prevResults.Clusters, &existing)
 	}
 
 	for {
@@ -445,7 +448,7 @@ func (s *consulStore) handlePCUpdates(concrete ConcreteSyncer, changes chan podC
 				// if no current cluster exists, but there is a previous cluster,
 				// it means we need to destroy this concrete cluster
 				s.logger.WithField("pc_id", change.previous.ID).Infof("Calling DeleteCluster with %v", change.previous)
-				err := concrete.DeleteCluster(change.previous)
+				err := concrete.DeleteCluster(change.previous.ID)
 				if err != nil {
 					s.logger.Errorf("Deletion of cluster failed! %v", err)
 				} else {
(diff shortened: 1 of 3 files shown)


pkg/kp/pcstore/consul_store.go
```diff
@@ -331,8 +331,26 @@ func (s *consulStore) WatchAndSync(syncer ConcreteSyncer, quit <-chan struct{})
...
+	// populate the initial clusters, if any provided
+	initial, err := syncer.GetInitialClusters()
+	if err != nil {
+		s.logger.Errorf("Error retrieving initial clusters: %v", err)
+		return err
+	}
+
 	var prevResults WatchedPodClusters
+	for _, id := range initial {
+		existing, err := s.Get(id)
```
mpuncel: you should

pkg/kp/pcstore/consul_store.go
```diff
@@ -331,8 +331,26 @@ func (s *consulStore) WatchAndSync(syncer ConcreteSyncer, quit <-chan struct{})
...
+		if err != nil {
+			s.logger.WithField("pc_id", id).Warnf("Could not find initial cluster %v, will call DeleteCluster momentarily", id)
```
mpuncel: you should check `err == ErrNoPodCluster` here

pkg/kp/pcstore/store.go
```diff
@@ -68,8 +69,35 @@ type ConcreteSyncerType string
...
+	// ConcreteSyncers will be called concurrently and must operate safely.
 	SyncCluster(pc *fields.PodCluster, pods []labels.Labeled) error
+
+	// DeleteCluster is called when a pod cluster is observed to have been removed
+	// from the store. DeleteCluster can be invoked in two circumstances: first,
+	// if the cluster was present in a first watch, then absent in a subsequent watch,
+	// then it is assumed it was deleted from the pcstore. Second, if the call
+	// to GetInitialClusters() returns a pod cluster ID that is not present in the very
+	// first watch result, DeleteCluster will be invoked with that pod cluster. In the
+	// latter case, only the pod cluster's ID will be set on the passed `pc` object.
```
mpuncel: `DeleteCluster` should probably ONLY ever receive the ID. depending on anything other than the ID would be a bug

mpuncel:
👍

pkg/kp/pcstore/store.go
```diff
@@ -68,8 +69,38 @@ type ConcreteSyncerType string
...
+	//
+	// If the passed ID is used to retrieve the pod cluster via store.Get(), it will
+	// return ErrNoPodCluster. Clients should track any relevant metadata to the pod
+	// cluster ID in the status store or in vendor-specific code.
+	DeleteCluster(pc fields.ID) error
+
+	// GetInitialClusters is called at the beginning of the WatchAndSync
+	// routine. See DeleteCluster() for an explanation of how its results are
+	// used. If the function results in an error, the WatchAndSync function will
+	// terminate immediately, forwarding the error.
```
rudle: ⭐ 📜

rudle:
🚢

PR #484, opened 2016-05-05
Add a NodeName type

rudle:
There's no cycle to work around in my use case, but I like the idea of having one package of types. Thoughts?

### DIFF
--- a/pkg/types/types.go
+++ b/pkg/types/types.go
@@ -3,7 +3,12 @@
 // wish to use pods.ID, an import cycle is created.
 package types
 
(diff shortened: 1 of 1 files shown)


mpuncel:
👍 i'm all for centralizing the types

mpuncel:
fwiw kubernetes does it too: https://github.com/kubernetes/kubernetes/blob/master/pkg/api/types.go

rudle:
Cool! I'll do a find and replace to increase type safety when I pull this in via godep.

PR #487, opened 2016-05-11
Implement cached health store

rudle:
This allows clients to lookup health without introducing load on the
datastore.

Also improve the channel semantics in &checker.WatchHealth().
  We value fresh values over the durability of messages in this channel.

The `types.NodeName` references are as small as possible now, I will follow up with a better rename in a subsequent PR.

Additionally, we have discussed changing the pattern of channels as arguments. Since this change will affect all of health on its clients it is a little risky. I will introduce this in a subsequent PR.

### DIFF
--- a/pkg/health/checker/checker.go
+++ b/pkg/health/checker/checker.go
@@ -125,7 +125,6 @@ func publishLatestHealth(inCh <-chan api.KVPairs, quitCh <-chan struct{}, result
 				}
 			default:
 			}
-
 			// Now we check the quit chan and try to write to our resultCh
 			select {
 			case <-quitCh:
--- a/pkg/health/checker/checker.go
+++ b/pkg/health/checker/checker.go
@@ -152,26 +152,24 @@ func (c consulHealthChecker) WatchHealth(
 	go consulutil.WatchPrefix("health/", c.kv, inCh, quitCh, watchErrCh)
 	publishErrCh := publishLatestHealth(inCh, quitCh, resultCh)
 
-	go func() {
+	select {
+	case <-quitCh:
+		return
+	case err := <-watchErrCh:
 		select {
+		case errCh <- err:
 		case <-quitCh:
 			return
-		case err := <-watchErrCh:
-			select {
-			case errCh <- err:
-			case <-quitCh:
-				return
-			default:
-			}
-		case err := <-publishErrCh:
-			select {
-			case errCh <- err:
-			case <-quitCh:
-				return
-			default:
-			}
+		default:
 		}
-	}()
+	case err := <-publishErrCh:
+		select {
+		case errCh <- err:
+		case <-quitCh:
+			return
+		default:
+		}
+	}
 }
 
 func (c consulHealthChecker) WatchService(
--- a/pkg/health/checker/checker.go
+++ b/pkg/health/checker/checker.go
@@ -129,9 +129,8 @@ func publishLatestHealth(inCh <-chan api.KVPairs, quitCh <-chan struct{}, result
 			select {
 			case <-quitCh:
 				return
-			default:
+			case resultCh <- results:
 			}
-			resultCh <- results
 		}
 	}()
 
--- a/pkg/health/checker/checker.go
+++ b/pkg/health/checker/checker.go
@@ -26,7 +26,7 @@ type ConsulHealthChecker interface {
 		errCh chan<- error,
 		quitCh <-chan struct{})
 	WatchHealth(
-		resultCh chan<- []*health.Result,
+		resultCh chan []*health.Result,
 		errCh chan<- error,
 		quitCh <-chan struct{})
 	Service(serviceID string) (map[string]health.Result, error)
@@ -84,8 +84,9 @@ func (c consulHealthChecker) WatchNodeService(
 }
 
 // Watch the health tree and write the whole subtree on the chan passed by caller
+// the result channel argument _must be buffered_
 func (c consulHealthChecker) WatchHealth(
-	resultCh chan<- []*health.Result,
+	resultCh chan []*health.Result,
 	errCh chan<- error,
 	quitCh <-chan struct{},
 ) {
@@ -112,6 +113,7 @@ func (c consulHealthChecker) WatchHealth(
 				}
 			} else {
 				select {
+				case <-resultCh: // we drain this channel prior to attempting a write so that the freshest value is always available
 				case resultCh <- healthResults:
 				default:
 				}
--- a/pkg/health/checker/checker.go
+++ b/pkg/health/checker/checker.go
@@ -140,33 +140,38 @@ func publishLatestHealth(inCh <-chan api.KVPairs, quitCh <-chan struct{}, result
 
 // Watch the health tree and write the whole subtree on the chan passed by caller
 // the result channel argument _must be buffered_
+// Any errors are passed, best effort, over errCh
 func (c consulHealthChecker) WatchHealth(
 	resultCh chan []*health.Result,
 	errCh chan<- error,
 	quitCh <-chan struct{},
 ) {
-	defer close(resultCh)
-
 	// closed by watchPrefix when we close quitWatch
 	inCh := make(chan api.KVPairs)
-	go consulutil.WatchPrefix("health/", c.kv, inCh, quitCh, errCh)
-	errCh = publishLatestHealth(inCh, quitCh, resultCh)
+	watchErrCh := make(chan error)
+	go consulutil.WatchPrefix("health/", c.kv, inCh, quitCh, watchErrCh)
+	publishErrCh := publishLatestHealth(inCh, quitCh, resultCh)
 
-	var open bool
-	for {
+	go func() {
 		select {
 		case <-quitCh:
 			return
-		case _, open = <-inCh:
-			if !open {
+		case err := <-watchErrCh:
+			select {
+			case errCh <- err:
+			case <-quitCh:
 				return
+			default:
 			}
-		case _, open = <-resultCh:
-			if !open {
+		case err := <-publishErrCh:
+			select {
+			case errCh <- err:
+			case <-quitCh:
 				return
+			default:
 			}
 		}
-	}
+	}()
 }
 
 func (c consulHealthChecker) WatchService(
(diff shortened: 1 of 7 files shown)


pkg/health/checker/checker.go
```diff
@@ -112,6 +113,7 @@ func (c consulHealthChecker) WatchHealth(
 				}
 			} else {
 				select {
+				case <-resultCh: // we drain this channel prior to attempting a write so that the freshest value is always available
```
mpuncel: you want this to be in a separate `select` because if this case is ready then you won't write healthResults at all

pkg/health/checker/checker.go
```diff
@@ -112,6 +113,7 @@ func (c consulHealthChecker) WatchHealth(
...
 				case resultCh <- healthResults:
 				default:
```
mpuncel: you shouldn't need this `default` if the channel has buffer headroom

pkg/health/checker/checker_test.go
```diff
@@ -115,3 +114,75 @@ func TestWatchHealth(t *testing.T) {
...
+		Node:    "node1.example.com",
+		Service: "Service",
+		Status:  newStatus,
+		Output:  "output",
+	}
+	dummyBufNew, err := json.Marshal(dummyHealthResultNew)
+	if err != nil {
+		t.Fatalf("json marshal err: %v", err)
+	}
+	_, _, err = fakeKV.CAS(&api.KVPair{Key: "health/service/node1.example.com", Value: dummyBufNew}, nil)
```
mpuncel: you should do this after running `WatchHealth` in a goroutine, otherwise you aren't proving that the old value is being skipped

pkg/health/checker/checker_test.go
```diff
@@ -115,3 +114,75 @@ func TestWatchHealth(t *testing.T) {
...
+	if err != nil {
+		t.Fatalf("error: %v", err)
+	}
+
+	go func() {
+		defer func() {
+			quitChan <- struct{}{}
+		}()
+		for {
```
mpuncel: don't think you need this `for` because all cases of the select result in test failure or exit

pkg/health/checker/checker_test.go
```diff
@@ -115,3 +114,75 @@ func TestWatchHealth(t *testing.T) {
...
+	dummyBufNew, err := json.Marshal(dummyHealthResultNew)
+	if err != nil {
+		t.Fatalf("json marshal err: %v", err)
+	}
+	_, _, err = fakeKV.CAS(&api.KVPair{Key: "health/service/node1.example.com", Value: dummyBufNew}, nil)
+	if err != nil {
+		t.Fatalf("error: %v", err)
+	}
+
+	go func() {
```
mpuncel: if you do `WatchHealth` in a goroutine i'd do this in the main thread

pkg/health/store/store.go
```diff
@@ -0,0 +1,94 @@
...
+
+	"github.com/square/p2/Godeps/_workspace/src/github.com/Sirupsen/logrus"
+
+	"github.com/square/p2/pkg/health"
+	"github.com/square/p2/pkg/health/checker"
+	"github.com/square/p2/pkg/logging"
+	"github.com/square/p2/pkg/types"
+)
+
+// Health can answer questions about the health of a particular pod on a node
```
mpuncel: s/Health/HealthStore ?

pkg/health/store/store.go
```diff
@@ -0,0 +1,94 @@
...
+
+	res, ok := hs.cachedHealth[cacheKey]
+	if !ok {
+		hs.logger.WithFields(logrus.Fields{"cachedHealth": len(hs.cachedHealth)}).Errorln("Cache miss!!")
+		return nil
+	}
+	return res
+}
+
+type cacheKey string
```
mpuncel: nit: you could make this a `struct{
podID types.PodID
node types.NodeName
}` and not have to do string stuff

pkg/health/store/store.go
```diff
@@ -0,0 +1,94 @@
...
+	lock          *sync.RWMutex
+	logger        logging.Logger
+}
+
+var _ HealthStore = &healthStore{}
+
+func NewHealthStore(healthChecker checker.ConsulHealthChecker) HealthStore {
+	return &healthStore{
+		healthChecker: healthChecker,
+		lock:          &sync.RWMutex{},
```
mpuncel: you should be able to make this not be a pointer and just use the zero value: "RWMutexes can be created as part of other structures; the zero value for a RWMutex is an unlocked mutex." (https://golang.org/pkg/sync/)

pkg/health/store/store.go
```diff
@@ -0,0 +1,94 @@
...
+	for {
+		select {
+		case updates := <-healthUpdates:
+			hs.cache(updates)
+		case <-quitCh:
+			hs.logger.Errorln("Quitting...")
+			return
+		case err := <-errCh:
+			hs.logger.WithError(err).Errorln("Consul Watch error")
+		default:
```
mpuncel: you don't want this `default` because that makes this a busy loop.

pkg/health/store/store_test.go
```diff
@@ -0,0 +1,79 @@
...
+}
+
+func (hc *FakeHealthChecker) WatchHealth(resultCh chan<- []*health.Result, errCh chan<- error, quitCh <-chan struct{}) {
+	for {
+		select {
+		case result := <-hc.healthResults:
+			resultCh <- result
+		case <-quitCh:
+			return
+		default:
```
mpuncel: kill the default

pkg/health/checker/checker.go
```diff
@@ -83,38 +83,87 @@ func (c consulHealthChecker) WatchNodeService(
...
 
+func publishLatestHealth(inCh <-chan api.KVPairs, quitCh <-chan struct{}, resultCh chan []*health.Result) chan error {
+	errCh := make(chan error)
+
+	go func() {
+		var listed api.KVPairs
+		var ok bool
+		var err error
+
+		for {
```
rudle: This came out looking a bit simpler than its cousin in rcstore. I think it's because of the use of continue and only reading from inCh in one place - this reduces mental bookkeeping overhead.

It's possible I missed something, so please review!

pkg/health/checker/checker.go
```diff
@@ -83,38 +83,87 @@ func (c consulHealthChecker) WatchNodeService(
...
+				}
+			default:
+			}
+			// Now we check the quit chan and try to write to our resultCh
+			select {
+			case <-quitCh:
+				return
+			default:
+			}
+			resultCh <- results
```
mpuncel: this should be inside the `select` above, otherwise there's a race where quit might fire before you've written resultCh. Also, you `default` is unnecessary if the channel is buffered as we expect but i suppose it's harmless

pkg/health/checker/checker_test.go
```diff
@@ -52,66 +51,60 @@ func TestService(t *testing.T) {
...
-		}()
-		for {
-			select {
-			case result := <-retChan:
-				res := result[0]
-				if res.ID != "ID" {
-					t.Fatalf("Expected ID to match")
-				}
+	// Basic test that publishLatestHealth drains the channels correctly
+	healthListChan <- api.KVPairs{oldKV}
```
mpuncel: nit: `select {}` on a `<-time.After` for these so the test doesn't hang forever if someone introduces a bug

pkg/health/checker/checker_test.go
```diff
@@ -52,66 +51,60 @@ func TestService(t *testing.T) {
...
-				if res.Output != "output" {
-					t.Fatalf("Expected Output to match")
-				}
-				return
-			case err := <-errChan:
-				t.Fatalf("unexpected error: %v", err)
-			case <-time.After(5 * time.Second):
-				t.Error("Timed out waiting for message")
-			}
+	select {
```
mpuncel: looks great!

rudle:
Tests are green! I had to be a bit cheeky to handle some async stuff, but I'm happy that I found a consistent solution.

pkg/health/checker/checker.go
```diff
@@ -83,38 +83,87 @@ func (c consulHealthChecker) WatchNodeService(
...
-			} else {
-				select {
-				case resultCh <- healthResults:
-				default:
-				}
+		case _, open = <-inCh:
+			if !open {
+				return
+			}
+		case _, open = <-resultCh:
```
mpuncel: doesn't this mean you're going to be stealing values from the caller of `WatchHealth`?

pkg/health/checker/checker.go
```diff
@@ -83,38 +83,87 @@ func (c consulHealthChecker) WatchNodeService(
...
-				select {
-				case errCh <- err:
-				default:
-				}
-			} else {
-				select {
-				case resultCh <- healthResults:
-				default:
-				}
+		case _, open = <-inCh:
```
mpuncel: ditto but stealing from publishLatestHealth

pkg/health/checker/checker.go
rudle: right. I can't even tell what this is supposed to be doing. 

I'll make it look like rcstore and fan-in the err channels.

rudle:
OK, I think this is finally ready to go.

pkg/health/checker/checker.go
```diff
@@ -83,41 +83,95 @@ func (c consulHealthChecker) WatchNodeService(
...
-	go consulutil.WatchPrefix("health/", c.kv, res, quitWatch, errCh)
+	// closed by watchPrefix when we close quitWatch
+	inCh := make(chan api.KVPairs)
+	watchErrCh := make(chan error)
+	go consulutil.WatchPrefix("health/", c.kv, inCh, quitCh, watchErrCh)
+	publishErrCh := publishLatestHealth(inCh, quitCh, resultCh)
 
-	var results api.KVPairs
-	for {
+	go func() {
```
mpuncel: nit: i'd prefer for this function to block since it's easy for a caller to call it in a goroutine if they so desire. @anthonybishopric once showed me a talk asserting that libraries should always be blocking

rudle:
This is ready for a final review! Hoping to get this merged today.

mpuncel:
looks good!

PR #488, opened 2016-05-11
add a List() function to pcstore to list all pod clusters

mpuncel:


pkg/kp/pcstore/consul_store_test.go
```diff
@@ -231,6 +231,57 @@ func TestDelete(t *testing.T) {
 	}
 }
 
+func TestList(t *testing.T) {
+	store := consulStoreWithFakeKV()
+	podID := types.PodID("pod_id")
+	az := fields.AvailabilityZone("us-west")
+	clusterName := fields.ClusterName("cluster_name")
+
+	selector := klabels.Everything().
```
rudle: You probably already know there's a helper for this, FindWhereLabeled()

rudle:
:ship:

pkg/kp/pcstore/consul_store_test.go
mpuncel: I don't actually care about the content of the selector here so i'll make it `klabels.Everything()`

mpuncel:
made the test use a klabels.Everything() to keep it more concise and added List() to fake pcstore

rudle:
:ship: :ship:

PR #489, opened 2016-05-12
Fx default log exec

anthonybishopric:
When we refactored the default log exec function, we accidentally omitted the leading command `p2-exec`. Run scripts for logs by default are generated with just the arguments:

```
#!/usr/bin/ruby
require 'yaml'
sleep 2
exec *YAML.load(DATA.read)
sleep 2
__END__
--- 
- -u
- nobody
- --
- svlogd
- -tt
- ./main
```

This change prepends the default `p2-exec` command to the log execution.


mpuncel:
👍

petertseng:
I'm unsure, but it may have been possible to do this with `init()` to avoid turning the thing into a function call - https://golang.org/ref/spec#Package_initialization . oh well, if what's in this PR works then it's alll good I think

PR #490, opened 2016-05-16
Configurable artifact signing and verification

anthonybishopric:
This PR adds a new configuration option to the preparer's manifest to allow it to verify installed artifacts in a number of ways, based on a PGP keyring containing entities from a trusted build system.

The type matches one of the `auth.Verify*` constants
```
"type: none"     - no artifact verification is done
"type: build"    - checks that builds have a corresponding signature
"type: manifest" - checks that builds have corresponding digest manifest and manifest signature files.
"type: either"   - checks that one of "build" or "manifest" strategies pass.
```

Except for `none`, all options take a `keyring` option, which is a path to a keyring containing *only* trusted build entities.

This supplements existing Merkle tree based verification via the signature options in the pod manifest.


bin/p2-bootstrap/bootstrap.go
```diff
@@ -87,7 +89,7 @@ func main() {
 
 func InstallConsul(consulPod *pods.Pod, consulManifest pods.Manifest) error {
 	// Inject servicebuilder?
-	err := consulPod.Install(consulManifest)
+	err := consulPod.Install(consulManifest, auth.NoopVerifier())
```
mpuncel: super duper nit but the `ioutil` package uses `Nop` instead of `Noop`: https://golang.org/pkg/io/ioutil/#NopCloser

mpuncel: totally cool with this though

pkg/auth/artifact_verifier.go
```diff
@@ -0,0 +1,162 @@
...
+// The manifest and signature files should match the following convention if the launchable
+// location is a full path to a file:
+//
+// If the artifact is located here:
+// https://foo.bar.baz/artifacts/myapp_abc123.tar.gz
+//
+// Then its build manifest is located here:
+// https://foo.bar.baz/artifacts/myapp_abc123.tar.gz.manifest
+//
+// artifact_sha: abc23456
```
mpuncel: are you missing a line above this?

pkg/auth/artifact_verifier.go
```diff
@@ -0,0 +1,162 @@
...
+	}
+	return &BuildManifestVerifier{
+		keyring: keyring,
+		fetcher: fetcher,
+		logger:  logger,
+	}, nil
+}
+
+// Returns an error if the stanza's artifact is not signed appropriately. Note that this
+// implementation does not use the pod manifest digest location options.
```
mpuncel: maybe specify that you're talking about `LaunchableStanza` here?

pkg/auth/artifact_verifier.go
```diff
@@ -0,0 +1,162 @@
...
+func (b *BuildManifestVerifier) VerifyHoistArtifact(localCopy *os.File, artifactLocation string) error {
+	u, err := url.Parse(artifactLocation)
+	if err != nil {
+		return err
```
mpuncel: wrap in a `util.Errorf`

pkg/auth/artifact_verifier.go
```diff
@@ -0,0 +1,162 @@
...
+	}
+	switch u.Scheme {
+	default:
+		return fmt.Errorf("%v does not have a recognized scheme, cannot verify manifest or signature", artifactLocation)
```
mpuncel: maybe put `u.Scheme` in here too

pkg/auth/artifact_verifier.go
```diff
@@ -0,0 +1,162 @@
...
+	case "http", "https", "file":
+		dir, err := ioutil.TempDir("", "artifact_verification")
+		if err != nil {
+			return fmt.Errorf("Could not create temporary directory for manifest file: %v", err)
```
mpuncel: `util.Errorf`

pkg/auth/artifact_verifier.go
```diff
@@ -0,0 +1,162 @@
...
+		manifestSrc := fmt.Sprintf("%v.manifest", artifactLocation)
+		manifestDst := filepath.Join(dir, "manifest")
+		err = b.fetcher.CopyLocal(manifestSrc, manifestDst)
+		if err != nil {
+			return fmt.Errorf("Could not download artifact manifest for %v: %v", artifactLocation, err)
+		}
+
+		signatureSrc := fmt.Sprintf("%v.sig", manifestSrc)
+		signatureDst := filepath.Join(dir, "signature")
+		if err = b.fetcher.CopyLocal(signatureSrc, signatureDst); err != nil {
```
mpuncel: nit: you're using different `err` checking syntax between the two `CopyLocal` calls

pkg/auth/artifact_verifier.go
```diff
@@ -0,0 +1,162 @@
...
+}
+
+// Returns an error if the stanza's artifact is not signed appropriately. Note that this
+// implementation does not use the pod manifest digest location options.
+func (b *BuildManifestVerifier) VerifyHoistArtifact(localCopy *os.File, artifactLocation string) error {
+	u, err := url.Parse(artifactLocation)
+	if err != nil {
+		return err
+	}
+	switch u.Scheme {
```
mpuncel: i don't think you need to check the scheme here, seems like something the `Fetcher` should do

mpuncel: and you're not using `u` for anything else

pkg/auth/artifact_verifier.go
```diff
@@ -0,0 +1,162 @@
...
+	}
+}
+
+func (b *BuildManifestVerifier) verifySigned(manifestBytes, signatureBytes []byte) error {
+	// permit an armored detached signature
+	block, err := armor.Decode(bytes.NewBuffer(signatureBytes))
+	if err == nil {
+		signatureBytes, err = ioutil.ReadAll(block.Body)
+		if err != nil {
+			return err
```
mpuncel: `util.Errorf`

pkg/auth/artifact_verifier.go
```diff
@@ -0,0 +1,162 @@
...
+		}
+	}
+	// check that the manifest was adequately signed by our signer
+	_, err = checkDetachedSignature(b.keyring, manifestBytes, signatureBytes)
```
mpuncel: do you need to check who the signer was here? or just that it was signed by something in the keyring

pkg/auth/digest_fetcher.go
```diff
@@ -0,0 +1,4 @@
+package auth
+
+type DigestFetcher interface {
```
mpuncel: is this used anywhere?

pkg/hoist/hoist_launchable.go
```diff
@@ -295,6 +297,18 @@ func (hl *Launchable) Install() error {
 	if err != nil {
 		return err
 	}
+	// rewind once so we can ask the verifier
+	_, err = artifactFile.Seek(0, os.SEEK_SET)
+	if err != nil {
+		return err
```
mpuncel: util.Errorf

bin/p2-verify-artifact/main.go
```diff
@@ -0,0 +1,68 @@
...
+var (
+	location         = kingpin.Arg("location", "The URI of the artifact.").Required().String()
+	originalLocation = kingpin.Flag("original-location", "The URI where the artifact which has already been downloaded came from. The primary location must be an existing file").String()
+	gpgKeyringPath   = kingpin.Flag("keyring", "The PGP keyring to use to verify the artifact").Required().ExistingFile()
+)
+
+func main() {
+	kingpin.Version(version.VERSION)
+	k := kingpin.Parse()
+	log.Println(k)
```
mpuncel: remove?

bin/p2-verify-artifact/main.go
```diff
@@ -0,0 +1,68 @@
...
+			log.Fatalf("Could not open local copy of the file %v: %v", *location, err)
+		}
+		locationForSignature = *originalLocation
+	}
+
+	if err = verifier.VerifyHoistArtifact(localCopy, locationForSignature); err != nil {
+		log.Fatalln(err)
+	}
+
+	log.Println("OK")
```
mpuncel: this one seems intentional but in case it's not

bin/p2-verify-artifact/main.go
```diff
@@ -22,13 +24,7 @@ var (
 
 func main() {
 	kingpin.Version(version.VERSION)
-	k := kingpin.Parse()
-	log.Println(k)
```
mpuncel: oh you got it

bin/p2-bootstrap/bootstrap.go
anthonybishopric: Good point, will change

pkg/auth/artifact_verifier.go
anthonybishopric: Yep

mpuncel:
looks good 👍

pkg/auth/artifact_verifier.go
anthonybishopric: The protocol-free version of artifact locations will eventually be repurposed for image shorthands (ie `myrepo/myapp:abc123`) and we'd want this code to do something differently in such a case - specifically, I'd prefer such a shorthand to use a discovery endpoint to find the signature URL, not just guess at it by suffixing it with an arbitrary string.

Can remove if you think it's confusing as is

pkg/auth/artifact_verifier.go
anthonybishopric: 👍  added here and all throughout this file where I used `fmt` instead

pkg/auth/artifact_verifier.go
mpuncel: it's not confusing so no need to remove

pkg/auth/digest_fetcher.go
anthonybishopric: Nope, will remove

pkg/auth/artifact_verifier_test.go
```diff
@@ -0,0 +1,29 @@
...
+	"path/filepath"
+	"runtime"
+	"testing"
+
+	"github.com/square/p2/pkg/logging"
+	"github.com/square/p2/pkg/uri"
+	"github.com/square/p2/pkg/util"
+)
+
+func TestManifestVerifierAuthorizesValidBuild(t *testing.T) {
```
petertseng: any interest at all in shoring it up with the following tests?

* reject a build that is signed by something, but not in the keyring
* reject a build not signed at all

anthonybishopric: @petertseng check the final commit, it adds a bit more coverage than this

PR #491, opened 2016-05-17
Use pod-wide lock in replication

anthonybishopric:
When using `p2-replicate` for large numbers of pods it quickly becomes burdensome to use a per-node pod lock. The original intent of per-node locking was to allow parallel deployments to multiple environments simultaneously. Now that our recommended deployment system is to use rolling updates + replication controllers, which do not use pkg/replication to achieve their work, we can focus on making `p2-replicate` effective for today's primary use cases (deploying agents).


mpuncel:
👍

rudle:
⚓ ⬆️

petertseng:
Oh nice, this should be good. 👍 , confirmed that p2-replicate is the only thing using `InitializeReplication` and/or `lockHosts`

PR #492, opened 2016-05-17
fix bug with initial conditions in pcstore.WatchAndSync

mpuncel:
Previously func handlePCUpdate() operated under the assumption that the
first update would always have a nil value for the previous state of a
pod cluster. However, now that the ConcreteSyncer interface includes
GetInitialClusters(), this is a faulty assumption.

This would result in a pod cluster pod membership watch not being
started.


anthonybishopric:
good catch :+1:

PR #493, opened 2016-05-18
Add helper for extracting node name and pod id from pod label key

mpuncel:
Due to returning a value of type types.NodeName from that function,
there was a lot of fallout of changing node names with string type to
use types.NodeName instead.


mpuncel:
The main change is in `pkg/labels/consul_applicator.go` and then the return type caused the rest of the fallout

mpuncel:
also pkg/rc/replication_controller and one other place i'm forgetting use the new function

rudle:
Looks fairly mechanical. Did you find any interesting errors?

:+1:

mpuncel:
@rudle nope it was very mechanical

PR #494, opened 2016-05-19
Only call `SyncCluster` if pods or cluster change.

anthonybishopric:
`SyncCluster` would previously be called on every return from the
label watch. Clients were implementing their own variation of rate
limiting and throttling to deal with the flow of calls. This change
reduces the frequency of unnecessary calls to `SyncCluster`.

`SyncCluster` is still expected to be idempotent. It is still guaranteed
to be called (possibly redundantly) when the watch against the pcstore
expires.


pkg/kp/pcstore/consul_store.go
```diff
@@ -496,6 +506,17 @@ func (s *consulStore) handlePCUpdates(concrete ConcreteSyncer, changes chan podC
 	}
 }
 
+func labeledEqual(left, right []labels.Labeled) bool {
+	leftSet, rightSet := sets.NewString(), sets.NewString()
```
mpuncel: nit: you can do `sets.NewString(left...), sets.NewString(right...)`

mpuncel: nope no you can't

mpuncel:
saweet 👍

PR #495, opened 2016-05-19
Make the pcstore constructor take api.Client as a pointer

mpuncel:
This makes it consistent with our other store types instead of being the
single case that takes a value instead of a pointer.


anthonybishopric:
👌

PR #497, opened 2016-05-20
Further optimize calls to SyncCluster

anthonybishopric:
We now only call `SyncCluster` when the pod cluster object itself
has changed. When `WatchAndSync` is first called it will proceed
to call `SyncCluster` for all returned pod clusters and will only
submit differences following that.

### DIFF
--- a/pkg/pc/fields/fields.go
+++ b/pkg/pc/fields/fields.go
@@ -2,6 +2,7 @@ package fields
 
 import (
 	"encoding/json"
+	"reflect"
 
 	"github.com/square/p2/pkg/kp/rcstore"
 	"github.com/square/p2/pkg/types"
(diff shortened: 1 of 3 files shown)


pkg/pc/fields/fields.go
```diff
@@ -58,6 +59,29 @@ type PodCluster struct {
 	Annotations Annotations
 }
 
+func (pc *PodCluster) Equals(other *PodCluster) bool {
+	if pc == nil && other == nil {
```
mpuncel: can you not just return `reflect.DeepEqual(*pc, *other)`?

anthonybishopric: You would still need the nil check to prevent a panic - I wanted to be explicit about the other properties I compared, but I'm not married to it

mpuncel: i'm fine with it, might just be nice for new fields being added to be automatically included in equality

mpuncel: although maybe some fields _shouldn't_ be included like if there's a status or something someday

mpuncel:
👍

PR #498, opened 2016-05-23
add a PodClusterAlreadyExists error type

mpuncel:
This allows callers to attempt to create pod clusters and have custom
handling for the case where a pod cluster creation fails due to the pod
cluster already existing. This is useful when relying on idempotent
creation for example.


anthonybishopric:
👍

PR #499, opened 2016-05-23
Fix buggy loop condition + watch pod clusters in p2-watch

anthonybishopric:


mpuncel:
👍

bin/p2-watch/main.go
```diff
@@ -32,32 +41,77 @@ func main() {
...
+		podCh := make(chan []kp.ManifestResult)
+		go store.WatchPods(podPrefix, types.NodeName(*nodeName), quit, errChan, podCh)
+		for {
+			select {
+			case results := <-podCh:
+				if len(results) == 0 {
+					fmt.Println(fmt.Sprintf("No manifests exist for %s under %s (they may have been deleted)", *nodeName, podPrefix))
+				} else {
+					for _, result := range results {
+						fmt.Println("")
```
petertseng: I realize this is just an indentation change, but I am a bit curious - what was this used for?

petertseng: oh I guess it makes it easier to understand where one output ends and one begins

anthonybishopric: i'm pretty sure the `fmt.Println` is a mistake and I'll remove it

PR #500, opened 2016-05-24
Hzeng.pc update

KoishiKomeiji:
To implement pod cluster update.

### DIFF
--- a/pkg/kp/pcstore/consul_store.go
+++ b/pkg/kp/pcstore/consul_store.go
@@ -117,7 +117,10 @@ func (s *consulStore) Create(
 	err = s.setLabelsForPC(pc)
 	if err != nil {
 		// TODO: what if this delete fails?
-		_ = s.Delete(pc.ID)
+		deleteErr := s.Delete(pc.ID)
+		if deleteErr != nil {
+			err = util.Errorf("%s\n%s", err, deleteErr)
+		}
 		return fields.PodCluster{}, err
 	}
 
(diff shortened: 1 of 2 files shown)


pkg/kp/pcstore/consul_store.go
```diff
@@ -113,6 +113,62 @@ func (s *consulStore) Create(
...
+	podID types.PodID,
+	availabilityZone fields.AvailabilityZone,
+	clusterName fields.ClusterName,
+	podSelector klabels.Selector,
+	annotations fields.Annotations,
+	session Session,
+	oldPC fields.PodCluster,
+	modifyIndex uint64,
+) (fields.PodCluster, error) {
+	newPC := fields.PodCluster{
```
anthonybishopric: Why not just let the caller pass a `fields.PodCluster` as `newPC` that already has the necessary ID?

KoishiKomeiji: Alright, I fixed the parameters

pkg/kp/pcstore/consul_store_test.go
```diff
@@ -18,6 +18,85 @@ import (
...
+	}
+
+	if newPc.Annotations["bar"] != "foo" {
+		t.Errorf("Annotations didn't match expected")
+	} else if _, ok := newPc.Annotations["foo"]; ok {
+		t.Errorf("Annotations didn't match expected")
+	}
+}
+
+func GetGoodCreatePodCluster(store *consulStore, t *testing.T) fields.PodCluster {
```
rudle: I think this makes more sense as a private method called `createPodCluster`. No need to denote it as "good" or say that we "get" it as a return type. Those facts are implied.

pkg/kp/pcstore/consul_store.go
```diff
@@ -113,6 +113,50 @@ func (s *consulStore) Create(
...
+	}
+
+	if !success {
+		return fields.PodCluster{}, util.Errorf("Could not set pod cluster at path '%s'", key)
+	}
+
+	err = s.setLabelsForPC(pc)
+	if err != nil {
+		// TODO: what if this delete fails?
+		_ = s.Delete(pc.ID)
```
rudle: This needs to be logged at the very least.

anthonybishopric: Agreed, although this code was copy/pasted from `Create`, which should also be fixed

mpuncel:
Something to think about, I kind of like the way `mutateRc` works: https://github.com/mpuncel/p2/blob/master/pkg/kp/rcstore/consul_store.go#L486

You don't need to know the full original structure, you just change the part you care about. This looks good though just a little more work for the caller arguably

anthonybishopric:
Mike's suggestion is really good, particularly because you can do things like [this](https://github.com/mpuncel/p2/blob/master/pkg/kp/rcstore/consul_store.go#L507) to make it so callers don't necessarily need to retrieve and pass the index value on their own

pkg/kp/pcstore/consul_store.go
KoishiKomeiji: Updated

pkg/kp/pcstore/consul_store_test.go
KoishiKomeiji: Updated

KoishiKomeiji:
The arguments for the update function have been updated.

mpuncel:
looks good to me, it may make sense to use the `mutatePC` model instead at some later time to encompass more types of mutations and require less work by the caller, but this looks good for now! 👍

anthonybishopric:
@KoishiKomeiji I think you'll have to implement the `Update` function using the pattern Mike suggests - or at least accept an `index` field in the call. As written, `Update` will take a `fields.PodCluster` which _may have changed_ from the one returned from the `Get` call done inside. This makes the safety guarantee provided by `CAS` not very useful.

KoishiKomeiji:
Updated, the Update function has been renamed to mutatePc. It now takes a fields.ID, mutator function, modifyIndex, and session instead.

pkg/kp/pcstore/consul_store.go
```diff
@@ -154,6 +157,76 @@ func (s *consulStore) Delete(id fields.ID) error {
...
 }
 
+// performs a safe (ie check-and-set) mutation of the pc with the given id,
+// using the given function
+// if the mutator returns an error, it will be propagated out
+// if the returned PC has id="", then it will be deleted
+func (s *consulStore) mutatePc(
+	id fields.ID,
+	mutator func(fields.PodCluster) (fields.PodCluster, error),
+	modifyIndex uint64,
```
mpuncel: sorry for further churn, but you should get the `modifyIndex` from your `Get()` call below, then the caller doesn't need to fetch the pod cluster beforehand at all!

pkg/kp/pcstore/consul_store.go
```diff
@@ -154,6 +157,76 @@ func (s *consulStore) Delete(id fields.ID) error {
...
+	if kvp == nil {
+		return fields.PodCluster{}, NoPodCluster
+	}
+
+	pc, err := kvpToPC(kvp)
+	if err != nil {
+		return fields.PodCluster{}, err
+	}
+
+	var success bool
```
mpuncel: nit: but this right before you use it

mpuncel:
Looks great, just need to get rid of that `modifyIndex` argument

pkg/kp/pcstore/consul_store.go
```diff
@@ -154,6 +157,76 @@ func (s *consulStore) Delete(id fields.ID) error {
 	return s.applicator.RemoveAllLabels(labels.PC, id.String())
 }
 
+// performs a safe (ie check-and-set) mutation of the pc with the given id,
+// using the given function
+// if the mutator returns an error, it will be propagated out
+// if the returned PC has id="", then it will be deleted
+func (s *consulStore) mutatePc(
```
mpuncel: also you're going to have to name this `MutatePC` otherwise it won't be exported outside this package

anthonybishopric: In Go, functions with lowercase first letters aren't exposed outside the package (in this case, the package is `pcstore`. We'll want to be calling this from our internal repo, so you want to name this `MutatePC`

KoishiKomeiji:
Alright, I've renamed the functions and am now using meta.LastIndex

mpuncel:
🚢

PR #501, opened 2016-05-24
Log bridge metrics

rudle:
logbridge needs a metric registry so we can understand log lines per unit time and log bytes per unit time.

### DIFF
--- a/pkg/logbridge/logbridge.go
+++ b/pkg/logbridge/logbridge.go
@@ -6,23 +6,47 @@ import (
 	"io"
 	"time"
 
+	"github.com/square/p2/Godeps/_workspace/src/github.com/rcrowley/go-metrics"
 	"github.com/square/p2/pkg/logging"
 )
+
+type LogBridge struct {
+	Reader        io.Reader
+	DurableWriter io.Writer
+	LossyWriter   io.Writer
+
+	logger  logging.Logger
+	metrics metrics.Registry
+}
+
+type MetricsRegistry interface {
+	Register(metricName string, metric interface{}) error
+}
+
+func NewLogBridge(r io.Reader, durableWriter io.Writer, lossyWriter io.Writer, logger logging.Logger, metricsRegistry metrics.Registry) *LogBridge {
+	return &LogBridge{
+		Reader:        r,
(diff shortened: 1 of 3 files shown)


pkg/logbridge/logbridge.go
```diff
@@ -6,50 +6,80 @@ import (
...
+type MetricsRegistry interface {
+	Register(metricName string, metric interface{}) error
+}
+
+func NewLogBridge(r io.Reader,
+	durableWriter io.Writer,
+	lossyWriter io.Writer,
+	logger logging.Logger,
+	metricsRegistry MetricsRegistry,
+	loglineMetricKeyName string,
```
anthonybishopric: We should standardize on a way to name metrics, as the convention elsewhere in this repo is to name the keys whatever we please when the registry is passed to us (ie they are constants). Check consul_aggregator.go for an example.

I'm not married by any means to that approach, in fact I'd almost prefer a central location for metric names - lets dedicate a few minutes of time to figuring that out.

anthonybishopric: Also not important that it's sorted in this PR per se, but I'd like to track it at the very least.

rudle: my primary concern about injecting here was not muddling up the constructor with: podID, az, clusterName and host. I agree that centralizing them would be good - but I think such belongs on the Square side of things.

anthonybishopric: There is probably a reasonable middle ground in letting this project define metric suffixes and letting the Square side provide a single prefix - that would get us the best of both worlds presumably

rudle: I don't know that we want a "single prefix" for all of Square. I would say that we want a single prefix for all pod-wise metrics like this one and another single prefix for all (for example) datastore watch metrics.

I think that we're agreeing, but I want to clarify the goal.

mpuncel: i agree we need to standardize metrics names somehow. but doesn't have to block this PR

mpuncel:
👍

rudle:
Thanks for the review!

PR #502, opened 2016-05-24
Collect metrics on SyncCluster() time

mpuncel:
For certain syncer types, it may take non-negligible amounts of time to
sync a pod cluster. This commit adds support for providing a metrics
registry to the pod cluster store. If a registry is set, a metric will
be added for pod cluster sync time and data points will be collected.


petertseng:
I consider myself inauthoritative on this, but I saw no problems. `GetType` may be less idiomatic than `Type`, I suspect

rudle:
👍

anthonybishopric:
🚢

PR #503, opened 2016-05-24
Log bridge rate limits

rudle:
I'm not sure what numbers to fill in here yet. Will need to deploy & observe the results of my previous PR to determine the constants.

### DIFF
--- a/pkg/logbridge/logbridge.go
+++ b/pkg/logbridge/logbridge.go
@@ -7,6 +7,7 @@ import (
 	"time"
 
 	"github.com/square/p2/Godeps/_workspace/src/github.com/rcrowley/go-metrics"
+	"github.com/square/p2/Godeps/_workspace/src/golang.org/x/time/rate"
 	"github.com/square/p2/pkg/logging"
 )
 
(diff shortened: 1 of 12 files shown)


pkg/logbridge/logbridge.go
```diff
@@ -63,6 +80,17 @@ func (lb *LogBridge) LossyCopy(r io.Reader, capacity int) {
 	var n int
 	var err error
 	for line := range lines {
+		err = lb.logLineRateLimit.WaitN(context.TODO(), 1)
```
anthonybishopric: for maximum awesome, should we extract this fake context out of the for loop?

rudle: allocations in the loop all day long

pkg/logbridge/logbridge.go
```diff
@@ -100,8 +129,9 @@ func (lb *LogBridge) lossyCopy(r io.Reader, lines chan []byte) {
 		case lines <- line:
 		default:
 			droppedLines++
+			lb.droppedLineCount.Inc(1)
```
anthonybishopric: Isn't this limit just based on the channel capacity and not the lovely rate limiter you've just added?

rudle: the rate limit will cause the channel (buffer) to fill up which will cause us to fall into this case

I agree that this isn't a precise metric, but I think it's better than nothing?

anthonybishopric: Discussed IRL, will switch to `ReserveN` in order to track a second gauge related to how long we were waiting

rudle:
@anthonybishopric PTAL I've changed this to use `Reserve…` vs `Wait…`

rudle:
Wondering how badly the histogram will become skewed in the case that the application under log stops logging to STDOUT. We can emit a heartbeat update of "0ms" via a separate go routine, but that would seem to imply synchronization inside logbridge's main loop (which is undesirable for performance reasons).

rudle:
This uses a simple counter to address the concern I highlighted in: https://github.com/square/p2/pull/503#issuecomment-222839086

bonus: the tests are passing.

anthonybishopric:
👍

rudle:
thanks for the review. Gonna fixup the godeps (removing net/context :key:) squash and merge.

PR #505, opened 2016-05-24
Fix measurement of cluster sync time

mpuncel:
Previously we were only measuring the time it takes for the
SyncCluster() call to execute from start to finish. It's more
interesting to track the time between a change being noticed and
SyncCluster() being finished.

In other words, we want to include the time it takes for handlePCUpdates
to actually trigger the SyncCluster() call as well.


rudle:
:shipit:

PR #506, opened 2016-05-25
Time out hook execution after 60s

cuzelac:
- Create `Hook` struct; use it for executing hooks
- Add `RunWithTimeout` method to `Hook`
    - Test it with some temp files
- Default timeout to 5s

It feels like the hook package is ready for a refactor to take into account the Hook type (which feels really natural for this package), but I feel that's out of scope of this PR.

Please double-check my style and adherence to convention.  I'm still picking up the idioms of Go.

### DIFF
--- a/pkg/hooks/hooks.go
+++ b/pkg/hooks/hooks.go
@@ -7,6 +7,7 @@ import (
 	"os"
 	"os/exec"
 	"path"
+	"time"
 
 	"github.com/square/p2/Godeps/_workspace/src/github.com/Sirupsen/logrus"
 	"github.com/square/p2/pkg/logging"
@@ -23,6 +24,8 @@ const (
 	HOOKED_POD_MANIFEST_ENV_VAR = "HOOKED_POD_MANIFEST"
 	HOOKED_CONFIG_PATH_ENV_VAR  = "HOOKED_CONFIG_PATH"
 	HOOKED_ENV_PATH_ENV_VAR     = "HOOKED_ENV_PATH"
+
+	DefaultTimeout = 60 * time.Second
 )
 
 type Pod interface {
@@ -74,6 +77,7 @@ func Hooks(dirpath string, logger *logging.Logger) *HookDir {
 	return &HookDir{dirpath, logger}
 }
 
+// runDirectory executes all executable files in a given directory path.
 func runDirectory(dirpath string, environment []string, logger logging.Logger) error {
 	entries, err := ioutil.ReadDir(dirpath)
 	if os.IsNotExist(err) {
@@ -94,27 +98,88 @@ func runDirectory(dirpath string, environment []string, logger logging.Logger) e
 		if f.IsDir() {
 			continue
 		}
-		logger.WithField("path", fullpath).Infof("Executing hook %s", f.Name())
-		cmd := exec.Command(fullpath)
-		hookOut := &bytes.Buffer{}
-		cmd.Stdout = hookOut
-		cmd.Stderr = hookOut
-		cmd.Env = environment
-		err := cmd.Run()
-		if err != nil {
+
+		h := Hook{fullpath, f.Name(), DefaultTimeout, environment, logger}
+
+		err := h.RunWithTimeout()
+		if htErr, ok := err.(HookTimeoutError); ok {
+			logger.WithErrorAndFields(htErr, logrus.Fields{
+				"path":      h.Path,
+				"hook_name": h.Name,
+				"timeout":   h.Timeout,
+			}).Warnln(htErr.Error())
+			// we intentionally swallow timeout HookTimeoutErrors
+		} else if err != nil {
 			logger.WithErrorAndFields(err, logrus.Fields{
-				"path":   fullpath,
-				"output": hookOut.String(),
-			}).Warnf("Could not execute hook %s", f.Name())
-		} else {
-			logger.WithFields(logrus.Fields{
-				"path":   fullpath,
-				"output": hookOut.String(),
-			}).Debugln("Executed hook")
+				"path":      h.Path,
+				"hook_name": h.Name,
+			}).Warningf("Unknown error in hook %s: %s", h.Name, err)
 		}
 	}
 
 	return nil
+}
+
+type Hook struct {
+	Path    string // path to hook's executable
+	Name    string // human-readable name of Hook
+	Timeout time.Duration
+	env     []string // unix environment in which the hook should be executed
+	logger  logging.Logger
+}
+
+// HookTimeoutError is returned when a Hook's execution times out
+type HookTimeoutError struct {
+	Hook Hook
+}
+
+func (e HookTimeoutError) Error() string {
+	sec := e.Hook.Timeout / time.Millisecond
+	return fmt.Sprintf("Hook %s timed out after %dms", e.Hook.Name, sec)
+}
+
+// RunWithTimeout runs the hook but returns a HookTimeoutError when it exceeds its timeout
+//
+// Necessary because Run() hangs if the command double-forks without properly
+// re-opening its fd's and exec.Start() will dutifully wait on any unclosed fd
+//
+// NB: in the event of a timeout this will leak descriptors
+func (h *Hook) RunWithTimeout() error {
+	finished := make(chan struct{})
+	go func() {
+		h.Run()
+		close(finished)
+	}()
+
+	select {
+	case <-finished:
+	case <-time.After(h.Timeout):
+		return HookTimeoutError{*h}
+	}
+
+	return nil
+}
+
+// Run executes the hook in the context of its environment and logs the output
+func (h *Hook) Run() {
+	h.logger.WithField("path", h.Path).Infof("Executing hook %s", h.Name)
+	cmd := exec.Command(h.Path)
+	hookOut := &bytes.Buffer{}
+	cmd.Stdout = hookOut
+	cmd.Stderr = hookOut
+	cmd.Env = h.env
+	err := cmd.Run()
+	if err != nil {
+		h.logger.WithErrorAndFields(err, logrus.Fields{
+			"path":   h.Path,
+			"output": hookOut.String(),
+		}).Warnf("Could not execute hook %s", h.Name)
+	} else {
+		h.logger.WithFields(logrus.Fields{
+			"path":   h.Path,
+			"output": hookOut.String(),
+		}).Debugln("Executed hook")
+	}
 }
 
 func (h *HookDir) runHooks(dirpath string, hType HookType, pod Pod, podManifest pods.Manifest, logger logging.Logger) error {
--- a/pkg/hooks/hooks_test.go
+++ b/pkg/hooks/hooks_test.go
@@ -4,7 +4,9 @@ import (
 	"io/ioutil"
 	"os"
 	"path"
+	"path/filepath"
 	"testing"
+	"time"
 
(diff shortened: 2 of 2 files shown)


pkg/hooks/hooks.go
```diff
@@ -94,29 +96,83 @@ func runDirectory(dirpath string, environment []string, logger logging.Logger) e
...
-				"path":   fullpath,
-				"output": hookOut.String(),
-			}).Debugln("Executed hook")
+
+		timeout := 5
+		h := Hook{fullpath, f.Name(), environment, logger, timeout}
+
+		err := h.RunWithTimeout()
+		htErr, ok := err.(HookTimeoutError)
+		if !ok {
```
petertseng: (what does this do if `err == nil`? will it warn?

pkg/hooks/hooks.go
```diff
@@ -94,29 +96,83 @@ func runDirectory(dirpath string, environment []string, logger logging.Logger) e
...
+			logger.WithField("path", h.Path).Warnln(htErr.Error())
```
petertseng: someone kno knows better than I will have to confirm, but is [`WithError`](https://github.com/square/p2/blob/master/pkg/logging/logging.go#L62) and/or  [`WithErrorAndFields`](https://github.com/square/p2/blob/master/pkg/logging/logging.go#L81) be appropriate?

pkg/hooks/hooks.go
```diff
@@ -94,29 +96,83 @@ func runDirectory(dirpath string, environment []string, logger logging.Logger) e
...
 
 	return nil
 }
 
+type Hook struct {
+	Path    string   // path to hook's executable
+	Name    string   // human-readable name of Hook
+	env     []string // unix environment in which the hook should be executed
+	logger  logging.Logger
+	timeout int // number of seconds before hook's execution times out
```
mpuncel: nit: make this a `time.Duration` and then you don't have to remember that this is seconds and do the multiplication below

pkg/hooks/hooks.go
```diff
@@ -94,29 +96,83 @@ func runDirectory(dirpath string, environment []string, logger logging.Logger) e
...
+}
+
+// HookTimeoutError is returned when a Hook's execution times out
+type HookTimeoutError struct {
+	H Hook
```
petertseng: can this simply be `Hook`?

pkg/hooks/hooks.go
```diff
@@ -94,29 +96,83 @@ func runDirectory(dirpath string, environment []string, logger logging.Logger) e
...
-				"path":   fullpath,
-				"output": hookOut.String(),
-			}).Warnf("Could not execute hook %s", f.Name())
-		} else {
-			logger.WithFields(logrus.Fields{
-				"path":   fullpath,
-				"output": hookOut.String(),
-			}).Debugln("Executed hook")
+
+		timeout := 5
```
mpuncel: if you make my `time.Duration` suggestion below this becomes `5 * time.Second`

pkg/hooks/hooks.go
petertseng: can this simply be `Hook`?

When accessing the member you would have to use `Hook` instead of `H`, but perhaps good for clarity, and it is not such a big loss

pkg/hooks/hooks_test.go
```diff
@@ -77,3 +79,62 @@ func testManifest() pods.Manifest {
...
+func TestHookRunWithTimeout(t *testing.T) {
+	type sleepTest struct {
+		sleep      int
+		timeout    int
+		shouldPass bool
+	}
+
+	var sleepTests = []sleepTest{
+		{1, 2, true},
+		{2, 1, false},
```
mpuncel: this will be kind of a long running test, i wonder if there's a way to cut it down. Maybe if you abstract out the use of `time.After` as the channel selected on you can just write immediately do it

pkg/hooks/hooks.go
mpuncel: yeah those would be appropriate, that way we get line number and file information as well

pkg/hooks/hooks.go
cuzelac: I had considered this, but I left it this way in case we ever want to allow hooks to configure their own timeouts, which would probably be in integer seconds.

I'm not at all tied to this and am happy to change it.  Let me know.

anthonybishopric: Mike's suggestion is more idiomatic in this codebase. And who can pass up an opportunity to use `time.ParseDuration`?

pkg/hooks/hooks_test.go
anthonybishopric: If you use a `time.Duration` you could make this test run in acceptable time with `100ms` vs `50ms` or similar. Mike's suggestion is strictly less flimsy than mine though :)

pkg/hooks/hooks.go
anthonybishopric: I'm pretty sure this isn't correct - if `!ok`, then `htErr` is `nil` because the type assertion will have failed. More likely you want to enter this condition `if ok {`. You'll also want to verify that `err` is not nil.

pkg/hooks/hooks_test.go
mpuncel: i'd settle for tests that short, definitely easier that way. although you'd have to use a command other than `sleep` and then we're getting into race condition territory

pkg/hooks/hooks.go
cuzelac: Absolutely.  I'll make the change.

pkg/hooks/hooks_test.go
cuzelac: I had the solution @anthonybishopric proposed in my prototype and it works well enough (I could use nanoseconds as the unit as long as the difference between the two was at least 3 orders of magnitude).  This particular test is complicated by the fact that a Hook has to shell out and uses `sleep`.

`sleep` isn't implemented the same way on every system and sometimes only allows integers so this solves for the least common denominator.  However, `sleep` on macos and RHEL6 *do* support decimal values.  I'm not sure what travis-ci uses, but I can give it a try.

pkg/hooks/hooks.go
cuzelac: Works for me.  Adding it to my todo list.

pkg/hooks/hooks.go
cuzelac: Added to my todo list.

pkg/hooks/hooks.go
cuzelac: I thought I double-checked this, as I'm pretty fuzzy with the type assertion in Go and this took a couple tries to get right.  But if I did, it's lost to a git rebase.

I'll verify and clarify or fix it.

cuzelac:
Collecting my todos here.

* [test should run faster](https://github.com/square/p2/pull/506#discussion-diff-64654724)
* [rename var](https://github.com/square/p2/pull/506/files#r64654589)
* [use a `time.Duration` in the `Hook` struct](https://github.com/square/p2/pull/506#discussion-diff-64654289)
    * [change timeout here too](https://github.com/square/p2/pull/506#discussion-diff-64654328)
* [pass in error to logger](https://github.com/square/p2/pull/506#discussion-diff-64654203)
* [verify that type assertion is correct; fix if necessary](https://github.com/square/p2/pull/506#discussion-diff-64653727)

pkg/hooks/hooks_test.go
anthonybishopric: It doesn't matter if `sleep` takes seconds or not, if you can use a command like `sleep 10` and then time it out after `50ms` then that's fast. If you then make the other test just `echo hi`, it should also run quickly. It's okay if this test is only slow if the underlying code is broken, since presumably it won't be broken most of the time.

pkg/hooks/hooks.go
cuzelac: Yes this is a logic error.  Thanks for catching @anthonybishopric 

> what does this do if err == nil? will it warn?

@petertseng - it's actually kind of odd to me that this even works, isn't it?  I was puzzled so I wrote some example code to figure out the behavior, but it turns out that it's right in the [language reference](https://golang.org/ref/spec#Type_assertions):

> A type assertion used in an assignment or initialization of the special form
>     v, ok = x.(T)
>     v, ok := x.(T)
>     var v, ok = x.(T)
> yields an additional untyped boolean value. The value of ok is true if the assertion holds. Otherwise it is false and the value of v is the zero value for type T. No run-time panic occurs in this case.

pkg/hooks/hooks_test.go
cuzelac: That's a good point.  I was concerned with testing the working case too, but that's clearly done in the above tests in the file.

pkg/hooks/hooks.go
```diff
@@ -94,29 +96,85 @@ func runDirectory(dirpath string, environment []string, logger logging.Logger) e
...
-				"path":   fullpath,
-				"output": hookOut.String(),
-			}).Warnf("Could not execute hook %s", f.Name())
-		} else {
-			logger.WithFields(logrus.Fields{
-				"path":   fullpath,
-				"output": hookOut.String(),
-			}).Debugln("Executed hook")
+
+		timeout := 5 * time.Second
```
anthonybishopric: Extract to a `const`

anthonybishopric: Preferably called `DefaultTimeout` since we will eventually make this configurable. Also, the value 5 seconds is probably too short for some of our internal hooks. Would recommend extending to 60 seconds.

pkg/hooks/hooks.go
```diff
@@ -94,29 +96,85 @@ func runDirectory(dirpath string, environment []string, logger logging.Logger) e
...
+		h := Hook{fullpath, f.Name(), &timeout, environment, logger}
+
+		err := h.RunWithTimeout()
```
anthonybishopric: Although there's no way to successfully return an error from `RunWithTimeout` that's _not_ a `HookTimeoutError`, it would be nice to see a `else if err != nil` case below

pkg/hooks/hooks.go
cuzelac: Done.

pkg/hooks/hooks.go
cuzelac: Would you like it to return the error, or just log it (and continue to the next hook)?

anthonybishopric: Logging the error is sufficient, hook executions should be independent of each other

anthonybishopric:
🚢 !

pkg/hooks/hooks_test.go
```diff
@@ -77,3 +79,44 @@ func testManifest() pods.Manifest {
 	builder.SetID(podId)
 	return builder.GetManifest()
 }
+
+func TestHookRunWithTimeout(t *testing.T) {
```
petertseng: I remember the initial version had a run that would succeed (2 second timeout, 1 second sleep) - did we consider this not worth keeping? I might have liked it! Perhaps just a simple run of `true` with some nonzero timeout which should end quickly (I believe that was suggested before)

cuzelac: The case gets exercised above in `TestExecutableHooksAreRun`, so I left it out here.

I didn't refresh before merging, so I just saw this.  I don't mind adding it back in a quick PR if you like.

petertseng: oh okay, I think as long as it is in fact exercised I'll be OK then.

PR #509, opened 2016-05-27
Add desired_replicas and minimum_replicas to RU logger

mpuncel:


anthonybishopric:
👍

petertseng:
On the one hand, I think in some messages emitted by `Run`, we don't care, but I suspect we'd much rather have these everywhere (including places we don't need) than manually go through and figure out each and every place where we might want to show these two values. So 👍

KoishiKomeiji:
👍

PR #510, opened 2016-05-27
Fix a bug where a label watch wouldn't be started for some pod clusters

mpuncel:
pcstore.WatchAndSync() calls the concrete syncer's GetInitialClusters()
function in order to notify the syncers of pod cluster deletions that
occurred when the syncer wasn't running.

However, handlePCUpdates only begins a label query on the pod cluster's
pod selector if something has changed, and there was no apparent change
since we were seeding the data with the GetInitialClusters() call.

This commit fixes that by only seeding the pod cluster changes with the
ID of the pod cluster if it was returned in GetInitialClusters().

The behavior change of this is that a watch will be started for every
pod cluster, and a SyncCluster() call will also be called once per pod
cluster even if the view of the world hasn't changed, perhaps resulting
in some performance degradation when the syncer starts up.

### DIFF
--- a/pkg/kp/pcstore/consul_store.go
+++ b/pkg/kp/pcstore/consul_store.go
@@ -490,17 +490,13 @@ func (s *consulStore) getInitialClusters(syncer ConcreteSyncer) (WatchedPodClust
 	}
 
 	for _, id := range initial {
-		existing, err := s.Get(id)
-		if err == NoPodCluster {
-			s.logger.WithField("pc_id", id).Warnf("Could not find initial cluster %v, will call DeleteCluster momentarily", id)
-			existing = fields.PodCluster{
-				ID: id,
-			}
-		} else if err != nil {
-			s.logger.WithField("pc_id", id).Errorln("Error retrieving pod cluster from consul")
-			return prevResults, err
-		}
-		prevResults.Clusters = append(prevResults.Clusters, &existing)
+		// We only populate the existing cluster with the ID. It's a little strange
+		// but this means that there will definitely be a "change" for the cluster
+		// when the watch is started, resulting in the label watch being started
+		// as well
+		prevResults.Clusters = append(prevResults.Clusters, &fields.PodCluster{
+			ID: id,
+		})
 	}
 	return prevResults, nil
 }
--- a/pkg/kp/pcstore/consul_store_test.go
+++ b/pkg/kp/pcstore/consul_store_test.go
@@ -913,10 +913,127 @@ func TestClosedChangeChannelResultsInTermination(t *testing.T) {
 	}
 }
 
+type RecordingSyncer struct {
+	InitialClusters []fields.ID
+
+	SyncClusterCalls   []fields.ID
+	DeleteClusterCalls []fields.ID
+
+	// Used to signal when SyncCluster is called so tests can have non-racey
+	// timeouts
+	SyncSignal chan<- struct{}
+}
+
+func (r *RecordingSyncer) GetInitialClusters() ([]fields.ID, error) {
+	return r.InitialClusters, nil
+}
+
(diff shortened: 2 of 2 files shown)


pkg/kp/pcstore/consul_store_test.go
```diff
@@ -913,10 +913,125 @@ func TestClosedChangeChannelResultsInTermination(t *testing.T) {
...
+		case <-syncSignal:
+			actualSyncCount++
+		}
+	}
+
+	for _, expectedID := range []fields.ID{pc1.ID, pc2.ID} {
+		found := false
+		for _, id := range syncer.SyncClusterCalls {
+			if id == expectedID {
+				found = true
```
petertseng: optionally, can `break` here, but I don't care

pkg/kp/pcstore/consul_store_test.go
```diff
@@ -913,10 +913,125 @@ func TestClosedChangeChannelResultsInTermination(t *testing.T) {
...
+			t.Fatalf("Couldn't start WatchAndSync(): %s", err)
+		}
+	}()
+
+	expectedSyncCount := 2
+	actualSyncCount := 0
+	for actualSyncCount < expectedSyncCount {
+		select {
+		case <-time.After(1 * time.Second):
+			t.Fatalf("Timed out waiting for sync to happen, there were %d syncs", actualSyncCount)
```
petertseng: it may be useful if we are debugging a test failure to show the contents of SyncClusterCalls, if that's not too hard?

pkg/kp/pcstore/consul_store_test.go
```diff
@@ -913,10 +913,125 @@ func TestClosedChangeChannelResultsInTermination(t *testing.T) {
...
+	for _, expectedID := range []fields.ID{pc1.ID, pc2.ID} {
+		found := false
+		for _, id := range syncer.SyncClusterCalls {
+			if id == expectedID {
+				found = true
+			}
+		}
+
+		if !found {
+			t.Fatalf("Expected sync to be called for %s but it wasn't", expectedID)
```
petertseng: also maybe show contents of SyncClusterCalls here?

pkg/kp/pcstore/consul_store_test.go
```diff
@@ -913,10 +913,125 @@ func TestClosedChangeChannelResultsInTermination(t *testing.T) {
...
+}
+
+func (r *RecordingSyncer) Type() ConcreteSyncerType {
+	return "recording"
+}
+
+func TestWatchAndSync(t *testing.T) {
+	store := consulStoreWithFakeKV()
+	quit := make(chan struct{})
+	syncSignal := make(chan struct{})
```
petertseng: does someone close this, and should someone?

pkg/kp/pcstore/consul_store.go
```diff
@@ -500,7 +500,14 @@ func (s *consulStore) getInitialClusters(syncer ConcreteSyncer) (WatchedPodClust
 			s.logger.WithField("pc_id", id).Errorln("Error retrieving pod cluster from consul")
 			return prevResults, err
 		}
-		prevResults.Clusters = append(prevResults.Clusters, &existing)
+
```
anthonybishopric: I think this whole function could actually just be simplified - I don't see any reason to individually retrieve each pod cluster anymore to test for existence, given that the value we put in `prevResults` will always be the same. We should just fill `prevResults` with `&fields.PodCluster{ID: id}` in this loop and call it a day.

Is there any reason not to?

mpuncel: oh you're right that would be simpler

anthonybishopric:
👍

PR #511, opened 2016-05-31
Use a metrics.Gauge to measure pod cluster sync time

mpuncel:
Previously a metrics.Timer was used, but it is more straightforward to
use a Gauge. The most recent processing time can be extracted with
gauge.Value()


KoishiKomeiji:
:+1:

PR #512, opened 2016-05-31
Use a histogram instead of a gauge for recording pod cluster sync time

mpuncel:
A histogram makes more sense here because pod cluster updates will be
frequent and fast, while a metrics graphing library may only sample
metrics every 30 seconds for example. A histogram metric will provide a
more useful average with an exponentially decaying window to smooth out
outliers.


KoishiKomeiji:
:+1: As long as the numbers for the exponential decay work out.
There's also a pkg/kp/pcstore/output in there too, don't know if it's important.

mpuncel:
good catch, removed

PR #514, opened 2016-05-31
Support a configured pod blacklist for logbridge

rudle:
apparently svlogdExec was unused before this change. The type change
did not cause any compiler error.

### DIFF
--- a/pkg/preparer/orchestrate.go
+++ b/pkg/preparer/orchestrate.go
@@ -17,7 +17,8 @@ import (
 // Used because the preparer special-cases itself in a few places.
 const POD_ID = types.PodID("p2-preparer")
 
-const svlogdExec = "svlogd -tt ./main"
+// slice literals are not const
+var svlogdExec = []string{"svlogd", "-tt", "./main"}
 
 type Pod interface {
 	hooks.Pod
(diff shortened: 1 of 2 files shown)


pkg/preparer/orchestrate.go
```diff
@@ -189,7 +190,18 @@ func (p *Preparer) handlePods(podChan <-chan ManifestPair, quit <-chan struct{})
 				if pod.Id == POD_ID {
 					pod.DefaultTimeout = time.Duration(0)
 				}
-				pod.SetLogBridgeExec(p.logExec)
+
+				effectiveLogBridgeExec := p.logExec
+				// pods that are in the blacklist for this preparer shall not use the
+				// preparer's log exec. Instead, they will use the default svlogd logexec.
+				for _, podID := range p.logBridgeBlacklist {
+					if string(pod.Id) == podID {
```
mpuncel: nit: `pod.Id.String()`. This makes it easier for the compiler to validate things if we ever change the underlying type although I admit that's unlikely

mpuncel:
🆗

pkg/preparer/orchestrate.go
rudle: fair point.

PR #517, opened 2016-06-09
To commit fields for daemon sets along with the create, get, and list…

KoishiKomeiji:
… functions

### DIFF
--- a/pkg/ds/fields/fields.go
+++ b/pkg/ds/fields/fields.go
@@ -1,0 +1,121 @@
+package fields
+
+import (
+	"encoding/json"
+
+	"k8s.io/kubernetes/pkg/labels"
+
+	"github.com/square/p2/pkg/pods"
+	"github.com/square/p2/pkg/types"
+)
+
+// ID is a named type alias for DaemonSet IDs
+type ID string
+
+func (id ID) String() string {
+	return string(id)
+}
+
+// Cluster name is where the DaemonSet lives on
+type ClusterName string
+
+// DaemonSet holds the runtime state of a Daemon Set as saved in Consul.
+type DaemonSet struct {
+	// UUID for this DaemonSet
+	ID ID
+
+	// When disabled, this DaemonSet will not make any scheduling changes
+	Disabled bool
+
+	// The pod manifest that should be scheduled on nodes
+	Manifest pods.Manifest
+
+	// Minimum health for nodes when scheduling
+	MinHealth int
+
+	// DaemonSet's environment name
+	Name ClusterName
+
+	// Defines the set of nodes on which the manifest can be scheduled
+	NodeSelector labels.Selector
+
+	// PodID to deploy
+	PodID types.PodID
+}
+
+// RawDaemonSet defines the JSON format used to store data into Consul
+type RawDaemonSet struct {
+	ID           ID          `json:"id"`
+	Disabled     bool        `json:"disabled"`
+	Manifest     string      `json:"manifest"`
+	MinHealth    int         `json:"min_health"`
+	Name         ClusterName `json:"cluster_name"`
+	NodeSelector string      `json:"node_selector"`
+	PodID        types.PodID `json:"pod_id"`
+}
+
+// MarshalJSON implements the json.Marshaler interface for serializing the DS
+// to JSON format. Lets json interface know how to marshal a DaemonSet
+// Can be called using json.Marshal
+func (ds DaemonSet) MarshalJSON() ([]byte, error) {
+	var manifest []byte
+	var err error
+	if ds.Manifest != nil {
+		manifest, err = ds.Manifest.Marshal()
+		if err != nil {
+			return nil, err
+		}
+	}
+
+	var nodeSelector string
+	if ds.NodeSelector != nil {
+		nodeSelector = ds.NodeSelector.String()
+	}
+
+	return json.Marshal(RawDaemonSet{
+		ID:           ds.ID,
+		Disabled:     ds.Disabled,
+		Manifest:     string(manifest),
+		MinHealth:    ds.MinHealth,
+		Name:         ds.Name,
+		NodeSelector: nodeSelector,
+		PodID:        ds.PodID,
+	})
+}
+
+// Assert DaemonSet.MashalJSON is implemented in json.Marshaler
+var _ json.Marshaler = DaemonSet{}
+
+// UnmarshalJSON implements the json.Unmarshaler interface for deserializing the JSON
+// representation of an DS. Lets json interface know how to unmarshal a DaemonSet
+// Can be called using json.Unmarshal
+func (ds *DaemonSet) UnmarshalJSON(b []byte) error {
+	var rawDS RawDaemonSet
+	if err := json.Unmarshal(b, &rawDS); err != nil {
+		return err
+	}
+
+	podManifest, err := pods.ManifestFromBytes([]byte(rawDS.Manifest))
+	if err != nil {
+		return err
+	}
+
+	nodeSelector, err := labels.Parse(rawDS.NodeSelector)
+	if err != nil {
+		return err
+	}
+
+	*ds = DaemonSet{
+		ID:           rawDS.ID,
+		Disabled:     rawDS.Disabled,
+		Manifest:     podManifest,
+		MinHealth:    rawDS.MinHealth,
+		Name:         rawDS.Name,
+		NodeSelector: nodeSelector,
+		PodID:        rawDS.PodID,
+	}
+	return nil
+}
+
+// Assert DaemonSet.UnmarshalJSON is implemented in json.Unmarshaler
+var _ json.Unmarshaler = &DaemonSet{}
--- a/pkg/kp/dsstore/consul_store.go
+++ b/pkg/kp/dsstore/consul_store.go
@@ -1,0 +1,198 @@
+package dsstore
+
+import (
+	"encoding/json"
+	"errors"
+	"fmt"
+	"path"
+
+	"github.com/hashicorp/consul/api"
+	"github.com/pborman/uuid"
+	klabels "k8s.io/kubernetes/pkg/labels"
+
+	"github.com/square/p2/pkg/ds/fields"
+	"github.com/square/p2/pkg/kp/consulutil"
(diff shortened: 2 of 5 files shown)


pkg/ds/fields/fields.go
```diff
@@ -0,0 +1,117 @@
...
+
+// ID is a named type alias for Daemon Set IDs
+type ID string
+
+// Cluster name is where the daemon set lives on
+type ClusterName string
+
+// DaemonSet holds the runtime state of a Daemon Set as saved in Consul.
+type DaemonSet struct {
+	// UUID for this controller
```
rudle: Please update these comments to say "DaemonSet"

pkg/ds/fields/fields.go
```diff
@@ -0,0 +1,117 @@
...
+	Disabled bool
+
+	// The pod manifest that should be scheduled on nodes
+	Manifest pods.Manifest
+
+	// Minimum health for nodes when scheduling
+	MinHealth int
+
+	// Daemon Set's environment name
+	Name ClusterName
```
rudle: A daemon set could potentially reference nodes across many Clusters. How is this field used?

pkg/ds/fields/fields.go
```diff
@@ -0,0 +1,117 @@
...
+	// PodID to deploy
+	PodID types.PodID
+}
+
+// RawDaemonSet defines the JSON format used to store data into Consul
+type RawDaemonSet struct {
+	ID           ID          `json:"id"`
+	Disabled     bool        `json:"disabled"`
+	Manifest     string      `json:"manifest"`
+	MinHealth    int         `json:"MinHealth"`
```
rudle: `min_health` would match the other fields. Is there a good reason for this to be camelcased?

pkg/kp/dsstore/consul_store.go
```diff
@@ -0,0 +1,207 @@
+package dsstore
+
+import (
+	"encoding/json"
+	"errors"
+	"fmt"
+	"path"
+	"square/up/util"
```
rudle: I think you mean `pkg/util` here. This is why the build is failing.

pkg/ds/fields/fields.go
KoishiKomeiji: @anthonybishopric said that we may want to choose to deploy this on staging or production.

pkg/ds/fields/fields.go
KoishiKomeiji: updated

KoishiKomeiji:
Updated.

pkg/ds/fields/fields.go
mpuncel: @anthonybishopric do you not think that this should be on the pod cluster level?

mpuncel: while rolling out you'd have two DaemonSets with the same Name?

pkg/ds/fields/fields.go
```diff
@@ -0,0 +1,117 @@
...
+
+	// PodID to deploy
+	PodID types.PodID
+}
+
+// RawDaemonSet defines the JSON format used to store data into Consul
+type RawDaemonSet struct {
+	ID           ID          `json:"id"`
+	Disabled     bool        `json:"disabled"`
+	Manifest     string      `json:"manifest"`
```
mpuncel: does the manifest type really not marshal cleanly? not important for this PR but something we should fix.

pkg/kp/dsstore/consul_store.go
```diff
@@ -0,0 +1,207 @@
...
+	"github.com/square/p2/pkg/ds/fields"
+	"github.com/square/p2/pkg/kp/consulutil"
+	"github.com/square/p2/pkg/labels"
+	"github.com/square/p2/pkg/logging"
+	"github.com/square/p2/pkg/pods"
+	"github.com/square/p2/pkg/types"
+	"github.com/square/p2/pkg/util"
+)
+
+type kvPair struct {
```
mpuncel: this looks unused

pkg/kp/dsstore/consul_store.go
```diff
@@ -0,0 +1,207 @@
...
+type CASError string
+
+func (e CASError) Error() string {
+	return fmt.Sprintf("Could not check-and-set key %q", string(e))
+}
+
+// Make sure functions declared in the Store interface have the proper contract
+var _ Store = &consulStore{}
+
+func NewConsul(client *api.Client, retries int, logger *logging.Logger) *consulStore {
```
mpuncel: usually I return the interface type (e.g. `Store`) from these, but since it's an unexported type i'm not sure what will happen?

pkg/kp/dsstore/consul_store.go
```diff
@@ -0,0 +1,207 @@
...
+	name fields.ClusterName,
+	nodeSelector klabels.Selector,
+	podID types.PodID,
+) (fields.DaemonSet, error) {
+	if err := checkManifestPodID(podID, manifest); err != nil {
+		return fields.DaemonSet{}, err
+	}
+
+	ds, err := s.innerCreate(manifest, minHealth, name, nodeSelector, podID)
+
```
mpuncel: nit: get rid of this empty line

pkg/kp/dsstore/consul_store.go
```diff
@@ -0,0 +1,207 @@
...
+	if err != nil {
+		return fields.DaemonSet{}, util.Errorf("Could not marshal DS as json: %s", err)
+	}
+
+	success, _, err := s.kv.CAS(&api.KVPair{
+		Key:         dsPath,
+		Value:       rawDS,
+		ModifyIndex: 0,
+	}, nil)
+
```
mpuncel: nit: no newline

pkg/kp/dsstore/consul_store.go
```diff
@@ -0,0 +1,207 @@
...
+	}
+	if !success {
+		return fields.DaemonSet{}, CASError(dsPath)
+	}
+	return ds, nil
+}
+
+func (s *consulStore) Get(id fields.ID) (fields.DaemonSet, *api.QueryMeta, error) {
+	dsPath, err := s.dsPath(id)
+	var metadata *api.QueryMeta
```
mpuncel: move this above the call and put the error checking next to the function call

pkg/kp/dsstore/consul_store.go
```diff
@@ -0,0 +1,207 @@
...
+
+	if err != nil {
+		return fields.DaemonSet{}, metadata, err
+	}
+
+	kvp, metadata, err := s.kv.Get(dsPath, nil)
+	if err != nil {
+		return fields.DaemonSet{}, metadata, consulutil.NewKVError("get", dsPath, err)
+	}
+	if metadata == nil {
```
mpuncel: is this a thing that can happen?

pkg/kp/dsstore/consul_store.go
```diff
@@ -0,0 +1,207 @@
...
+	if kvp == nil {
+		// ID didn't exist
+		return fields.DaemonSet{}, metadata, NoDaemonSet
+	}
+
+	ds, err := kvpToDS(kvp)
+	if err != nil {
+		return fields.DaemonSet{}, metadata, err
+	}
+	return ds, metadata, err
```
mpuncel: you don't need the `err` check above the way this is written, or you can return `nil` for the error here

pkg/kp/dsstore/store.go
```diff
@@ -0,0 +1,38 @@
...
+
+	"k8s.io/kubernetes/pkg/labels"
+
+	"github.com/hashicorp/consul/api"
+	"github.com/square/p2/pkg/ds/fields"
+	"github.com/square/p2/pkg/pods"
+	"github.com/square/p2/pkg/types"
+)
+
+const dsTree string = "daemon_set"
```
mpuncel: pluralize this

pkg/kp/dsstore/consul_store.go
```diff
@@ -0,0 +1,207 @@
...
+	}
+	return nil
+}
+
+func (s *consulStore) dsPath(dsID fields.ID) (string, error) {
+	if dsID == "" {
+		return "", util.Errorf("Path requested for empty DS id")
+	}
+
+	return path.Join(dsTree, string(dsID)), nil
```
mpuncel: `dsID.String()`. That way if that type ever changes we won't be blindly casting it to string, the compiler will warn us

mpuncel: (looks like you'll have to implement that first tho)

pkg/kp/dsstore/consul_store_test.go
```diff
@@ -0,0 +1,224 @@
...
+	"github.com/square/p2/pkg/pods"
+	"github.com/square/p2/pkg/types"
+
+	klabels "k8s.io/kubernetes/pkg/labels"
+)
+
+func TestCreate(t *testing.T) {
+	store := consulStoreWithFakeKV()
+	createDaemonSet(store, t)
+	createBadDaemonSet(store, t)
```
mpuncel: it seems unlikely another test would want to call this one, maybe inline it.

mpuncel:
looks pretty good! For extra credit, you could provide a `dsstoretest` package that has an in-memory implementation of a `dsstore.Store` that can be used in tests that depend on a daemon-set store. Not a blocker though, I don't want to slow you down.

pkg/kp/dsstore/consul_store.go
KoishiKomeiji: It is possible for this to happen if the api response does not have a status code of 404 or 200.
But if err != nil, metadata will also but nil. So this does not run as long as you're checking the err beforehand.

KoishiKomeiji: I'll keep it for now just in case.

KoishiKomeiji:
Updated.

mpuncel:
Looks good to me, just want to make sure having `ClusterName` on daemon sets is the right choice, the alternative would be to make them be underneath pod clusters

mpuncel:
discussed it, i'm happy with this 👍

KoishiKomeiji:
Thanks for the PR! Will squash changes and merge.

PR #518, opened 2016-06-09
To commit update and delete functions for daemon sets

KoishiKomeiji:


pkg/kp/dsstore/consul_store.go
```diff
@@ -119,6 +120,24 @@ func (s *consulStore) innerCreate(
 	return ds, nil
 }
 
+func (s *consulStore) Delete(id fields.ID) error {
+	dsPath, err := s.dsPath(id)
+	if err != nil {
+		return err
+	}
+
+	_, _, err = s.Get(id)
```
mpuncel: this probably isn't necessary, it's okay if we try to delete something that doesn't exist and i'm not sure how that would happen

pkg/kp/dsstore/consul_store_test.go
```diff
@@ -80,6 +82,7 @@ func createDaemonSet(store *consulStore, t *testing.T) {
 	Assert(t).AreEqual(ds.PodID, podID, "Daemon set pod id was not set correctly")
 	Assert(t).AreEqual(ds.MinHealth, minHealth, "Daemon set minimum health was not set correctly")
 	Assert(t).AreEqual(ds.Name, clusterName, "Daemon set cluster name was not set correctly")
+	Assert(t).AreEqual(ds.Disabled, false, "Daemon set disabled field was not set correctly")
```
mpuncel: you could use `IsFalse()`

pkg/kp/dsstore/consul_store_test.go
```diff
@@ -90,13 +93,36 @@ func createDaemonSet(store *consulStore, t *testing.T) {
...
+
+	return ds
+}
+
+func TestDelete(t *testing.T) {
+	store := consulStoreWithFakeKV()
+	ds := createDaemonSet(store, t)
+
+	if err := store.Delete("bad_id"); err == nil {
+		t.Errorf("Expected to encounter an error while deleting a daemon set which does not exist")
```
mpuncel: you could use `Error` for these

KoishiKomeiji:
Updated.

mpuncel:
🚢

KoishiKomeiji:
Thanks!

pkg/kp/dsstore/consul_store_test.go
```diff
@@ -139,23 +165,30 @@ func TestGet(t *testing.T) {
...
+		t.Fatal("Unable to retrieve SHA from manifest")
 	}
 	getSHA, err := getDS.Manifest.SHA()
 	if err != nil {
-		t.Errorf("Unable to retrieve SHA from manifest retrieved from daemon set")
+		t.Fatal("Unable to retrieve SHA from manifest retrieved from daemon set")
 	}
 	Assert(t).AreEqual(originalSHA, getSHA, "Daemon set shas were not equal")
+
+	// Invalid get opertaion
```
petertseng: operation

PR #519, opened 2016-06-13
Add ability to fetch from Google Storage

drcapulet:
Wasn't sure of a great way to test this, could potentially cache the result of either of the calls to `google.DefaultClient` or `storage.New(client)` and inject that as part of testing?


anthonybishopric:
BasicFetcher should probably have a google `Client` and that should be provided by the test. The default fetcher can just use `google.DefaultClient`.

mpuncel:
@drcapulet you can extract the critical section to a function and do a `defer f.mu.Unlock()` immediately after locking it

anthonybishopric:
@drcapulet @mpuncel can we just do this in `init()` for that file, just for the default instance? We don't need to introduce all the mutex logic that way.

mpuncel:
yeah that works

mpuncel:
alternatively we could use `sync.Once`

anthonybishopric:
If it's not necessary to introduce concurrency primitives I'd like to avoid it altogether.

drcapulet:
Updated & added a test. However, the call to `google.DefaultClient` will fail for people w/o the Google config setup (somehow it passes on Travis and was passing locally since I'm correctly setup).

I'm thinking about moving this into another file so that you can enable this feature at compile time via a build tag, thoughts?

anthonybishopric:
I'd like to leave this as is, for now.

anthonybishopric:
🚢

drcapulet:
Ping, everything look good? I don't have merge access.

drcapulet:
Updated since I noticed `artifact_verifier.go` also had a list of protocols.

drcapulet:
Ping @anthonybishopric, this look good?

rudle:
Is PR still desirable? Please close if not.

rudle:
Sorry, but this PR must be closed because it is stale.

drcapulet:
Desirable but probably better implemented as an artifact repository.

PR #521, opened 2016-06-15
To commit a Watch for daemon sets to watch create, updates, and deletes

KoishiKomeiji:
This is a watch to watch all create, updates, and deletes. If a watch is needed for a single daemon set, then it will be implemented as needed.

### DIFF
--- a/pkg/kp/consulutil/watch.go
+++ b/pkg/kp/consulutil/watch.go
@@ -1,6 +1,7 @@
 package consulutil
 
 import (
+	"bytes"
 	"time"
 
 	"github.com/hashicorp/consul/api"
@@ -205,6 +206,98 @@ func WatchNewKeys(pairsChan <-chan api.KVPairs, onNewKey NewKeyHandler, done <-c
 				}
 				delete(keys, key)
 			}
+		}
+	}
+}
+
+type WatchedChanges struct {
+	Created api.KVPairs
+	Updated api.KVPairs
+	Deleted api.KVPairs
+}
+
+// WatchDiff watches a Consul prefix for changes and categorizes them
+// into create, update, and delete
+func WatchDiff(
+	prefix string,
+	clientKV ConsulLister,
+	outChanges chan<- *WatchedChanges,
+	done <-chan struct{},
+	outErrors chan<- error,
+) {
+	defer close(outChanges)
+
+	// Keep track of what we have seen so that we know when something was changed
+	keys := make(map[string]*api.KVPair)
+
+	var currentIndex uint64
+	timer := time.NewTimer(time.Duration(0))
+
+	for {
+		select {
+		case <-done:
+			return
+		case <-timer.C:
+		}
+		timer.Reset(250 * time.Millisecond) // upper bound on request rate
+
+		pairs, queryMeta, err := SafeList(clientKV, done, prefix, &api.QueryOptions{
+			WaitIndex: currentIndex,
+		})
+
+		// A copy used to keep track of what was deleted
+		mapCopy := make(map[string]*api.KVPair)
+		for key, val := range keys {
+			mapCopy[key] = val
+		}
+
+		outgoingChanges := &WatchedChanges{}
+		for _, val := range pairs {
+			if _, ok := keys[val.Key]; !ok {
+				// If it is not in the map, then it was a create
+				outgoingChanges.Created = append(outgoingChanges.Created, val)
+				keys[val.Key] = val
+
+			} else if !bytes.Equal(keys[val.Key].Value, val.Value) {
+				// If is in the map and the values are the not same, then it was an update
+				// TODO: Should use something else other than comparing values
+				outgoingChanges.Updated = append(outgoingChanges.Updated, val)
+				if _, ok := mapCopy[val.Key]; ok {
+					delete(mapCopy, val.Key)
+				}
+				keys[val.Key] = val
+
+			} else {
+				// Otherwise it is in the map and the values are equal, so it was not an update
+				if _, ok := mapCopy[val.Key]; ok {
+					delete(mapCopy, val.Key)
+				}
+			}
+		}
+
+		// If it was not observed, then it was a delete
+		for key, val := range mapCopy {
+			outgoingChanges.Deleted = append(outgoingChanges.Deleted, val)
+			if _, ok := keys[key]; ok {
+				delete(keys, key)
+			}
+		}
+
+		switch err {
+		case CanceledError:
+			return
+		case nil:
+			currentIndex = queryMeta.LastIndex
+			select {
+			case <-done:
+			case outChanges <- outgoingChanges:
+			}
+		default:
+			select {
+			case <-done:
+			case outErrors <- err:
+			}
+			timer.Reset(2 * time.Second) // backoff
 		}
 	}
 }
--- a/pkg/kp/consulutil/watch_test.go
+++ b/pkg/kp/consulutil/watch_test.go
@@ -5,6 +5,7 @@ import (
 	"sync"
 	"testing"
 
+	. "github.com/anthonybishopric/gotcha"
 	"github.com/hashicorp/consul/api"
 
 	"github.com/square/p2/pkg/consultest"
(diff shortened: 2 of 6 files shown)


pkg/kp/consulutil/watch.go
```diff
@@ -208,3 +209,95 @@ func WatchNewKeys(pairsChan <-chan api.KVPairs, onNewKey NewKeyHandler, done <-c
...
+	Updated api.KVPairs
+	Deleted api.KVPairs
+}
+
+// WatchDiff watches a Consul prefix for changes and categorizes them
+// into create, update, and delete
+func WatchDiff(
+	prefix string,
+	clientKV ConsulLister,
+	outChanges chan<- *WatchedChanges,
```
mpuncel: I know you're imitating `WatchPrefix`, but I'd actually prefer that this function initializes `outChanges` and returns it, doing the work inside of a goroutine. That's more conventional in Go, and we were just bad at following conventions back in the day :)

pkg/kp/consulutil/watch.go
```diff
@@ -208,3 +209,95 @@ func WatchNewKeys(pairsChan <-chan api.KVPairs, onNewKey NewKeyHandler, done <-c
...
+		select {
+		case <-done:
+			return
+		case <-timer.C:
+		}
+		timer.Reset(250 * time.Millisecond) // upper bound on request rate
+
+		pairs, queryMeta, err := SafeList(clientKV, done, prefix, &api.QueryOptions{
+			WaitIndex: currentIndex,
+		})
```
mpuncel: check `err` here

pkg/kp/consulutil/watch.go
```diff
@@ -208,3 +209,95 @@ func WatchNewKeys(pairsChan <-chan api.KVPairs, onNewKey NewKeyHandler, done <-c
...
+
+		// If it was not observed, then it was a delete
+		for key, val := range mapCopy {
+			outgoingChanges.Deleted = append(outgoingChanges.Deleted, val)
+			if _, ok := keys[key]; ok {
+				delete(keys, key)
+			}
+		}
+
+		switch err {
```
mpuncel: ah here's the `err` check, I think it should be immediately after the call, and you do a select on `<-quit` and `outChanges <- outgoingChange` here

pkg/kp/consulutil/watch_test.go
```diff
@@ -448,3 +449,82 @@ func TestWatchNewKeysExistingData(t *testing.T) {
...
+	pairs := kvToMap(changes.Created)
+	if !kvMatch(pairs, kv1a) {
+		t.Error("existing data not recognized")
+	}
+	Assert(t).AreEqual(len(changes.Created), 1, "Unexpected number of creates watched")
+	Assert(t).AreEqual(len(changes.Updated), 0, "Unexpected number of updates watched")
+	Assert(t).AreEqual(len(changes.Deleted), 0, "Unexpected number of deletes watched")
+
+	// Get an updates when the data changes (create, modify, delete)
+	f.Client.KV().Put(kv1b, nil)
```
mpuncel: is there an `err` to check here?

pkg/kp/consulutil/watch_test.go
```diff
@@ -448,3 +449,82 @@ func TestWatchNewKeysExistingData(t *testing.T) {
...
+	kv3a := &api.KVPair{Key: "something", Value: []byte("different")}
+
+	// Process existing data
+	var changes *WatchedChanges
+	watchedCh := make(chan *WatchedChanges)
+
+	f.Client.KV().Put(kv1a, nil)
+	go WatchDiff("prefix/", f.Client.KV(), watchedCh, done, testLogger(t))
+
+	changes = <-watchedCh
```
mpuncel: probably select on this with a `<-time.After(1*time.Second)` so that if a value never arrives, the test doesn't hang

pkg/kp/dsstore/consul_store.go
```diff
@@ -214,6 +214,84 @@ func (s *consulStore) MutateDS(
...
+		var kvps *consulutil.WatchedChanges
+		for {
+			// Populate kvps by reading inCh and checking for errors
+			select {
+			case <-quitCh:
+				return
+			case err := <-errCh:
+				s.logger.WithError(err).Errorf("WatchDiff returned error, recovered.")
+			case kvps = <-inCh:
+				if kvps == nil {
```
mpuncel: is this a thing that can happen? i guess it doesn't hurt to check for it but maybe log it since it would be very odd

pkg/kp/dsstore/consul_store_test.go
```diff
@@ -376,6 +377,120 @@ func TestMutate(t *testing.T) {
...
+			Assert(t).AreEqual(watchedDS.PodID, someOtherDS.PodID, "Daemon sets should have equal pod ids")
+		} else {
+			t.Errorf("Expected to find id '%s' among watch results, but was not present", watchedDS.ID)
+		}
+	}
+
+	//
+	// Make sure Watch does not output any daemon sets something gets deleted
+	//
+	store.Delete(someOtherDS.ID)
```
mpuncel: err to check?

pkg/kp/dsstore/dsstoretest/fake_dsstore.go
```diff
@@ -10,9 +10,22 @@ import (
 	klabels "k8s.io/kubernetes/pkg/labels"
 )
 
+const (
+	created = "created"
```
mpuncel: mind making these a `type WatchOperation string` type?

mpuncel: oh this is a test file never mind

pkg/kp/dsstore/dsstoretest/fake_dsstore.go
```diff
@@ -78,13 +109,67 @@ func (s *FakeDSStore) MutateDS(
...
+		for {
+			select {
+			case <-quitCh:
+				return
+			default:
+			}
+
+			outgoingChanges := dsstore.WatchedDaemonSets{}
+			// Reads in new changes that are sent to FakeDSStore.watchers
+			for _, ch := range s.watchers {
```
mpuncel: where is anything ever added to `s.watchers`

mpuncel:
looks pretty good! just a few questions

pkg/kp/dsstore/dsstoretest/fake_dsstore.go
KoishiKomeiji: No, nothing is added to s.watchers in this function but things do get added in the create, update, and delete.

KoishiKomeiji:
Updated, what questions do you have?

pkg/kp/consulutil/watch.go
```diff
@@ -208,3 +209,102 @@ func WatchNewKeys(pairsChan <-chan api.KVPairs, onNewKey NewKeyHandler, done <-c
...
+// into create, update, and delete, please note that if a kvPair was
+// create and modified before this starts watching, this watch will
+// treat it as a create
+func WatchDiff(
+	prefix string,
+	clientKV ConsulLister,
+	quitCh <-chan struct{},
+	outErrors chan<- error,
+) <-chan *WatchedChanges {
+	outCh := make(chan *WatchedChanges)
```
petertseng: should this function close `outCh`?

KoishiKomeiji: Since this function returns outCh, outCh should automatically be closed when the caller is done using the channel, or the caller can also explicitly close it.

petertseng: > outCh should automatically be closed when the caller is done using the channel

I didn't realize this happens! What makes it happen?

rudle: We've been trying to follow the pattern of: those who open channels are responsible for ensuring that they are closed. This might mean that this function should instead take a (chan *WatchedChanges) as an argument vs returning one to the caller.

KoishiKomeiji: Actually, you're right it doesn't automatically close. I'll add line on 232 saying
`defer close(outCh)` in a future commit on daemon sets.

mpuncel: @rudle That's how it was before, and I told him to change it to a return argument :)

KoishiKomeiji: Alright I'll fix it in a commit soon.

PR #522, opened 2016-06-16
Hzeng.dsctl

KoishiKomeiji:
CLI commands for create, get, list, enable, disable, delete, and update.

### DIFF
--- a/bin/p2-dsctl/main.go
+++ b/bin/p2-dsctl/main.go
@@ -1,0 +1,237 @@
+package main
+
+import (
+	"encoding/json"
+	"fmt"
+	"log"
+	"strconv"
+
+	"gopkg.in/alecthomas/kingpin.v2"
+
+	"github.com/Sirupsen/logrus"
+	ds_fields "github.com/square/p2/pkg/ds/fields"
+	"github.com/square/p2/pkg/kp"
+	"github.com/square/p2/pkg/kp/dsstore"
+	"github.com/square/p2/pkg/kp/flags"
+	"github.com/square/p2/pkg/logging"
+	pc_fields "github.com/square/p2/pkg/pc/fields"
+	"github.com/square/p2/pkg/pods"
+	"github.com/square/p2/pkg/types"
+	"github.com/square/p2/pkg/util"
+	"k8s.io/kubernetes/pkg/labels"
+)
+
+const (
+	CmdCreate  = "create"
+	CmdGet     = "get"
+	CmdList    = "list"
+	CmdEnable  = "enable"
+	CmdDisable = "disable"
+	CmdDelete  = "delete"
+	CmdUpdate  = "update"
+)
+
+var (
+	cmdCreate       = kingpin.Command(CmdCreate, "Create a daemon set.")
+	createAZ        = cmdCreate.Flag("az", "The availability zone of the pod cluster").Required().String()
+	createManifest  = cmdCreate.Flag("manifest", "Complete manifest - must parse as JSON!").Required().String()
+	createMinHealth = cmdCreate.Flag("minhealth", "The minimum health of the daemon set").Required().String()
+	createName      = cmdCreate.Flag("name", "The cluster name (ie. staging, production)").Required().String()
+	createPodID     = cmdCreate.Flag("pod", "The pod ID on the daemon set").Required().String()
+
+	cmdGet = kingpin.Command(CmdGet, "Show a daemon set.")
+	getID  = cmdGet.Arg("id", "The uuid for the daemon set").Required().String()
+
+	cmdList = kingpin.Command(CmdList, "List daemon sets.")
+
+	cmdEnable = kingpin.Command(CmdEnable, "Enable daemon set.")
+	enableID  = cmdEnable.Arg("id", "The uuid for the daemon set").Required().String()
+
+	cmdDisable = kingpin.Command(CmdDisable, "Disable daemon set.")
+	disableID  = cmdDisable.Arg("id", "The uuid for the daemon set").Required().String()
+
+	cmdDelete = kingpin.Command(CmdDelete, "Delete daemon set.")
+	deleteID  = cmdDelete.Arg("id", "The uuid for the daemon set").Required().String()
+
+	cmdUpdate       = kingpin.Command(CmdUpdate, "Update a daemon set.")
+	updateID        = cmdUpdate.Arg("id", "The uuid for the daemon set").Required().String()
+	updateAZ        = cmdUpdate.Flag("az", "The availability zone of the pod cluster").String()
+	updateManifest  = cmdUpdate.Flag("manifest", "Complete manifest - must parse as JSON!").String()
+	updateMinHealth = cmdUpdate.Flag("minhealth", "The minimum health of the daemon set").String()
+	updateName      = cmdUpdate.Flag("name", "The cluster name (ie. staging, production)").String()
+	updatePodID     = cmdUpdate.Flag("pod", "The pod ID on the daemon set").String()
+)
+
+func main() {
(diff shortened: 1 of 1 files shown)


bin/p2-dsctl/main.go
```diff
@@ -0,0 +1,237 @@
...
+
+	cmdDisable = kingpin.Command(CmdDisable, "Disable daemon set.")
+	disableID  = cmdDisable.Flag("id", "The uuid for the daemon set").Required().String()
+
+	cmdDelete = kingpin.Command(CmdDelete, "Delete daemon set.")
+	deleteID  = cmdDelete.Flag("id", "The uuid for the daemon set").Required().String()
+
+	cmdUpdate       = kingpin.Command(CmdUpdate, "Update a daemon set.")
+	updateAZ        = cmdUpdate.Flag("az", "The availability zone of the pod cluster").String()
+	updateID        = cmdUpdate.Flag("id", "The uuid for the daemon set").Required().String()
```
mpuncel: i feel that this shouldn't be a flag but an arg, so I can do `dsctl update abcd-3813 --az uswest` for example

bin/p2-dsctl/main.go
```diff
@@ -0,0 +1,237 @@
...
+var (
+	cmdCreate       = kingpin.Command(CmdCreate, "Create a daemon set.")
+	createAZ        = cmdCreate.Flag("az", "The availability zone of the pod cluster").Required().String()
+	createManifest  = cmdCreate.Flag("manifest", "Complete manifest - must parse as JSON!").Required().String()
+	createMinHealth = cmdCreate.Flag("minhealth", "The minimum health of the daemon set").Required().String()
+	createName      = cmdCreate.Flag("name", "The cluster name (ie. staging, production)").Required().String()
+	createPodID     = cmdCreate.Flag("pod", "The pod ID on the daemon set").Required().String()
+
+	cmdGet = kingpin.Command(CmdGet, "Show a daemon set.")
+	getID  = cmdGet.Flag("id", "The uuid for the daemon set").Required().String()
```
mpuncel: same here, make this an arg

bin/p2-dsctl/main.go
```diff
@@ -0,0 +1,237 @@
...
+
+	cmdList = kingpin.Command(CmdList, "List daemon sets.")
+
+	cmdEnable = kingpin.Command(CmdEnable, "Enable daemon set.")
+	enableID  = cmdEnable.Flag("id", "The uuid for the daemon set").Required().String()
```
mpuncel: and all these other "id" flags as well

bin/p2-dsctl/main.go
```diff
@@ -0,0 +1,237 @@
...
+	client := kp.NewConsulClient(consulOpts)
+	logger := logging.NewLogger(logrus.Fields{})
+	dsstore := dsstore.NewConsul(client, 3, &logger)
+
+	switch cmd {
+	case CmdCreate:
+		az := pc_fields.AvailabilityZone(*createAZ)
+		minHealth, err := strconv.Atoi(*createMinHealth)
+		if err != nil {
+			log.Fatalf("Invalid value for minimum health, expected integer")
```
mpuncel: put the `err` somewhere in here

bin/p2-dsctl/main.go
```diff
@@ -0,0 +1,237 @@
...
+		}
+		fmt.Printf("The daemon set '%s' has been successfully updated in consul", id.String())
+		fmt.Println()
+
+	default:
+		log.Fatalf("Unrecognized command %v", cmd)
+	}
+}
+
+func selectorFrom(az pc_fields.AvailabilityZone) labels.Selector {
```
mpuncel: side note: it's interesting to me that availability zone is expected to be in the node selector but is not present on the daemon set itself, the former suggests that you'd want a daemon set to span more than a single availability zone, so maybe you want to support a node selector using `labels.InOperator` to support selecting nodes from multiple availability zones. Not a blocker for this PR though, we'll watch what use cases users have with daemon sets.

mpuncel:
👍

KoishiKomeiji:
Updated, also using labels.InOperator instead of labels.EqualsOperator.
Will squash and merge when tests pass.

bin/p2-dsctl/main.go
```diff
@@ -0,0 +1,237 @@
...
+func main() {
+	cmd, consulOpts := flags.ParseWithConsulOptions()
+	client := kp.NewConsulClient(consulOpts)
+	logger := logging.NewLogger(logrus.Fields{})
+	dsstore := dsstore.NewConsul(client, 3, &logger)
+
+	switch cmd {
+	case CmdCreate:
+		az := pc_fields.AvailabilityZone(*createAZ)
+		minHealth, err := strconv.Atoi(*createMinHealth)
```
petertseng: Is it possible to use `Integer()` instead of `String()` here to automatically get the integer parse? Any reason why we must use `String()`?

petertseng: `Int` not `Integer`, will edit my earlier comment.

bin/p2-dsctl/main.go
```diff
@@ -0,0 +1,237 @@
...
+		}
+		fmt.Printf("The daemon set '%s' has been successfully enabled in consul", id.String())
+		fmt.Println()
+
+	case CmdDisable:
+		id := ds_fields.ID(*disableID)
+
+		mutator := func(ds ds_fields.DaemonSet) (ds_fields.DaemonSet, error) {
+			if ds.Disabled {
+				return ds, util.Errorf("Daemon set has already been enabled")
```
petertseng: already been disabled

bin/p2-dsctl/main.go
```diff
@@ -0,0 +1,237 @@
...
+			ds.Disabled = false
+			return ds, nil
+		}
+
+		_, err := dsstore.MutateDS(id, mutator)
+		if err != nil {
+			log.Fatalf("err: %v", err)
+		}
+		fmt.Printf("The daemon set '%s' has been successfully enabled in consul", id.String())
+		fmt.Println()
```
petertseng: do we like this explicit `Println` better than appending a `\n` to previous message?

I don't care, but curious

bin/p2-dsctl/main.go
```diff
@@ -0,0 +1,237 @@
...
+
+			if *updateAZ != "" {
+				selector := selectorFrom(pc_fields.AvailabilityZone(*updateAZ))
+				if ds.NodeSelector.String() != selector.String() {
+					changed = true
+					ds.NodeSelector = selectorFrom(pc_fields.AvailabilityZone(*updateAZ))
+				}
+			}
+			if *updateMinHealth != "" {
+				minHealth, err := strconv.Atoi(*updateMinHealth)
```
petertseng: another possible use of `Int()` rather than `String()`?

bin/p2-dsctl/main.go
KoishiKomeiji: I did this because `updateMinHealth` uses `String()` so it is consistent.
`updateMinHealth` uses `String()` because I want to check `*updateMinHealth==""` so that I would know if the user did not specify the optional argument.

petertseng: aha! OK, seems like a good reason.

bin/p2-dsctl/main.go
petertseng: cancel, explained above.

bin/p2-dsctl/main.go
KoishiKomeiji: Thanks!

petertseng:
Yup 👍

KoishiKomeiji:
Thanks! Squashing now.

PR #523, opened 2016-06-20
fix p2-pcctl session creation

mpuncel:
Previously, p2-pcctl was using kv.NewUnmanagedSession which
(confusingly) takes a string argument for a session ID, however it
expects the session ID to correspond with a valid session in consul.

This commit uses kv.NewSession() instead which actually creates a
session in consul.


rudle:
:ship:

PR #525, opened 2016-06-21
Make launchable.Location a url.URL type instead of a string

mpuncel:
This paves the way for an upcoming change which will allow an
alternative method of specifying artifact download.

### DIFF
--- a/bin/p2-bin2pod/main.go
+++ b/bin/p2-bin2pod/main.go
@@ -39,7 +39,7 @@ var (
     cat example.json | jq '.["tar_path"] | xargs -I {} cp {} ~/test_file_server_dir
     cat example.json | jq '.["manifest_path"] | xargs -I {} curl -X PUT https://consul.dev:8500/api/v1/kv/intent/$(hostname)/example -d {}
     `)
-	executable    = bin2pod.Arg("executable", "the executable to turn into a hoist artifact + pod manifest. The format of executable is of a URL.").Required().String()
+	executable    = bin2pod.Arg("executable", "the executable to turn into a hoist artifact + pod manifest. The format of executable is of a URL.").Required().URL()
 	id            = bin2pod.Flag("id", "The ID of the pod. By default this is the name of the executable passed").String()
 	location      = bin2pod.Flag("location", "The location where the outputted tar will live. The characters {} will be replaced with the unique basename of the tar, including its SHA. If not provided, the location will be a file path to the resulting tar from the build, which is included in the output of this script. Users must copy the resultant tar to the new location if it is different from the default output path.").String()
 	workDirectory = bin2pod.Flag("work-dir", "A directory where the results will be written.").ExistingDir()
@@ -56,7 +56,7 @@ func podId() types.PodID {
 	if *id != "" {
 		return types.PodID(*id)
 	}
-	return types.PodID(path.Base(*executable))
+	return types.PodID(path.Base((*executable).Path))
 }
 
 func activeDir() string {
@@ -130,25 +130,30 @@ func makeTar(workingDir string) (string, error) {
 	if err != nil {
 		return "", fmt.Errorf("Couldn't make a new working directory %s for tarring: %s", tarContents, err)
 	}
+
 	err = os.MkdirAll(path.Join(tarContents, "bin"), 0744)
 	if err != nil {
 		return "", fmt.Errorf("Couldn't make bin directory in %s: %s", tarContents, err)
 	}
 	launchablePath := path.Join(tarContents, "bin", "launch")
+
 	err = uri.URICopy(*executable, launchablePath)
 	if err != nil {
 		return "", fmt.Errorf("Couldn't copy from %s.: %s", *executable, err)
 	}
+
 	err = os.Chmod(launchablePath, 0755) // make file executable by all.
 	if err != nil {
 		return "", fmt.Errorf("Couldn't make %s executable: %s", launchablePath, err)
 	}
-	tarPath := path.Join(workingDir, fmt.Sprintf("%s_%s.tar.gz", path.Base(*executable), randomSuffix()))
+
+	tarPath := path.Join(workingDir, fmt.Sprintf("%s_%s.tar.gz", path.Base((**executable).Path), randomSuffix()))
 	cmd := exec.Command("tar", "-czvf", tarPath, "-C", tarContents, ".")
 	err = cmd.Run()
 	if err != nil {
 		return "", fmt.Errorf("Couldn't build tar: %s", err)
 	}
+
 	return tarPath, nil
 }
 
--- a/bin/p2-verify-artifact/main.go
+++ b/bin/p2-verify-artifact/main.go
@@ -17,8 +17,8 @@ import (
 )
 
 var (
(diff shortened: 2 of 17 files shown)


mpuncel:
This is so we can specify tags as URI params programatically

rudle:
I don't see how this relates to a new location key. Do you expect that launchable will get a new field representing a version struct in addition to this URL key?

Did you think about taking the bigger bite and replacing launchable.Location with a new type (ie. something more abstract than a URL). I'm suggesting this based on what we discussed earlier… this type could be passed around inside P2 and turned into a URL at the edge.

mpuncel:
@rudle I was thinking that tags could be part of the url query, do you think that isn't flexible enough for what we're planning?

anthonybishopric:
👍 but I'll leave it up to you to decide if you want to merge this or not, given that it's not really related to the work we ultimately need to do

bin/p2-bin2pod/main.go
```diff
@@ -130,25 +131,35 @@ func makeTar(workingDir string) (string, error) {
...
 	}
+
 	err = os.MkdirAll(path.Join(tarContents, "bin"), 0744)
 	if err != nil {
 		return "", fmt.Errorf("Couldn't make bin directory in %s: %s", tarContents, err)
 	}
 	launchablePath := path.Join(tarContents, "bin", "launch")
-	err = uri.URICopy(*executable, launchablePath)
+
+	url, err := url.Parse(*executable)
```
bwester: kingpin provides a [URL()](https://godoc.org/gopkg.in/alecthomas/kingpin.v2#ArgClause.URL) argument type that will parse the URL for you.

bin/p2-verify-artifact/main.go
```diff
@@ -35,24 +36,34 @@ func main() {
 	localCopyPath := filepath.Join(dir, "temp.tar.gz")
 
 	var localCopy *os.File
-	var locationForSignature string
+	var locationForSignature url.URL
```
bwester: Based on what I've seen in the standard library, it's idiomatic to pass around `*url.URL` values instead of `url.URL` structs directly.

pkg/auth/artifact_verifier_test.go
```diff
@@ -45,7 +46,13 @@ func testVerifiedWithFiles(t *testing.T, files []testFile, verifier ArtifactVeri
 	if err != nil {
 		t.Fatal(err)
 	}
-	err = verifier.VerifyHoistArtifact(localCopy, fmt.Sprintf("file://%v", filePath))
+
+	url, err := url.Parse(fmt.Sprintf("file://%v", filePath))
```
bwester: Instead of constructing a string and parsing it, can you just create a `*url.URL` object yourself?

pkg/hoist/hoist_launchable.go
```diff
@@ -322,10 +322,16 @@ func (hl *Launchable) Install(verifier auth.ArtifactVerifier) error {
 	return err
 }
 
-// The version of the artifact is currently derived from the location, using
-// the naming scheme <the-app>_<unique-version-string>.tar.gz
+func (hl *Launchable) tempFilePath() string {
```
bwester: Why do this?  I don't readily see the impact of this change.

pkg/auth/artifact_verifier_test.go
mpuncel: I did it this way to ensure that the behavior of the test doesn't change, but given that they pass now it should be okay to do what you're suggesting

pkg/hoist/hoist_launchable.go
mpuncel: the `path.Base(hl.location)` call doesn't make sense anymore now that `Location` is a url.URL. I could use the path but in the future the artifact URL path may not be suitable for making a unique temp directory path

bwester: That makes sense.

pkg/hoist/hoist_launchable.go
```diff
@@ -322,10 +322,16 @@ func (hl *Launchable) Install(verifier auth.ArtifactVerifier) error {
...
+	return fmt.Sprintf("%s_%s.tar.gz", hl.Id, hl.Version())
```
bwester: `Version()` is everything in the file name except the extension, so this is going to produce temp file paths like `foo_foo_1234`. Is that fine?

bin/p2-bin2pod/main.go
```diff
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"log"
 	"math/rand"
+	// "net/url"
```
bwester: This file needs some cleanup.

mpuncel: removing

pkg/hoist/hoist_launchable.go
mpuncel: removing the `hl.Id` part

mpuncel: actually that just puts me back to the same behavior as before, i'll put it back for now and we can change the temp file path as part of a future PR when url end stops being acceptable.

pkg/auth/artifact_verifier_test.go
```diff
@@ -45,8 +47,16 @@ func testVerifiedWithFiles(t *testing.T, files []testFile, verifier ArtifactVeri
...
-	err = verifier.VerifyHoistArtifact(localCopy, fmt.Sprintf("file://%v", filePath))
+
+	url := &url.URL{
+		Scheme: "file",
+		Path:   filePath,
+	}
+
+	err = verifier.VerifyHoistArtifact(localCopy, url)
 	if err != nil {
+		fmt.Println(err)
```
bwester: Clean up?

pkg/digest/verification.go
```diff
@@ -113,22 +114,20 @@ func VerifyDir(root string, digest map[string]string) error {
...
 	}
 	defer digest.Close()
 	var signature io.ReadCloser
-	if signatureUri != "" {
-		signature, err = fetcher.Open(signatureUri)
-		if err != nil {
-			return Digest{}, err
-		}
-		defer signature.Close()
+	signature, err = fetcher.Open(signatureUri)
```
bwester: Did you intent to make signatures mandatory?  If so, I'd like to see this in its own commit instead of hitching a ride with a commit that seems like a pure refactoring of types.

mpuncel: i did not intend to do that, fixed

bwester:
🚢 pending Travis CI completion

bin/p2-bin2pod/main.go
```diff
@@ -56,7 +56,7 @@ func podId() types.PodID {
 	if *id != "" {
 		return types.PodID(*id)
 	}
-	return types.PodID(path.Base(*executable))
+	return types.PodID(path.Base((*executable).Path))
```
petertseng: I hate to nit but can this be just `executable.Path` since pointers inherit the method sets of their pointees?

Not that I would refuse to merge on the basis of this, but it's a question that makes me curious

mpuncel: it's actually a `**url.URL` here, so you have to dereference it once

PR #526, opened 2016-06-22
Add mutually exclusive ID command line flag

rudle:
This better matches the use cases we've encountered in reality.

Some duplicate code here, but I'm ok with it because it seems very unlikely that we'll change it in the near future. I wouldn't be surprised if we delete the label parameters for Get, Update, Delete

### DIFF
--- a/bin/p2-pcctl/main.go
+++ b/bin/p2-pcctl/main.go
@@ -35,20 +35,23 @@ var (
 	createAnnotations = cmdCreate.Flag("annotations", "Complete set of annotations - must parse as JSON!").String()
 
 	cmdGet   = kingpin.Command(cmdGetText, "Show a pod cluster. ")
-	getPodID = cmdGet.Flag("pod", "The pod ID on the pod cluster").Required().String()
-	getAZ    = cmdGet.Flag("az", "The availability zone of the pod cluster").Required().String()
-	getName  = cmdGet.Flag("name", "The cluster name (ie. staging, production)").Required().String()
+	getPodID = cmdGet.Flag("pod", "The pod ID on the pod cluster").String()
+	getAZ    = cmdGet.Flag("az", "The availability zone of the pod cluster").String()
+	getName  = cmdGet.Flag("name", "The cluster name (ie. staging, production)").String()
+	getID    = cmdGet.Flag("id", "The cluster UUID. This option is mutually exclusive with pod,az,name").String()
 
 	cmdDelete   = kingpin.Command(cmdDeleteText, "Delete a pod cluster. ")
-	deletePodID = cmdDelete.Flag("pod", "The pod ID on the pod cluster").Required().String()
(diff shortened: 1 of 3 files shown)


bin/p2-pcctl/main.go
```diff
@@ -36,20 +36,23 @@ var (
...
 	cmdUpdate         = kingpin.Command(CmdUpdate, "Update a pod cluster. ")
-	updatePodID       = cmdUpdate.Flag("pod", "The pod ID on the pod cluster").Required().String()
-	updateAZ          = cmdUpdate.Flag("az", "The availability zone of the pod cluster").Required().String()
-	updateName        = cmdUpdate.Flag("name", "The cluster name (ie. staging, production)").Required().String()
-	updateAnnotations = cmdUpdate.Flag("annotations", "JSON string representing the complete update ").Required().String()
+	updatePodID       = cmdUpdate.Flag("pod", "The pod ID on the pod cluster").String()
+	updateAZ          = cmdUpdate.Flag("az", "The availability zone of the pod cluster").String()
+	updateName        = cmdUpdate.Flag("name", "The cluster name (ie. staging, production)").String()
+	updateAnnotations = cmdUpdate.Flag("annotations", "JSON string representing the complete update ").String()
+	updateID          = cmdUpdate.Flag("ID", "The cluster UUID. This option is mutually exclusive with pod,az,name").String()
```
petertseng: why is this ID uppercase

pkg/pc/control/control.go
```diff
@@ -42,7 +44,26 @@ func NewPodCluster(
...
+	return pc
+}
+
 func (pccontrol *PodCluster) All() ([]fields.PodCluster, error) {
+	if pccontrol.ID != "" {
+		pc, err := pccontrol.pcStore.Get(pccontrol.ID)
+		if err != nil {
+			return nil, err
+		}
+		return []fields.PodCluster{pc}, err
```
petertseng: maybe clearer to say `nil` instead of `err` here since if we get here `err` is `nil`

pkg/pc/control/control_test.go
```diff
@@ -137,3 +137,45 @@ func TestUpdate(t *testing.T) {
...
+	}
+
+	errs := pcControllerFromID.Delete()
+	if len(errs) > 0 {
+		t.Fatalf("%v", errs)
+	}
+
+	notFoundPC, err := pcControllerFromID.Get()
+	if err == nil {
+		t.Errorf("Expected not found error, got none")
```
petertseng: since the error message says you are looking for a not-found, would you like to check that it is `pcstore.NoPodCluster` ?

petertseng:
this all seems reasonable, :+1:

anthonybishopric:
🚢

rudle:
Pushed a new commit to kick travis.ci. This build: https://travis-ci.org/square/p2/builds/139807632 refuses to be rebuilt. 

💻 🙅

PR #528, opened 2016-06-23
Fix lint errors in binaries

bwester:
Fixes a bunch of warnings raised by running "gometalinter" on all the
binary packages. Changes mostly eliminate dead code, check error
return values, and adjust variable names/cases.

### DIFF
--- a/bin/p2-bin2pod/main.go
+++ b/bin/p2-bin2pod/main.go
@@ -46,13 +46,13 @@ var (
 	config        = bin2pod.Flag("config", "a list of key=value assignments. Each key will be set in the config section.").Strings()
 )
 
-type Result struct {
+type result struct {
 	TarPath       string `json:"tar_path"`
 	ManifestPath  string `json:"manifest_path"`
 	FinalLocation string `json:"final_location"`
 }
 
-func podId() types.PodID {
+func podID() types.PodID {
 	if *id != "" {
 		return types.PodID(*id)
 	}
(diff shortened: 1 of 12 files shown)


bin/p2-bin2pod/main.go
```diff
@@ -100,7 +100,7 @@ func main() {
 	}
 
 	manifestBuilder.SetLaunchables(map[string]pods.LaunchableStanza{
-		string(podId()): stanza,
+		string(podID()): stanza,
```
mpuncel: while you're at it, can we make this `podID().String()`

bin/p2-rctl/main.go
```diff
@@ -34,57 +34,57 @@ import (
...
-	CMD_SCHEDUP  = "schedule-update"
+	cmdCreateText   = "create"
+	cmdDeleteText   = "delete"
+	cmdReplicasText = "set-replicas"
+	cmdListText     = "list"
+	cmdGetText      = "get"
+	cmdEnableText   = "enable"
+	cmdDisableText  = "disable"
+	cmdRollText     = "rolling-update"
+	cmdSchedupText  = "schedule-update"
```
petertseng: perhaps `SchedUp`, but I'm a bit unsure whether that's a material improvement

mpuncel:
👍

petertseng:
Seems OK, 👍

rudle:
:+1:

bin/p2-bin2pod/main.go
bwester: Sounds good; will do.

bin/p2-rctl/main.go
bwester: Elsewhere, there is already a `cmdSchedup`, and this just provides the text of the command. Mostly I just care about removing CAPS and not having two variables with the same name that vary only in the casing pattern used.

bwester:
@mpuncel: I expect this to conflict with #525. Are you close to committing that one?

bwester:
Updated: use `podID.String()`, rebase onto latest master. Will commit after tests pass.

PR #529, opened 2016-06-23
Fix bug in flaky unit test

bwester:
The TestPublishLatestRCs unit test failed with the following output:

    Unexpected error on errCh: <nil>

The probable immediate cause is that the channel was closed while the
unit test was stopping. Once `quitCh` is closed, the goroutine that
feeds into `errCh` will eventually close that channel too. If the
error-monitoring goroutine exists after that happens, and if the
select statement receives from `errCh` instead of `quitCh`, the test
will unnecessarily fail.


petertseng:
okay, yeah. 👍

mpuncel:
🍧

PR #530, opened 2016-06-23
Update dependencies

bwester:
Updates some of P2's dependencies. The latest tagged version is used, or if the project doesn't use tags, the latest commit. These are just the packages I found to be easy to update.


anthonybishopric:
👍

PR #533, opened 2016-06-27
Move scheduler to its own package

KoishiKomeiji:
Move rc/scheduler.go to its own package

scheduler.go used to be in the rc package, but because it will now
be used with daemon sets, it would make more sense for it to be
in its own package.

This will also fix the issues with the watch for daemon sets.

### DIFF
--- a/pkg/kp/consulutil/watch.go
+++ b/pkg/kp/consulutil/watch.go
@@ -229,6 +229,8 @@ func WatchDiff(
 	outCh := make(chan *WatchedChanges)
 
 	go func() {
+		defer close(outCh)
+
 		// Keep track of what we have seen so that we know when something was changed
 		keys := make(map[string]*api.KVPair)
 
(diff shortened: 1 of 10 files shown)


mpuncel:
👍

pkg/kp/consulutil/watch.go
```diff
@@ -229,6 +229,8 @@ func WatchDiff(
 	outCh := make(chan *WatchedChanges)
 
 	go func() {
+		defer close(outCh)
```
petertseng: should have gone in its own commit

KoishiKomeiji: I think it was in its own commit, but I squashed it, so it's looks that way now.

petertseng: recommend no squash on those if they are separate changes. each commit should be one logical change

KoishiKomeiji: Alright, thanks!

PR #534, opened 2016-06-28
Mpuncel/artifact downloader

mpuncel:


pkg/artifact/downloader.go
```diff
@@ -0,0 +1,82 @@
...
+	if err != nil {
+		return util.Errorf("Could not copy artifact locally: %v", err)
+	}
+	// rewind once so we can ask the verifier
+	_, err = artifactFile.Seek(0, os.SEEK_SET)
+	if err != nil {
+		return util.Errorf("Could not reset artifact file position for verification: %v", err)
+	}
+
+	err = l.verifier.VerifyHoistArtifact(artifactFile, l.location)
```
petertseng: we OK with embedding this in? One could argue that now the downloader is doing more than just downloading. On the other hand, I don't have a good idea of what the alternative looks like (Separate downloader and verifier, then you download then verify?), so I don't yet have a better to suggest

mpuncel: the alternative would be to pass the verifier to `DownloadTo()`, and therefore pass it to `launchable.Install()` as well. I don't think either way is more complex, and i think it does make sense to "verify" everything we download even if it's a `NopVerifier`

pkg/pods/pod.go
```diff
@@ -397,7 +399,24 @@ func (pod *Pod) Install(manifest Manifest, artifactVerifier auth.ArtifactVerifie
...
+			}
+		}
+		locationURL, err := url.Parse(launchableStanza.Location)
+		if err != nil {
+			pod.logLaunchableError(launchable.ServiceID(), err, "Unable to install launchable")
+			return util.Errorf("Couldn't parse launchable location '%s' as a URL: %s", launchableStanza.Location, err)
+		}
+
+		downloader := artifact.NewLocationDownloader(locationURL, uri.DefaultFetcher, verifier)
+		err = launchable.Install(downloader)
```
anthonybishopric: I'd kill `launchable.Install` altogether - I'm not sure it's adding much

pkg/pods/pod.go
```diff
@@ -660,14 +679,24 @@ func (pod *Pod) getLaunchable(launchableStanza LaunchableStanza, runAsUser strin
...
+
 	if launchableStanza.LaunchableType == "hoist" {
 		ret := &hoist.Launchable{
-			Location:         locationURL,
+			Version:          version,
 			Id:               launchableStanza.LaunchableId,
 			ServiceId:        serviceId,
 			RunAs:            runAsUser,
 			PodEnvDir:        pod.EnvDir(),
-			Fetcher:          uri.DefaultFetcher,
```
anthonybishopric: Why remove this? If anything we may add it back if we decide to add MTLS support to the configuration

anthonybishopric:
Looks like a good change. In the follow-up PR, I'd suggest extracting some component that creates one of these:

```
struct LocationData {
    Location net.URL
    SigLocation net.URL
    ManifestLocation net.URL
    ManifestSigLocation net.URL
}
``` 

and change the signature of `auth.ArtifactVerifier` to take this instead. As it currently stands, it interpolates the magic suffixes inline, which will conflict with the abstraction you've added. I'm open to better solutions

pkg/pods/pod.go
mpuncel: the launchable doesn't need to fetch anything anymore so i'm not sure why it needs to be there

pkg/pods/pod.go
mpuncel: it's currently useful for the opencontainer launchable, so i'm leaving it

mpuncel:
will merge on green

PR #535, opened 2016-06-28
Optimizations to outstanding connection counts in pkg/replication.

anthonybishopric:
`pkg/replication` is primarily used by human operators attempting to
deploy many pods simultaneously (in the absence of daemon sets, which
are being actively worked on). For this reason, we want to be considerate
of the number of open file descriptors used by the replication package
at any given time.

This PR switches reality observations into periodic polls (20 + rand(10) seconds)
and changes health to use a single service-wide aggregation. Previously on my
Macbook I was unable to execute any deploy with a parallelism factor over 100. With
this change I was able to perform a deployment at a 200 node parallelism factor without
issue.


pkg/replication/replication.go
```diff
@@ -273,66 +275,48 @@ func (r replication) updateOne(node types.NodeName, done chan<- types.NodeName,
...
 		case <-quitCh:
 			return
-		case err := <-healthErr:
-			nodeLogger.WithError(err).Errorln("Could not read health check")
-			select {
-			case r.errCh <- err:
-			case <-quitCh:
+		case <-time.After(1 * time.Second):
+			res, err := aggregateHealth.GetHealth(node)
+			if err != nil {
```
mpuncel: can the health value be empty if the node hasn't been deployed yet?

mpuncel: i mean it's been deployed but maybe no health value yet

mpuncel:
👍 looks good, it was kind of hard to wrap my head around the `sync.Cond` for a while even thought he outcome is simple, I wonder if there's a simpler way to write it given that it's only needed as an intial condition. Using a channel would be even harder though so and I can't think of another solution

PR #536, opened 2016-06-29
Return a TLS config that trusts nothing instead of nil to prevent NPE

drcapulet:
Prevents an NPE on `tlsConfig.InsecureSkipVerify` in the following code in `pkg/preparer/setup.go` by returning a `tls.Config` that trusts nothing instead of `nil`:

```
func (c *PreparerConfig) GetInsecureClient(cxnTimeout time.Duration) (*http.Client, error) {
	tlsConfig, err := getTLSConfig(c.CertFile, c.KeyFile, c.CAFile)
	if err != nil {
		return nil, err
	}
	tlsConfig.InsecureSkipVerify = true
	...
```

### DIFF
--- a/pkg/preparer/setup.go
+++ b/pkg/preparer/setup.go
@@ -192,10 +192,6 @@ func getTLSConfig(certFile, keyFile, caFile string) (*tls.Config, error) {
 		}
 	}
 
-	if len(certs) == 0 && cas == nil {
-		return nil, nil
-	}
-
 	tlsConfig := &tls.Config{
 		Certificates: certs,
 		ClientCAs:    cas,


pkg/preparer/setup.go
```diff
@@ -179,9 +179,8 @@ func getTLSConfig(certFile, keyFile, caFile string) (*tls.Config, error) {
 		certs = append(certs, cert)
 	}
 
-	var cas *x509.CertPool
+	cas := x509.NewCertPool()
```
bwester: When the root CAs are unspecified, the system default CAs are used. That seems like preferable behavior to trusting nothing except what is explicitly configured.

Can you revert the change to these two lines? Removing the `if len(certs)...` block should be sufficient to prevent the exception.

drcapulet: Yeap, updated!

bwester:
👍

PR #537, opened 2016-06-30
Write pods to intent tree for daemon sets

KoishiKomeiji:
This commit will:
Provide functionality for writing to the intent tree for daemon sets
Provide unit tests for this functionality
Fix kp/dsstore/fake_dsstore.go so that in-memory tests will work

### DIFF
--- a/pkg/ds/daemon_set.go
+++ b/pkg/ds/daemon_set.go
@@ -282,9 +282,9 @@ func (ds *daemonSet) schedule(node types.NodeName) error {
 
 	// Will apply the following label on the key <labels.POD>/<node>/<ds.Manifest.ID()>:
 	// 	{ DSIDLabel : ds.ID() }
-	// eg node/127.0.0.1[daemon_set_id] := test_ds_id
+	// eg node/127.0.0.1/test_pod[daemon_set_id] := test_ds_id
 	// This is for indicating that this pod path belongs to this daemon set
-	id := MakePodLabelKey(node, ds.Manifest.ID())
+	id := labels.MakePodLabelKey(node, ds.Manifest.ID())
 	err := ds.applicator.SetLabel(labels.POD, id, DSIDLabel, ds.ID().String())
 	if err != nil {
 		return util.Errorf("Error setting label: %v", err)
@@ -312,7 +312,7 @@ func (ds *daemonSet) unschedule(node types.NodeName) error {
 
 	// Will remove the following label on the key <labels.POD>/<node>/<ds.Manifest.ID()>: DSIDLabel
 	// This is for indicating that this pod path no longer belongs to this daemon set
-	id := MakePodLabelKey(node, ds.Manifest.ID())
+	id := labels.MakePodLabelKey(node, ds.Manifest.ID())
 	err = ds.applicator.RemoveLabel(labels.POD, id, DSIDLabel)
 	if err != nil {
 		return util.Errorf("Error removing label: %v", err)
@@ -344,8 +344,4 @@ func (ds *daemonSet) CurrentPods() (PodLocations, error) {
 	}
 
 	return result, nil
-}
-
-func MakePodLabelKey(node types.NodeName, podID types.PodID) string {
-	return node.String() + "/" + podID.String()
 }
--- a/pkg/ds/daemon_set.go
+++ b/pkg/ds/daemon_set.go
@@ -1,0 +1,351 @@
+package ds
+
+import (
+	"time"
+
+	"github.com/square/p2/pkg/util"
+
+	klabels "k8s.io/kubernetes/pkg/labels"
+
+	"github.com/square/p2/pkg/ds/fields"
+	"github.com/square/p2/pkg/kp"
+	"github.com/square/p2/pkg/kp/dsstore"
+	"github.com/square/p2/pkg/labels"
+	"github.com/square/p2/pkg/logging"
+	"github.com/square/p2/pkg/pods"
+	"github.com/square/p2/pkg/scheduler"
+	"github.com/square/p2/pkg/types"
+)
+
+const (
+	// This label is applied to pods owned by an RC.
+	DSIDLabel = "daemon_set_id"
+)
+
+type DaemonSet interface {
+	ID() fields.ID
+
+	WatchDesires(
+		quitCh <-chan struct{},
+		updatedCh <-chan fields.DaemonSet,
+		deletedCh <-chan struct{},
+		nodesChangedCh <-chan struct{},
+	) <-chan error
+
+	// CurrentPods() returns all nodes that are scheduled by this daemon set
+	CurrentPods() (PodLocations, error)
+}
+
+type PodLocation struct {
+	Node  types.NodeName
+	PodID types.PodID
+}
+type PodLocations []PodLocation
+
+// Nodes returns a list of just the locations' nodes.
+func (l PodLocations) Nodes() []types.NodeName {
+	nodes := make([]types.NodeName, len(l))
+	for i, pod := range l {
+		nodes[i] = pod.Node
+	}
+	return nodes
+}
+
+// These methods are the same as the methods of the same name in kp.Store.
+// This interface makes writing unit tests easier as daemon sets do not
+// use any functions not listed here
+type kpStore interface {
+	SetPod(
+		podPrefix kp.PodPrefix,
+		nodeName types.NodeName,
+		manifest pods.Manifest,
+	) (time.Duration, error)
+
+	DeletePod(podPrefix kp.PodPrefix,
+		nodeName types.NodeName,
+		manifestID types.PodID,
+	) (time.Duration, error)
+}
+
+type daemonSet struct {
+	fields.DaemonSet
+
+	logger     logging.Logger
+	kpStore    kpStore
+	scheduler  scheduler.Scheduler
+	dsStore    dsstore.Store
+	applicator labels.Applicator
+}
+
+func New(
+	fields fields.DaemonSet,
+	dsStore dsstore.Store,
+	kpStore kpStore,
+	applicator labels.Applicator,
+	logger logging.Logger,
+) DaemonSet {
+	return &daemonSet{
+		DaemonSet: fields,
+
+		dsStore:    dsStore,
+		kpStore:    kpStore,
+		logger:     logger,
+		applicator: applicator,
+		scheduler:  scheduler.NewApplicatorScheduler(applicator),
+	}
+}
+
+func (ds *daemonSet) ID() fields.ID {
+	return ds.DaemonSet.ID
+}
+
+// WatchDesires watches for changes to its daemon set, then schedule/unschedule
+// pods to to the nodes that it is responsible for
+//
+// Whatever calls WatchDesires is responsible for sending signals for whether
+// the daemon set updated or deleted
+//
+// When this is first called, it assumes that the daemon set is created
+//
+// The caller is responsible for sending signals when something has been changed
+func (ds *daemonSet) WatchDesires(
+	quitCh <-chan struct{},
+	updatedCh <-chan fields.DaemonSet,
+	deletedCh <-chan struct{},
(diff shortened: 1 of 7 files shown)


pkg/ds/daemon_set.go
```diff
@@ -0,0 +1,353 @@
...
+	return nil
+}
+
+func (ds *daemonSet) schedule(node types.NodeName) error {
+	ds.logger.NoFields().Infof("Scheduling on %s", node)
+
+	// Will apply the following label on the key <labels.NODE>/<node.String()>:
+	// 	{ DSIDLabel : ds.ID().String() }
+	// eg node/127.0.0.1[daemon_set_id] := test_ds_id
+	err := ds.nodeApplicator.SetLabel(labels.NODE, node.String(), DSIDLabel, ds.ID().String())
```
mpuncel: i haven't been privy to a lot of the discussion around daemon sets, do we really want this to be a label on the `Node` or the `Pod`

mpuncel: i suspect the latter, because there might be two daemon sets scheduling different pods on the same node, and they would overwrite each other's labels

pkg/ds/daemon_set.go
```diff
@@ -0,0 +1,353 @@
...
+			case <-quitCh:
+				return
+			}
+		}
+	}()
+
+	return errCh
+}
+
+func (ds *daemonSet) meetDesires() error {
```
mpuncel: looks like this function is unused and doesn't do anything

pkg/ds/daemon_set.go
```diff
@@ -0,0 +1,353 @@
...
+
+	// Will remove the following key:
+	// <kp.INTENT_TREE>/<node>/<ds.Manifest.ID()>
+	_, err := ds.kpStore.DeletePod(kp.INTENT_TREE, node, ds.Manifest.ID())
+	if err != nil {
+		return util.Errorf("Unable to delete pod from intent tree: %v", err)
+	}
+
+	// Will remove the following label on the key <labels.NODE>/<node>: DSIDLabel
+	err = ds.nodeApplicator.RemoveLabel(labels.NODE, node.String(), DSIDLabel)
```
mpuncel: again, I think you probably want `labels.POD` label here

pkg/ds/daemon_set.go
```diff
@@ -0,0 +1,353 @@
...
+			case newDS := <-updatedCh:
+				ds.logger.NoFields().Infof("Received daemon set update signal: %v", newDS)
+				if ds.ID() != newDS.ID {
+					err = util.Errorf("Expected uuid to be the same, expected '%v', got '%v'", ds.ID(), newDS.ID)
+					continue
+				}
+				ds.DaemonSet = newDS
+
+				if ds.Disabled {
+					err = ds.clearPods()
```
mpuncel: do you really want to delete all of the pods when a daemon set is disabled? I suspect the behavior we want is for it to just stop doing anything

pkg/ds/daemon_set.go
```diff
@@ -0,0 +1,353 @@
...
+		if err != nil {
+			return util.Errorf("Error scheduling node: %v", err)
+		}
+	}
+	return nil
+}
+
+// clearPods unschedules pods for all the nodes that have been scheduled by
+// this daemon set by using CurrentNodes()
+// This should only be used when a daemon set is deleted
```
mpuncel: the comment here suggests `clearPods()` should only be called when deleting a daemon set, but above it is used when disabled

pkg/ds/daemon_set.go
```diff
@@ -0,0 +1,353 @@
...
+				}
+				err = ds.addPods()
+				if err != nil {
+					err = util.Errorf("Unable to add pods to intent tree: %v", err)
+					continue
+				}
+
+			case <-deletedCh:
+				ds.logger.NoFields().Infof("Received daemon set delete signal")
+				err = ds.clearPods()
```
mpuncel: i think we probably only want to do this if `!ds.Disabled`, but i'm not totally aware of the intended design

pkg/ds/daemon_set.go
```diff
@@ -0,0 +1,353 @@
...
+					continue
+				}
+				err := ds.removePods()
+				if err != nil {
+					err = util.Errorf("Unable to remove pods from intent tree: %v", err)
+					continue
+				}
+				err = ds.addPods()
+				if err != nil {
+					err = util.Errorf("Unable to add pods to intent tree: %v", err)
```
mpuncel: a general comment on errors, you're wrapping many errors in `util.Errorf` recursively, which is going to result in very long error messages with lots of line numbers. Usually I only wrap errors from 3rd party code with `util.Errorf` and everywhere an error is being generated by OUR code, I simply `return err`. This means we just get a line number at the border of our code. There may be some cases where a second tier is helpful (for example if there are many layers of abstraction and it's hard to reason about what behavior the error is preventing).

pkg/ds/daemon_set.go
```diff
@@ -0,0 +1,353 @@
...
+
+	// Do something whenever something is changed
+	go func() {
+		var err error
+		defer close(errCh)
+
+		// Try to schedule pods when this begins watching
+		if !ds.Disabled {
+			ds.logger.NoFields().Infof("Received new daemon set: %v", *ds)
+			err = ds.addPods()
```
mpuncel: why not `removePods()` here as well

pkg/ds/daemon_set.go
```diff
@@ -0,0 +1,353 @@
...
+	return nil
+}
+
+func (ds *daemonSet) CurrentNodes() ([]types.NodeName, error) {
+	// Changing DaemonSet.ID is not permitted, so as long as there is no
+	// uuid collision, this will always get the current nodes that this daemon set
+	// had scheduled on
+	selector := klabels.Everything().Add(DSIDLabel, klabels.EqualsOperator, []string{ds.ID().String()})
+
+	nodeMatches, err := ds.nodeApplicator.GetMatches(selector, labels.NODE)
```
mpuncel: `labels.POD` here too, replication controller likely has prior art for parsing the resulting strings since it's `<node>/<pod_id>`

pkg/ds/daemon_set_test.go
```diff
@@ -0,0 +1,332 @@
...
+	"github.com/square/p2/pkg/scheduler"
+	"github.com/square/p2/pkg/types"
+
+	. "github.com/anthonybishopric/gotcha"
+	ds_fields "github.com/square/p2/pkg/ds/fields"
+	klabels "k8s.io/kubernetes/pkg/labels"
+)
+
+type fakeKpStore struct {
+	manifests map[string]pods.Manifest
```
mpuncel: nit: it might be neater to use a `struct` containing prefix, nodename, manifest.id as the map key so you don't have to do `path.Join`

pkg/ds/daemon_set_test.go
```diff
@@ -0,0 +1,332 @@
...
+	"github.com/square/p2/pkg/pods"
+	"github.com/square/p2/pkg/scheduler"
+	"github.com/square/p2/pkg/types"
+
+	. "github.com/anthonybishopric/gotcha"
+	ds_fields "github.com/square/p2/pkg/ds/fields"
+	klabels "k8s.io/kubernetes/pkg/labels"
+)
+
+type fakeKpStore struct {
```
mpuncel: actually, this looks exactly the same as `FakePodStore` in `pkg/kp/kptest`, you can probably use that instead so we only have one implementation

pkg/ds/daemon_set.go
KoishiKomeiji: The deleted signal is sent when the daemon set is deleted in the store, so we shouldn't need to worry about what its fields were. Which is why we should unschedule it.

pkg/ds/daemon_set.go
KoishiKomeiji: `removePods()` will remove nodes(soon to be changed to pods) that have been labelled with the ds's uuid but not selected. Since nodes are first scheduled before the for loop at the beginning here, there shouldn't be any pods that have been labelled with the ds's uuid.

pkg/ds/daemon_set_test.go
KoishiKomeiji: Updated.

pkg/ds/daemon_set_test.go
KoishiKomeiji: Fixed below.

pkg/ds/daemon_set.go
KoishiKomeiji: Fixed above.

pkg/ds/daemon_set.go
KoishiKomeiji: Updated

pkg/ds/daemon_set.go
KoishiKomeiji: Updated

pkg/ds/daemon_set.go
KoishiKomeiji: Updated

pkg/ds/daemon_set.go
KoishiKomeiji: Fixed above.

pkg/ds/daemon_set.go
KoishiKomeiji: Updated

pkg/ds/daemon_set.go
mpuncel: 👍

pkg/ds/daemon_set.go
mpuncel: 👍

pkg/ds/daemon_set.go
```diff
@@ -0,0 +1,351 @@
...
+			return nil, err
+		}
+		result[i].Node = node
+		result[i].PodID = podID
+	}
+
+	return result, nil
+}
+
+func MakePodLabelKey(node types.NodeName, podID types.PodID) string {
```
mpuncel: bonus points if you put this in `pkg/labels/consul_applicator.go` and make replication controllers use it as well

KoishiKomeiji: Updated

mpuncel:
🚢

KoishiKomeiji:
wat

petertseng:
We only need to close and reopen if Travis didn't pick up the build. Once Travis has it, use the rebuild button at e.g. https://travis-ci.org/square/p2/builds/141494863 to make it build again

KoishiKomeiji:
Squashed changes.

pkg/ds/daemon_set.go
```diff
@@ -0,0 +1,347 @@
...
+		updatedCh <-chan fields.DaemonSet,
+		deletedCh <-chan struct{},
+		nodesChangedCh <-chan struct{},
+	) <-chan error
+
+	// CurrentPods() returns all nodes that are scheduled by this daemon set
+	CurrentPods() (PodLocations, error)
+}
+
+type PodLocation struct {
```
petertseng: Should we deduplicate this type and the `Nodes` function bweteen the `rc` package and this one by moving it somewhere?

KoishiKomeiji: I'll de-duplicate it in a later commit for now.

pkg/labels/consul_applicator.go
```diff
@@ -259,6 +259,14 @@ func (c *consulApplicator) WatchMatches(selector labels.Selector, labelType Type
 	return aggregator.Watch(selector, quitCh)
 }
 
+// these utility functions are used primarily while we exist in a mutable
+// deployment world. We will need to figure out how to replace these with
+// different datasources to allow RCs and DSs to continue to function correctly
+// in the future.
+func MakePodLabelKey(node types.NodeName, podID types.PodID) string {
```
petertseng: Hmm, why duplicate this from `rc` package, rather than use the one from `rc`?

petertseng: oh never mind you moved it!

PR #538, opened 2016-06-30
Launch launchables even if post-activate fails

mpuncel:
Previously the behavior was to refrain from launching a launchable if
its post-activate script failed, but still consider the launch of the
pod successful. This results in a confusing user experience when the pod
is partially started due to post-activate failures.

This commit simply logs the post-activate failure but proceeds to launch
each launchable anyway. The launchable status check will hopefully
capture any actual problems resulting from post-activate failing to run,
which will cause a deploy to stall via rolling update.


anthonybishopric:
👍

PR #540, opened 2016-07-01
Modify auth.ArtifactVerifier interface to take VerificationData struct

mpuncel:
We're planning to support two methods of specifying the location from
which a launchable may be downloaded:

1) using the "location" field in LaunchableStanza
2) using the "version" field in LaunchableStanza. This method is
currently not implemented to keep this commit smaller.

Previously all artifact downloads used the "location" field, and when
artifact verification is desired, the supplementary files needed to
verify the artifact (such as manifest and signatures) were inferred from
the artifact location by adding some special suffixes.

When the "version" scheme is used, an artifact registry will instead be
queried using the artifact metadata under the "version" key (as well as
other data) and the location of the artifact will be returned along with
the locations of the supplementary verification files. As a result, this
commit pulls the suffix-using logic out of artifact verification code.

Now the artifact package has a new function that takes a
LaunchableStanza and returns the location of the launchable and a
VerificationData struct that can be used for artifact verification.
Querying the artifact registry when using a "version" scheme is not yet
implemented.

### DIFF
--- a/pkg/auth/artifact_verifier.go
+++ b/pkg/auth/artifact_verifier.go
@@ -23,15 +23,26 @@ const VerifyManifest = "manifest"
 const VerifyBuild = "build"
 const VerifyEither = "either"
 
+// Contains URLs to extra files needed to verify the artifact. Not all verification
+// strategies make use of each field.
+type VerificationData struct {
+	// Used by BuildManifestVerifier
+	ManifestLocation          *url.URL
+	ManifestSignatureLocation *url.URL
+
+	// Used by BuildVerifier
+	BuildSignatureLocation *url.URL
+}
+
 // The artifact verifier is responsible for checking that the artifact
 // was created by a trusted entity.
 type ArtifactVerifier interface {
-	VerifyHoistArtifact(localCopy *os.File, artifactLocation *url.URL) error
+	VerifyHoistArtifact(localCopy *os.File, verificationData VerificationData) error
 }
 
 type nopVerifier struct{}
 
-func (n *nopVerifier) VerifyHoistArtifact(_ *os.File, _ *url.URL) error {
+func (n *nopVerifier) VerifyHoistArtifact(_ *os.File, _ VerificationData) error {
 	return nil
 }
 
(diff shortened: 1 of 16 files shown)


pkg/types/types.go
```diff
@@ -67,3 +69,27 @@ func (n NodeSet) PopAny() (NodeName, bool) {
 	node, ok := n.String.PopAny()
 	return NodeName(node), ok
 }
+
+type LaunchableVersion struct {
+	ID   string            `yaml:"id"`
+	Tags map[string]string `yaml:"tags"`
+}
+
+type LaunchableStanza struct {
```
mpuncel: this was to avoid an import cycle

pkg/auth/artifact_verifier_test.go
```diff
@@ -37,6 +37,26 @@ func buildTestFileTree(t *testing.T, files []testFile) string {
 	return tempDir
 }
 
+// This is copied from pkg/artifact/location_data.go to avoid an import cycle
+func VerificationDataForLocation(location *url.URL) VerificationData {
```
mpuncel: I anticipate this function being defined on an `ArtifactRegistry` struct in the near future

pkg/artifact/location_data.go
```diff
@@ -0,0 +1,64 @@
...
+		location, err := url.Parse(stanza.Location)
+		if err != nil {
+			return nil, auth.VerificationData{}, util.Errorf("Couldn't parse launchable url '%s': %s", stanza.Location, err)
+		}
+
+		verificationData := VerificationDataForLocation(location)
+		return location, verificationData, nil
+	}
+
+	return nil, auth.VerificationData{}, util.Errorf("Version key in launchable stanza not implemented")
```
anthonybishopric: I think this should not be a package level function. It's very likely we will want to back the artifact discovery functionality with a struct that has an http client, among other things. For some minor future proofing, can we put this on a `artifact.Registry` or something?

pkg/auth/artifact_verifier.go
```diff
@@ -115,50 +126,60 @@ func NewBuildManifestVerifier(keyringPath string, fetcher uri.Fetcher, logger *l
...
+	manifestSignatureLocation := verificationData.ManifestSignatureLocation
+	if manifestSignatureLocation == nil {
+		return util.Errorf("Manifest verification failed: manifest signature location not provided")
+	}
 
-		manifestDst := filepath.Join(dir, "manifest")
+	if manifestLocation.Scheme != "http" &&
+		manifestLocation.Scheme != "https" &&
+		manifestLocation.Scheme != "file" {
+		return util.Errorf("%v does not have a recognized scheme '%v', cannot verify manifest or signature", manifestLocation.String(), manifestLocation.Scheme)
```
anthonybishopric: Why are we adding this protocol check here and not letting the `uri.Fetcher` enforce it?

anthonybishopric: I realize that it was present before in the `switch` statement, but now that we've decided on the `version` key, instead of a `image://` protocol, I don't think it's necessary

pkg/types/types.go
anthonybishopric: What was the cycle? This definitely feels weird - this is a description of how a pod manifest will be structured. There are going to be multiple versions of pod manifest schemas, which is orthogonal to the idea of a launchable version and resolving it to artifact data. This `types` package does not seem like the right place to be putting these.

mpuncel: `pkg/pods` depends on `auth` in `pod.go`, and `auth` now depends on `LaunchableStanza` which was in `pkg/pods`

mpuncel: maybe we want a `pkg/manifest` or something that both `pods` and `auth` depend on?

pkg/artifact/location_data.go
mpuncel: sure, i did make a PR comment saying this is the intended future but I suppose I can do that now even if it doesn't do anything interesting yet

mpuncel:
addressed comments except import cycle one (awaiting further discussion)

anthonybishopric:
I like the idea of a `manifest` package to break the cycle, but that can happen in a future PR 🚢

PR #542, opened 2016-07-06
Provide functionality to notice changes in node tree

KoishiKomeiji:
Minor changes:
Share the PodLocation structure between RCs and DSs
Ask the caller to WatchDesires to send a daemon set when a change has been made so that you can verify it
Fix tests to for the above change to WatchDesires

Major changes:
Add a WatchMatchDiff function for each daemon set process to detect changes in node tree
Provide and modify tests for the above WatchDiff function

### DIFF
--- a/pkg/ds/daemon_set.go
+++ b/pkg/ds/daemon_set.go
@@ -29,7 +29,6 @@ type DaemonSet interface {
 		quitCh <-chan struct{},
 		updatedCh <-chan *fields.DaemonSet,
 		deletedCh <-chan *fields.DaemonSet,
-		nodesChangedCh <-chan struct{},
 	) <-chan error
 
 	// CurrentPods() returns all nodes that are scheduled by this daemon set
@@ -97,9 +96,9 @@ func (ds *daemonSet) WatchDesires(
 	quitCh <-chan struct{},
 	updatedCh <-chan *fields.DaemonSet,
 	deletedCh <-chan *fields.DaemonSet,
-	nodesChangedCh <-chan struct{},
 ) <-chan error {
 	errCh := make(chan error)
+	nodesChangedCh := ds.applicator.WatchMatchDiff(ds.NodeSelector, labels.NODE, quitCh)
 
 	// Do something whenever something is changed
 	go func() {
@@ -120,6 +119,7 @@ func (ds *daemonSet) WatchDesires(
 				select {
 				case errCh <- err:
 				case <-quitCh:
+					return
 				}
 			}
 
@@ -179,23 +179,16 @@ func (ds *daemonSet) WatchDesires(
 				}
 				return
 
-			case _, ok := <-nodesChangedCh:
+			case labeledChanges, ok := <-nodesChangedCh:
 				if !ok {
 					// channel closed
 					return
 				}
-				ds.logger.NoFields().Infof("Received node update signal")
 				if ds.Disabled {
 					continue
 				}
-				err := ds.removePods()
+				err = ds.handleNodeChanges(labeledChanges)
 				if err != nil {
-					err = util.Errorf("Unable to remove pods from intent tree: %v", err)
-					continue
-				}
-				err = ds.addPods()
-				if err != nil {
-					err = util.Errorf("Unable to add pods to intent tree: %v", err)
 					continue
 				}
 
@@ -206,6 +199,40 @@ func (ds *daemonSet) WatchDesires(
 	}()
 
 	return errCh
+}
+
+// Watch for changes to nodes and sends update and delete signals
+func (ds *daemonSet) handleNodeChanges(changes *labels.LabeledChanges) error {
+	if len(changes.Updated) > 0 {
+		ds.logger.NoFields().Infof("Received node change signal")
+		err := ds.removePods()
+		if err != nil {
+			return util.Errorf("Unable to remove pods from intent tree: %v", err)
+		}
+		err = ds.addPods()
+		if err != nil {
+			return util.Errorf("Unable to add pods to intent tree: %v", err)
+		}
+		return nil
+	}
+
+	if len(changes.Created) > 0 {
+		ds.logger.NoFields().Infof("Received node create signal")
+		err := ds.addPods()
+		if err != nil {
+			return util.Errorf("Unable to add pods to intent tree: %v", err)
+		}
+	}
+
+	if len(changes.Deleted) > 0 {
+		ds.logger.NoFields().Infof("Received node delete signal")
+		err := ds.removePods()
+		if err != nil {
+			return util.Errorf("Unable to remove pods from intent tree: %v", err)
+		}
+	}
+
+	return nil
 }
 
 // addPods schedules pods for all unscheduled nodes selected by ds.nodeSelector
--- a/pkg/ds/daemon_set.go
+++ b/pkg/ds/daemon_set.go
@@ -27,8 +27,8 @@ type DaemonSet interface {
 
 	WatchDesires(
 		quitCh <-chan struct{},
-		updatedCh <-chan fields.DaemonSet,
-		deletedCh <-chan struct{},
+		updatedCh <-chan *fields.DaemonSet,
+		deletedCh <-chan *fields.DaemonSet,
 		nodesChangedCh <-chan struct{},
 	) <-chan error
 
@@ -95,8 +95,8 @@ func (ds *daemonSet) ID() fields.ID {
 // The caller is responsible for sending signals when something has been changed
 func (ds *daemonSet) WatchDesires(
 	quitCh <-chan struct{},
-	updatedCh <-chan fields.DaemonSet,
-	deletedCh <-chan struct{},
+	updatedCh <-chan *fields.DaemonSet,
+	deletedCh <-chan *fields.DaemonSet,
 	nodesChangedCh <-chan struct{},
 ) <-chan error {
 	errCh := make(chan error)
@@ -124,13 +124,21 @@ func (ds *daemonSet) WatchDesires(
 			}
 
 			select {
-			case newDS := <-updatedCh:
+			case newDS, ok := <-updatedCh:
+				if !ok {
+					// channel closed
+					return
+				}
 				ds.logger.NoFields().Infof("Received daemon set update signal: %v", newDS)
+				if newDS == nil {
+					ds.logger.NoFields().Fatal("Unexpected a nil daemon set during update", *ds)
+					return
+				}
 				if ds.ID() != newDS.ID {
 					err = util.Errorf("Expected uuid to be the same, expected '%v', got '%v'", ds.ID(), newDS.ID)
 					continue
 				}
-				ds.DaemonSet = newDS
+				ds.DaemonSet = *newDS
 
 				if ds.Disabled {
 					continue
@@ -146,8 +154,21 @@ func (ds *daemonSet) WatchDesires(
 					continue
 				}
 
-			case <-deletedCh:
-				ds.logger.NoFields().Infof("Received daemon set delete signal")
+			case deleteDS, ok := <-deletedCh:
+				if !ok {
+					// channel closed
+					return
+				}
+				ds.logger.NoFields().Infof("Received daemon set delete signal: %v", deleteDS)
+				if deleteDS == nil {
+					ds.logger.NoFields().Fatal("Unexpected a nil daemon set during delete")
+					return
+				}
+				if ds.ID() != deleteDS.ID {
+					err = util.Errorf("Expected uuid to be the same, expected '%v', got '%v'", ds.ID(), deleteDS.ID)
+					continue
+				}
+
 				err = ds.clearPods()
 				if err != nil {
 					err = util.Errorf("Unable to clear pods from intent tree: %v", err)
@@ -158,7 +179,11 @@ func (ds *daemonSet) WatchDesires(
 				}
 				return
 
-			case <-nodesChangedCh:
+			case _, ok := <-nodesChangedCh:
+				if !ok {
+					// channel closed
+					return
+				}
 				ds.logger.NoFields().Infof("Received node update signal")
 				if ds.Disabled {
 					continue
@@ -263,7 +288,7 @@ func (ds *daemonSet) clearPods() error {
 }
 
 func (ds *daemonSet) schedule(node types.NodeName) error {
-	ds.logger.NoFields().Infof("Scheduling on %s", node)
+	ds.logger.NoFields().Infof("Scheduling on %s with daemon set uuid %s", node, ds.ID())
 
 	// Will apply the following label on the key <labels.POD>/<node>/<ds.Manifest.ID()>:
 	// 	{ DSIDLabel : ds.ID() }
@@ -286,7 +311,7 @@ func (ds *daemonSet) schedule(node types.NodeName) error {
 }
 
 func (ds *daemonSet) unschedule(node types.NodeName) error {
-	ds.logger.NoFields().Infof("Unscheduling from %s", node)
+	ds.logger.NoFields().Infof("Unscheduling from %s with daemon set uuid %s", node, ds.ID())
 
 	// Will remove the following key:
 	// <kp.INTENT_TREE>/<node>/<ds.Manifest.ID()>
--- a/pkg/ds/daemon_set.go
+++ b/pkg/ds/daemon_set.go
@@ -33,22 +33,7 @@ type DaemonSet interface {
 	) <-chan error
 
 	// CurrentPods() returns all nodes that are scheduled by this daemon set
-	CurrentPods() (PodLocations, error)
-}
-
-type PodLocation struct {
-	Node  types.NodeName
-	PodID types.PodID
-}
-type PodLocations []PodLocation
-
-// Nodes returns a list of just the locations' nodes.
-func (l PodLocations) Nodes() []types.NodeName {
-	nodes := make([]types.NodeName, len(l))
-	for i, pod := range l {
-		nodes[i] = pod.Node
-	}
-	return nodes
+	CurrentPods() (types.PodLocations, error)
 }
 
 // These methods are the same as the methods of the same name in kp.Store.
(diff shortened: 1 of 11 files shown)


pkg/ds/daemon_set.go
```diff
@@ -161,8 +150,17 @@ func (ds *daemonSet) WatchDesires(
 					continue
 				}
 
-			case <-deletedCh:
-				ds.logger.NoFields().Infof("Received daemon set delete signal")
+			case deleteDS := <-deletedCh:
```
mpuncel: is the caller of `WatchDesires()` going to be able to know the contents of the deleted daemon set? I would think all we'll know is the uuid of it

KoishiKomeiji: The current implementation `WatchDiff` will allow the caller to know what was deleted because it keeps a copy of the daemon set in memory. Since the farm uses `WatchDiff` and `WatchDesires`, it will know the contents.

mpuncel: what happens if a daemon set is deleted while the process isn't running?

KoishiKomeiji: Currently, if a daemon set is deleted, we will lose track of it, I plan to include a check in the daemon set farm during initialization that will unschedule nodes that are labelled with a daemon set uuid when their daemon set does not exist in the tree.

pkg/ds/daemon_set.go
```diff
@@ -141,11 +126,15 @@ func (ds *daemonSet) WatchDesires(
 			select {
 			case newDS := <-updatedCh:
 				ds.logger.NoFields().Infof("Received daemon set update signal: %v", newDS)
+				if newDS == nil {
+					ds.logger.NoFields().Infof("Very odd, recieved a nil daemon set during update for %v, recovered", *ds)
```
anthonybishopric: Remove "Very odd,"

pkg/ds/daemon_set_test.go
```diff
@@ -101,17 +103,54 @@ func watchDSChanges(
...
 					ds.logger.NoFields().Infof("Watched daemon set get updated: %v", *changedDS)
-					updatedCh <- *changedDS
+					updatedCh <- changedDS
 				}
 			}
 			for _, changedDS := range watched.Deleted {
 				if ds.ID() == changedDS.ID {
-					ds.logger.NoFields().Infof("Watched daemon set get deleted: %v", changedDS)
-					deletedCh <- struct{}{}
+					ds.logger.NoFields().Infof("Watched daemon set get deleted: %v", *changedDS)
```
anthonybishopric: s/get/was

pkg/labels/fake_applicator.go
```diff
@@ -117,6 +118,79 @@ func (app *fakeApplicator) WatchMatches(selector labels.Selector, labelType Type
 	return ch
 }
 
+func (app *fakeApplicator) WatchDiff(labelType Type, quitCh <-chan struct{}) <-chan *LabeledChanges {
```
anthonybishopric: Why is this necessary? Shouldn't watching a daemon set's node selector be sufficient for detecting node changes?

KoishiKomeiji: Updated, all daemon set processes will now use a WatchMatchDiff function which uses the WatchMatches as a helper function to detect changes.

pkg/ds/daemon_set_test.go
KoishiKomeiji: Updated

pkg/ds/daemon_set.go
KoishiKomeiji: Updated.

pkg/labels/http_applicator.go
```diff
@@ -95,7 +95,21 @@ func (h *httpApplicator) GetMatches(selector labels.Selector, labelType Type) ([
...
+func (h *httpApplicator) WatchMatchDiff(
+	selector labels.Selector,
+	labelType Type,
+	quitCh <-chan struct{},
+) <-chan *LabeledChanges {
+	panic("Not implemented")
+}
+
+/*
+func (h *httpApplicator) WatchDiff(labelType Type, quitCh <-chan struct{}) <-chan *LabeledChanges {
```
KoishiKomeiji: Oops, I'll remove this tomorrow.

pkg/labels/consul_applicator.go
```diff
@@ -259,6 +259,88 @@ func (c *consulApplicator) WatchMatches(selector labels.Selector, labelType Type
...
+			select {
+			case <-quitCh:
+				return
+			case labelsList = <-inCh:
+				if labelsList == nil {
+					c.logger.Errorf("Unexpected nil value received from WatchMatches, recovered")
+					continue
+				}
+			}
+			break
```
mpuncel: rather than using `for {  break }` just do the `select{}` statement

pkg/labels/fake_applicator.go
```diff
@@ -102,21 +104,124 @@ func (app *fakeApplicator) GetMatches(selector labels.Selector, labelType Type)
...
+			select {
+			case <-quitCh:
+				return
+			case labelsList = <-inCh:
+				if labelsList == nil {
+					fmt.Println("Unexpected nil value received from WatchMatches, recovered")
+					continue
+				}
+			}
+			break
```
mpuncel: same

pkg/labels/fake_applicator.go
```diff
@@ -102,21 +104,124 @@ func (app *fakeApplicator) GetMatches(selector labels.Selector, labelType Type)
...
+				if _, ok := oldLabels[id]; !ok {
+					// If it is not observed, then it was created
+					outgoingChanges.Created = append(outgoingChanges.Created, Labeled{
+						ID:        id,
+						LabelType: labelType,
+						Labels:    copySet(labeled.Labels),
+					})
+					oldLabels[id] = labeled
+
+				} else if oldLabels[id].Labels.String() != labeled.Labels.String() {
```
mpuncel: nit: it seems like there's a lot of code-reuse here. Given that it's in a fake implementation I don't care that much but consider extracting this to a function that both real and fake can use

pkg/labels/fake_applicator_test.go
```diff
@@ -2,6 +2,7 @@ package labels
```
mpuncel: am I understanding correctly that this is testing `WatchMatchDiff()` on the fake applicator but there's no test for `WatchMatchDiff` on the real one?

KoishiKomeiji: Yes.

pkg/labels/consul_applicator.go
KoishiKomeiji: Updated

pkg/labels/fake_applicator.go
KoishiKomeiji: Updated

pkg/labels/fake_applicator.go
KoishiKomeiji: Updated.

pkg/labels/fake_applicator_test.go
KoishiKomeiji: Updated to add test to the real `WatchMatchDiff`

bwester:
This is a really large PR.  There are a couple problems with large PRs that come to mind when reading through this:
1. When there is more code to review, it takes reviewers longer to understand and consider it all.
2. Uncontroversial changes, like moving `PodLocation` to a different package, get held up by reviewing all the other changes.
3. It's easier to understand one commit that proposes a single logical change instead of several where you implement something one way only to change it later in the same PR. (E.g., 0d2cfe82 and ffbffb16.)

I would like to suggest that you should break this one large PR into several smaller chunks for code review.

bwester:
It's important that your change request include not only the code being changed but the *context* of the change.  Does it solve a problem?  Is there a specific bug it fixes?  Why does this code need changing?  Why is this the right way to change it? What is your intent?

All of this may be obvious to everyone right now, but someone looks back at this code in 6 months, a clear English description will make the code much easier understand.

Please add some context to your git commit messages.

KoishiKomeiji:
Updated, rebased into less commits and added more context to commit messages.

bwester:
From the commit message to 3eff464:
> The tests in WatchDiff have been changed ...
> This will remove the code ...
> This will also change the tests ...
> There are also changes ...
> Also change ...

I want to reiterate that you should work to put one logical change into each commit. Each commit should be a self-contained bundle of changes that accomplish *one* goal and can stand on its own.

Why go to the effort? It's relevant right now because it makes reviewing easier.  When I read through a PR, I have to learn how the code works while also figuring out which parts are related to watching their own nodes, improving type safety, fixing error messages, fixing other bugs, etc.  *You can do that last part for me!* After having written it, you know your own code better than anyone else—you can split apart separate concerns faster and more precisely than I can.

If I don't spend time having to disentangle parts your code, I can spend it checking the code for completeness and correctness. Your review will go faster, you will get higher quality feedback, and everyone will have higher confidence that this code is good. I suspect that failing to do this is one reason why this PR is now a couple weeks old without being approved on an otherwise-fast-moving team.

This isn't simply about the number of commits. Github's UI merges everything together anyway. Instead, it's about you presenting a clear view of your work in a way that makes it understandable to others.

bwester:
:shipit: for commit 7a16654.

You still have the problem that I called out earlier: 3eff464 adds some logging messages that are then changed in b330869. Present it right the first time.

KoishiKomeiji:
Updated, commits are now broken down into separate and more logical parts.

pkg/labels/applicator.go
```diff
@@ -86,5 +92,13 @@ type Applicator interface {
 	// the quit channel - this will be indicated by the closing of the result channel. For
 	// this reason, callers should **always** verify that the channel is closed by checking
 	// the "ok" boolean or using `range`.
-	WatchMatches(selector labels.Selector, labelType Type, quitCh chan struct{}) chan []Labeled
+	WatchMatches(selector labels.Selector, labelType Type, quitCh <-chan struct{}) chan []Labeled
```
bwester: Can you put this change in its own commit or PR?

pkg/ds/daemon_set_test.go
```diff
@@ -33,9 +33,9 @@ func waitForNodes(
 	ds DaemonSet,
 	desired int,
 	desiresErrCh <-chan error,
-	changesErrCh <-chan error,
+	dsChangesErrCh <-chan error,
 ) int {
-	timeout := time.After(1 * time.Second)
+	timeout := time.After(2 * time.Second)
```
bwester: Why do we need a larger timeout?

KoishiKomeiji: My test failed due to a timeout issue once, so I'm a little paranoid it might happen in the future.
It doesn't timeout using one second anymore though.

pkg/labels/consul_applicator.go
```diff
@@ -259,6 +259,131 @@ func (c *consulApplicator) WatchMatches(selector labels.Selector, labelType Type
...
+					return
+				}
+			}
+
+			newLabels := make(map[string]Labeled)
+			for _, labeled := range results {
+				newLabels[labeled.ID] = labeled
+			}
+
+			//outgoingChanges := diffLabels(oldLabels, newLabels)
```
bwester: Remove commented code

pkg/labels/fake_applicator.go
```diff
@@ -102,21 +105,73 @@ func (app *fakeApplicator) GetMatches(selector labels.Selector, labelType Type)
...
 	ch := make(chan []Labeled)
+	timer := time.NewTimer(time.Duration(0))
 	go func() {
 		for {
+			select {
+			case <-quitCh:
+				return
+			case <-timer.C:
+			}
+			// upper bound on request rate
```
bwester: I don't think we need to worry about request rate in `fakeApplicator`

pkg/labels/fake_applicator.go
```diff
@@ -102,21 +105,73 @@ func (app *fakeApplicator) GetMatches(selector labels.Selector, labelType Type)
...
+) <-chan *LabeledChanges {
+	outCh := make(chan *LabeledChanges)
+
+	go func() {
+		defer close(outCh)
+
+		logger := logging.DefaultLogger
+		inCh := app.WatchMatches(selector, labelType, quitCh)
+		diffCh := watchDiffLabels(inCh, quitCh, logger)
+		for {
```
bwester: What does this loop do? As far as I can tell, it copies from one channel to another. Can you just return `diffCh`?

pkg/labels/consul_applicator.go
```diff
@@ -259,6 +259,131 @@ func (c *consulApplicator) WatchMatches(selector labels.Selector, labelType Type
...
+	quitCh <-chan struct{},
+) <-chan *LabeledChanges {
+	outCh := make(chan *LabeledChanges)
+
+	go func() {
+		defer close(outCh)
+
+		inCh := c.WatchMatches(selector, labelType, quitCh)
+		diffCh := watchDiffLabels(inCh, quitCh, c.logger)
+		for {
```
bwester: As in my other comment, what benefit does this loop provide?

pkg/labels/fake_applicator.go
KoishiKomeiji: We won't need to worry about it for performance, but I'll need it so the changes are all captured together for the tests to work consistently.

bwester: Unit tests execute in a variety of environments and should not rely on timing for correctness. If you want to make sure that a group of changes are found by this method, you should add different synchronization here.

However, I suspect that what you really want is to test a consumer of the `chan []Labeled`. If so, your unit tests should construct the `[]Labeled` batches you want your consumer to see and send them over a channel you control.

pkg/labels/consul_applicator.go
KoishiKomeiji: Updated, it's been removed now.

pkg/labels/fake_applicator.go
KoishiKomeiji: Updated.

pkg/labels/fake_applicator.go
KoishiKomeiji: Updated, tests are configured to pass without the rate now.

pkg/ds/daemon_set.go
```diff
@@ -97,14 +96,15 @@ func (ds *daemonSet) WatchDesires(
 	quitCh <-chan struct{},
 	updatedCh <-chan *fields.DaemonSet,
 	deletedCh <-chan *fields.DaemonSet,
-	nodesChangedCh <-chan struct{},
 ) <-chan error {
 	errCh := make(chan error)
+	nodesChangedCh := make(chan struct{})
```
bwester: Why not `nodesChangedCh := ds.watchNodeChanges(quitCh)` and remove the goroutine you added to the end of the method? Would that accomplish the same thing?

pkg/ds/daemon_set.go
```diff
@@ -205,9 +206,62 @@ func (ds *daemonSet) WatchDesires(
...
+	}()
+
 	return errCh
 }
 
+// Watch for changes to nodes and sends update and delete signals
+func (ds *daemonSet) watchNodeChanges(
+	quitCh <-chan struct{},
+) <-chan struct{} {
+	nodesChangedCh := make(chan struct{})
```
bwester: What owns this channel?

pkg/ds/daemon_set.go
```diff
@@ -205,9 +206,62 @@ func (ds *daemonSet) WatchDesires(
...
+				}
+				changes = val
+				if changes == nil {
+					ds.logger.Error("Unexpected nil value received")
+					continue
+				}
+			}
+
+			if len(changes.Created) != 0 || len(changes.Updated) != 0 || len(changes.Deleted) != 0 {
+				nodesChangedCh <- struct{}{}
```
bwester: It seems like a wasted opportunity to meticulously gather a list of the created, updated, and deleted nodes in the query, then send a message signaling no more than, "Hey! Something changed!"

Is there any way to make use of the data you have?

pkg/ds/daemon_set.go
KoishiKomeiji: This function owns the channel, I'll add a
`defer close(nodesChangedCh)`

pkg/ds/daemon_set.go
KoishiKomeiji: Updated.

pkg/ds/daemon_set.go
KoishiKomeiji: Updated.

bwester:
I think I'm OK with this PR now.

pkg/ds/daemon_set.go
KoishiKomeiji: Updated to ds smarter scheduling based on whether there was an create, update, or delete.

pkg/ds/daemon_set.go
```diff
@@ -135,17 +119,26 @@ func (ds *daemonSet) WatchDesires(
...
 			select {
-			case newDS := <-updatedCh:
+			case newDS, ok := <-updatedCh:
+				if !ok {
+					// channel closed
+					return
+				}
 				ds.logger.NoFields().Infof("Received daemon set update signal: %v", newDS)
+				if newDS == nil {
+					ds.logger.NoFields().Fatal("Unexpected a nil daemon set during update", *ds)
```
petertseng: can I just ask whether it's OK for this to be fatal? Who will be running it and is it OK with stopping the entire process?

pkg/ds/daemon_set.go
```diff
@@ -161,8 +154,21 @@ func (ds *daemonSet) WatchDesires(
...
-			case <-deletedCh:
-				ds.logger.NoFields().Infof("Received daemon set delete signal")
+			case deleteDS, ok := <-deletedCh:
+				if !ok {
+					// channel closed
+					return
+				}
+				ds.logger.NoFields().Infof("Received daemon set delete signal: %v", deleteDS)
+				if deleteDS == nil {
+					ds.logger.NoFields().Fatal("Unexpected a nil daemon set during delete")
```
petertseng: same q about fatal here

anthonybishopric: Definitely do not use Fatal - we do not want the entire server to rash.

PR #543, opened 2016-07-06
Move the Manifest type from pkg/pods to pkg/manifest

mpuncel:
This change allows the LaunchableStanza type to move out of the
pkg/types package which is a dumping ground for commonly imported types
that result in import cycles when in a more "logical" location.

The immediate motivation was the Pod type (in pkg/pods) having a
reference to the pkg/auth (ArtifactVerifier) type, and the artifact
verifier type having a reference to pkg/pods.LaunchableStanza, resulting
in it moving to the types package.

### DIFF
--- a/pkg/pods/pod_test.go
+++ b/pkg/pods/pod_test.go
@@ -15,6 +15,7 @@ import (
 	"github.com/square/p2/pkg/artifact"
 	"github.com/square/p2/pkg/hoist"
 	"github.com/square/p2/pkg/launch"
+	"github.com/square/p2/pkg/manifest"
 	"github.com/square/p2/pkg/runit"
 	"github.com/square/p2/pkg/types"
 	"github.com/square/p2/pkg/util"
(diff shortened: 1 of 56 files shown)


petertseng:
in what may be a follow up PR, consider renaming `manifest.ManifestBuilder` to `manifest.Builder` to avoid stutter.

also with `manifest.NewManifestBuilder` can probably be called `manifest.NewBuilder`

pkg/pods/pod_test.go
```diff
@@ -216,43 +217,35 @@ func TestLogInfo(t *testing.T) {
...
+	existing := getTestPodManifest(t).GetBuilder()
+	updated := getUpdatedManifest(t).GetBuilder()
 
 	poddir, err := ioutil.TempDir("", "poddir")
 	Assert(t).IsNil(err, "couldn't create tempdir")
 	pod := NewPod(types.PodID("testPod"), poddir)
 
 	// set the RunAs user to the user running the test, because when we
-	// write files we need an owner. Unset them after the write so that the
-	// manifests match the manifests on disk
```
mpuncel: I'm a little puzzled by the changes that were required here. It seems that the test as written before was expecting the `existing.Marshal()` call to not write the `RunAs:` user out. The test looks sane in its current form but i'm not sure why some of this was needed before and not now.

petertseng:
Seems like a simple enough move/rename, I think the only behavior that changes is to add `SetRestartPolicy`, so seems OK. I don't have a good answer for why the test required the changes it did, heh.

PR #544, opened 2016-07-06
Support launchables with locations that don't match typical format

mpuncel:
Normally we expect a launchable to have a location URI whose base is of
the form <launchable_id>_<version>.tar.gz. Prior to this commit, errors
would occur during artifact installation if the location did not match
these scheme, since no version can be parsed out of the location.

This commit simply logs a warning if this occurs, but otherwise will
proceed without the version information for backwards compatibility with
launchables that don't fit the format.

Version is primarily used to determine the name of the directory the
launchable will be installed in, using <launchable_id>_<version>. If
version is blank, the directory will just be <launchable_id>. This means
that launchables that do not have version information in their location
will fail to install new updates as the preparer will think it is
already installed.


anthonybishopric:
👍

PR #545, opened 2016-07-11
Add ability to query artifact registry based on launchable version field

mpuncel:
Previously launchables had to be fetched from a URL specified in the
"location" field in the launchable stanza, and the "version" field was
unimplemented. This commit makes the artifact registry query the
configured registry URL and adds any artifact tags to the query string.

We intend to add OS information as an extra tag in a future commit.


pkg/artifact/registry.go
```diff
@@ -55,7 +64,92 @@ func (a registry) LocationDataForLaunchable(stanza manifest.LaunchableStanza) (*
...
+}
+
+type RegistryResponse struct {
+	ArtifactURL          string `json:"artifact_url"`
+	ManifestURL          string `json:"manifest_url"`
+	ManifestSignatureURL string `json:"manifest_signature_url"`
+	BuildSignatureURL    string `json:"build_signature_url"`
+}
+
+func (a registry) fetchRegistryData(launchableID manifest.LaunchableID, version manifest.LaunchableVersion) (*url.URL, auth.VerificationData, error) {
```
mpuncel: note: this function isn't yet complete, pending knowing the query structure we want for the registry fetch.

1) launchableID not used yet
2) no OS version tag yet

mpuncel:
I just broke this PR up into smaller commits to make review easier

pkg/manifest/manifest_test.go
```diff
@@ -23,7 +23,10 @@ func TestPodManifestCanBeRead(t *testing.T) {
 	Assert(t).AreEqual("hello", string(manifest.ID()), "Id read from manifest didn't have expected value")
 	Assert(t).AreEqual(manifest.GetLaunchableStanzas()["app"].Location, "hoisted-hello_def456.tar.gz", "Location read from manifest didn't have expected value")
 	Assert(t).AreEqual("hoist", manifest.GetLaunchableStanzas()["app"].LaunchableType, "LaunchableType read from manifest didn't have expected value")
-	Assert(t).AreEqual("app", manifest.GetLaunchableStanzas()["app"].LaunchableId, "LaunchableId read from manifest didn't have expected value")
+
+	if manifest.GetLaunchableStanzas()["app"].LaunchableId != "app" {
```
petertseng: what if  `manifest.GetLaunchableStanzas()["app"].LaunchableId.String()` were to be used in the assert instead?

mpuncel: that would work, I'd just rather have 1) compiler safety nagging me if the types changes again here and 2) not have confusing test failure messages

pkg/pods/pod_test.go
```diff
@@ -314,6 +327,59 @@ func TestBuildRunitServices(t *testing.T) {
...
+
+	pod := Pod{
+		Id:      "testPod",
+		path:    testPodDir,
+		logger:  Log.SubLogger(logrus.Fields{"pod": "testPod"}),
+		Fetcher: fetcher,
+	}
+
+	err = pod.Install(manifest, auth.NopVerifier(), artifact.NewRegistry(nil, uri.DefaultFetcher, osversion.DefaultDetector))
+	Assert(t).IsNil(err, "there should not have been an error when installing")
```
mpuncel: on travis i get an error here looking up the "hello" user, but i've clearly set the user on line 350 😖

pkg/hoist/hoist_launchable.go
```diff
@@ -14,6 +14,7 @@ import (
 	"github.com/square/p2/pkg/auth"
 	"github.com/square/p2/pkg/cgroups"
 	"github.com/square/p2/pkg/launch"
+	"github.com/square/p2/pkg/manifest"
```
anthonybishopric: I don't think launchables shouldn't know about pod manifests per previous discussions around versioning of manifests and separation of concerns.

mpuncel: moving `LaunchableID` into `pkg/launch` should fix this

pkg/manifest/manifest.go
```diff
@@ -449,10 +419,10 @@ func ValidManifest(m Manifest) error {
 			return fmt.Errorf("'%s': launchable must contain a 'launchable_type'", key)
 		case stanza.LaunchableId == "":
 			return fmt.Errorf("'%s': launchable must contain a 'launchable_id'", key)
-		case stanza.Version.ID != "":
-			return fmt.Errorf("'%s': 'version' launchable key not yet supported", key)
-		case stanza.Location == "":
-			return fmt.Errorf("'%s': launchable must contain a 'location'", key)
+		case stanza.Location == "" && stanza.Version.ID == "":
+			return fmt.Errorf("'%s': launchable must contain a 'location' or 'version'", key)
+		case stanza.Location != "" && stanza.Version.ID != "":
```
rudle: Based on how I read launchable.go, this case isn't strictly necessary but I appreciate it being here nonetheless.

rudle:
👍  looks good. Merge after verifying in staging

PR #546, opened 2016-07-11
add list function to p2-pcctl

mpuncel:


KoishiKomeiji:
👍

petertseng:
The code generally looks good, but when I try to run it I get a `Could not create session: session.go:40: Could not create session` - did I do something wrong

petertseng:
Right, need to either pass the headers or shim it, not just run it out of the directory. Works.

👍

PR #547, opened 2016-07-11
Fix misleading log message in RU

mpuncel:
The message was "could not lock update", when in reality it is the
replication controller lock failure that is causing the message. This
commit changes that to "could not lock rcs"


KoishiKomeiji:
👍

petertseng:
easy, 👍

PR #548, opened 2016-07-11
Adds convenience functions for marshaling RC and PC types as JSON

mpuncel:
Due to the fact that the RC and PC structs contain custom types that do
not marshal to JSON automatically, representing them as JSON involves
going through an intermediary type named RawRC and RawPC, respectively.
This commit adds convenience functions for converting an RC or PC to the
corresponding json-friendly type.


KoishiKomeiji:
👍

petertseng:
Hmm, code looks all right. Is the idea that I may call `ToRaw`, do something with the raw struct, and only then will I marshal it? Or is it something else?

mpuncel:
yes that's correct. In particular this is useful if you need to pass a JSON marshal-able object to another library, e.g. graphql library

PR #550, opened 2016-07-18
Add a configurable "pause" time duration to consulutil.WatchPrefix

mpuncel:
The pause signifies the amount of time to wait after the consul server
returns a watch result. Clients that can tolerate a high degree of
staleness of returned data can specify a large pause interval to reduce
QPS on the consul store. This is still preferable to polling because the
client can remain responsive when data turnover is slow, but can
significantly reduce QPS if data turnover is high.


bwester:
Where do you intend to use this?

bwester:
🚢 btw

PR #551, opened 2016-07-19
clarify error message

anthonybishopric:
This adds some words that might make this error clearer to users of P2 not familiar with our objects' acronyms.


mpuncel:
🚀

PR #552, opened 2016-07-20
Make version detection from location URL stricter

mpuncel:
Prior to this commit, the version detection code would do so incorrectly
on some artifact names. Now the regex enforces that the version is
a 40 character hexadecimal string with an optional hyphenated suffix.

This resulted in some bad install directory names. For example, a
launchable with id "foo_launchable.tar.gz" would be installed in a
directory named "foo_launchable_launchable/" because the launchable ID
was concatenated with the supposed version "launchable". The desired
behavior would have been for versionFromLocation() to return an error
and the ID of "foo_launchable" alone would be used to determine the
directory name.

### DIFF
--- a/pkg/pods/pod.go
+++ b/pkg/pods/pod.go
@@ -8,6 +8,7 @@ import (
 	"os"
 	"path"
 	"path/filepath"
+	"regexp"
 	"strings"
 	"time"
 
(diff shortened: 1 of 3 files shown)


pkg/pods/pod.go
```diff
@@ -717,18 +718,25 @@ func (pod *Pod) getLaunchable(launchableStanza manifest.LaunchableStanza, runAsU
...
-	parts := strings.Split(filename, "_")
-	if len(parts) < 2 {
+	parts := locationBaseRegex.FindStringSubmatch(filename)
+	if parts == nil {
 		return "", util.Errorf("Malformed filename in URL: %s", filename)
 	}
 
-	versionAndExtension := parts[len(parts)-1]
+	version := parts[1]
+	if len(parts) >= 3 && parts[2] != "" {
```
bwester: If there is a match, `len(parts)` will not vary based on the input string. The length is defined by the number of opening parenthesis in the regular expression.

https://play.golang.org/p/pWKJJI8QeE

The point is, you shouldn't need to ever check the length.

pkg/pods/pod.go
```diff
@@ -717,18 +718,25 @@ func (pod *Pod) getLaunchable(launchableStanza manifest.LaunchableStanza, runAsU
...
+		version = version + parts[2]
```
bwester: If you want to include the "extra" part of the version spec, I would suggest you just change your regular expression to capture that part too.

pkg/pods/pod.go
```diff
@@ -717,18 +718,25 @@ func (pod *Pod) getLaunchable(launchableStanza manifest.LaunchableStanza, runAsU
 
 // Uses the assumption that all locations have a Path component ending in
 // /<launchable_id>_<version>.tar.gz, which is intended to be phased out in
-// favor of explicit launchable versions specified in pod manifests
-// Version may not contain an underscore
+// favor of explicit launchable versions specified in pod manifests.
+// The version expected to be a 40 character hexadecimal string with an
+// optional hexadecimal suffix after a hyphen
+
+var locationBaseRegex = regexp.MustCompile(`^[a-z0-9-_]+_([a-f0-9]{40})(\-([a-z0-9]+))?\.tar\.gz$`)
```
bwester: re my other comment:

    `^[a-z0-9-_]+_([a-f0-9]{40}(\-[a-z0-9]+)?)\.tar\.gz$`

anthonybishopric:
👍

PR #553, opened 2016-07-22
Add a minimally working farm that instantiates daemon sets

KoishiKomeiji:


pkg/ds/farm.go
```diff
@@ -176,8 +178,73 @@ func (dsf *Farm) handleDSChanges(quitCh <-chan struct{}, changes *dsstore.Watche
...
+		return "", false, nil
+	}
+	// Get all eligible nodes for this daemon set by looking at the labes.NODE tree
+	eligibleNodes, err := dsf.scheduler.EligibleNodes(dsFields.Manifest, dsFields.NodeSelector)
+	if err != nil {
+		return "", false, util.Errorf("Error retrieving eligible nodes for daemon set: %v", err)
+	}
+
+	// If this daemon set has a node selector set to Everything, check the labels
+	// of other daeon sets
```
rudle: typo: daeon -> daemon

pkg/ds/farm.go
```diff
@@ -176,8 +178,73 @@ func (dsf *Farm) handleDSChanges(quitCh <-chan struct{}, changes *dsstore.Watche
...
+	for _, child := range dsf.children {
+		everythingSelector := klabels.Everything().String()
+		if !child.ds.IsDisabled() && child.ds.PodID() == dsFields.PodID && child.ds.ID() != dsFields.ID {
+			// Naively check if both selectors are the Everything selector plus
+			// something else or if both selectors are the same
+			//
+			// It is possible that both selectors won't contentd if they both have
```
rudle: typo: contentd -> content

pkg/ds/farm.go
```diff
@@ -0,0 +1,342 @@
...
+							return
+						}
+
+						// TODO: Since this mirrors the unschedule function in daemon_set.go,
+						// We should find a nice way to couple them together
+						dsf.logger.NoFields().Infof("Unscheduling from %s with dangling daemon set uuid %s", nodeName, dsID)
+
+						_, err = dsf.kpStore.DeletePod(kp.INTENT_TREE, nodeName, podID)
+						if err != nil {
+							dsf.logger.NoFields().Errorf("Unable to delete pod from intent tree: %v", err)
```
rudle: Is the error handling strategy here something like: "We'll attempt this same DeletePod call on the next iteration of the loop if we can't delete it now for some reason?"

That's fine, but I'd like to see an ErrorReporter call happen here also.

pkg/ds/daemon_set_test.go
```diff
@@ -82,7 +82,7 @@ func watchDSChanges(
 		defer close(errCh)
 
 		for {
-			var watched dsstore.WatchedDaemonSets
+			var watched *dsstore.WatchedDaemonSets
```
rudle: I don't understand the rationale for this change. To me, pointers are scary because it's more likely to lead to uncaught runtime errors that crash the program. Can you point to some code elsewhere that benefits from this change? I see only additional error checks in my preliminary scan of this code.

pkg/ds/farm.go
```diff
@@ -0,0 +1,342 @@
...
+				if !ok {
+					// child error channel closed
+					return
+				}
+				dsf.logger.Fatalf("An fatal error has occured in ds: '%v':, %v", child.ds, err)
+			default:
+			}
+		}
+
+		err := dsf.handleDSChanges(quitCh, changes)
```
rudle: Should we copy/pass the value of the `changes` pointer here instead? This code will pass a copy of the pointer. We could avoid the copy on line 143 in that case.

pkg/ds/farm.go
```diff
@@ -0,0 +1,342 @@
...
+
+func (dsf *Farm) handleDSChanges(quitCh <-chan struct{}, changes *dsstore.WatchedDaemonSets) error {
+	if changes.Err != nil {
+		dsf.logger.Infof("An error has occrred while watching daemon sets: %v", changes.Err)
+	}
+
+	if len(changes.Created) > 0 {
+		dsf.logger.Infof("The following daemon sets have been created:")
+		for _, dsFields := range changes.Created {
+			dsf.logger.Infof("%v", *dsFields)
```
rudle: move this into the symmetric loop below

pkg/ds/farm.go
```diff
@@ -0,0 +1,342 @@
...
+			dsIDContended, isContended, err := dsf.dsContends(dsFields)
+			if err != nil {
+				return util.Errorf("Fatal error occured when trying to check for daemon set contention: %v", err)
+			}
+
+			if isContended {
+				dsf.logger.Errorf("Created daemon set '%s' contends with %s", dsFields.ID, dsIDContended)
+				newDS, err := dsf.dsStore.Disable(dsFields.ID)
+				if err != nil {
+					return util.Errorf("Fatal error occured when trying to delete daemon set: %v", err)
```
rudle: I think that this means any contended DaemonSet can stop the progress of all other DaemonSets. Would a `continue` statement be more robust?

pkg/ds/daemon_set.go
```diff
@@ -25,6 +25,14 @@ const (
...
 
+	IsDisabled() bool
+
+	PodID() types.PodID
+
+	GetNodeSelector() klabels.Selector
+
+	EligibleNodes() ([]types.NodeName, error)
+
 	WatchDesires(
```
mpuncel: nit: usually we write comments documenting the functions on an interface, at least the non-obvious ones. Probably `EligibleNodes()` and `WatchDesires()` at least are good candidates for that

pkg/ds/farm.go
```diff
@@ -0,0 +1,358 @@
...
+
+		for {
+			select {
+			case <-quitCh:
+				return
+			case <-timer.C:
+			}
+			timer.Reset(cleanupInterval)
+
+			allDaemonSets, err := dsf.dsStore.List()
```
mpuncel: check `err` here

pkg/ds/farm.go
```diff
@@ -0,0 +1,358 @@
...
+			dsIDMap := make(map[string]ds_fields.DaemonSet)
```
mpuncel: if you make this `map[fields.ID]` instead of `map[string]` then you don't need to do the `.String()` cast below and you're benefitting from the type safety of having that custom type in the first place

pkg/ds/farm.go
```diff
@@ -0,0 +1,358 @@
...
+			for _, dsFields := range allDaemonSets {
+				dsIDMap[dsFields.ID.String()] = dsFields
+			}
+
+			allPods, err := dsf.applicator.GetMatches(klabels.Everything(), labels.POD)
+			if err != nil {
+				dsf.logger.NoFields().Fatal(err)
```
mpuncel: I think we want `Error` instead of `Fatal` here. `Fatal` calls `os.Exit(1)` which will kill the whole farm. Considering that this is inside of a `for {}` loop i'd think you should log the error and then `continue`

pkg/ds/farm.go
```diff
@@ -0,0 +1,358 @@
...
+				return
```
mpuncel: `continue` instead of `return` so you can recover from transient errors (reset the timer first though to avoid a busy loop)

mpuncel: nvm you already reset the timer, just `continue`

pkg/ds/farm.go
```diff
@@ -0,0 +1,358 @@
...
+			}
+
+			for _, podLabels := range allPods {
+				// Only check if it is a pod scheduled by a daemon set
+				if podLabels.Labels.Has(DSIDLabel) {
```
mpuncel: if you're doing this label check on every pod, why aren't you passing the appropriate label selector to the `GetMatches()` call instead?

mpuncel: the `ExistsOperator` will be useful here

pkg/ds/farm.go
```diff
@@ -0,0 +1,358 @@
...
+					dsID := podLabels.Labels.Get(DSIDLabel)
+
+					// Check if the daemon set exists, if it doesn't unschedule the pod
+					if _, ok := dsIDMap[dsID]; !ok {
+						nodeName, podID, err := labels.NodeAndPodIDFromPodLabel(podLabels)
+						if err != nil {
+							dsf.logger.NoFields().Fatal(err)
+							return
```
mpuncel: change this to `Error()` and `continue` as well

pkg/ds/farm.go
mpuncel: you should at least have a `continue` here, we don't want to remove the label if the pod deletion failed

pkg/ds/farm.go
```diff
@@ -0,0 +1,358 @@
...
+
+						_, err = dsf.kpStore.DeletePod(kp.INTENT_TREE, nodeName, podID)
+						if err != nil {
+							dsf.logger.NoFields().Errorf("Unable to delete pod from intent tree: %v", err)
+						}
+
+						id := labels.MakePodLabelKey(nodeName, podID)
+						err = dsf.applicator.RemoveLabel(labels.POD, id, DSIDLabel)
+						if err != nil {
+							dsf.logger.NoFields().Errorf("Error removing label: %v", err)
```
mpuncel: should probably say the pod id the label couldn't be removed for as well

pkg/ds/farm.go
```diff
@@ -0,0 +1,358 @@
...
+			if !ok {
+				return
+			}
+			if changes == nil {
+				dsf.logger.Error("Unexpected nil value received on channel")
+				return
+			}
+		}
+
+		for _, child := range dsf.children {
```
mpuncel: Mind putting a comment explaining the rationale behind this loop? looks like all it does is scan all of the children for closed error channels and crashes the farm if any of them are closed. I'm curious under what circumstances you'd expect one to be closed, and why it's necessary to guarantee they're all open.

pkg/ds/farm.go
```diff
@@ -0,0 +1,358 @@
...
+			select {
+			case <-quitCh:
+				return
+			case err, ok = <-child.errCh:
+				if !ok {
+					// child error channel closed
+					return
+				}
+				dsf.logger.Fatalf("An fatal error has occurred in ds: '%v':, %v", child.ds, err)
```
mpuncel: does this need to be a `Fatal`? that will exit the entire farm. I don't entirely understand what it means for the errCh to be closed

pkg/ds/farm.go
```diff
@@ -0,0 +1,358 @@
...
+			default:
+			}
+		}
+		dsf.handleDSChanges(quitCh, changes)
+	}
+}
+
+func (dsf *Farm) handleDSChanges(quitCh <-chan struct{}, changes *dsstore.WatchedDaemonSets) {
+	if changes.Err != nil {
+		dsf.logger.Infof("An error has occurred while watching daemon sets: %v", changes.Err)
```
mpuncel: you should probably return here? doesn't make sense to look at the rest of `changes` if there was an error right?

pkg/ds/farm.go
```diff
@@ -0,0 +1,358 @@
...
+			dsf.logger.Infof("%v", *dsFields)
+			if child, ok := dsf.children[dsFields.ID]; ok {
+				child.deletedCh <- dsFields
+			}
+		}
+	}
+}
+
+// Naive implementation of a guard, this checks if any of the scheduled nodes
+// are used two daemon sets, does not pre-emptively catch overlaps by selector
```
mpuncel: i think you mean "are used by two daemon sets"

pkg/ds/farm.go
```diff
@@ -0,0 +1,358 @@
...
+	// If this daemon set has a node selector set to Everything, check the labels
+	// of other daemon sets
+	for _, child := range dsf.children {
+		everythingSelector := klabels.Everything().String()
+		if !child.ds.IsDisabled() && child.ds.PodID() == dsFields.PodID && child.ds.ID() != dsFields.ID {
+			// Naively check if both selectors are the Everything selector plus
+			// something else or if both selectors are the same
+			//
+			// It is possible that both selectors won't contend if they both have
+			// something like { az = zone_one , az != zone_one } and that this
```
mpuncel: it looks to me like this comment isn't true, this is just checking for klabels.Everything() which I believe is represented with an empty string

mpuncel: printing klabels.Everything() and a conficting selector gives me: ```''
'foo=bar,foo!=bar'```

pkg/ds/farm.go
```diff
@@ -0,0 +1,358 @@
...
+			if dsFields.NodeSelector.String() == child.ds.GetNodeSelector().String() {
+				dsf.raiseContentionAlert(child.ds.ID(), dsFields.ID)
+				return child.ds.ID(), true, nil
+			}
+		}
+	}
+
+	for _, child := range dsf.children {
+		// Check child if it is not disabled and that if has the same PodID
+		if !child.ds.IsDisabled() && child.ds.PodID() == dsFields.PodID && child.ds.ID() != dsFields.ID {
```
mpuncel: this condition looks the same as in the other `for` loop, you could probably move the checks into one loop to make this function more readable

pkg/ds/farm.go
```diff
@@ -0,0 +1,358 @@
...
+			}
+		}
+	}
+
+	if len(changes.Deleted) > 0 {
+		dsf.logger.Infof("The following daemon sets have been deleted:")
+		for _, dsFields := range changes.Deleted {
+			dsf.logger.Infof("%v", *dsFields)
+			if child, ok := dsf.children[dsFields.ID]; ok {
+				child.deletedCh <- dsFields
```
mpuncel: you probably want to `delete(dsf.children, dsFields.ID)` here as well

mpuncel: also, you should probably close the channel along with child.updatedCh, and child.quitCh

mpuncel: also, select on `quitCh`

pkg/ds/farm.go
```diff
@@ -0,0 +1,358 @@
...
+	if alertErr := dsf.alerter.Alert(alerting.AlertInfo{
+		Description: fmt.Sprintf("New ds '%v', contends with '%v'", newDSID, oldDSID),
+		IncidentKey: "preemptive_ds_contention",
+	}); alertErr != nil {
+		dsf.logger.WithError(alertErr).Errorln("Unable to deliver alert!")
+	}
+}
+
+// Creates a functioning daemon set that will watch and write to the pod tree
+func (dsf *Farm) spawnDaemonSet(dsFields *ds_fields.DaemonSet, quitCh <-chan struct{}) *childDS {
```
mpuncel: `quitCh` appears unused, perhaps you want to pass that to `childDS` instead of making a new one

pkg/ds/farm.go
KoishiKomeiji: Updated to avoid the copy

pkg/ds/farm.go
KoishiKomeiji: Updated to use a continue

pkg/ds/farm.go
KoishiKomeiji: The loop will return when the `quitCh` is closed by the caller to `mainLoop()` .
This has been updated to remove the child and stop its process when it gets an error and continue so it doesn't kill the farm.

pkg/ds/farm.go
KoishiKomeiji: Updated and removed the parameter to the function. Still making a new quit channel.

pkg/ds/farm.go
KoishiKomeiji: Updated.

pkg/ds/farm.go
KoishiKomeiji: Updated.

pkg/ds/farm.go
KoishiKomeiji: Updated, using a removing child from map followed by a continue now, `errCh` is closed when the `WatchDesires` function returns, so keeping the child in the map will not be necessary now.

pkg/ds/farm.go
KoishiKomeiji: Updated.

pkg/ds/farm.go
KoishiKomeiji: Updated

pkg/ds/daemon_set_test.go
KoishiKomeiji: Updated to avoid using a pointer.

KoishiKomeiji:
Updated and comments have been addressed now.

pkg/ds/farm.go
```diff
@@ -0,0 +1,401 @@
...
+	}
+
+	if len(changes.Deleted) > 0 {
+		dsf.logger.Infof("The following daemon sets have been deleted:")
+		for _, dsFields := range changes.Deleted {
+			dsf.logger.Infof("%v", *dsFields)
+			if child, ok := dsf.children[dsFields.ID]; ok {
+				child.deletedCh <- dsFields
+				select {
+				case err := <-child.errCh:
```
mpuncel: you need to select on `quitCh` here too

pkg/ds/farm.go
mpuncel: you still need to put this in the `select`

KoishiKomeiji: Updated

pkg/ds/farm.go
KoishiKomeiji: Updated

pkg/ds/farm.go
```diff
@@ -0,0 +1,412 @@
...
+						if alertErr := dsf.alerter.Alert(alerting.AlertInfo{
+							Description: fmt.Sprintf("Error occurred when deleting spawned daemon set '%v': %v", dsFields.ID, err),
+							IncidentKey: "child_ds_error",
+						}); alertErr != nil {
+							dsf.logger.WithError(alertErr).Errorln("Unable to deliver alert!")
+						}
+					}
+					dsf.closeChild(dsFields.ID)
+				case child.deletedCh <- dsFields:
+					close(child.quitCh)
```
mpuncel: shouldn't you `dsf.closeChild(dsFields.ID)` here

KoishiKomeiji: Thanks! Updated

pkg/ds/daemon_set.go
```diff
@@ -131,7 +153,7 @@ func (ds *daemonSet) WatchDesires(
 				}
 				ds.logger.NoFields().Infof("Received daemon set update signal: %v", newDS)
 				if newDS == nil {
-					ds.logger.NoFields().Fatal("Unexpected a nil daemon set during update", *ds)
+					ds.logger.NoFields().Error("Unexpected a nil daemon set during update", *ds)
```
rudle: replace this with Errorf

pkg/ds/daemon_set.go
```diff
@@ -161,7 +183,7 @@ func (ds *daemonSet) WatchDesires(
 				}
 				ds.logger.NoFields().Infof("Received daemon set delete signal: %v", deleteDS)
 				if deleteDS == nil {
-					ds.logger.NoFields().Fatal("Unexpected a nil daemon set during delete")
+					ds.logger.NoFields().Error("Unexpected a nil daemon set during delete")
```
rudle: Replace this with Errorf

pkg/kp/dsstore/consul_store.go
```diff
@@ -214,6 +214,31 @@ func (s *consulStore) MutateDS(
...
+func (s *consulStore) Disable(id fields.ID) (fields.DaemonSet, error) {
+	s.logger.Infof("Attempting to disable '%s' in store now", id)
+
+	mutator := func(dsToUpdate fields.DaemonSet) (fields.DaemonSet, error) {
+		dsToUpdate.Disabled = true
+		return dsToUpdate, nil
+	}
+	newDS, err := s.MutateDS(id, mutator)
+
+	// Delete the daemon set because there was an error during mutation
```
rudle: Is Delete() the safest way for this function to fail? I understand the following:

`Disable` has the semantic: "Don't schedule this pod onto any more nodes"
`Delete` has the semantic: "Remove this DS from consul and _unschedule all of its pods_"

If my characterization is correct, it's not at all safe to Delete the DaemonSet. This function should probably do nothing in case of Mutation error.

pkg/ds/daemon_set.go
mpuncel: also get rid of the `a`

pkg/ds/daemon_set.go
```diff
@@ -290,6 +314,14 @@ func (ds *daemonSet) removePods() error {
 	return nil
 }
 
+func (ds *daemonSet) EligibleNodes() ([]types.NodeName, error) {
+	eligible, err := ds.scheduler.EligibleNodes(ds.Manifest, ds.NodeSelector)
+	if err != nil {
+		return eligible, util.Errorf("Error retrieving eligible nodes for daemon set: %v", err)
```
mpuncel: returning `nil` for the first return value would be more conventional

pkg/ds/farm.go
```diff
@@ -0,0 +1,411 @@
...
+	dsf.mainLoop(quitCh)
+}
+
+const cleanupInterval = 60 * time.Second
+
+// This function removes all pods with a DSIDLabel where the daemon set id does
+// not exist in the store at every interval specified because it is possible
+// that the farm will unexpectedly crash or someone deletes or modifies a node
+func (dsf *Farm) cleanupDaemonSetPods(quitCh <-chan struct{}) {
+	go func() {
```
mpuncel: nit: i'd get rid of this `go func () { }` in this function and just do `go dsf.cleanupDaemonSetPods()` in `Start()`. 

Reasoning: it's easy to make a synchronous function asynchronous in Go but the reverse is difficult

pkg/ds/farm.go
```diff
@@ -0,0 +1,411 @@
...
+			select {
+			case <-quitCh:
+				return
+			case <-timer.C:
+			}
+			timer.Reset(cleanupInterval)
+
+			allDaemonSets, err := dsf.dsStore.List()
+			if err != nil {
+				dsf.logger.NoFields().Errorf("Unable to get daemon sets from intent tree: %v", err)
```
mpuncel: I suspect we're going to want some fields on all log messages, such as daemon set ID, pod ID, availability zone, etc. Otherwise it will be hard to determine the impact of these errors

mpuncel: sorry this is a farm i'm confused, ignore

pkg/ds/farm.go
```diff
@@ -0,0 +1,411 @@
...
+			for _, dsFields := range allDaemonSets {
+				dsIDMap[dsFields.ID] = dsFields
+			}
+
+			dsIDLabelSelector := klabels.Everything().
+				Add(DSIDLabel, klabels.ExistsOperator, []string{})
+
+			allPods, err := dsf.applicator.GetMatches(dsIDLabelSelector, labels.POD)
+			if err != nil {
+				dsf.logger.NoFields().Error(err)
```
mpuncel: this log message could use more context

pkg/ds/farm.go
mpuncel: it might be nice to make it clear that these log messages are from a daemon set farm though, given that this process will live with RC and RU farms most likely

pkg/ds/farm.go
```diff
@@ -0,0 +1,411 @@
...
+				continue
+			}
+
+			for _, podLabels := range allPods {
+				// Only check if it is a pod scheduled by a daemon set
+				dsID := podLabels.Labels.Get(DSIDLabel)
+
+				// Check if the daemon set exists, if it doesn't unschedule the pod
+				if _, ok := dsIDMap[fields.ID(dsID)]; !ok {
```
mpuncel: nit: if you rewrote this as `if ok { continue } ` it might read a little better

pkg/ds/farm.go
```diff
@@ -0,0 +1,411 @@
...
+					nodeName, podID, err := labels.NodeAndPodIDFromPodLabel(podLabels)
+					if err != nil {
+						dsf.logger.NoFields().Error(err)
+						continue
+					}
+
+					// TODO: Since this mirrors the unschedule function in daemon_set.go,
+					// We should find a nice way to couple them together
+					dsf.logger.NoFields().Infof("Unscheduling from '%s' with dangling daemon set uuid '%s'", nodeName, dsID)
```
mpuncel: included pod id in this message

pkg/ds/farm.go
```diff
@@ -0,0 +1,411 @@
...
+
+					_, err = dsf.kpStore.DeletePod(kp.INTENT_TREE, nodeName, podID)
+					if err != nil {
+						dsf.logger.NoFields().Errorf("Unable to delete pod id '%v' from intent tree: %v", dsID, err)
```
mpuncel: include node name here

pkg/ds/farm.go
```diff
@@ -0,0 +1,411 @@
...
+
+						if alertErr := dsf.alerter.Alert(alerting.AlertInfo{
+							Description: fmt.Sprintf("Unable to delete the following pod from store: %v", podID),
```
mpuncel: indicate in here that this is a daemon set farm

pkg/ds/farm.go
```diff
@@ -0,0 +1,411 @@
...
+							IncidentKey: "pod_store_delete_error",
```
mpuncel: prefix with `ds_` maybe

mpuncel: also, it's proved very helpful to include a `Details` struct in the RC alerts, pod id, node name, and hostname of the farm instance itself might be useful context

mpuncel: now that i think about it i'm not sure we want to page people if a pod fails to delete though, especially since it gets retried

pkg/ds/farm.go
```diff
@@ -0,0 +1,411 @@
...
+			case <-quitCh:
+				return
+			case err, ok = <-child.errCh:
+				if !ok {
+					// child error channel closed
+					dsf.closeChild(dsID)
+					continue
+				}
+				dsf.logger.Errorf("An error has occurred in spawned ds '%v':, %v", child.ds, err)
+				if alertErr := dsf.alerter.Alert(alerting.AlertInfo{
```
mpuncel: i don't think we want to page people for this either, logging is good enough

pkg/ds/farm.go
```diff
@@ -0,0 +1,411 @@
...
+
+func (dsf *Farm) handleDSChanges(changes dsstore.WatchedDaemonSets, quitCh <-chan struct{}) {
+	if changes.Err != nil {
+		dsf.logger.Infof("An error has occurred while watching daemon sets: %v", changes.Err)
+		return
+	}
+
+	if len(changes.Created) > 0 {
+		dsf.logger.Infof("The following daemon sets have been created:")
+		for _, dsFields := range changes.Created {
```
mpuncel: the daemon set key should be locked somewhere in this function so concurrent farm instances don't try to handle the same daemon set

pkg/ds/farm.go
```diff
@@ -0,0 +1,411 @@
...
+			if err != nil {
+				dsf.logger.Errorf("Error occurred when trying to check for daemon set contention: %v", err)
+				continue
+			}
+
+			if isContended {
+				dsf.logger.Errorf("Created daemon set '%s' contends with %s", dsFields.ID, dsIDContended)
+				newDS, err := dsf.dsStore.Disable(dsFields.ID)
+				if err != nil {
+					dsf.logger.Errorf("Error occurred when trying to delete daemon set: %v", err)
```
mpuncel: s/delete/disable

pkg/ds/farm.go
```diff
@@ -0,0 +1,411 @@
...
+			dsf.logger.Infof("%v", *dsFields)
+			if child, ok := dsf.children[dsFields.ID]; ok {
+				select {
+				case <-quitCh:
+					return
+				case err := <-child.errCh:
+					if err != nil {
+						dsf.logger.Errorf("Error occurred when deleting spawned daemon set '%v': %v", dsFields.ID, err)
+
+						if alertErr := dsf.alerter.Alert(alerting.AlertInfo{
```
mpuncel: don't alert on this

pkg/ds/farm.go
```diff
@@ -0,0 +1,411 @@
...
+				return child.ds.ID(), true, nil
+			}
+		}
+	}
+
+	return "", false, nil
+}
+
+func (dsf *Farm) raiseContentionAlert(oldDSID ds_fields.ID, newDSID ds_fields.ID) {
+	if alertErr := dsf.alerter.Alert(alerting.AlertInfo{
```
mpuncel: i'm okay with this alert for now but we may want to tune this quickly if it becomes a problem.

pkg/ds/farm.go
```diff
@@ -0,0 +1,411 @@
...
+		Description: fmt.Sprintf("New ds '%v', contends with '%v'", newDSID, oldDSID),
+		IncidentKey: "preemptive_ds_contention",
```
mpuncel: consider a `Details` struct with pod id, availability zone, and other information for diagnostics

pkg/ds/farm.go
KoishiKomeiji: Updated, will write tests for it soon.

pkg/ds/farm.go
```diff
@@ -294,6 +349,35 @@ func (dsf *Farm) handleDSChanges(changes dsstore.WatchedDaemonSets, quitCh <-cha
...
+		"ds":  dsFields.ID,
+		"pod": dsFields.Manifest.ID(),
+	})
+}
+
+func (dsf *Farm) makeSessionExpiryAlert(dsFields ds_fields.DaemonSet, dsLogger logging.Logger, err error) {
+	dsLogger.WithError(err).Errorln("Got error while locking daemon set in farm - session may be expired")
+
+	if alertErr := dsf.alerter.Alert(alerting.AlertInfo{
+		Description: fmt.Sprintf("Got error while locking daemon set - session may be expired"),
```
rudle: nit, no need for Sprintf here if you're note going to use a variable.

mpuncel:
looks good! excited to see this in action 🚢

PR #554, opened 2016-07-22
Mpuncel/farm error reporting

mpuncel:


pkg/rc/farm.go
```diff
@@ -65,22 +67,27 @@ func NewFarm(
 	logger logging.Logger,
 	rcSelector klabels.Selector,
 	alerter alerting.Alerter,
+	errorReporter error_reporter.Reporter,
```
mpuncel: alternatively I could make the constructor place a no-op reporter on the struct and then add a SetErrorReporter function. that might be a little nicer than spreading `nil` everywhere

rudle: I don't see any precedent for that style of injection. I don't mind the extra nil in constructors fwiw

rudle:
👍

pkg/rc/farm.go
mpuncel: i did do that for metrics registry in the farms I think, but i also don't mind the `nil`

PR #556, opened 2016-07-25
Add p2-cgroup-info command

bwester:
Add a new p2-cgroup-info command that displays information about which
cgroups contain running P2 launchables.


mpuncel:
👍

PR #558, opened 2016-07-26
Add support for the "entry_points" field in launchable stanza

mpuncel:
This field is only used by hoist launchables, and enumerates the path to
a file or directory of files that should be used as entry points for
runit services. If there are no entry points enumerated, a default value
of bin/launch is used which matches the behavior of all hoist
launchables prior to this commit.

A constraint is enforced that no two runit services may have the same
name, which translates to entry points necessarily having differening
file names. For example, a manifest specifying ["bin/launch",
"bin/start"] for entry point is invalid if bin/start and
bin/launch/start are files in the tarball because those would result in
runit services with the same name.


rudle:
This looks good to me! What have you to test this? Because I'm lazy, I wouldn't mind seeing the full runit subtree for a service with multiple entry points. In particular, I'm wondering how the log agent scripts work.

PR #560, opened 2016-07-28
Do not block main update loop if individual pod is stuck.

anthonybishopric:
In cases where a pod's launch or shutdown time is very long, we
do not want this wait time to block other ongoing deploys. While
pods own their update channel, it is possible for the primary
intent watching channel to get stuck trying to notify a pod routine
of a new manifest.

Another fix is required to time out long or stuck installations during
a self-update or graceful shutdown scenario, present at orchestrate.go:133
but achieving this safely is less clear.


rudle:
A regression test would be suitable in a follow up PR.

👍

PR #561, opened 2016-07-28
Enable replication in ds farm

KoishiKomeiji:
This will allow you to run ./bin/p2-ds-farm to call `replication` to schedule nodes

### DIFF
--- a/bin/p2-ds-farm/main.go
+++ b/bin/p2-ds-farm/main.go
@@ -1,0 +1,67 @@
+package main
+
+import (
+	"fmt"
+	"os"
+	"os/signal"
+	"time"
+
+	"github.com/Sirupsen/logrus"
+	"github.com/hashicorp/consul/api"
+	"github.com/square/p2/pkg/health/checker"
+	"github.com/square/p2/pkg/kp"
+	"github.com/square/p2/pkg/kp/consulutil"
+	"github.com/square/p2/pkg/kp/dsstore"
+	"github.com/square/p2/pkg/kp/flags"
+	"github.com/square/p2/pkg/labels"
+	"github.com/square/p2/pkg/logging"
+	"github.com/square/p2/pkg/util/stream"
+
+	ds_farm "github.com/square/p2/pkg/ds"
+)
+
+// SessionName returns a node identifier for use when creating Consul sessions.
+func SessionName() string {
+	hostname, err := os.Hostname()
+	if err != nil {
+		hostname = "unknown hostname"
+	}
+	// Current time of "Jan 2, 2006, 15:04:05" turns into "2006-01-02-15-04-05"
+	timeStr := time.Now().Format("2006-01-02-15-04-05")
+	return fmt.Sprintf("p2-ds-farm:%s:%s", hostname, timeStr)
+}
+
+func main() {
+	_, consulOpts := flags.ParseWithConsulOptions()
+	client := kp.NewConsulClient(consulOpts)
+	logger := logging.NewLogger(logrus.Fields{})
+	dsStore := dsstore.NewConsul(client, 3, &logger)
+	kpStore := kp.NewConsulStore(client)
+	applicator := labels.NewConsulApplicator(client, 3)
+	healthChecker := checker.NewConsulHealthChecker(client)
+
+	sessions := make(chan string)
+	go consulutil.SessionManager(api.SessionEntry{
+		Name:      SessionName(),
+		LockDelay: 5 * time.Second,
+		Behavior:  api.SessionBehaviorDelete,
+		TTL:       "15s",
+	}, client, sessions, nil, logger)
+	pub := stream.NewStringValuePublisher(sessions, "")
+	rcSub := pub.Subscribe(nil)
+
+	dsf := ds_farm.NewFarm(kpStore, dsStore, applicator, rcSub.Chan(), logger, nil, &healthChecker)
+
+	quitCh := make(chan struct{})
+
+	go func() {
+		// clear lock immediately on ctrl-C
+		signals := make(chan os.Signal, 1)
+		signal.Notify(signals, os.Interrupt)
+		<-signals
+		close(quitCh)
+		os.Exit(1)
+	}()
+
+	dsf.Start(quitCh)
+}
--- a/pkg/ds/daemon_set.go
+++ b/pkg/ds/daemon_set.go
@@ -1,8 +1,15 @@
 package ds
 
 import (
+	"fmt"
+	"os"
+	"os/user"
+	"sync"
 	"time"
 
+	"github.com/square/p2/pkg/health"
+	"github.com/square/p2/pkg/health/checker"
+	"github.com/square/p2/pkg/replication"
 	"github.com/square/p2/pkg/util"
 
 	klabels "k8s.io/kubernetes/pkg/labels"
@@ -12,7 +19,6 @@ import (
 	"github.com/square/p2/pkg/kp/dsstore"
 	"github.com/square/p2/pkg/labels"
 	"github.com/square/p2/pkg/logging"
-	"github.com/square/p2/pkg/manifest"
 	"github.com/square/p2/pkg/scheduler"
 	"github.com/square/p2/pkg/types"
 )
@@ -54,49 +60,53 @@ type DaemonSet interface {
 
 	// CurrentPods() returns all nodes that are scheduled by this daemon set
 	CurrentPods() (types.PodLocations, error)
-}
 
-// These methods are the same as the methods of the same name in kp.Store.
-// This interface makes writing unit tests easier as daemon sets do not
-// use any functions not listed here
-type kpStore interface {
-	SetPod(
-		podPrefix kp.PodPrefix,
-		nodeName types.NodeName,
-		manifest manifest.Manifest,
-	) (time.Duration, error)
-
-	DeletePod(podPrefix kp.PodPrefix,
-		nodeName types.NodeName,
-		manifestID types.PodID,
-	) (time.Duration, error)
+	// Schedules pods by using replication, this will automatically
+	// cancel any current replications and re-enact
+	PublishToReplication() error
 }
 
 type daemonSet struct {
 	fields.DaemonSet
 
-	logger     logging.Logger
-	kpStore    kpStore
-	scheduler  scheduler.Scheduler
-	dsStore    dsstore.Store
-	applicator labels.Applicator
+	contention    dsContention
+	logger        logging.Logger
+	kpStore       kp.Store
+	scheduler     scheduler.Scheduler
+	dsStore       dsstore.Store
+	applicator    labels.Applicator
+	healthChecker *checker.ConsulHealthChecker
+
+	// This is the current replication enact go routine that is running
+	currentReplication replication.Replication
+
+	// This locks are used to make sure only one replication is going on at the same time
+	replicationLock sync.Mutex
+}
+
+type dsContention struct {
+	contendedWith fields.ID
+	isContended   bool
 }
 
 func New(
 	fields fields.DaemonSet,
 	dsStore dsstore.Store,
-	kpStore kpStore,
+	kpStore kp.Store,
 	applicator labels.Applicator,
 	logger logging.Logger,
+	healthChecker *checker.ConsulHealthChecker,
 ) DaemonSet {
 	return &daemonSet{
 		DaemonSet: fields,
 
-		dsStore:    dsStore,
-		kpStore:    kpStore,
-		logger:     logger,
-		applicator: applicator,
-		scheduler:  scheduler.NewApplicatorScheduler(applicator),
+		dsStore:            dsStore,
+		kpStore:            kpStore,
+		logger:             logger,
+		applicator:         applicator,
+		scheduler:          scheduler.NewApplicatorScheduler(applicator),
+		healthChecker:      healthChecker,
+		currentReplication: nil,
 	}
 }
 
@@ -136,6 +146,7 @@ func (ds *daemonSet) WatchDesires(
 	go func() {
 		var err error
 		defer close(errCh)
+		defer ds.CancelReplication()
 
 		// Try to schedule pods when this begins watching
 		if !ds.Disabled {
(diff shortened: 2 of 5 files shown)


bin/p2-ds-farm/main.go
```diff
@@ -0,0 +1,56 @@
...
+
+// SessionName returns a node identifier for use when creating Consul sessions.
+func SessionName() string {
+	hostname, err := os.Hostname()
+	if err != nil {
+		hostname = "unknown hostname"
+	}
+	// Current time of "Jan 2, 2006, 15:04:05" turns into "2006-01-02-15-04-05"
+	timeStr := time.Now().Format("2006-01-02-15-04-05")
+	return fmt.Sprintf("p2-rctl-server:%s:%s", hostname, timeStr)
```
mpuncel: i think you mean `p2-ds-farm`

KoishiKomeiji: Oops thanks!

pkg/ds/daemon_set.go
```diff
@@ -54,49 +62,49 @@ type DaemonSet interface {
...
-		podPrefix kp.PodPrefix,
-		nodeName types.NodeName,
-		manifest manifest.Manifest,
-	) (time.Duration, error)
-
-	DeletePod(podPrefix kp.PodPrefix,
-		nodeName types.NodeName,
-		manifestID types.PodID,
-	) (time.Duration, error)
+	PublishToReplication() error
```
mpuncel: a comment explaining this would be 👍

pkg/ds/daemon_set.go
```diff
@@ -293,6 +303,14 @@ func (ds *daemonSet) addPods() error {
 			return util.Errorf("Error scheduling node: %v", err)
 		}
 	}
+
+	if len(toScheduleSorted) > 0 {
+		err = ds.PublishToReplication()
```
mpuncel: `return ds.PublishToReplication()`

pkg/ds/daemon_set.go
```diff
@@ -315,12 +333,24 @@ func (ds *daemonSet) removePods() error {
...
+
 	for _, node := range toUnscheduleSorted {
 		err := ds.unschedule(node)
 		if err != nil {
 			return util.Errorf("Error unscheduling node: %v", err)
 		}
 	}
+
+	if len(podLocations)-len(toUnscheduleSorted) > 0 {
+		err = ds.PublishToReplication()
```
mpuncel: `return ds.PublishToReplication()`

pkg/ds/daemon_set.go
```diff
@@ -360,14 +395,6 @@ func (ds *daemonSet) schedule(node types.NodeName) error {
 	if err != nil {
 		return util.Errorf("Error setting label: %v", err)
 	}
-
-	// Will set the following the value, ds.Manifest, to the following key:
-	// <kp.INTENT_TREE>/<node>/<ds.Manifest.ID()>
-	// eg intent/127.0.0.1/test = test_pod
-	_, err = ds.kpStore.SetPod(kp.INTENT_TREE, node, ds.Manifest)
```
mpuncel: did you mean to remove this? now the function only sets a label and does not set the pod?

pkg/ds/daemon_set.go
```diff
@@ -392,6 +413,111 @@ func (ds *daemonSet) unschedule(node types.NodeName) error {
...
+	lockMessage := fmt.Sprintf("%q from %q at %q", thisUser.Username, thisHost, time.Now())
+	repl, err := replication.NewReplicator(
+		ds.DaemonSet.Manifest,
+		ds.logger,
+		nodes,
+		len(nodes)-ds.DaemonSet.MinHealth,
+		ds.kpStore,
+		ds.applicator,
+		*ds.healthChecker,
+		health.HealthState("warning"),
```
mpuncel: why are you using warning and not passing? also, it's preferable to use a constant like `health.Passing` so that typos are harder to introduce

pkg/ds/daemon_set.go
KoishiKomeiji: Yes, the `PublishToReplication()` will be setting this for us in its `Enact()`, but the labels will be used to keep track of which nodes are handled by which daemon sets.

pkg/ds/daemon_set.go
```diff
@@ -392,6 +413,111 @@ func (ds *daemonSet) unschedule(node types.NodeName) error {
...
+		lockMessage,
+	)
+	if err != nil {
+		for _, node := range nodes {
+			ds.unschedule(node)
```
mpuncel: I don't think you want to wipe out all of the pods if the replication fails, just bail out

pkg/ds/daemon_set.go
```diff
@@ -392,6 +413,111 @@ func (ds *daemonSet) unschedule(node types.NodeName) error {
...
+		}
+		log.Fatalf("Unable to initialize replication, pods have been unscheduled: %s", err)
+	}
+
+	ds.logger.Info("Replication initialized")
+
+	// auto-drain this channel
+	go func() {
+		for range errCh {
```
mpuncel: log these probably

pkg/ds/daemon_set.go
```diff
@@ -392,6 +413,111 @@ func (ds *daemonSet) unschedule(node types.NodeName) error {
...
+	ds.logger.Info("New replicator was made")
+
+	replication, errCh, err := repl.InitializeReplication(
+		false,
+		false,
+		replication.DefaultConcurrentReality,
+	)
+	if err != nil {
+		for _, node := range nodes {
+			ds.unschedule(node)
```
mpuncel: same, don't unschedule them just halt

pkg/ds/daemon_set.go
```diff
@@ -392,6 +413,111 @@ func (ds *daemonSet) unschedule(node types.NodeName) error {
...
+		}
+		log.Fatalf("Unable to initialize replication, pods have been unscheduled: %s", err)
```
mpuncel: don't use `Fatal` because it will kill the whole server

pkg/ds/daemon_set.go
```diff
@@ -392,6 +413,111 @@ func (ds *daemonSet) unschedule(node types.NodeName) error {
...
+		ds.applicator,
+		*ds.healthChecker,
+		health.HealthState("warning"),
+		lockMessage,
+	)
+	if err != nil {
+		for _, node := range nodes {
+			ds.unschedule(node)
+		}
+		log.Fatalf("Could not initialize replicator, pods have been unscheduled: %s", err)
```
mpuncel: use `Errorf` not `Fatalf`

pkg/ds/daemon_set.go
mpuncel: i'd suggest renaming this function to `labelPod` or something like that then

pkg/ds/daemon_set.go
KoishiKomeiji: Updated

pkg/ds/daemon_set.go
```diff
@@ -392,6 +413,105 @@ func (ds *daemonSet) unschedule(node types.NodeName) error {
...
+			ds.replicationOngoing = false
+		}
+		// Set a new replication
+		ds.currentReplication = replication
+		ds.replicationOngoing = true
+
+		ds.replicationLock.Unlock()
+
+		// This function is blocking, so we can't defer the unlocks
+		replication.Enact()
```
mpuncel: do we want to do this inside of a goroutine? is it possible to end up with multiple of these running or are we depending on the locking mechanism

pkg/ds/daemon_set.go
```diff
@@ -54,49 +62,51 @@ type DaemonSet interface {
...
-	applicator labels.Applicator
+	contention    dsContention
+	logger        logging.Logger
+	kpStore       kp.Store
+	scheduler     scheduler.Scheduler
+	dsStore       dsstore.Store
+	applicator    labels.Applicator
+	healthChecker *checker.ConsulHealthChecker
+
+	currentReplication replication.Replication
```
mpuncel: comments on these variables would be useful, particularly that `replicationLock` is used to synchronize access to `currentReplication and `replicationOngoing`

pkg/ds/daemon_set.go
```diff
@@ -392,6 +413,105 @@ func (ds *daemonSet) unschedule(node types.NodeName) error {
...
+	ds.logger.Info("Replication enacted")
+
+	return nil
+}
+
+func (ds *daemonSet) CancelReplication() {
+	ds.replicationLock.Lock()
+	defer ds.replicationLock.Unlock()
+
+	if ds.replicationOngoing == true {
```
mpuncel: i wonder if `ds.currentReplication != nil` would accomplish the same thing without using a `bool`, but not a big deal if that adds too much work to test

pkg/ds/daemon_set.go
```diff
@@ -392,6 +413,105 @@ func (ds *daemonSet) unschedule(node types.NodeName) error {
 	return nil
 }
 
+func (ds *daemonSet) PublishToReplication() error {
+	// TODO: We need to specifically turn this on to work
+	if ds.healthChecker == nil {
```
rudle: Is it time to change this?

pkg/ds/daemon_set.go
```diff
@@ -293,6 +302,11 @@ func (ds *daemonSet) addPods() error {
 			return util.Errorf("Error scheduling node: %v", err)
```
mpuncel: now that the scheduling is being done in replication, you should rename this error message to "labeling" or someting along those lines (and rename `schedule()` to label)

pkg/ds/daemon_set.go
```diff
@@ -392,6 +413,105 @@ func (ds *daemonSet) unschedule(node types.NodeName) error {
...
+	if err != nil {
+		return util.Errorf("Error retrieving pod locations from daemon set: %v", err)
+	}
+	nodes := podLocations.Nodes()
+
+	ds.logger.Infof("Preparing to publish the following nodes: %v", nodes)
+
+	thisHost, err := os.Hostname()
+	if err != nil {
+		log.Fatalf("Could not retrieve hostname: %s", err)
```
rudle: this will crash the whole server. log.Errorf and `return err` instead

pkg/ds/daemon_set.go
```diff
@@ -315,12 +329,21 @@ func (ds *daemonSet) removePods() error {
...
+
 	for _, node := range toUnscheduleSorted {
 		err := ds.unschedule(node)
 		if err != nil {
 			return util.Errorf("Error unscheduling node: %v", err)
 		}
 	}
+
+	if len(podLocations)-len(toUnscheduleSorted) > 0 {
+		return ds.PublishToReplication()
```
mpuncel: seems odd to do this in the `removePods()` function. the `addPods()` function should handle this for you right?

pkg/ds/daemon_set.go
```diff
@@ -339,12 +362,17 @@ func (ds *daemonSet) clearPods() error {
 	toUnscheduleSorted := types.NewNodeSet(currentNodes...).ListNodes()
 	ds.logger.NoFields().Infof("Need to unschedule %d nodes", len(toUnscheduleSorted))
 
+	if len(toUnscheduleSorted) > 0 {
+		ds.CancelReplication()
```
mpuncel: you can call this unconditionally here, the daemon set is being deleted and `CancelReplication` protects itself against there not being a replication in progress

pkg/ds/daemon_set.go
```diff
@@ -392,6 +413,105 @@ func (ds *daemonSet) unschedule(node types.NodeName) error {
...
+	ds.logger.Info("Replication initialized")
+
+	// auto-drain this channel
+	go func() {
+		for range errCh {
+		}
+	}()
+
+	go func() {
+		// clear lock immediately on ctrl-C
```
rudle: This signal trap is useful in bin/p2-ds-farm but won't be useful when this is running in our server. Either: localize it to the code that's in `bin/` or remove it

pkg/ds/daemon_set.go
KoishiKomeiji: I'm keeping this in here for now so that the tests work because I haven't faked the health yet.

pkg/ds/daemon_set.go
KoishiKomeiji: Updated

pkg/ds/daemon_set.go
KoishiKomeiji: When I cancel the replication a few lines above, everything will be canceled and we won't know what got scheduled and what didn't by replication.  It is possible that we only need to unschedule one pod but then `ds.CancelReplication()` will halt all scheduling.
So it's possible that not all pods have been scheduled and we will need to reschedule this.

mpuncel: right, but `addPods()` gets called right after `removePods()` in `WatchDesires()` so you shouldn't need it here right?

KoishiKomeiji: Right, but `handleNodeChanges()` behaves a little differently when a node specifically gets deleted.

pkg/ds/daemon_set.go
KoishiKomeiji: Updated.

pkg/ds/daemon_set.go
mpuncel: i'm not sure I understand but i'll defer to you since you've been the one implementing it

bin/p2-ds-farm/main.go
```diff
@@ -0,0 +1,66 @@
...
+	}, client, sessions, nil, logger)
+	pub := stream.NewStringValuePublisher(sessions, "")
+	rcSub := pub.Subscribe(nil)
+
+	dsf := ds_farm.NewFarm(kpStore, dsStore, applicator, rcSub.Chan(), logger, nil, &healthChecker)
+
+	quitCh := make(chan struct{})
+	dsf.Start(quitCh)
+
+	go func() {
```
mpuncel: i think you need to do this before the `dsf.Start()` since it blocks

mpuncel:
looks good to me once the final comment about the interrupt capture is addressed 👍

KoishiKomeiji:
Thanks for the review!

PR #562, opened 2016-07-29
Modify watch labels diff to notice the initial state

KoishiKomeiji:
Modify labels to notice the initial state of the labels store
    
Also removed an unnecessary nil value check on the values because  a check for the closed channel is already made

How this notices the intial state of the labels store is that by not
taking the initial state, the diff loop will believe that any initial
labels in the store are created and they will be signalled


mpuncel:
🚢

KoishiKomeiji:
Thanks!

PR #563, opened 2016-07-29
Hzeng.ds replication

KoishiKomeiji:
This is to add in tests in the daemon set farm for to test for multiple farms running

### DIFF
--- a/pkg/ds/farm.go
+++ b/pkg/ds/farm.go
@@ -238,21 +238,26 @@ func (dsf *Farm) handleDSChanges(changes dsstore.WatchedDaemonSets, quitCh <-cha
 		for _, dsFields := range changes.Created {
 			dsf.logger.Infof("%v", *dsFields)
 
+			var dsUnlocker consulutil.Unlocker
+			var err error
 			dsLogger := dsf.makeDSLogger(*dsFields)
-			dsUnlocker, err := dsf.dsStore.LockForOwnership(dsFields.ID, dsf.session)
 
-			if _, ok := err.(consulutil.AlreadyLockedError); ok {
-				// Lock was either already acquired by another farm or it was Acquired
-				// by this farm
-				if _, ok := dsf.children[dsFields.ID]; !ok {
-					dsLogger.Debugln("Lock on daemon set was already acquired by another farm")
+			// If it is not in our map, then try to acquire the lock
+			if _, ok := dsf.children[dsFields.ID]; !ok {
+				dsUnlocker, err = dsf.dsStore.LockForOwnership(dsFields.ID, dsf.session)
+				if _, ok := err.(consulutil.AlreadyLockedError); ok {
+					// Lock was either already acquired by another farm or it was Acquired
+					// by this farm
+					dsf.logger.Infof("Lock on daemon set '%v' was already acquired by another farm", dsFields.ID)
 					continue
+
+				} else if err != nil {
+					// The session probably either expired or there was probably a network
+					// error, so the rest will probably fail
+					dsf.makeSessionExpiryAlert(*dsFields, dsLogger, err)
+					return
 				}
-			} else if err != nil {
-				// The session probably either expired or there was probably a network
-				// error, so the rest will probably fail
-				dsf.makeSessionExpiryAlert(*dsFields, dsLogger, err)
-				return
+				dsf.logger.Infof("Lock on daemon set '%v' acquired", dsFields.ID)
 			}
 
 			// If the daemon set contends with another daemon set, disable it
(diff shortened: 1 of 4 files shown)


pkg/ds/farm.go
```diff
@@ -245,15 +245,18 @@ func (dsf *Farm) handleDSChanges(changes dsstore.WatchedDaemonSets, quitCh <-cha
 				// Lock was either already acquired by another farm or it was Acquired
 				// by this farm
 				if _, ok := dsf.children[dsFields.ID]; !ok {
-					dsLogger.Debugln("Lock on daemon set was already acquired by another farm")
+					dsf.logger.Infof("Lock on daemon set '%v' was already acquired by another farm", dsFields.ID)
 					continue
 				}
+				dsf.logger.Info("Lock on daemon set '%v' was already acquired by this farm", dsFields.ID)
```
mpuncel: you won't get an err from `LockForOwnership` if your session holds the lock

pkg/ds/farm.go
```diff
@@ -286,13 +289,16 @@ func (dsf *Farm) handleDSChanges(changes dsstore.WatchedDaemonSets, quitCh <-cha
 
 			if _, ok := err.(consulutil.AlreadyLockedError); ok {
 				if _, ok := dsf.children[dsFields.ID]; !ok {
-					dsLogger.Debugln("Lock on daemon set was already acquired by another farm")
+					dsf.logger.Infof("Lock on daemon set '%v' was already acquired by another farm", dsFields.ID)
 					continue
 				}
+				dsf.logger.Info("Lock on daemon set '%v' was already acquired by this farm", dsFields.ID)
```
mpuncel: same

pkg/ds/farm.go
```diff
@@ -245,15 +245,18 @@ func (dsf *Farm) handleDSChanges(changes dsstore.WatchedDaemonSets, quitCh <-cha
 				// Lock was either already acquired by another farm or it was Acquired
 				// by this farm
 				if _, ok := dsf.children[dsFields.ID]; !ok {
```
mpuncel: you don't even need this check if everything is working properly

pkg/ds/farm.go
```diff
@@ -286,13 +289,16 @@ func (dsf *Farm) handleDSChanges(changes dsstore.WatchedDaemonSets, quitCh <-cha
 
 			if _, ok := err.(consulutil.AlreadyLockedError); ok {
 				if _, ok := dsf.children[dsFields.ID]; !ok {
```
mpuncel: same here re: don't need the check if things are working properly. In fact, if you _did_ have that daemon set as a child it would be prudent to remove it if another farm holds it

pkg/ds/farm.go
```diff
@@ -306,9 +312,12 @@ func (dsf *Farm) handleDSChanges(changes dsstore.WatchedDaemonSets, quitCh <-cha
 					dsf.logger.Errorf("Updated daemon set '%s' contends with %s", dsFields.ID, dsIDContended)
 					newDS, err := dsf.dsStore.Disable(dsFields.ID)
 					if err != nil {
-						dsf.logger.Errorf("Error occurred when trying to delete daemon set: %v", err)
+						dsf.logger.Errorf("Error occurred when trying to disable daemon set: %v", err)
 						continue
 					}
+					if dsFields.ID != newDS.ID {
```
mpuncel: have you seen this happen? seems like a weird thing to detect and `dsStore.Disable()` should guarantee that this can't happen

pkg/ds/farm.go
```diff
@@ -327,13 +336,16 @@ func (dsf *Farm) handleDSChanges(changes dsstore.WatchedDaemonSets, quitCh <-cha
 
 			if _, ok := err.(consulutil.AlreadyLockedError); ok {
 				if _, ok := dsf.children[dsFields.ID]; !ok {
-					dsLogger.Debugln("Lock on daemon set was already acquired by another farm")
+					dsf.logger.Infof("Lock on daemon set '%v' was already acquired by another farm", dsFields.ID)
 					continue
 				}
+				dsf.logger.Info("Lock on daemon set '%v' was already acquired by this farm", dsFields.ID)
```
mpuncel: same comment as the other `AlreadyLockedError` checks, you shouldn't need to check if it's a child or you wouldn't be trying to lock it

mpuncel:
no functionality flaws here, just some unnecessary stuff 👍

pkg/ds/farm.go
KoishiKomeiji: Updated, the code now follows this pattern, it will only acquire the lock if the child does not exist it its map.

pkg/ds/farm.go
KoishiKomeiji: Updated to be removed now.

KoishiKomeiji:
Thanks! Will merge after CI passes.

pkg/ds/farm.go
```diff
@@ -284,14 +287,16 @@ func (dsf *Farm) handleDSChanges(changes dsstore.WatchedDaemonSets, quitCh <-cha
 			dsLogger := dsf.makeDSLogger(*dsFields)
 			_, err := dsf.dsStore.LockForOwnership(dsFields.ID, dsf.session)
```
mpuncel: same here

pkg/ds/farm.go
```diff
@@ -241,18 +241,21 @@ func (dsf *Farm) handleDSChanges(changes dsstore.WatchedDaemonSets, quitCh <-cha
 			dsLogger := dsf.makeDSLogger(*dsFields)
 			dsUnlocker, err := dsf.dsStore.LockForOwnership(dsFields.ID, dsf.session)
```
KoishiKomeiji: Updated

pkg/ds/farm.go
KoishiKomeiji: Updated

PR #564, opened 2016-07-29
replication: Fix health aggregation

bwester:
The replication.AggregateHealth() call would return a podHealth object whose
health values would never change from those gathered in an initial query. As
used, the replication package would update a node and then only observe its
health from before the update, not after the update. This could allow a bad
update to be ~~pusehd~~ pushed to all nodes. The p2-replicate command and daemon set
updates are affected.

This bug happened because intended mutable shared state (a podHealth object)
was being passed by copy. The fix implemented in this commit is to pass by
reference.

There is also a potential for deadlock if the update thread encountered any
Consul error. It would send that error on a channel that was never read. This
mitigates the error by draining the channel.

### DIFF
--- a/pkg/replication/health_aggregator.go
+++ b/pkg/replication/health_aggregator.go
@@ -1,61 +1,84 @@
 package replication
 
 import (
+	"sync"
+
 	"github.com/square/p2/pkg/health"
 	"github.com/square/p2/pkg/health/checker"
 	"github.com/square/p2/pkg/types"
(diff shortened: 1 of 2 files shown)


pkg/replication/health_aggregator.go
```diff
@@ -1,61 +1,84 @@
 package replication
```
mpuncel: nit: s/pusehd/pushed in commit message

anthonybishopric:
🚢

PR #565, opened 2016-07-30
cgroups: Remove traces of cgexec

bwester:
P2's reliance on the "cgexec" command was removed in commit 004cafc.
A few unreferenced traces remained and are now expunged.


anthonybishopric:
easy 👍

mpuncel:
🚢

PR #566, opened 2016-07-30
cgroups: Allow hierarchical names

bwester:
Add support for hierarchical cgroups to be created by including the "/"
character in a cgroup's name when passed to p2-exec.


anthonybishopric:
👍

mpuncel:
🚢

PR #567, opened 2016-08-01
Make cgroup name explicit in hoist.Launchable

bwester:
Refactor the hoist.Launchable struct so that the name of the cgroup it
will use is explicitly configured instead of being implicitly the same
as the service ID. A future change will alter the cgroup naming scheme.


anthonybishopric:
👍

mpuncel:
🚢

PR #569, opened 2016-08-01
Restore the Installed() check before installing a launchable

mpuncel:
Prior to commit f6635afd588c9a3d79e137565166f0cc8a399a66, each
launchable type would check exit the Install() function early if the
launchable was already installed. When this code was moved to the pods
package, the behavior was not preserved, and installation is always
attempted which would result in errors due to files already being
present.

These errors would result in the whole install directory being deleted
and the installation being reattempted, at which point it would succeed.


bwester:
:shipit: Go for it.

rudle:
:ship:

PR #571, opened 2016-08-01
Protect against empty strings in status keys in status store

mpuncel:


rudle:
LGTM. The build failed on the rcstore package, which seems suspicious

bwester:
Could you add a description to your commit?

How did you find this problem? What are its symptoms?  What could cause a resource type/ID/namespace to be empty? Does this fix the root problem?

mpuncel:
@bwester this wasn't found due to a problem, but I was implementing a `ResourceType` for pods and realized that it isn't obvious why a namespace is necessary for this use case, however I do want to enforce that the use case is not an empty string. Figured I may as well enforce all of the keys so we don't end up with badly named consul keys

bwester:
Thanks for the context!

Add that to the commit message.

bwester:
🚢

PR #572, opened 2016-08-01
Mpuncel/pod id uuid transition

mpuncel:


pkg/types/types.go
```diff
@@ -4,12 +4,89 @@
...
+// functionality will only be supported for pods with UUID unique keys, hence
+// the need for the bool.
+type PodUniqueKey struct {
+	// The actual key, e.g. "example.com/mysql" under the old format or a uuid
+	// under the new format
+	id string
+
+	// Tracks whether the key is a uuid or the node/pod_id concatenation. Some
+	// features will require isUUID to be true.
+	isUUID bool
```
mpuncel: alternatively I could make `PodUniqueKey` _always_ be a uuid, and then in `ManifestResult` it would be a pointer that is nil if the pod is using the old key format.

pkg/types/types.go
```diff
@@ -4,12 +4,89 @@
...
+// "mysql" is running on a node "example.com".
+//
+// P2 will begin using uuids instead of the previous format to better support
+// running multiple pods with the same pod id on the same node. Certain new
+// functionality will only be supported for pods with UUID unique keys, hence
+// the need for the bool.
+type PodUniqueKey struct {
+	// The actual key, e.g. "example.com/mysql" under the old format or a uuid
+	// under the new format
+	id string
```
bwester: Considering the role of this structure, could you make these fields public and cut out the accessors?

mpuncel: fixed

pkg/types/types.go
```diff
@@ -4,12 +4,81 @@
...
+		}, nil
+	}
+
+	return PodUniqueKey{
+		ID:     path.Join(keyParts[1], keyParts[2]),
+		IsUUID: false,
+	}, nil
+}
+
+// PodKey represents the location
```
bwester: What is this for?

mpuncel: accidentally didn't remove this, will fix

pkg/types/types.go
```diff
@@ -4,12 +4,78 @@
...
+		return PodUniqueKey{}, util.Errorf("Malformed key '%s'", consulPath)
+	}
+
+	// Unforunately we can't use kp.INTENT_TREE and kp.REALITY_TREE here because of an import cycle
+	if keyParts[0] != "intent" && keyParts[0] != "reality" {
+		return PodUniqueKey{}, util.Errorf("Unrecognized key tree '%s' (must be intent or reality)", keyParts[0])
+	}
+
+	// Parse() returns nil if the input string does not match the uuid spec
+	if uuid.Parse(keyParts[2]) != nil {
```
bwester: Does this mean you're expecting the key `intent/myhost.example.com/123e4567-e89b-12d3-a456-426655440000` to contain a PGP-signed manfiest?

mpuncel: you're right that the code expects that but that's not our intention. I'll rework things so that a "uuid key not supported" error is generated for uuid keys given that nothing writes them yet and I don't want this PR to need to include the rest of the uuid implementation

bwester:
Also, something is not right with the split of this PR into two separate commits. For instance, 149f3fa contains a bunch of imports in pkg/types/types.go without any changes.

mpuncel:
rebased on master

pkg/kp/kv.go
```diff
@@ -404,3 +410,71 @@ func HealthPath(service string, node types.NodeName) string {
 func (c consulStore) NewHealthManager(node types.NodeName, logger logging.Logger) HealthManager {
 	return c.newSessionHealthManager(node, logger)
 }
+
+// Custom error type for returning errors related to uuid consul keys being
+// partiall implemented.  This is useful for things that list keys that would
```
anthonybishopric: nit: spelling

mpuncel:
fixed a typo and added a "hooks" entry in the table test

pkg/types/types.go
```diff
@@ -4,12 +4,90 @@
 package types
 
 import (
+	"path"
+	"strings"
+
+	"github.com/square/p2/pkg/util"
+
+	"github.com/pborman/uuid"
```
anthonybishopric: Was this dependency already present in our godeps?

pkg/types/types.go
```diff
@@ -4,12 +4,90 @@
...
+			ID:     consulPath,
+			IsUUID: false,
+		}, nil
+	}
+
+	if len(keyParts) != 3 {
+		return PodUniqueKey{}, util.Errorf("Malformed key '%s'", consulPath)
+	}
+
+	// Unforunately we can't use kp.INTENT_TREE and kp.REALITY_TREE here because of an import cycle
```
anthonybishopric: Why wouldn't we put this function in the `kp` package instead? `types` doesn't seem very related to Consul interaction, and conceptually shouldn't, right?

mpuncel: good suggestion 👍

anthonybishopric:
:+1: modulo nit

pkg/types/types.go
mpuncel: missed this question, but yeah it's what we use to generate uuids for RCs and pod clusters

PR #574, opened 2016-08-01
Change scheduling to labelling in daemon set

KoishiKomeiji:
This is because it is not actually scheduling pods yet but rather
labelling them, this will reduce confusion in the log messages.


anthonybishopric:
:shipit:

PR #575, opened 2016-08-03
Modify daemon sets to skip the preparer check

KoishiKomeiji:
Will also return an error when initialization fails

### DIFF
--- a/pkg/replication/replicator.go
+++ b/pkg/replication/replicator.go
@@ -20,6 +20,14 @@ type Replicator interface {
 		ignoreControllers bool,
 		concurrentRealityNodes int,
 	) (Replication, chan error, error)
+
+	// Same as InitializeReplication but can skip the checkPreparers
+	InitializeReplicationWithCheck(
+		overrideLock bool,
+		ignoreControllers bool,
+		concurrentRealityRequests int,
+		checkPreparers bool,
+	) (Replication, chan error, error)
 }
 
 // Replicator creates replications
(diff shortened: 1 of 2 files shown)


pkg/replication/replicator.go
```diff
@@ -110,6 +117,58 @@ func (r replicator) InitializeReplication(
...
 
+// Initializes a replication after performing some initial validation.
+// Validation errors are returned immediately, and asynchronous errors are
+// passed on the returned channel
+func (r replicator) InitializeReplicationWithCheck(
+	overrideLock bool,
+	ignoreControllers bool,
+	concurrentRealityRequests int,
+	checkPreparers bool,
+) (Replication, chan error, error) {
```
rudle: This function is copy pasted from `InitializeReplication`. I'd like to see `InitializeReplication` call this function with a checkPreparers argument of  `true` so that we can reduce code duplication.

KoishiKomeiji: Updated

rudle:
:ship:

KoishiKomeiji:
Thanks!

pkg/replication/replicator.go
```diff
@@ -71,9 +79,41 @@ func (r replicator) InitializeReplication(
...
+	return r.initializeReplicationWithCheck(
+		overrideLock,
+		ignoreControllers,
+		concurrentRealityRequests,
+		checkPreparers,
+	)
+}
+
+func (r replicator) initializeReplicationWithCheck(
```
mpuncel: if you export this function you can get rid of the one above it looks like

bwester:
It's too late to provide this context in the commit message, but why exactly do daemon sets need to skip the preparer check?

KoishiKomeiji:
Daemon sets need to skip the preparer check because not all nodes would necessarily have a preparer on it, when this happens there would be an error and replication can't be enacted.

bwester:
What good does it do to schedule a pod on a node without a preparer?

rudle:
To me, it sounds like it would be desirable to filter out nodes without preparers instead of writing bogus intent entries for them.

In an ideal world, our node labels tree would be 100% consistent with the reality of the datacenter but such a world is not achievable at the moment.

KoishiKomeiji:
Alright, I will filter out nodes without preparers.

PR #576, opened 2016-08-03
Replication bugfixes and logging

KoishiKomeiji:
This is mainly for fixing some bugs with how locks are handled in the farm.
There is also a fix to cancel the replication when a daemon set is disabled.
There are also additional logging and removing an unnecessary check for whether a daemon set is in the map or not.
This also includes a change in the fake dsstore to imitate the real dsstore's locking mechanism.


rudle:
🚢

KoishiKomeiji:
Thanks!

PR #577, opened 2016-08-03
Remove the old dsctl's availability zone with a raw selector

KoishiKomeiji:
The createAZ and updateAZ has been removed and replaced with createSelector and updateSelector.

### DIFF
--- a/bin/p2-dsctl/main.go
+++ b/bin/p2-dsctl/main.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"log"
 	"strconv"
+	"strings"
 	"time"
 
 	"gopkg.in/alecthomas/kingpin.v2"
@@ -14,34 +15,43 @@ import (
 	"github.com/square/p2/pkg/kp"
 	"github.com/square/p2/pkg/kp/dsstore"
 	"github.com/square/p2/pkg/kp/flags"
+	"github.com/square/p2/pkg/labels"
 	"github.com/square/p2/pkg/logging"
 	"github.com/square/p2/pkg/manifest"
-	pc_fields "github.com/square/p2/pkg/pc/fields"
 	"github.com/square/p2/pkg/types"
 	"github.com/square/p2/pkg/util"
-	"k8s.io/kubernetes/pkg/labels"
+	klabels "k8s.io/kubernetes/pkg/labels"
 )
 
 const (
-	CmdCreate  = "create"
-	CmdGet     = "get"
-	CmdList    = "list"
-	CmdEnable  = "enable"
-	CmdDisable = "disable"
-	CmdDelete  = "delete"
-	CmdUpdate  = "update"
+	CmdCreate       = "create"
+	CmdGet          = "get"
+	CmdList         = "list"
+	CmdEnable       = "enable"
+	CmdDisable      = "disable"
+	CmdDelete       = "delete"
+	CmdUpdate       = "update"
+	CmdTestSelector = "test-selector"
 
 	TimeoutNotSpecified = time.Duration(-1)
 )
 
+func flagUsed(marker *bool) kingpin.Action {
+	return func(*kingpin.ParseContext) error {
+		*marker = true
+		return nil
+	}
+}
+
 var (
-	cmdCreate       = kingpin.Command(CmdCreate, "Create a daemon set.")
-	createAZ        = cmdCreate.Flag("az", "The availability zone of the pod cluster").Required().String()
-	createManifest  = cmdCreate.Flag("manifest", "Complete manifest - must parse as JSON!").Required().String()
-	createMinHealth = cmdCreate.Flag("minhealth", "The minimum health of the daemon set").Required().String()
-	createName      = cmdCreate.Flag("name", "The cluster name (ie. staging, production)").Required().String()
-	createPodID     = cmdCreate.Flag("pod", "The pod ID on the daemon set").Required().String()
-	createTimeout   = cmdCreate.Flag("timeout", "Non-zero timeout for replicating hosts. e.g. 1m2s for 1 minute and 2 seconds").Required().Duration()
+	cmdCreate        = kingpin.Command(CmdCreate, "Create a daemon set.")
+	createSelector   = cmdCreate.Flag("selector", "The node selector, uses the same syntax as the test-selector command").Required().String()
+	createManifest   = cmdCreate.Flag("manifest", "Complete manifest - must parse as JSON!").Required().String()
+	createMinHealth  = cmdCreate.Flag("minhealth", "The minimum health of the daemon set").Required().String()
+	createName       = cmdCreate.Flag("name", "The cluster name (ie. staging, production)").Required().String()
+	createPodID      = cmdCreate.Flag("pod", "The pod ID on the daemon set").Required().String()
+	createTimeout    = cmdCreate.Flag("timeout", "Non-zero timeout for replicating hosts. e.g. 1m2s for 1 minute and 2 seconds").Required().Duration()
+	createEverywhere = cmdCreate.Flag("everywhere", "Sets selector to match everything regardless of its value").Bool()
 
 	cmdGet = kingpin.Command(CmdGet, "Show a daemon set.")
 	getID  = cmdGet.Arg("id", "The uuid for the daemon set").Required().String()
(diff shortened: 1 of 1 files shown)


bin/p2-dsctl/main.go
```diff
@@ -22,18 +22,19 @@ import (
...
+	CmdDisable          = "disable"
+	CmdDelete           = "delete"
+	CmdUpdate           = "update"
+	CmdOverrideSelector = "overrideSelector"
 )
 
 var (
 	cmdCreate       = kingpin.Command(CmdCreate, "Create a daemon set.")
-	createAZ        = cmdCreate.Flag("az", "The availability zone of the pod cluster").Required().String()
+	createAZ        = cmdCreate.Flag("az", "The availability zone of the pod cluster").String()
```
bwester: What does this AZ change have to do with: "Add a an option to override the ds node selector"?

bin/p2-dsctl/main.go
```diff
@@ -22,18 +22,19 @@ import (
...
-	CmdDelete  = "delete"
-	CmdUpdate  = "update"
+	CmdCreate           = "create"
+	CmdGet              = "get"
+	CmdList             = "list"
+	CmdEnable           = "enable"
+	CmdDisable          = "disable"
+	CmdDelete           = "delete"
+	CmdUpdate           = "update"
+	CmdOverrideSelector = "overrideSelector"
```
bwester: Unix commands typically use dashes to join words in flags/commands. So this would be `override-selector`.

But I don't think calling this an "override" is a good descriptor of what this command does. "Override" usually means something like: "Ignore that saved value, I'll supply the value to use right now." That isn't what's happening here. This command sets the DS's node selector. I think `set-selector` would be easier to understand.

bin/p2-dsctl/main.go
```diff
@@ -60,6 +61,14 @@ var (
 	updateMinHealth = cmdUpdate.Flag("minhealth", "The minimum health of the daemon set").String()
 	updateName      = cmdUpdate.Flag("name", "The cluster name (ie. staging, production)").String()
 	updatePodID     = cmdUpdate.Flag("pod", "The pod ID on the daemon set").String()
+
+	cmdOverrideSelector = kingpin.Command(CmdOverrideSelector, `
+		The raw selector for the daemon set represented as a string.
```
bwester: Could you change this help text to describe what the command does? Arguments already have their own descriptions below.

bin/p2-dsctl/main.go
```diff
@@ -60,6 +61,14 @@ var (
...
+		The selector string uses same syntax as the kubernetes selectors or p2-select without flags.
+		It is highly that recommended that you test it on p2-select first`,
```
bwester: > It is highly that recommended that you test it on p2-select first

I'm glad you recognize that this operation could have huge consequences, but I don't think a warning in the help text is enough. What do you think about showing a change summary?

```
$ p2-dsctl set-selector 01-23-45-67-89 sq/datacenter=sjc1,sq/zone=web-data
Changing slug deployment from sq/datacenter=sjc1,sq/zone=web-data:
removing:    -43 hosts (awa0.sjc1.square, ...)
adding:     +339 hosts (awd139.sjc1.square, ...)
no change:     0 hosts
Continue? [yn]: 
````

Comment (mpuncel) [inline bin/p2-dsctl/main.go]: furthermore, `p2-select` is not an open source binary so we shouldn't reference it in open source code

File: bin/p2-dsctl/main.go [hunk 5]
```diff
@@ -226,6 +241,31 @@ func main() {
...
+		id := ds_fields.ID(*overrideSelectorID)
+
+		selector, err := labels.Parse(*overrideSelectorString)
+		if err != nil {
+			log.Fatalf("Malformed selector: %v", err)
+			return
+		}
+
+		mutator := func(ds ds_fields.DaemonSet) (ds_fields.DaemonSet, error) {
+			if ds.NodeSelector.String() != selector.String() {
```
mpuncel: this might read better as ```
if ds.NodeSelector.String() == selector.String() {
  return ds, util.Errorf()
}

ds.NodeSelector = selector
```

File: bin/p2-dsctl/main.go [hunk 6]
```diff
@@ -70,14 +79,20 @@ func main() {
 
 	switch cmd {
 	case CmdCreate:
-		az := pc_fields.AvailabilityZone(*createAZ)
+		var selector labels.Selector
```
rudle: How hard would it be to allow a raw selector to be expressed during the create operation? This tool is not very useful to Square without such an option.

bin/p2-dsctl/main.go
KoishiKomeiji: Updated

bin/p2-dsctl/main.go
KoishiKomeiji: Updated

bin/p2-dsctl/main.go
KoishiKomeiji: Removed.

bin/p2-dsctl/main.go
KoishiKomeiji: In progress.

bin/p2-dsctl/main.go
KoishiKomeiji: Updated

bin/p2-dsctl/main.go
KoishiKomeiji: Updated

bin/p2-dsctl/main.go
```diff
@@ -13,27 +13,31 @@ import (
...
+	CmdCreate       = "create"
+	CmdGet          = "get"
+	CmdList         = "list"
+	CmdEnable       = "enable"
+	CmdDisable      = "disable"
+	CmdDelete       = "delete"
+	CmdUpdate       = "update"
+	CmdTestSelector = "test-selector"
+
+	SelectorNotSpecified = "thereisnowaysomeonewouldtypethis"
```
rudle: I don't know if seeing this in a DS would tell me all I needed to know. Instead, we should try to make this self-documenting. How about: "defaultNullSelector", "null", or ø

bin/p2-dsctl/main.go
```diff
@@ -70,14 +82,18 @@ func main() {
 
 	switch cmd {
 	case CmdCreate:
-		az := pc_fields.AvailabilityZone(*createAZ)
+		selector, err := klabels.Parse(*createSelector)
+		if err != nil {
+			log.Fatalf("Malformed selector: %v", err)
+			return
```
rudle: this is unreachable code, I'm surprised go vet didn't catch it.

bin/p2-dsctl/main.go
```diff
@@ -226,12 +245,31 @@ func main() {
 		fmt.Printf("The daemon set '%s' has been successfully updated in consul", id.String())
 		fmt.Println()
 
+	case CmdTestSelector:
+		applicator := labels.NewConsulApplicator(client, 3)
+		selector, err := klabels.Parse(*testSelectorString)
+		if err != nil {
+			log.Fatalf("Malformed selector: %v", err)
+			return
```
rudle: also unreachable code here and on line 258

bin/p2-dsctl/main.go
```diff
@@ -226,12 +245,31 @@ func main() {
...
+
 	default:
 		log.Fatalf("Unrecognized command %v", cmd)
 	}
 }
 
-func selectorFrom(az pc_fields.AvailabilityZone) labels.Selector {
-	return labels.Everything().
-		Add(pc_fields.AvailabilityZoneLabel, labels.InOperator, []string{az.String()})
+func selectorFrom(az pc_fields.AvailabilityZone) klabels.Selector {
```
rudle: Is this function used? I can't see any references to it

bin/p2-dsctl/main.go
```diff
@@ -226,12 +245,31 @@ func main() {
...
+	return klabels.Everything().
+		Add(pc_fields.AvailabilityZoneLabel, klabels.InOperator, []string{az.String()})
+}
+
+func parseSelectorWithInfo(selectorString string) (klabels.Selector, error) {
```
rudle: This function also looks unused - perhaps my editor is misconfigured…

bin/p2-dsctl/main.go
```diff
@@ -55,11 +59,19 @@ var (
...
-	updateAZ        = cmdUpdate.Flag("az", "The availability zone of the pod cluster").String()
+	updateSelector  = cmdUpdate.Flag("selector", "The node selector, uses the same syntax as the test-selector command").Default(SelectorNotSpecified).String()
 	updateManifest  = cmdUpdate.Flag("manifest", "Complete manifest - must parse as JSON!").String()
 	updateMinHealth = cmdUpdate.Flag("minhealth", "The minimum health of the daemon set").String()
 	updateName      = cmdUpdate.Flag("name", "The cluster name (ie. staging, production)").String()
 	updatePodID     = cmdUpdate.Flag("pod", "The pod ID on the daemon set").String()
+
+	cmdTestSelector = kingpin.Command(CmdTestSelector, `
+		This will output the hosts that match the selector,
+		The selector string uses same syntax as the kubernetes selectors or p2-select without flags.
```
rudle: I think it's strange to reference `p2-select` in the open source code

bin/p2-dsctl/main.go
```diff
@@ -226,12 +245,31 @@ func main() {
...
+			log.Fatalf("Malformed selector: %v", err)
+			return
+		}
+		matches, err := applicator.GetMatches(selector, "")
+		if err != nil {
+			log.Fatalf("Error getting matching labels: %v", err)
+			return
+		}
+		fmt.Println(matches)
+		fmt.Println(klabels.Nothing().String())
```
rudle: I don't understand why this line is here

bin/p2-dsctl/main.go
```diff
@@ -55,11 +59,19 @@ var (
 
 	cmdUpdate       = kingpin.Command(CmdUpdate, "Update a daemon set.")
 	updateID        = cmdUpdate.Arg("id", "The uuid for the daemon set").Required().String()
-	updateAZ        = cmdUpdate.Flag("az", "The availability zone of the pod cluster").String()
+	updateSelector  = cmdUpdate.Flag("selector", "The node selector, uses the same syntax as the test-selector command").Default(SelectorNotSpecified).String()
```
anthonybishopric: The default value is not required. Kingpin string arguments are always pointers to strings (`*string`). This means you can test the pointer if it's equal to `nil` to decide if the user has passed it or not.

bin/p2-dsctl/main.go
```diff
@@ -163,11 +179,14 @@ func main() {
 		mutator := func(ds ds_fields.DaemonSet) (ds_fields.DaemonSet, error) {
 			changed := false
 
-			if *updateAZ != "" {
-				selector := selectorFrom(pc_fields.AvailabilityZone(*updateAZ))
+			if *updateSelector != SelectorNotSpecified {
```
anthonybishopric: `if updateSelector != nil` -> then you know the selector was passed on the command line.

bin/p2-dsctl/main.go
bwester: @anthonybishopric: [Kingpin flags don't work that way](https://github.com/alecthomas/kingpin/blob/v2.2.2/values_generated.go#L77). `updateSelector` is a `*string`, but it will always be non-nil, regardless of whether the user passed the flag.

bwester: Would it make sense to have a DS without any node selector?  I.e., will `--selector ""` ever be valid?  If not, don't specify a default and just test for `updateSelector == ""` to see if the flag was given.

Otherwise, be sure to use [`.Placeholder()`](https://godoc.org/gopkg.in/alecthomas/kingpin.v2#FlagClause.PlaceHolder) for this flag, or your improbable default value becomes visible with `--help`.

anthonybishopric: Empty selectors are valid in the case where you want a daemon to go to every node in the label tree, and I suspect there is a need to detect when the flag was passed or not. Sorry about the kingpin confusion @KoishiKomeiji

mpuncel: I think it would be awkward to express an empty string as an arg anyway, maybe we make a `--everywhere` flag and disallow empty selector?

KoishiKomeiji: It doesn't seem to behave like that, I seem to be getting an address when running
`fmt.Println(updateSelector)`

KoishiKomeiji: I'll do what @mpuncel suggested

KoishiKomeiji:
Updated.

rudle:
I've used this and it works well for me. Going to merge it.

PR #579, opened 2016-08-03
Mask *api.Client behind a ConsulClient interface in stores

mpuncel:
This will allow providing a fake *api.Client to stores for use in tests
that provides all of the necessary functionality. The main motivation
for this commit is to add more tests to the kp.consulStore struct.

Unfortunately this change required a bit of gymnastics around Go's type
system. *api.Client is a struct with KV() and Session() methods that
return structs providing functionality for interacting with consul keys
and consul sessions, respectively.

The ConsulClient interface captures the KV() and Session() functions
that return sub-interfaces that we use, since tests will ultimately want
to fake out the KV and session portions of consul functionality.

Unfortunately, a raw *api.Client does not implement the ConsulClient
interface because the return types of its KV() and Session() functions
are structs, which is not an exact match to what the ConsulClient
interface requires. As a result, a consulutil.ConsulClientFromRaw()
function was added which returns an implementation of the ConsulClient
interface based on a passed *api.Client.

Since most consul client initialization code goes through
the kp.NewConsulClient function, this function simply calls the wrapper
function, so there shouldn't be much extra burden added to callers
initializing a store.

### DIFF
--- a/pkg/kp/config.go
+++ b/pkg/kp/config.go
@@ -4,6 +4,8 @@ import (
 	"net/http"
 	"time"
 
+	"github.com/square/p2/pkg/kp/consulutil"
+
 	"github.com/hashicorp/consul/api"
 )
 
@@ -24,7 +26,7 @@ type Options struct {
 	WaitTime time.Duration
 }
 
-func NewConsulClient(opts Options) *api.Client {
+func NewConsulClient(opts Options) consulutil.ConsulClient {
 	conf := api.DefaultConfig()
 	if opts.Address != "" {
 		conf.Address = opts.Address
@@ -41,6 +43,6 @@ func NewConsulClient(opts Options) *api.Client {
 	}
 
 	// error is always nil
-	ret, _ := api.NewClient(conf)
-	return ret
+	client, _ := api.NewClient(conf)
+	return consulutil.ConsulClientFromRaw(client)
 }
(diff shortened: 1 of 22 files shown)


pkg/kp/config.go
```diff
@@ -41,6 +43,6 @@ func NewConsulClient(opts Options) *api.Client {
 	}
 
 	// error is always nil
-	ret, _ := api.NewClient(conf)
-	return ret
+	client, _ := api.NewClient(conf)
```
rudle: this code is pulling a fast one on errcheck. Either that, or errcheck is able to read documentation ;)

pkg/kp/consulutil/consultest.go
```diff
@@ -1,7 +1,7 @@
-// The consultest package contains common routines for setting up a live Consul server for
+// This file contains common routines for setting up a live Consul server for
```
rudle: If you're changing this, I'd prefer it be changed to match the convention. The godoc spec decrees this style: "package consulutil contains"

rudle:
just nits, 👍

pkg/kp/config.go
mpuncel: doesn't errcheck only complain if you don't assign it to a variable (even if that variable is _)

PR #580, opened 2016-08-03
Hzeng.replicator timeout

KoishiKomeiji:
Adds a timeout to daemon sets when replicating.
Also modified p2-ds-farm to gracefully shutdown and remove its locks.

### DIFF
--- a/bin/p2-dsctl/main.go
+++ b/bin/p2-dsctl/main.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"log"
 	"strconv"
+	"time"
 
 	"gopkg.in/alecthomas/kingpin.v2"
 
@@ -29,6 +30,8 @@ const (
 	CmdDisable = "disable"
 	CmdDelete  = "delete"
 	CmdUpdate  = "update"
+
+	TimeoutNotSpecified = time.Duration(-1)
 )
 
 var (
@@ -38,6 +41,7 @@ var (
 	createMinHealth = cmdCreate.Flag("minhealth", "The minimum health of the daemon set").Required().String()
 	createName      = cmdCreate.Flag("name", "The cluster name (ie. staging, production)").Required().String()
 	createPodID     = cmdCreate.Flag("pod", "The pod ID on the daemon set").Required().String()
+	createTimeout   = cmdCreate.Flag("timeout", "Non-zero timeout for replicating hosts. e.g. 1m2s for 1 minute and 2 seconds").Required().Duration()
 
 	cmdGet = kingpin.Command(CmdGet, "Show a daemon set.")
 	getID  = cmdGet.Arg("id", "The uuid for the daemon set").Required().String()
@@ -60,6 +64,7 @@ var (
 	updateMinHealth = cmdUpdate.Flag("minhealth", "The minimum health of the daemon set").String()
 	updateName      = cmdUpdate.Flag("name", "The cluster name (ie. staging, production)").String()
 	updatePodID     = cmdUpdate.Flag("pod", "The pod ID on the daemon set").String()
+	updateTimeout   = cmdUpdate.Flag("timeout", "Non-zero timeout for replicating hosts. e.g. 1m2s for 1 minute and 2 seconds").Default(TimeoutNotSpecified.String()).Duration()
 )
 
 func main() {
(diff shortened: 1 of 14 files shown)


bin/p2-dsctl/main.go
```diff
@@ -194,6 +201,14 @@ func main() {
 					ds.PodID = podID
 				}
 			}
+			// If the timeout is not specified, the default timeout is time.Duration(0)
+			// We don't want a timeout to be 0 anyways because nothing will be scheduled
+			if *updateTimeout != time.Duration(0) {
```
mpuncel: if createTimeout is 0, you exit the program but if updateTimeout is 0, you just don't update it. Probably exit here too?

pkg/replication/replicator.go
```diff
@@ -14,6 +16,10 @@ import (
 
 const DefaultConcurrentReality = 3
 
+// Normal replications will have no timeout, but daemon sets will
+// because it is unlikely that all hosts are healthy at all times
```
mpuncel: i'm not sure I understand the justification, shouldn't a daemon set keep trying forever to sync out the version it is configured to sync?

pkg/replication/replication.go
```diff
@@ -278,8 +286,34 @@ func (r replication) updateOne(node types.NodeName, done chan<- types.NodeName,
...
+		select {
+		case <-completedChan:
+		case <-time.After(r.timeout):
+			r.timedOutReplications = append(r.timedOutReplications, node)
+			r.logger.Errorf("The host '%v' timed out during replication for pod '%v'", node.String(), r.manifest.ID())
+		case <-quitCh:
+			return
+		}
+	}
+	done <- node
```
mpuncel: `select` on `quitCh` here

pkg/replication/replicator.go
KoishiKomeiji: @abishopric asked to specify a time limit because it was possible that some hosts will never be healthy and to log the hosts which did not meet the timeout.

KoishiKomeiji:
Updated.

pkg/replication/replication.go
```diff
@@ -278,8 +286,37 @@ func (r replication) updateOne(node types.NodeName, done chan<- types.NodeName,
...
+		case <-completedChan:
+		case <-quitCh:
+			return
+		}
+	} else {
+		// Wait until either completion, timeout, or a quit
+		select {
+		case <-completedChan:
+		case <-time.After(r.timeout):
+			r.timedOutReplications = append(r.timedOutReplications, node)
```
rudle: `r.timedOutReplications` is shared state - I'm pretty sure this line needs a mutex around it

pkg/replication/replicator.go
anthonybishopric: If nothing else, the node will have had the new manifest scheduled.

anthonybishopric:
Agreed with the mutex around `r.timedOutReplications`, otherwise 🚢

pkg/replication/replication.go
```diff
@@ -74,6 +74,12 @@ type replication struct {
 	// Semaphore that sets a maximum value on the number of concurrent
 	// reality requests that can be fired simultaneously.
 	concurrentRealityRequests chan struct{}
+
+	// Used to timeout daemon set replications
+	timeout time.Duration
+
+	// Used to log replications that have timed out
+	timedOutReplications []types.NodeName
```
bwester: Why are you collecting this list? I don't see where it's being used.

KoishiKomeiji: Not being used it, but it will be used in the future in the status tree.

bin/p2-dsctl/main.go
```diff
@@ -60,6 +64,7 @@ var (
 	updateMinHealth = cmdUpdate.Flag("minhealth", "The minimum health of the daemon set").String()
 	updateName      = cmdUpdate.Flag("name", "The cluster name (ie. staging, production)").String()
 	updatePodID     = cmdUpdate.Flag("pod", "The pod ID on the daemon set").String()
+	updateTimeout   = cmdUpdate.Flag("timeout", "Non-zero timeout for replicating hosts. e.g. 1m2s for 1 minute and 2 seconds").Default(TimeoutNotSpecified.String()).Duration()
```
KoishiKomeiji: I believe that `Default()` will provide the behavior I need.
Placeholder doesn't seem to give me the default value of `TimeoutNotSpecified` for me.

pkg/replication/replication.go
KoishiKomeiji: Updated.

KoishiKomeiji:
Updated, will merge when CI passes.

pkg/replication/replication.go
```diff
@@ -348,13 +388,6 @@ func (r *replication) ensureHealthy(
 				nodeLogger.WithFields(logrus.Fields{"check": id, "health": status}).Infoln("Node is not healthy")
 			} else {
 				r.logger.WithField("node", node).Infoln("Node is current and healthy")
```
bwester: @KoishiKomeiji: can you take another look at `ensureHealthy()`?  In particular, when a node has been successfully replicated, how does this method end?

KoishiKomeiji: Thanks for the catch, I looked at the history, and it should return here.

PR #581, opened 2016-08-04
Mpuncel/pod status

mpuncel:


pkg/kp/statusstore/podstatus/status.go
```diff
@@ -0,0 +1,69 @@
...
+	ExitCode int         `json:"exit_code"`
+}
+
+// Encapsulates information regarding the state of a service. This includes
+// information about a service that might still be running as well as its most
+// recent exit. This is because many services under P2 are always restarted
+// after exit by runit, and both the status of the most recent exit as well as
+// the current status of the service are useful to expose
+type ServiceStatus struct {
+	Name     string       `json:"name"`
```
rudle: I can't see how the name field will be used. Does this match up with a launchable ID?

pkg/kp/statusstore/podstatus/status.go
```diff
@@ -0,0 +1,69 @@
...
+	State    ServiceState `json:"state"`
+	LastExit *ExitStatus  `json:"last_exit"`
```
rudle: A text field `LastExitMessage` might be useful here especially when it comes to debugging. I can imagine, for example, storing a subset of STDERR in here.

pkg/kp/statusstore/podstatus/status.go
mpuncel: it will be the full runit service ID, e.g. `podid__launchableid__scriptname`. Perhaps there's a type I can use to make that more obvious (or i'll at least add a comment)

mpuncel: there will also be some kind of `uuid` somewhere for pods named with uuids, probably `podid-uuid` in the first part

pkg/kp/statusstore/podstatus/status.go
mpuncel: is it alright if that gets added in a future PR? Should be very easy to do and it will probably be more clear what the shape of it should be etc when we're down in the weeds

rudle: sure thing!

rudle:
📦

PR #582, opened 2016-08-04
Mpuncel/dual pod write

mpuncel:


rudle:
I've reviewed this several times but neglected to express my approval. Godspeed you brave code.

👍

mpuncel:
rebased onto master

PR #583, opened 2016-08-04
Make statusstore use a ConsulClient instead of *api.Client

mpuncel:
This makes it easier for 3rd party code to use the status store, because
kp.NewConsulClient now returns a ConsulClient interface backed by an
*api.Client instead of the client itself.


rudle:
:shipit:

PR #584, opened 2016-08-04
Fix unscheduling bug and daemon set update race conditions

KoishiKomeiji:
When unscheduling some pods, the replication needs to be cancelled, but
some pods were being scheduled even though they weren't supposed to
because of a counting error due to currentNodes not being updated.

Previously, an update to daemon sets would quickly cancel the
replication and re-enact it, but it does not wait for replication to
gracefully release the session locks, this commit will fix that problem.

The health has also been faked in the daemon set and daemon set farm test. 
The locks in replication have also been faked, but is a lazy implementation. 
Because of this, we can now remove the nil health checker condition for daemon set unit tests.

### DIFF
--- a/pkg/ds/daemon_set.go
+++ b/pkg/ds/daemon_set.go
@@ -343,7 +343,7 @@ func (ds *daemonSet) removePods() error {
 	}
 
 	ds.logger.Infof("Need to schedule %v nodes", len(currentNodes))
-	if len(currentNodes) > 0 {
+	if len(currentNodes)-len(toUnscheduleSorted) > 0 {
 		return ds.PublishToReplication()
 	}
 
--- a/pkg/ds/daemon_set.go
+++ b/pkg/ds/daemon_set.go
@@ -4,7 +4,6 @@ import (
 	"fmt"
 	"os"
 	"os/user"
-	"sync"
 	"time"
 
 	"github.com/square/p2/pkg/health"
@@ -79,9 +78,6 @@ type daemonSet struct {
 
 	// This is the current replication enact go routine that is running
 	currentReplication replication.Replication
-
-	// This locks are used to make sure only one replication is going on at the same time
-	replicationLock sync.Mutex
 }
 
 type dsContention struct {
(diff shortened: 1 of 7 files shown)


pkg/ds/daemon_set.go
```diff
@@ -343,7 +343,7 @@ func (ds *daemonSet) removePods() error {
 	}
 
 	ds.logger.Infof("Need to schedule %v nodes", len(currentNodes))
-	if len(currentNodes) > 0 {
+	if len(currentNodes)-len(toUnscheduleSorted) > 0 {
 		return ds.PublishToReplication()
```
anthonybishopric: Is there a test to reproduce this and the effectiveness of this fix?

rudle:
Can you demonstrate this case with a test? The description is a bit nuanced

KoishiKomeiji:
There isn't a test to demonstrate it, but it is because the currentNodes wasn't being updated from above, so if current nodes says `[a b c]` and we unscheduled `[a b c]`, this will think that we still have the nodes `[a b c]` to schedule.

bwester:
How hard is it to write a test that fails without this fix and passes with it?

KoishiKomeiji:
I can reproduce it, but I'll need to make a fake health to test it.

KoishiKomeiji:
Updated, tests have been added to the tests for daemon sets and the daemon set farm to test out the replicator. The health has also been faked and so have the locks.

pkg/ds/daemon_set.go
```diff
@@ -187,7 +187,7 @@ func (ds *daemonSet) WatchDesires(
 					ds.CancelReplication()
 					continue
 				}
-				err := ds.removePods()
+				err = ds.removePods()
```
rudle: Is this done just to save the allocation? I can't see any behavioural impact, but I want to be sure.

rudle:
Looks good. Can you squash the replication lock removal commit (ie. `20b4237` into any other commits that churn that section of code? I don't see a good reason to clutter up git history with that.

KoishiKomeiji:
Alright I'll do that.

pkg/ds/daemon_set.go
KoishiKomeiji: This is a bug, this is in its own namespace and will not be reported in the 
```
			if err != nil {
				select {
				case errCh <- err:
				case <-quitCh:
					return
				}
			}
```
above.

rudle:
:ship:

PR #585, opened 2016-08-04
rc, ds: Permit unmarshaling zero values

bwester:
When unmarshaling a JSON-encoded RC or DaemonSet, skip creating a
manifest object if the Manifest field is empty. This allows the zero
value for the RC and DaemonSet types to be decoded without being a parse
error. It was already OK for a zero value to be encoded. This leads to
strange errors where an object that contains an RC/DS can be encoded but
cannot be decoded.

RCs/DSs without a Manifest are still not valid, and their main Store
interfaces will still reject objects that do not contain a Manifest.
This preserves any existing consistency guarantees.

### DIFF
--- a/pkg/ds/fields/fields.go
+++ b/pkg/ds/fields/fields.go
@@ -100,9 +100,13 @@ func (ds *DaemonSet) UnmarshalJSON(b []byte) error {
 		return err
 	}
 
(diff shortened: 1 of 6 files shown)


mpuncel:
🚢

PR #586, opened 2016-08-05
Add functionality for daemon sets to retry replication on error

KoishiKomeiji:
This will make daemon sets more robust to pod lock contention during
replication


rudle:
:ship:

PR #587, opened 2016-08-08
Allow the daemon set to act on changes when an error occurs

KoishiKomeiji:
Previously, when the daemon set encounters an error during replication,
it would sleep for 5 minutes and then retry. It was not able to take in
updates or deletes until after the retry happens.

Now with this change it can take in updates in between this time period
and if the updates produce no error on replication it would cancel the
retry.

### DIFF
--- a/pkg/ds/daemon_set.go
+++ b/pkg/ds/daemon_set.go
@@ -145,6 +145,11 @@ func (ds *daemonSet) WatchDesires(
 		defer close(errCh)
 		defer ds.cancelReplication()
 
+		// Make a timer and stop it so the receieve from channel does not occur
+		// until a reset happens
+		timer := time.NewTimer(time.Duration(0))
+		timer.Stop()
+
 		// Try to schedule pods when this begins watching
 		if !ds.Disabled {
 			ds.logger.NoFields().Infof("Received new daemon set: %v", *ds)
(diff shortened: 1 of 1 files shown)


pkg/ds/daemon_set.go
```diff
@@ -158,14 +163,23 @@ func (ds *daemonSet) WatchDesires(
 			if err != nil {
 				select {
 				case errCh <- err:
-					// Retry the replication for extra robustness
-					time.Sleep(RetryInterval)
-					err = ds.PublishToReplication()
+					ds.logger.Infof("An error has occurred in the daemon set, retrying if no changes are made in %v", RetryInterval)
```
anthonybishopric: This should probably be a `Warnf`

pkg/ds/daemon_set.go
```diff
@@ -158,14 +163,23 @@ func (ds *daemonSet) WatchDesires(
...
+					// Retry the replication in the RetryInterval's duration
+					timer.Reset(RetryInterval)
+					// This is required in case the user disables the daemon set
+					// so that the timer would be stopped after
+					err = nil
```
anthonybishopric: Do we ever log the error?

KoishiKomeiji: Yeah, it gets logged by the farm, the `Infof` a few lines above are a little more verbose in saying a retry will occur if nothing else happens.

rudle:
:ship:

PR #588, opened 2016-08-08
Create nested cgroup names

bwester:
Add new config option to preparers that changes how they name cgroups.
Instead of creating a singe directory in the cpu and memory resource
controller named "{pod}__{launchable}", the new scheme uses a nested
hierarchy of "p2/{node}/{pod}/{launchable}".

To use the new cgroup name, set the "nested_cgroups" parameter in
the preparer configuration.


mpuncel:
👍

PR #589, opened 2016-08-08
Include response output in remote API call errors

bwester:
It is a hopeless task to debug a JSON decode error without knowing the
input document. Many JSON documents exist in P2 in a durable format
where they can be examined after a failures, e.g., a file on a disk or a
key in Consul. However, the response to a remote API call isn't always
reproducible, so this changes the error messages to include the first 80
characters of the response when querying the artifact registry and an
HTTP label applicator. I am guessing that 80 chars is enough to show
what kind of response was received without overwhelming the logs.

### DIFF
--- a/pkg/artifact/registry.go
+++ b/pkg/artifact/registry.go
@@ -127,7 +127,15 @@ func (a registry) fetchRegistryData(launchableID launch.LaunchableID, version la
 	var registryResponse RegistryResponse
 	err = json.Unmarshal(respBytes, &registryResponse)
 	if err != nil {
-		return nil, auth.VerificationData{}, util.Errorf("Could not unmarshal JSON response from artifact registry: %s", err)
+		l := len(respBytes)
+		if l > 80 {
+			l = 80
+		}
+		return nil, auth.VerificationData{}, util.Errorf(
+			"bad response from artifact registry: %s: %q",
+			err,
+			string(respBytes[:l]),
+		)
 	}
 
 	// Require artifact URL to be present, other fields are optional but returned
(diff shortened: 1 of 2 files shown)


pkg/artifact/registry.go
```diff
@@ -127,7 +127,15 @@ func (a registry) fetchRegistryData(launchableID launch.LaunchableID, version la
 	var registryResponse RegistryResponse
 	err = json.Unmarshal(respBytes, &registryResponse)
 	if err != nil {
-		return nil, auth.VerificationData{}, util.Errorf("Could not unmarshal JSON response from artifact registry: %s", err)
+		l := len(respBytes)
```
anthonybishopric: I really like this pattern. Should we add a very small `json` utility package that can always return errors containing this extra context?

rudle:
:ship:

pkg/artifact/registry.go
bwester: I don't think I fully follow your proposal. Would you want to route all JSON (un)marshaling through a wrapper module that captures its input in its error messages?

I only did this to the API calls because those are the ones where the JSON document is truly ephemeral. And it's more likely that an HTTP response will be something else, like a text or HTML error message.

I suspect the complexity of creating a `p2json` module outweights the benefits of changing the error message.  If you're concerned about repeating the code to do `x[:Min(80, len(x))]`, that could be factored out, but I didn't think in this case it was worth creating a new module for it.

anthonybishopric: Yeah the idea is mostly to abstract the code to extract the message. Fair point about it being specifically useful for HTTP responses. Cool with not extracting it.

anthonybishopric:
👍

PR #591, opened 2016-08-08
Mpuncel/pod-factory

mpuncel:


bwester:
Sorry, my earlier PR also changes the Pod type, hence the conflicts.

bin/p2-restart/main.go
```diff
@@ -24,8 +24,8 @@ $ p2-restart --pod-dir /custom/pod/home
 
 `)
 
-	podName  = restart.Arg("pod-name", fmt.Sprintf("The name of the pod to be restarted. Looks in the default pod home '%s' for the pod", pods.DEFAULT_PATH)).String()
 	nodeName = restart.Flag("node-name", "The name of this node (default: hostname)").String()
+	podName  = restart.Arg("pod-name", fmt.Sprintf("The name of the pod to be restarted. Looks in the default pod home '%s' for the pod", pods.DefaultPath)).String()
```
bwester: nit: Could you keep the two `Flag` definitions together?

mpuncel:
rebased on master

pkg/preparer/orchestrate_test.go
```diff
@@ -85,7 +85,7 @@ func (t *TestPod) Node() types.NodeName {
 	return "hostname"
 }
 
-func (t *TestPod) Path() string {
+func (t *TestPod) Home() string {
```
petertseng: I have a suspicion this would belong in the other commit

bin/p2-launch/main.go
```diff
@@ -55,7 +55,8 @@ func main() {
 		log.Fatalf("%s", err)
 	}
 
-	pod := pods.NewPod(manifest.ID(), types.NodeName(*nodeName), pods.PodPath(*podRoot, manifest.ID()))
+	podFactory := pods.NewFactory(*podRoot, types.NodeName(*nodeName))
+	pod := podFactory.NewPod(manifest.ID())
```
petertseng: I could see this being one line to make it clear this is the only use of this factory. unless you think there will be more later. I wouldn't insist on this before shipping though.

pkg/hooks/hooks_test.go
```diff
@@ -27,8 +27,13 @@ func TestExecutableHooksAreRun(t *testing.T) {
 
 	ioutil.WriteFile(path.Join(tempDir, "test1"), []byte("#!/bin/sh\necho $HOOKED_POD_ID > $(dirname $0)/output"), 0755)
 
+	// So PodFromPodHome doesn't bail out, write a minimal current_manifest.yaml
+	ioutil.WriteFile(path.Join(podDir, "current_manifest.yaml"), []byte("id: my_hook"), 0755)
```
petertseng: and will this get removed after the test?

answered own question. yes. entire podDir gets removed.

petertseng: if this becomes too painful (I do see it's repeated three times) I might suggest putting it in a function

pkg/pods/pod_test.go
```diff
@@ -29,7 +28,9 @@ import (
 )
 
 func getTestPod() *Pod {
-	return NewPod(types.PodID("hello"), "testNode", "/data/pods/test")
+	// Can't use public method because pod home is different than pod id
```
petertseng: what is the public method that you can't use? I only see exported methods being used here, so I think I don't understand the comment

pkg/pods/pod_test.go
```diff
@@ -136,7 +137,8 @@ config:
 
 	podTemp, _ := ioutil.TempDir("", "pod")
 
-	pod := NewPod(manifest.ID(), "testNode", PodPath(podTemp, manifest.ID()))
+	podFactory := NewFactory(podTemp, "testNode")
+	pod := podFactory.NewPod(manifest.ID())
```
petertseng: I can see all these potentially being a single line as well, but I don't care

pkg/pods/pod_test.go
mpuncel: i don't recall what I meant by that so i'll delete it lol

petertseng:
I think I've said all that needs to be said 👍

PR #592, opened 2016-08-08
Un-flake daemon set tests and and fix replication bug

KoishiKomeiji:
Added a waiting function that can wait up to a total of 5 seconds for the farm to finish running when waiting for the ds farm to apply its labels.

I am going to test this on travis CI a couple times to make sure it works.

### DIFF
--- a/pkg/ds/daemon_set.go
+++ b/pkg/ds/daemon_set.go
@@ -24,8 +24,11 @@ import (
 
 const (
 	// This label is applied to pods owned by an RC.
-	DSIDLabel     = "daemon_set_id"
-	RetryInterval = time.Duration(5 * time.Minute)
+	DSIDLabel = "daemon_set_id"
+)
+
+var (
+	retryInterval = time.Duration(5 * time.Minute)
 )
 
 type DaemonSet interface {
@@ -163,9 +166,9 @@ func (ds *daemonSet) WatchDesires(
 			if err != nil {
 				select {
 				case errCh <- err:
-					ds.logger.Warnf("An error has occurred in the daemon set, retrying if no changes are made in %v", RetryInterval)
+					ds.logger.Warnf("An error has occurred in the daemon set, retrying if no changes are made in %v", retryInterval)
 					// Retry the replication in the RetryInterval's duration
-					timer.Reset(RetryInterval)
+					timer.Reset(retryInterval)
 					// This is required in case the user disables the daemon set
 					// so that the timer would be stopped after
 					err = nil
--- a/pkg/ds/daemon_set.go
+++ b/pkg/ds/daemon_set.go
@@ -251,7 +251,17 @@ func (ds *daemonSet) WatchDesires(
 				}
 
 			case <-timer.C:
-				err = ds.PublishToReplication()
+				// Account for any operations that could have failed and retry the replication
+				err = ds.removePods()
+				if err != nil {
+					err = util.Errorf("Unable to remove pods from intent tree: %v", err)
+					continue
+				}
+				err = ds.addPods()
+				if err != nil {
+					err = util.Errorf("Unable to add pods to intent tree: %v", err)
+					continue
+				}
 
 			case <-quitCh:
 				return
--- a/pkg/ds/farm_test.go
+++ b/pkg/ds/farm_test.go
@@ -5,6 +5,9 @@ import (
 	"testing"
 	"time"
 
+	"github.com/square/p2/pkg/replication"
+	"github.com/square/p2/pkg/util"
+
 	"github.com/Sirupsen/logrus"
 	"github.com/square/p2/pkg/alerting"
 	"github.com/square/p2/pkg/kp"
@@ -14,7 +17,6 @@ import (
 	"github.com/square/p2/pkg/logging"
 	"github.com/square/p2/pkg/manifest"
 	"github.com/square/p2/pkg/pods"
-	"github.com/square/p2/pkg/replication"
 	"github.com/square/p2/pkg/scheduler"
 	"github.com/square/p2/pkg/types"
 
@@ -25,15 +27,15 @@ import (
 	klabels "k8s.io/kubernetes/pkg/labels"
 )
 
-func waitForFarm() {
-	// If the tests fail, increasing this value may fix your problem
-	// One reason is beacause of the request rate put on the the
-	// consulutil/watch's WatchDiff function which the farm --> dsStore uses
-	time.Sleep(250 * time.Millisecond)
-}
+const (
+	replicationTimeout    = replication.NoTimeout
+	testFarmRetryInterval = time.Duration(1000 * time.Millisecond)
+)
 
 // Tests dsContends for changes to both daemon sets and nodes
 func TestContendNodes(t *testing.T) {
+	retryInterval = testFarmRetryInterval
+
 	//
 	// Instantiate farm
 	//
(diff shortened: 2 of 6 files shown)


pkg/ds/farm_test.go
```diff
@@ -798,7 +792,42 @@ func TestMultipleFarms(t *testing.T) {
...
+		}
+		return util.Errorf("Expected hasIDLabel to be '%v', got '%v'", hasDSIDLabel, val)
+	}
+
+	err = waitForCondition(condition)
+	return labeled, err
+}
+
+func waitForCondition(condition func() error) error {
+	timeout := time.After(5 * time.Second)
```
rudle: do we still see test flakes if this is set to 1 second? 0.5 seconds? This kinds of delays are very large compared to how quickly most of our test suite runs.

KoishiKomeiji: I'll set it to 1 second check the results.

mpuncel:
It's great you have tests around the farm (I don't think the others do). Typically I think if you're polling for something to occur asynchronously rather than waiting on a channel that's a sign we should restructure it, but I think it's fine here given how difficult it would be to rework the farm. I suspect that would mean providing a custom `Ticker` or something like that which tells the farm loop to execute (instead of using `for { }`). Real implementations could use a fast ticker and tests could control the ticks to make sure the right action is taken.

If this test has a dramatic effect on running the test suite, I'd consider marking it as not-short using https://golang.org/pkg/testing/#Short

KoishiKomeiji:
Updated

pkg/ds/farm_test.go
```diff
@@ -798,7 +812,42 @@ func TestMultipleFarms(t *testing.T) {
...
+}
+
+func waitForCondition(condition func() error) error {
+	timeout := time.After(1 * time.Second)
+	timedOut := false
+	err := condition()
+
+	for err != nil && !timedOut {
+		select {
+		case <-time.Tick(100 * time.Millisecond):
```
bwester: Use `time.After()` instead

pkg/ds/farm_test.go
```diff
@@ -740,6 +751,8 @@ func TestMultipleFarms(t *testing.T) {
 	}
 	_, err = dsStore.MutateDS(anotherDSData.ID, mutator)
 	Assert(t).IsNil(err, "Expected no error mutating daemon set")
+	// Wait double the time because testing multiple farms is that slow
```
bwester: It doesn't seem right to wait an arbitrary amount of time for the farm to be available. Eventually (probably on Travis-CI) you're going to have test runs with high CPU contention and this timeout value isn't going to matter. Is there property you can poll to verify that the farm is in the state you want?

KoishiKomeiji:
Updated, it is now polling instead.

pkg/ds/daemon_set.go
```diff
@@ -24,8 +24,11 @@ import (
 
 const (
 	// This label is applied to pods owned by an RC.
-	DSIDLabel     = "daemon_set_id"
-	RetryInterval = time.Duration(5 * time.Minute)
+	DSIDLabel = "daemon_set_id"
+)
+
+var (
+	retryInterval = time.Duration(5 * time.Minute)
```
mpuncel: nit: could ahve left this in the `const( ... )` block

KoishiKomeiji: I change this value in the tests, so it needs to be in a var.

mpuncel: fair enough

pkg/replication/replicator.go
```diff
@@ -143,10 +143,13 @@ func (r replicator) initializeReplicationWithCheck(
 		errCh:     errCh,
 		replicationCancelledCh: make(chan struct{}),
 		replicationDoneCh:      make(chan struct{}),
+		enactedCh:              make(chan struct{}),
 		quitCh:                 make(chan struct{}),
 		concurrentRealityRequests: make(chan struct{}, concurrentRealityRequests),
 		timeout:                   r.timeout,
 	}
+	// To make a closed channel
+	close(replication.enactedCh)
```
mpuncel: why this change?

KoishiKomeiji: This is for the `Enact()` function in `replication.go` to figure out when a replication is in progress.

pkg/replication/replication.go
```diff
@@ -378,6 +407,8 @@ func (r *replication) ensureHealthy(
 		select {
 		case <-quitCh:
 			return
+		case <-r.replicationCancelledCh:
```
mpuncel: why is this necessary? it looks like the deferred func in `handleRenewalErrors` will close the `quitCh`

pkg/replication/replication.go
```diff
@@ -347,6 +374,8 @@ func (r *replication) ensureInReality(node types.NodeName, quitCh <-chan struct{
...
```
mpuncel: ditto below comment

pkg/replication/replication.go
```diff
@@ -288,6 +312,13 @@ func (r replication) updateOne(node types.NodeName, done chan<- types.NodeName,
 		r.manifest,
 	)
 	for err != nil {
+		select {
+		case <-r.quitCh:
+			return
+		case <-r.replicationCancelledCh:
```
mpuncel: dont' think you need this per other comments

pkg/replication/replication.go
```diff
@@ -278,6 +288,20 @@ func (r replication) handleRenewalErrors(session kp.Session, renewalErrCh chan e
 
 // note: logging should be delegated somehow
 func (r replication) updateOne(node types.NodeName, done chan<- types.NodeName, quitCh <-chan struct{}, aggregateHealth *podHealth) {
+	defer func() {
```
mpuncel: why this change? you don't want to pass a node to `done` unless you've actually updated it right? there are plenty of scenarios (e.g. a quit) where that wouldn't have happened

pkg/replication/replication.go
```diff
@@ -278,6 +288,20 @@ func (r replication) handleRenewalErrors(session kp.Session, renewalErrCh chan e
...
+		select {
+		case done <- node:
+		case <-quitCh:
+		}
+	}()
+
+	select {
+	case <-r.quitCh:
+		return
+	case <-r.replicationCancelledCh:
```
mpuncel: don't think you need to select on cancelledCh (because handleRenewalErrors will close `quitCh` after a cancellation

pkg/replication/replication.go
KoishiKomeiji: This is because the `close(r.quitCh)` is deferred in `handleRenewalErrors`

pkg/replication/replication.go
```diff
@@ -266,6 +272,10 @@ func (r replication) handleRenewalErrors(session kp.Session, renewalErrCh chan e
 	select {
 	case <-r.replicationDoneCh:
 	case <-r.replicationCancelledCh:
+		// If the replication is enacted, wait for it to exit
+		select {
+		case <-r.enactedCh:
```
mpuncel: this is equivalent to `<-r.enactedCh`. I think you probably meant to select on `r.quitCh` in here as well

pkg/replication/replication.go
```diff
@@ -235,7 +241,7 @@ func (r replication) Enact() {
 		select {
 		case <-done:
 			doneIndex++
-			r.logger.Infof("Nodes done (or timed out): %v", doneIndex)
+			r.logger.Infof("Nodes done, timed out, or cancelled: %v", doneIndex)
```
mpuncel: oh, this explains why you did the `defer` change below, although I'm not sure why timed out and canceled nodes would be lumped together

pkg/replication/replication.go
KoishiKomeiji: Actually this `quitCh` is passed by a parameter, it's actually a quitCh for the timeout, it's not related to `r.quitCh`

pkg/kp/kptest/fake_preparer.go
```diff
@@ -0,0 +1,97 @@
...
+						kp.REALITY_TREE,
+						manifestResult.PodLocation.Node,
+						manifestResult.Manifest,
+					)
+					if err != nil {
+						f.logger.Errorf("Error setting pod: %v", err)
+					}
+					delete(podsToDelete, manifestResult)
+				}
+				// Delete pods that are misisng from intent
```
mpuncel: typo on "missing"

pkg/ds/farm_test.go
```diff
@@ -777,28 +781,176 @@ func TestMultipleFarms(t *testing.T) {
...
+			}
+		} else {
+			return util.Errorf("Farm does not have daemon set id: '%v'", ds.ID)
+		}
+		return nil
+	}
+	return waitForCondition(condition)
+}
+
+func waitForMutateSelectorFarms(firstFarm *Farm, secondFarm *Farm, ds ds_fields.DaemonSet) error {
```
mpuncel: comments would be nice above all these wait functions to know what they're intended to do

pkg/ds/farm_test.go
```diff
@@ -777,28 +781,176 @@ func TestMultipleFarms(t *testing.T) {
...
+		return nil
+	}
+
+	err = waitForCondition(condition)
+	return err
+}
+
+func waitForCreate(dsf *Farm, dsID ds_fields.ID) error {
+	condition := func() error {
+		if _, ok := dsf.children[dsID]; !ok {
```
mpuncel: this function (and the others) might be clearer when flipped, e.g. `if ok { return nil }`

pkg/replication/replication.go
KoishiKomeiji: I'll pass a struct to tell whether or not they've timed out or succeeded.

pkg/ds/farm_test.go
```diff
@@ -777,28 +781,176 @@ func TestMultipleFarms(t *testing.T) {
...
+		if dsf.children[newDS.ID].ds.IsDisabled() != isDisabled {
+			return util.Errorf("Unexpected disabled value in farm. Expected '%v', got '%v'",
+				dsf.children[newDS.ID].ds.IsDisabled(),
+				isDisabled,
+			)
+		}
+		return nil
+	}
+
+	err = waitForCondition(condition)
```
mpuncel: nit: just `return waitForCondition(condition)`

pkg/replication/replicator.go
mpuncel: isn't the `Enact()` function itself the thing that should close the channel?

KoishiKomeiji: It's made so that if someone tries to call `WaitForReplication()` they wouldn't block if they didn't call enact.

mpuncel: why would anyone call `WaitForReplication()` if they didn't call `Enact()`?

mpuncel: furthermore, if the channel starts out as closed, `Enact()` will panic because you can't close an already closed channel: https://play.golang.org/p/spvZ45hOlM

KoishiKomeiji: They shouldn't, but I guess it is reasonable to expect the user to use this properly.

pkg/replication/replication.go
mpuncel: you're right that it's a different quitCh, I was mistaken. however my suggestion doesn't change because `updateOne` closes that channel when there is a timeout or if _its_ quit channel is closed, which is done by `Enact()` when `r.quitCh` is closed

pkg/replication/replicator.go
KoishiKomeiji: It actually gets opened again when you call enact.

pkg/replication/replication.go
KoishiKomeiji: Updated, it is now passing a node and a status of what happened.

pkg/replication/replication.go
KoishiKomeiji: Updated, it now handles different cases.

pkg/replication/replication.go
KoishiKomeiji: r.quitCh will not close until the select block is done because it is being deferred.

pkg/ds/farm_test.go
KoishiKomeiji: Updated.

pkg/ds/farm_test.go
KoishiKomeiji: Updated.

pkg/ds/farm_test.go
KoishiKomeiji: Updated.

pkg/kp/kptest/fake_preparer.go
KoishiKomeiji: Updated.

pkg/ds/daemon_set.go
```diff
@@ -204,6 +207,13 @@ func (ds *daemonSet) WatchDesires(
 				err = ds.removePods()
 				if err != nil {
 					err = util.Errorf("Unable to remove pods from intent tree: %v", err)
+					// We must still try to add pods so that we can label the pods that
+					// potentially need to be scheduled because all updates are taken in
+					// only once
```
mpuncel: i'm having trouble understanding this comment, particularly the "all updates are taken in only once"

pkg/ds/daemon_set.go
```diff
@@ -268,7 +278,15 @@ func (ds *daemonSet) handleNodeChanges(changes *labels.LabeledChanges) error {
...
-			return util.Errorf("Unable to remove pods from intent tree: %v", err)
+			err = util.Errorf("Unable to remove pods from intent tree: %v", err)
+			// We must still try to add pods so that we can label the pods that
+			// potentially need to be scheduled because all updates are taken in
+			// only once
+			subErr := ds.addPods()
+			if subErr != nil {
+				err = util.Errorf("%v, Unable to add pods to intent tree: %v", err, subErr)
+			}
+			return err
```
mpuncel: not a part of this PR, but should the `return nil` on line 295 really be there? you don't delete any pods if `len(updated) > 0`?

pkg/replication/replication.go
mpuncel: in that case change this to just `<-r.enactedCh`

pkg/ds/daemon_set.go
KoishiKomeiji: It means that the WatchDiff function will calculate a diff and send out the changes, but it wouldn't ensure consistency between the daemon set labels and the pod labels because it sends a specific change only once.

So then we need to make sure that the pods that are scheduled for a replication will be labelled to belong to this daemon set so that a retryReplication would work.

Another way to handle the retryReplication is to do `removePods()`  and `addPods()`.

pkg/ds/daemon_set.go
KoishiKomeiji: This is because the `removePods()` function will account for any nodes that gets deleted and the `addPods()` will account for any nodes that gets created.

pkg/ds/daemon_set.go
KoishiKomeiji: The code has changed to update and remove pods as accordingly when retrying the replication.

pkg/replication/replicator.go
```diff
@@ -14,11 +15,24 @@ import (
...
+	NoTimeout = time.Duration(-1)
+)
+
+var (
+	errTimeout   = errors.New("Update timed out")
+	errCancelled = errors.New("Replication cancelled")
+	errQuit      = errors.New("Replication quit")
+)
+
+type nodeUpdated struct {
```
mpuncel: this type should be in `replication.go` because that's where it's used

pkg/replication/replication.go
```diff
@@ -370,14 +409,18 @@ func (r *replication) ensureInReality(node types.NodeName, quitCh <-chan struct{
...
+) error {
 	for {
 		select {
-		case <-quitCh:
-			return
+		case <-r.quitCh:
+			return errQuit
+		case <-timeoutCh:
+			return errTimeout
+		case <-r.replicationCancelledCh:
```
mpuncel: I think `replicationCancelled` is going to race with the quit scenario because `handleRenewalErrors` closes quit after a cancellation

pkg/replication/replication.go
```diff
@@ -287,50 +336,31 @@ func (r replication) updateOne(node types.NodeName, done chan<- types.NodeName,
...
+			return errCancelled
+		default:
+			nodeLogger.WithError(err).Errorln("Could not write intent store")
+			r.errCh <- err
+			time.Sleep(1 * time.Second)
```
mpuncel: I wonder if we should consider using an exponential backoff here, e.g. sleep for twice as long after each failure

mpuncel:
@KoishiKomeiji if you've observed the behavior of this PR and are satisfied with it then 👍 from me. It's a little hard to anticipate the ramifications of some of the pkg/replication changes

pkg/replication/replication.go
KoishiKomeiji: This function is called inside enact between the open and closed `r.enactedCh`
so if the replication gets cancelled, it will wait for `r.enactedCh` to close and so this shouldn't be a race case.

This should only close if the `renewalErrCh` gets an error in `handleRevewalErrors(...)`

pkg/replication/replication.go
KoishiKomeiji: Updated

pkg/replication/replicator.go
KoishiKomeiji: Updated

pkg/replication/replication.go
KoishiKomeiji: Updated, this will backup and double its time each pass.

PR #594, opened 2016-08-09
labels: Add "Accept" header to HttpApplicator requests

bwester:
Add an "Accept: application/json" header to the requests sent by the
httpApplicator type. Only JSON is accepted, and the server is only
allowed to return JSON, but having an explicit Accept header plays
nicely with HTTP middleware.


mpuncel:
👍

PR #596, opened 2016-08-10
Add a missing return statement in replication

KoishiKomeiji:
Replication was taking too long and was always timing out because a
return statement was missing.

https://github.com/KoishiKomeiji/p2/blob/087a0abcbadf77d40302db123637866addba310b/pkg/replication/replication.go


rudle:
⏳ 👍 

Good catch

PR #598, opened 2016-08-10
override locks

rudle:
Based on our experience with Daemon Sets in the wild, it is more
valuable to have multiple Daemon Sets executing replications at once
than it is to mutually exclude them to prevent operator error.

Considering the failure cases of overlapping Daemon Set selectors - they
are similar under values of this boolean. If we mutex replications per
pod, then we experience an undefined state of intent/ on any hosts in
the intersection. If we remove the mutex, then this undefined behavior
is more apparent (as it may change multiple times per minute) but it is
no worse or better than in the case using the mutex.


mpuncel:
👍

PR #599, opened 2016-08-10
pods: Write config files robustly

bwester:
When writing a pod's config and platform config files, make the
procedure a little more robust to rare errors:
* Fix nil pointer dereference when the open() fails
* close() earlier and detect its errors

An empty platform config file was observed on one host. I don't have any
particular insight into how that could happen, but these changes might
help.


mpuncel:
🚢

PR #600, opened 2016-08-10
Remove --pod arguments from p2-dsctl

mpuncel:
The argument was used in both the "create" and "update" commands when
creating a daemon set, allowing the daemon set's pod id to differ from
the pod ID in its manifest, which would be an invalid configuration.

This commit does the following:

1) when creating a daemon set, infer the daemon set's PodID field from
the pod ID of the manifest. This removes the possibility of errors where
the two wouldn't match, and also reduces the number of flags necessary
to create a daemon set.
2) When updating a daemon set's manifest, reject the update if the
manifest's ID does not match the pod ID of the daemon set. This reduces
the possibility of error of setting the manifest for a different pod
onto a daemon set.


bwester:
✔️

PR #601, opened 2016-08-10
Improve the "quitting" log messages in p2-preparer

mpuncel:
There are two routines that both had identical "Quitting..." log
messages, and it's unclear what they mean. Now the messages are more
specific to the functionality that is stopping.

Furthermore, the routine watching for pod updates would make it look
like the pod being managed itself is quitting, rather than the
p2-preparer quitting (usually for an update). This can cause confusion
as it makes it look like all of the preparer's pods are being shut down
at the same time.


bwester:
🚢

PR #602, opened 2016-08-10
hooks: Add convenience function to get node name in hooks

bwester:
Add the HookEnv.Node() method to allow hooks to easily get the name of
the node they're running on. This is similar to what is done in cdb4a9c.

Also refactors the PodFromDisk() method to use Node() (et al.) to avoid
repeating error messages.


mpuncel:
👍

PR #603, opened 2016-08-10
Log the cancellation of a replication.

mpuncel:
Daemon sets in particular start and cancel replication under a variety
of circumstances, it makes sense to log this event so that it is easier
to observe the behavior of a daemon set.


KoishiKomeiji:
We should remove this line
https://github.com/square/p2/blob/master/pkg/ds/daemon_set.go#L524

PR #604, opened 2016-08-11
Make the roll farm robust to RUs with missing RCs.

mpuncel:
If a rolling update exists in consul and is missing either its new or
old RC, the rolling update cannot be handled because the enable/disable
step at the beginning of the roll handler will fail.

This commit validates that both RCs in an RU exist before attempting to
run it. If one of the RCs is missing, the RU is deleted.

### DIFF
--- a/pkg/roll/farm.go
+++ b/pkg/roll/farm.go
@@ -1,6 +1,7 @@
 package roll
 
 import (
+	"fmt"
 	"sync"
 	"time"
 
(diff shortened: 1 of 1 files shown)


mpuncel:
Open question: If the old RC is missing, should this also validate that the new RC is enabled and if not, enable it?

mpuncel:
an alternative to this approach would be to push this logic into the up

mpuncel:
rebased on master

bwester:
This looks fine to me. 🚢

PR #605, opened 2016-08-11
Remove buggy check in hookenv.PodFromDisk()

mpuncel:
The PodFromDisk() function returned an error if there was no
current_manifest.yaml on the disk for that node. This was to protect
agains thooks loadinga pod struct from disk before it is placed.

However, some hooks are knowingly loading the pod after it has been
installed but before the current manifest exists.

This check has been replaced with a check that the hook event is
BEFORE_INSTALL, in which case the pod should not be loaded from disk.


bwester:
👍

PR #607, opened 2016-08-15
Minor updates to session publisher

bwester:
* p2-ds-farm: Remove string publisher

  One subscriber doesn't need a publisher

* stream: Remove user-provided channels for subscriptions

  It was never used

* p2-rctl-server: Move Subscribe() call near its use

  Good hygiene

### DIFF
--- a/pkg/util/stream/publisher_test.go
+++ b/pkg/util/stream/publisher_test.go
@@ -68,7 +68,7 @@ func TestLastValue(t *testing.T) {
 	input := make(chan string)
 	p := NewStringValuePublisher(input, "init")
 
-	s1 := p.Subscribe(nil)
+	s1 := p.Subscribe()
 	if v := <-s1.Chan(); v != "init" {
 		t.Error("did not receive initial value. got:", v)
 	}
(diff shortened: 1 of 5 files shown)


bwester:
Any thoughts on this PR? The changes are fairly small and I would expect non-controversial.

pkg/util/stream/publisher_test.go
```diff
@@ -140,16 +140,38 @@ func TestAlwaysClosed(t *testing.T) {
...
+		}
+		if val := <-s1.Chan(); val != "hello" {
+			t.Error("dropped second value, got", val)
+		}
+		s1.Unsubscribe()
+		close(unsubscribed)
+
+		// No more messages should be received, but we can't wait forever to verify that.
+		time.Sleep(20 * time.Millisecond)
+		select {
```
petertseng: it doesn't seem to necessarily be the case that this will run after the `input <- "world"` and `input <- "this shouldn't block"`. do you mean for it to?

petertseng:
s/Chane/Change/ on acfc2dd I think.

pkg/util/stream/publisher_test.go
bwester: I didn't originally plan on enforcing that ordering because the original didn't have that ordering. But since you point it out, that does make a lot of sense.

petertseng:
I think the `pkg/util/stream/publisher_test.go` changes should get added to the "stream:" commit rather than the "p2-rctl-server:" command (as I currently see them)

👍  if that is done

bwester:
git rebase fail. Thanks for noticing that

petertseng:
Right then, everything seems to be in order 👍

PR #608, opened 2016-08-16
stream: Buffer subscriber output

bwester:
Return a buffered output channel when subscribing to a
StringValuePublisher. A slow reader will skip some updates instead of
blocking all other subscribers.

This eliminates a potential problem when sharing one publisher among
many different users, not all of which are equally responsive. In
practice, the subscribers used in P2 are already written in such a way
that they don't care if they receive all updates.

Buffering output also makes the publisher code easier, since it doesn't
have to deal with sends potentially blocking.


rudle:
:+1:

petertseng:
👍  as well.

PR #609, opened 2016-08-16
Add functions to daemon sets for graphQL

KoishiKomeiji:


pkg/kp/dsstore/consul_store.go
```diff
@@ -326,6 +326,61 @@ func (s *consulStore) Watch(quitCh <-chan struct{}) <-chan WatchedDaemonSets {
...
+// Watch watches dsTree for changes and returns a blocking channel
+// where the client can read a WatchedDaemonSets object which contain
+// changed daemon sets
+func (s *consulStore) WatchAll(quitCh <-chan struct{}) <-chan WatchedDaemonSetList {
+	inCh := make(chan api.KVPairs)
+	outCh := make(chan WatchedDaemonSetList)
+	errCh := make(chan error, 1)
+
+	// Watch for changes in the dsTree and deletedDSTree
+	go consulutil.WatchPrefix(dsTree, s.kv, inCh, quitCh, errCh, time.Duration(0))
```
rudle: In the interest of self-documenting code, can you change the duration to match what's actually happening here? (ie. change this to 250 ms)

```	if pause < 250*time.Millisecond {
		pause = 250 * time.Millisecond
	}
``` is near the top of WatchPrefix

**Comment by rudle** [comment]
This looks pretty OK to me! 👍

**Comment by KoishiKomeiji** [comment]
Thanks for the review! I will merge when CI passes.

File: pkg/kp/dsstore/consul_store.go (hunk 1 above)
Comment (KoishiKomeiji) [inline pkg/kp/dsstore/consul_store.go]: Updated.

File: pkg/kp/dsstore/consul_store.go [hunk 2]
```diff
@@ -326,6 +326,61 @@ func (s *consulStore) Watch(quitCh <-chan struct{}) <-chan WatchedDaemonSets {
...
+		defer close(outCh)
+		defer close(errCh)
+
+		var kvp api.KVPairs
+		for {
+			select {
+			case <-quitCh:
+				return
+			case err := <-errCh:
+				s.logger.WithError(err).Errorf("WatchPrefix returned error, recovered.")
```
petertseng: in this case, the watch will send its current daemon sets on the channel, even though they probably haven't changed. is that OK?

KoishiKomeiji: Yes, this is what was intended, this is meant for graphQL to easily populate its cache so anyone can read what is on the tree without directly calling consul.

petertseng: Hmm, OK. As long as no scheduling decisions are made on it. Because if the first thing received is an error, then this will send an empty list of daemon sets, right? I just didn't want someone thinking "that means there are no daemon sets. I will unschedule all of them", but I guess they'd use `Watch` for that.

PR #611, opened 2016-08-19
Add p2-norm command

bwester:
Add the "p2-norm" command. It reads a pod manifest and prints it in a
normalized format. This command is useful when comparing manifests where
one or both are hand-edited and may differ in trivial ways, like the
ordering of keys or spacing.


mpuncel:
👍

rudle:
:+1:

PR #612, opened 2016-08-19
Add a WatchAllPods function

KoishiKomeiji:


mpuncel:
nit: our graphQL stuff is internal, don't mention it in public commit messages

pkg/kp/kv.go
```diff
@@ -378,6 +379,39 @@ func (c consulStore) WatchPods(
...
+func (c consulStore) WatchAllPods(
+	podPrefix PodPrefix,
+	quitChan <-chan struct{},
+	errChan chan<- error,
+	podChan chan<- []ManifestResult,
+) {
+	defer close(podChan)
+
+	kvPairsChan := make(chan api.KVPairs)
+	go consulutil.WatchPrefix(string(podPrefix), c.client.KV(), kvPairsChan, quitChan, errChan, 0)
```
mpuncel: you should expose the `pause` (i.e. `0`) parameter to `WatchPrefix` to the caller, our /intent tree is over 40MB which means the fetch might take on the order of 10s of seconds. We can tune consul load by pausing between updates as necessary

mpuncel:
looks good, once you expose the pause parameter

pkg/kp/kv.go
KoishiKomeiji: Alright I changed it to 10 seconds.

KoishiKomeiji:
It's been changed to take in a parameter now.

mpuncel:
👍

KoishiKomeiji:
Thanks!

PR #613, opened 2016-08-22
Remove IsUUID field from PodUniqueKey struct

mpuncel:
The field was obviously used to differentiate PodUniqueKey structs that
signify a UUID from those that don't. Prior to this commit, the ID field
would be the consul key e.g. /intent/<node>/<pod_id> if IsUUID was
false.

After making use of PodUniqueKey, it's becoming clear that this struct
is ONLY useful in contexts where it's signifying a UUID. As a result,
this commit removes the IsUUID field and uses a nil pointer to a
PodUniqueKey to signify a legacy pod that does not have a UUID.

### DIFF
--- a/pkg/kp/kv.go
+++ b/pkg/kp/kv.go
@@ -6,7 +6,6 @@ import (
 	"bytes"
 	"encoding/json"
 	"fmt"
-	"path"
 	"strings"
 	"time"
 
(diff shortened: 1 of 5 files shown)


rudle:
Values that can take on nil are generally scary to me as they introduce the potential for runtime errors. It does seem redundant to have a boolean that tracks whether the pointer is nil or not, so I can support this change.

rudle:
👍

PR #615, opened 2016-08-22
Add json tags to PodLocation

KoishiKomeiji:
This is done so that PodLocation can be used as an index in a map


mpuncel:
nit on commit message, the JSON tags don't let it be used as a map index, it's so they can be embedded in other structures intended to be marshaled as JSON

mpuncel:
otherwise 🚢

KoishiKomeiji:
Alright, fixed the commit message, will merge now.

PR #616, opened 2016-08-22
Allow filtering daemon sets by pod

drcapulet:


rudle:
I've been doing `| grep` but this works too

bin/p2-dsctl/main.go
```diff
@@ -56,6 +56,7 @@ var (
 	getID  = cmdGet.Arg("id", "The uuid for the daemon set").Required().String()
 
 	cmdList = kingpin.Command(CmdList, "List daemon sets.")
+	listPod = cmdList.Flag("pod", "The pod ID on the daemon set").String()
```
rudle: typo: "of the daemon set"

drcapulet:
I've been using `grep` as well but doesn't work as well when you've got pods that share a piece of the name.

mpuncel:
fine for now, labels might be a more elegant solution later on but easy to change

PR #617, opened 2016-08-23
Mpuncel/pod status improvements

mpuncel:


pkg/kp/statusstore/statusstoretest/fake_status_store.go
```diff
@@ -43,24 +50,25 @@ func (s *FakeStatusStore) SetStatus(
...
-) (statusstore.Status, error) {
+) (statusstore.Status, *api.QueryMeta, error) {
 	s.mu.Lock()
 	defer s.mu.Unlock()
 
 	identifier := StatusIdentifier{t, id, namespace}
 	status, ok := s.Statuses[identifier]
 	if !ok {
-		return statusstore.Status{}, statusstore.NoStatusError{identifier.String()}
+		return statusstore.Status{}, nil, statusstore.NoStatusError{identifier.String()}
```
bwester: `(*consulStore).GetStatus()` returns a valid QueryMeta object along with a `NoStatusError` error value. Can you make these behaviors consistent?

mpuncel:
made a few fixes to this today after observing it in practice. Particularly using `MutateStatus()` didn't work if the key didn't already exist

pkg/kp/statusstore/podstatus/status.go
```diff
@@ -23,6 +24,10 @@ const (
 	Pending  ServiceState = "pending"
 	Running  ServiceState = "running"
 	Finished ServiceState = "finished"
+
+	PodLaunched PodState = "launched"
```
rudle: I can't see where these are used in this PR so I must express my concern that the strings are duplicated with the above states. If it's possible for these to appear under the same field then that may confuse debugging.

mpuncel: I'm hoping that the types being different (`PodState` vs `ServiceState`) will prevent this. You're right though that one could be unmarshaled into the wrong type though. Should I modify the strings e.g. `pod_launched` `pod_pending`?

rudle: wow, that would be a heck of a bug! It's probably most safe to prefix them all with "pod_"

rudle:
I can't find much that's giving me trouble here. Given that this has a bit of battle testing already, this seems OK to merge

PR #618, opened 2016-08-25
Add command to reset reality for a host

anthonybishopric:
This allows a host that has been manipulated by hand to be fully restored to the state set in Consul. For discussion: should we not operate on the preparer itself?

### DIFF
--- a/pkg/kp/flags/kingpin.go
+++ b/pkg/kp/flags/kingpin.go
@@ -3,6 +3,8 @@
 package flags
 
 import (
+	"io/ioutil"
+	"log"
 	"net/http"
 
 	"github.com/square/p2/pkg/kp"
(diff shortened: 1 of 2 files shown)


mpuncel:
looks good to me, not sure what you mean by "should we not operate on the preparer itself".

pkg/kp/flags/kingpin.go
```diff
@@ -13,11 +14,20 @@ import (
...
 	token := kingpin.Flag("token", "The consul ACL token to use. Empty by default.").String()
+	tokenFile := kingpin.Flag("token-file", "The file containing the Consul ACL token").ExistingFile()
 	headers := kingpin.Flag("header", "An HTTP header to add to requests, in KEY=VALUE form. Can be specified multiple times.").StringMap()
 	https := kingpin.Flag("https", "Use HTTPS").Bool()
 	wait := kingpin.Flag("wait", "Maximum duration for Consul watches, before resetting and starting again.").Default("30s").Duration()
 
 	cmd := kingpin.Parse()
+
+	if *tokenFile != "" {
+		tokenBytes, err := ioutil.ReadAll(*tokenFile)
```
mpuncel: looks like you want `ReadFile()`

pkg/kp/flags/kingpin.go
```diff
@@ -13,11 +14,20 @@ import (
...
+		if err != nil {
+			panic(err)
```
mpuncel: `log.Fatal` is probably nicer than panic?

anthonybishopric:
As in, should the preparer be removed from reality as well? This may mean the preparer restarts itself. Our preparer's failsafe looks at intent results to decide whether it should take action, and experimentally we saw that this works as is, but I wanted to bring it up just in case I didn't consider a case.

pkg/kp/flags/kingpin.go
anthonybishopric: Yes.

pkg/kp/flags/kingpin.go
anthonybishopric: Good catch, thanks

mpuncel:
I think the preparer restarting is harmless at worst, if it didn't come up properly after a host reimage we'd probably want to know

mpuncel:
👍

eam:
👍

PR #620, opened 2016-08-26
Double the hook timeout

rudle:
some hooks stop services that take as long as 60s to cleanly shutdown.


anthonybishopric:
sad thumbs 👍

bwester:
Can you expose this in the preparer's configuration?

PR #621, opened 2016-08-26
add TLS key flags to CLI defaults

anthonybishopric:
This addresses an internal issue related to calling to Consul from a CLI.

### DIFF
--- a/pkg/kp/flags/kingpin.go
+++ b/pkg/kp/flags/kingpin.go
@@ -5,10 +5,11 @@ package flags
 import (
 	"io/ioutil"
 	"log"
+	"net"
 	"net/http"
 
 	"github.com/square/p2/pkg/kp"
-	"github.com/square/p2/pkg/util/net"
+	netutil "github.com/square/p2/pkg/util/net"
 	"gopkg.in/alecthomas/kingpin.v2"
 )
 
@@ -19,6 +20,9 @@ func ParseWithConsulOptions() (string, kp.Options) {
 	headers := kingpin.Flag("header", "An HTTP header to add to requests, in KEY=VALUE form. Can be specified multiple times.").StringMap()
 	https := kingpin.Flag("https", "Use HTTPS").Bool()
 	wait := kingpin.Flag("wait", "Maximum duration for Consul watches, before resetting and starting again.").Default("30s").Duration()
+	caFile := kingpin.Flag("tls-ca-file", "File containing the x509 PEM-encoded CA ").ExistingFile()
+	keyFile := kingpin.Flag("tls-key-file", "File containing the x509 PEM-encoded private key").ExistingFile()
+	certFile := kingpin.Flag("tls-cert-file", "File containing the x509 PEM-encoded public key certificate").ExistingFile()
 
 	cmd := kingpin.Parse()
 
@@ -29,10 +33,28 @@ func ParseWithConsulOptions() (string, kp.Options) {
 		}
 		*token = string(tokenBytes)
 	}
+	var transport http.RoundTripper
+	if *caFile != "" || *keyFile != "" || *certFile != "" {
+		tlsConfig, err := netutil.GetTLSConfig(*certFile, *keyFile, *caFile)
+		if err != nil {
+			log.Fatalln(err)
+		}
+
+		transport = &http.Transport{
+			TLSClientConfig: tlsConfig,
+			// same dialer as http.DefaultTransport
+			Dial: (&net.Dialer{
+				Timeout:   http.DefaultClient.Timeout,
+				KeepAlive: http.DefaultClient.Timeout,
+			}).Dial,
+		}
+	} else {
+		transport = http.DefaultTransport
+	}
 	return cmd, kp.Options{
 		Address:  *url,
 		Token:    *token,
-		Client:   net.NewHeaderClient(*headers, http.DefaultTransport),
+		Client:   netutil.NewHeaderClient(*headers, transport),
 		HTTPS:    *https,
 		WaitTime: *wait,
 	}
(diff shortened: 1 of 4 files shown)


pkg/util/net/client.go
```diff
@@ -31,7 +31,9 @@ func GetTransport(extras map[string]string, rt http.RoundTripper) *headerTranspo
 // on every request.
 func NewHeaderClient(extras map[string]string, rt http.RoundTripper) *http.Client {
 	if len(extras) == 0 {
-		return http.DefaultClient
+		return &http.Client{
```
bwester: This seems like an unrelated bug fix. Can you put it in its own commit and discuss the triggers/impact of the bug?

pkg/kp/flags/kingpin.go
```diff
@@ -5,10 +5,11 @@ package flags
 import (
 	"io/ioutil"
 	"log"
+	"net"
 	"net/http"
 
 	"github.com/square/p2/pkg/kp"
-	"github.com/square/p2/pkg/util/net"
+	netutil "github.com/square/p2/pkg/util/net"
```
bwester: I could swear a different alias for this package is in use elsewhere. Could you look for that alias and make this consistent?

pkg/kp/flags/kingpin.go
```diff
@@ -19,6 +20,9 @@ func ParseWithConsulOptions() (string, kp.Options) {
 	headers := kingpin.Flag("header", "An HTTP header to add to requests, in KEY=VALUE form. Can be specified multiple times.").StringMap()
 	https := kingpin.Flag("https", "Use HTTPS").Bool()
 	wait := kingpin.Flag("wait", "Maximum duration for Consul watches, before resetting and starting again.").Default("30s").Duration()
+	caFile := kingpin.Flag("tls-ca-file", "File containing the x509 PEM-encoded CA ").ExistingFile()
+	keyFile := kingpin.Flag("tls-key-file", "File containing the x509 PEM-encoded private key").ExistingFile()
```
bwester: Please verify that a private key file is x509 PEM-encoded.

pkg/kp/flags/kingpin.go
```diff
@@ -29,10 +33,31 @@ func ParseWithConsulOptions() (string, kp.Options) {
 		}
 		*token = string(tokenBytes)
 	}
+	var transport http.RoundTripper
+	if *caFile != "" || *keyFile != "" || *certFile != "" {
```
bwester: I'm not sure why you want all or nothing. It seems reasonable to me that one might use the default certificate authorities while supplying a client certificate.

pkg/kp/flags/kingpin.go
```diff
@@ -29,10 +33,31 @@ func ParseWithConsulOptions() (string, kp.Options) {
...
+		if *caFile == "" || *keyFile == "" || *certFile == "" {
```
bwester: GetTLSConfig already validates its arguments; this check seems redundant.

bin/p2-wipe-reality/main.go
```diff
@@ -19,6 +21,16 @@ a known state.
 `
 )
 
+func handleError(err error, message string) {
```
bwester: I can see the benefit in wanting to unpack the hidden error messages. But:

1. What if I eventually want to do something other than print the error? Can you unlock without side effects?
2. This is unrelated to adding TLS flags. Can you put it in its own commit?

pkg/kp/flags/kingpin.go
```diff
@@ -19,6 +20,9 @@ func ParseWithConsulOptions() (string, kp.Options) {
 	headers := kingpin.Flag("header", "An HTTP header to add to requests, in KEY=VALUE form. Can be specified multiple times.").StringMap()
 	https := kingpin.Flag("https", "Use HTTPS").Bool()
 	wait := kingpin.Flag("wait", "Maximum duration for Consul watches, before resetting and starting again.").Default("30s").Duration()
+	caFile := kingpin.Flag("tls-ca-file", "File containing the x509 PEM-encoded CA ").ExistingFile()
```
bwester: Having "tls-" seems wordy, and I don't want to ever type that.  What do you think about removing that prefix? Incidentally, would that make these flags the same as for curl?

bin/p2-wipe-reality/main.go
anthonybishopric: I can remove it, it was meant for debugging why this utility wasn't working. Dealing with the unsafe errors in Consul isn't something I care to solve in this PR.

pkg/kp/flags/kingpin.go
anthonybishopric: It's mapped to `p2net` in a different repo at Square. Still want it to be consistent?

pkg/kp/flags/kingpin.go
anthonybishopric: The curl options are `--cert`, `--cacert` and `--key`. For at least the last option, `key` is a fairly ambiguous option given that it could refer to a key in the kv store or a gpg signing key. I can remove the `tls-` prefix and I think that will make it sufficiently less verbose.

pkg/kp/flags/kingpin.go
anthonybishopric: Will the `util/net` helper not do that for me?

pkg/util/net/client.go
anthonybishopric: extracted

pkg/kp/flags/kingpin.go
bwester: Sorry, not sure what I was thinking. What you have will work just fine.

pkg/kp/flags/kingpin.go
bwester: You're right, there are a lot of other keys in the system. Sure, keep `tls-`.

pkg/kp/flags/kingpin.go
bwester: Perhaps the real solution is to not have our own package named "net".

Pick your favorite.

bwester:
:ship:

PR #622, opened 2016-08-26
Fix RU bug if reality manifest is missing.

mpuncel:
This commit fixes a bug that would occur when attempting to count the
number of healthy nodes in a replication controller if there is no
reality manifest for the pod in question.

This can easily occur if a pod has been scheduled for the first time
on a bad host which will result in the reality manifest remaining empty.

### DIFF
--- a/pkg/roll/run_update.go
+++ b/pkg/roll/run_update.go
@@ -15,6 +15,7 @@ import (
 	"github.com/square/p2/pkg/kp/rcstore"
 	"github.com/square/p2/pkg/labels"
 	"github.com/square/p2/pkg/logging"
+	"github.com/square/p2/pkg/pods"
 	"github.com/square/p2/pkg/rc"
 	rcf "github.com/square/p2/pkg/rc/fields"
 	"github.com/square/p2/pkg/roll/fields"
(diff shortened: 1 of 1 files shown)


pkg/roll/run_update.go
```diff
@@ -402,9 +403,11 @@ func (u *update) countHealthy(id rcf.ID, checks map[types.NodeName]health.Result
 		node := pod.Node
 		// TODO: is reality checking an rc-layer concern?
 		realManifest, _, err := u.kps.Pod(kp.REALITY_TREE, node, rcFields.Manifest.ID())
-		if err != nil {
+		if err != nil && err != pods.NoCurrentManifest {
 			return ret, err
 		}
+
+		// even if realManifest is nil, we'll get empty tring for SHA which is fine for comparison purposes.
```
anthonybishopric: nit `string`

mpuncel:
I tested this change including with rolling over an already-running app to P2

pkg/roll/run_update.go
```diff
@@ -402,9 +403,11 @@ func (u *update) countHealthy(id rcf.ID, checks map[types.NodeName]health.Result
 		node := pod.Node
 		// TODO: is reality checking an rc-layer concern?
 		realManifest, _, err := u.kps.Pod(kp.REALITY_TREE, node, rcFields.Manifest.ID())
-		if err != nil {
+		if err != nil && err != pods.NoCurrentManifest {
 			return ret, err
 		}
+
+		// even if realManifest is nil, we'll get empty string for SHA which is fine for comparison purposes.
```
bwester: `realManifest` has type `manifest.Manifest`, an interface, and you can't call any method on a `nil` interface value.

mpuncel: I think you're right. I was confused because of situations like https://play.golang.org/p/ABUMhhPWL1, but in that case it's a `nil` with a concrete type that can handle a nil pointer.

bwester:
:ship: pending a successful CI run (failure looks unrelated to me)

PR #623, opened 2016-08-30
Skip hardlinks to themselves, as created by gnu tar

mcpherrinm:
There's not much else we could do with a tar file like this.


bwester:
This seems reasonable to me.

PR #626, opened 2016-08-30
Increase WatchPrefix latency where appropriate

rudle:
We observe a lot of network and disk read IO and we suspect it's related
to hungry watches on big trees. Presently, our consul trees are large:
/labels = 7 Megabytes
/intent,/reality = 57 Megabytes
/replication_controllers = 2.5 Megabytes

We need to be more cognizant about the costs of this data access pattern as it relates to our current datastore. I note, without comment, that differential watches are not coming soon to consul.

### DIFF
--- a/pkg/kp/rcstore/consul_store.go
+++ b/pkg/kp/rcstore/consul_store.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"path"
 	"strings"
+	"time"
 
 	"github.com/hashicorp/consul/api"
 	"github.com/pborman/uuid"
@@ -184,7 +185,7 @@ func (s *consulStore) WatchNew(quit <-chan struct{}) (<-chan []fields.RC, <-chan
 	inCh := make(chan api.KVPairs)
 
 	outCh, errCh := publishLatestRCs(inCh, quit)
-	go consulutil.WatchPrefix(rcTree+"/", s.kv, inCh, quit, errCh, 0)
+	go consulutil.WatchPrefix(rcTree+"/", s.kv, inCh, quit, errCh, 1*time.Second)
 
 	return outCh, errCh
 }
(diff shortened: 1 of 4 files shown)


pkg/kp/rcstore/consul_store.go
```diff
@@ -184,7 +184,7 @@ func (s *consulStore) WatchNew(quit <-chan struct{}) (<-chan []fields.RC, <-chan
 	inCh := make(chan api.KVPairs)
 
 	outCh, errCh := publishLatestRCs(inCh, quit)
-	go consulutil.WatchPrefix(rcTree+"/", s.kv, inCh, quit, errCh, 0)
+	go consulutil.WatchPrefix(rcTree+"/", s.kv, inCh, quit, errCh, 0) // TODO
```
rudle: I am unsure about what to do here!

pkg/kp/rcstore/consul_store.go
```diff
@@ -214,7 +214,7 @@ func (s *consulStore) WatchNewWithRCLockInfo(quit <-chan struct{}) (<-chan []RCL
 	combinedErrCh := make(chan error)
 
 	rcCh, rcErrCh := publishLatestRCs(inCh, quit)
-	go consulutil.WatchPrefix(rcTree+"/", s.kv, inCh, quit, rcErrCh, 0)
+	go consulutil.WatchPrefix(rcTree+"/", s.kv, inCh, quit, rcErrCh, 0) // TODO
```
rudle: here also!
It may be OK to leave these as the RC tree is not huge.

pkg/kp/rcstore/consul_store.go
mpuncel: you could include it as an argument to `WatchNew`, I actually think 2.5M is enough to drive high bandwidth usage given the RC tree triggers many times per RU

rudle: I think I'll set it to 1 second here. There are too many callsites to manage

rudle:
This is now reviewable: commits are specific and TODOs have been addressed.

mpuncel:
👍 my one question is whether it makes sense to let the caller adjust these parameters but we can tweak that later if we deem it necessary

bwester:
🚢

PR #627, opened 2016-08-31
Mpuncel/check roll quit channel

mpuncel:

PR #628, opened 2016-08-31
Mpuncel/rewrite watch prefix

mpuncel:


mpuncel:
This PR is sort of a proposal, I think we have a lot of non-idiomatic control flows in our codes that lead to bugs like the one (imperfectly) addressed by https://github.com/square/p2/pull/627

pkg/kp/consulutil/watch.go
```diff
@@ -64,6 +65,69 @@ func WatchPrefix(
...
 }
 
+type PairsResult struct {
+	Pairs api.KVPairs
+	Err   error
+}
+
+// Like WatchPrefix(), but has a different and better control flow. WatchPrefix() takes its
+// two output channels (outPairs and outErrors) as arguments, as well as a quit channel.
+// This means that the caller of WatchPrefix is responsible both for closing the output
```
bwester: `WatchPrefix()` does close its output channel [[L15](https://github.com/square/p2/pull/628/files#diff-bd93e88e13ecea60af2f8f61eddfc30aL15), [L27](https://github.com/square/p2/pull/628/files#diff-bd93e88e13ecea60af2f8f61eddfc30aL27)] when signaled to stop watching.

bwester:
Since you reference #627, how would `WatchPrefixNew()` have prevented that error?

bwester:
> This means callers of WatchPrefixNew() can simply exit when the output channel is closed, for example using "range" inside of a "for".

Callers could do that anyway. Checking for another quit signal is a means for closing down many goroutines concurrently instead of making their closed output channels cascade.

bwester:
I think there is a pervasive problem with error handling in P2's code: error reporting is performed by writing error messages into a `chan<- error`. As you pointed out, this complicates callers. It also obscures what happens to those errors. If you trace through all the indirection, inevitably those errors are simply printed to the log.

If you want to clean up the interface, perhaps we should be passing around logger instances instead of `chan<- error`.

Using `chan struct{Value, error}` would make sense when the error affects the caller in some way, but usually errors are merely logged and execution continues as if nothing bad happened.

bwester:
`WatchPrefix()` is blocking and accepts an output channel so that it doesn't have to construct the channel (buffered or not) or impose asynchronous operation on the caller. This follows many golang library patterns (look at the I/O library).

But I do recognize that the only way `WatchPrefix()` has been used in practice has been asynchronous with a new, unbuffered channel. For this use case, perhaps you could wrap up the boilerplate into a `WatchPrefixAsync()` method?

PR #629, opened 2016-08-31
Add a random sleep in session conroller before reestablishing a session

mpuncel:
This aims to mitigate a thundering herd problem after a consul server
invalidates a large amounts of sessions. Every p2-preparer instance as
well as any deployed farms that lose their sessions will attempt to
establish a new session around the same time, which can put extreme
write pressure on consul.

### DIFF
--- a/pkg/kp/consulutil/session_controller.go
+++ b/pkg/kp/consulutil/session_controller.go
@@ -1,6 +1,7 @@
 package consulutil
 
 import (
+	"math/rand"
 	"time"
 
 	"github.com/Sirupsen/logrus"
@@ -35,6 +36,8 @@ func SessionManager(
 	logger logging.Logger,
 ) {
 	logger.NoFields().Info("session manager: starting up")
+
+	timer := time.NewTimer(0)
 	for {
 		// Check for exit signal
 		select {
@@ -42,7 +45,7 @@ func SessionManager(
 			logger.NoFields().Info("session manager: shutting down")
 			close(output)
 			return
-		default:
+		case <-timer.C:
 		}
 		// Establish a new session
 		id, _, err := client.Session().CreateNoChecks(&config, nil)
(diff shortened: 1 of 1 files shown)


mpuncel:
Happy to discuss this. Is 10 seconds long enough? is 10 seconds too long? any smarter way to tell when to back off? testing? etc

pkg/kp/consulutil/session_controller.go
```diff
@@ -65,6 +68,11 @@ func SessionManager(
 				sessionLogger.NoFields().Info("session manager: released session")
 			}
 			output <- ""
+
+			// Wait a random amount of time between 0 and 10 seconds before attempting to
+			// re-establish a session. This protects against a thundering herd of sessions
+			// attempting to be re-establish
```
rudle: `rand.Seed(…)` needs to happen somewhere around here or this won't be all that random. `time.Now()` is a popular choice though of course it's not suitable to use here at this line of code because time.Now() might be the same for all callers!

rudle:
Assuming a uniform distribution, the way I'd figure out if 10s was enough would be `n / t = x` where x is the desired write rate at consul and n is the number of sessions we expect to lose (maybe ~1000 in our case). 10s would result in between 100-200 writes per sec which seems a bit high.

Further, I can't tell what distribution go uses for the function you call in this PR though I didn't do all that much research.

pkg/kp/consulutil/session_controller.go
bwester: `rand.Seed()` should only need to be called once per process to initialize the package's default entropy source. I suggest this call should go in `bin/p2-preparer/main.go` and be called with a hash of the current time and the node name or hostname.

bwester:
I think this change makes it possible (nondeterministically) for a new session to be created after the `done` channel has been closed.

bwester:
There is already a different backoff that happens after a failed call to create a session. Should these follow the same strategy? Is that one insufficient?

bwester:
Be careful not to delay the session manager from quitting when it wants to terminate the session.

bwester:
If your only goal is to prevent a thundering herd, perhaps an exponential backoff would be useful in addition to the jitter. I found a nice discussion here: https://www.awsarchitectureblog.com/2015/03/backoff.html

PR #630, opened 2016-08-31
Mpuncel/ds rate limit

mpuncel:


rudle:
What's the desired effect here? Based on my understanding of this goroutine inside of a goroutine  (>_<) we'll fork a call to `updateOne()` inside of all outer goroutines whenever the ticker fires. This means we'll update 50 pods simultaneously but not more than once a second.

Another approach is to use a semaphore channel of some size and limit the number of in flight requests to consul based on that.

WDYT? I'm leaning towards the latter but what's here is probably safer than what we have currently.

mpuncel:
you're right, in my haste i'm pulling a value off the channel at the wrong point

rudle:
I've convinced myself that this is an OK pattern. https://play.golang.org/p/BR2Qfkds3h

rudle:
LGTM. Please clean up the commits before merging.

pkg/replication/replication.go
```diff
@@ -305,6 +315,7 @@ func (r *replication) handleRenewalErrors(session kp.Session, renewalErrCh chan
 	defer func() {
 		close(r.quitCh)
 		close(r.errCh)
+		r.rateLimiter.Stop()
```
bwester: Can you guarantee that all workers (those created in [L219](https://github.com/square/p2/pull/630/files#diff-7ead7087026253e988173f0362b9bd81L219)) have started by the time you stop the timer? If not, they never will.

pkg/replication/replication.go
```diff
@@ -221,6 +225,12 @@ func (r *replication) Enact() {
 				exitCh := make(chan struct{})
 				timeoutCh := make(chan struct{})
 
+				// Wait until we can read off the rate limit channel before
+				// updating the node
+				_, ok := <-r.rateLimiter.C
+				if !ok {
```
bwester: A `Ticker`'s channel is never closed. The presence of this check suggests otherwise and is misleading.

pkg/replication/replicator.go
```diff
@@ -143,17 +150,24 @@ func (r replicator) initializeReplicationWithCheck(
 		concurrentRealityRequests = DefaultConcurrentReality
 	}
 
+	// NewTicker() requires the value passed to it to be positive.
+	if rateLimitInterval <= 0 {
+		rateLimitInterval = 1 * time.Nanosecond
```
bwester: From testing on my laptop, the presence of a 1ns ticker brings one CPU to 50% utilization (https://play.golang.org/p/yzDjBIL6PI). **This is a dangerous default.**

Perhaps a 1s default timeout would be better, or you could selectively wait for the ticker if it exists.

mpuncel:
i'll fix the comments here. I'm learning a lesson about hasty pull requests :(

PR #631, opened 2016-09-01
daemon_set: skip locking

petertseng:
InitializeReplicationWithCheck's first argument now controls whether
locks are skipped completely, rather than overridden (now they always
are).

InitializeReplicationWithCheck's first argument now controls whether
locks are skipped completely, rather than overridden (now they always
are).

This is the only place in the code where InitializeReplicationWithCheck
is called.
The behavior change: Now, daemon sets will ignore locks.

This change is desired:
Consider the case where two daemon sets with the same pod ID are
started.
Currently, the second one will override the lock of the first one, and
the first one will stop replicating completely.
That means some nodes do not get the pod installed (if the second one
does not cover all desired nodes)

Since this is undesirable, instead we will skip locking, and trust that
daemon sets will have disjoint node selectors.
The consequence of this assumption being violated is that two daemon
sets would fight over control of any nodes that belong to both of them.

We know that current daemon sets are all disjoint, and currently daemon
set modifications are gated through us so we can maintain this as true.
In addition, we do perform a disjointness check at *creation* time (note
that labels can still change after daemon set creation such that some
node belongs to two daemon sets after they are both created).

The idea is that we will make this change and see if it becomes the case
that we have daemon sets that end up fighting, and take action then.

### DIFF
--- a/pkg/ds/daemon_set.go
+++ b/pkg/ds/daemon_set.go
@@ -503,11 +503,8 @@ func (ds *daemonSet) PublishToReplication() error {
 	// on an overlapped node from thrashing. Therefore, it makes sense for
 	// daemon sets to ignore this locking mechanism and always try to
 	// converge nodes to the specified manifest
-	replication, errCh, err := repl.InitializeReplicationWithCheck(
-		true, // override locks
-		true, // ignore Replication Controllers
+	replication, errCh, err := repl.InitializeDaemonSetReplication(
 		replication.DefaultConcurrentReality,
-		false, // Ignore missing preparers by writing intent/ anyway
 		ds.rateLimitInterval,
 	)
 	if err != nil {
--- a/pkg/replication/replication.go
+++ b/pkg/replication/replication.go
@@ -104,10 +104,10 @@ type replication struct {
 // operations for this pod ID will not be able to take place.
 // if overrideLock is true, will destroy any session holding any of the keys we
 // wish to lock
-func (r *replication) lockHosts(overrideLock bool, lockMessage string) error {
+func (r *replication) lockHosts(overrideLock bool, lockMessage string) (kp.Session, chan error, error) {
 	session, renewalErrCh, err := r.store.NewSession(lockMessage, nil)
 	if err != nil {
-		return err
+		return nil, nil, err
 	}
 
 	lockPath := kp.ReplicationLockPath(r.manifest.ID())
@@ -117,12 +117,10 @@ func (r *replication) lockHosts(overrideLock bool, lockMessage string) error {
 	_, err = r.lock(session, lockPath, overrideLock)
 	if err != nil {
 		_ = session.Destroy()
-		return err
+		return nil, nil, err
 	}
 
-	go r.handleRenewalErrors(session, renewalErrCh)
-
-	return nil
+	return session, renewalErrCh, nil
 }
 
 // Attempts to claim a lock. If the overrideLock is set, any existing lock holder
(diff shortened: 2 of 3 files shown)


pkg/replication/replicator.go
```diff
@@ -108,20 +108,22 @@ func (r replicator) InitializeReplication(
 		ignoreControllers,
 		concurrentRealityRequests,
 		true,
+		false,
 	)
 }
 
 func (r replicator) InitializeReplicationWithCheck(
```
petertseng: if daemon_sets are the only caller (and they only call it once, with mostly fixed args), why not just say `InitializeDaemonSetReplication` and remove most of the args? The only arg that needs to remain in this call is `concurrentRealityRequests`.

petertseng: Just `InitializeDaemonSetReplication(concurrentRealityRequests int)` would maybe make a lot more sense than these arguments. Why does it have to be the case that you can't pass in `overrideLock`?

petertseng:
I'm going to take the suggestion I made for myself since I see no reason not to. You don't need to review until I do that.

pkg/ds/daemon_set.go
```diff
@@ -499,7 +499,7 @@ func (ds *daemonSet) PublishToReplication() error {
 	// daemon sets to ignore this locking mechanism and always try to
 	// converge nodes to the specified manifest
 	replication, errCh, err := repl.InitializeReplicationWithCheck(
```
petertseng: this ain't what I asked for

pkg/replication/replicator.go
```diff
@@ -164,9 +163,11 @@ func (r replicator) initializeReplicationWithCheck(
 	// To make a closed channel
 	close(replication.enactedCh)
 
-	err = replication.lockHosts(overrideLock, r.lockMessage)
-	if err != nil {
-		return nil, errCh, err
+	if !skipLocking {
+		err = replication.lockHosts(overrideLock, r.lockMessage)
```
rudle: saved you a click: this function really creates a lock based on podID. It has nothing to do with hosts anymore

rudle:
:+1:

petertseng:
Unfortunately I think the test failure is actually caused by the change, which surprises me greatly. I'll deal with it later.

Note for me: f4631ffa23425b22e2507753609a52f3215e4c5f pass locally, f4fa6f62f2ea4b8c9e664c4e87d9994de67099b0 fail locally

pkg/replication/replicator.go
```diff
@@ -178,9 +179,13 @@ func (r replicator) initializeReplicationWithCheck(
 	// To make a closed channel
 	close(replication.enactedCh)
 
-	err = replication.lockHosts(overrideLock, r.lockMessage)
-	if err != nil {
-		return nil, errCh, err
+	if skipLocking {
+		go replication.handleRenewalErrors(nil, nil)
```
petertseng: hey maybe this should be called something other than `handleRenewalErrors`, yeah?

petertseng:
Okay, extracted handleReplicationEnd from lockHosts, hopefully this makes things clearer.

pkg/replication/replication.go
```diff
@@ -315,14 +313,16 @@ func (r *replication) WaitForReplication() {
 // Listen for errors in lock renewal. If the lock can't be renewed, we need to
 // 1) stop replication and 2) communicate the error up a level
 // If replication finishes, destroy the lock
-func (r *replication) handleRenewalErrors(session kp.Session, renewalErrCh chan error) {
+func (r *replication) handleReplicationEnd(session kp.Session, renewalErrCh chan error) {
```
petertseng: bikeshedding on this name accepted

pkg/replication/replication.go
```diff
@@ -315,14 +313,16 @@ func (r *replication) WaitForReplication() {
 // Listen for errors in lock renewal. If the lock can't be renewed, we need to
```
rudle: This function still does two things… maybe this comment should call that out.

Of course in an ideal world all functions would do precisely one thing, but that may not be possible given the way the channels flow around here.

petertseng: a new comment appears

rudle:
This looks good to me!

PR #632, opened 2016-09-01
[WIP] only replicate to nodes that we need to

rudle:
We scan reality before attempting to schedule any pods on any host
whatsoever. This will improve the performance and scalability of daemon
sets triggered by host restarts or changes in node label matches (ie. imaging).

This is a proof of concept - I still need to write tests for it.

### DIFF
--- a/pkg/replication/replication.go
+++ b/pkg/replication/replication.go
@@ -334,7 +334,28 @@ func (r *replication) updateOne(
 	nodeLogger := r.logger.SubLogger(logrus.Fields{"node": node})
 	nodeLogger.WithField("sha", targetSHA).Infoln("Updating node")
 
-	_, err := r.store.SetPod(
+	nodeReality, err := r.queryReality(node)
+	if err != nil || nodeReality == nil {
+		nodeLogger.WithError(err).Errorln("Could not read Reality for this node. Will proceed to schedule onto it.")
+	}
+
+	if nodeReality != nil {
+		nodeRealitySHA, err := nodeReality.SHA()
+		if err != nil {
+			nodeLogger.WithError(err).Errorln("Unable to compute manifest SHA for this node. Attempting to schedule anyway")
+		}
+		replicationRealitySHA, err := r.manifest.SHA()
+		if err != nil {
+			nodeLogger.WithError(err).Errorln("Unable to compute manifest SHA for this daemon set. Attempting to schedule anyway")
+		}
+
+		if nodeRealitySHA == replicationRealitySHA {
+			nodeLogger.Info("Reality for this node matches this DS. No action required.")
+			return nil
+		}
+	}
+
+	_, err = r.store.SetPod(
 		kp.INTENT_TREE,
 		node,
 		r.manifest,


pkg/ds/daemon_set.go
```diff
@@ -563,3 +567,32 @@ func (ds *daemonSet) CurrentPods() (types.PodLocations, error) {
...
+		realityByNode[result.PodLocation] = result.Manifest
+	}
+
+	toUpdate := make(types.PodLocations, 0, len(superset))
+	desiredSHA, err := ds.Manifest.SHA()
+	if err != nil {
+		return nil, err
+	}
+	for _, node := range superset {
+		realitySHA, err := realityByNode[node].SHA()
```
mpuncel: we should make sure this does what you want if `realityByNode[node]` is `nil`

mpuncel:
I like the change. We'll have to test this carefully though because you're downloading ~40Mb of data in that `AllPods()` call, maybe we want to keep a cache around

rudle:
Good point re AllPods() - it was the closest function in kp that did what I want. I think the most prudent thing to do is to add a function that has a signature `AllPodsBy(podID types.PodID)` since we already have a symmetric one that filters by Node.

mpuncel:
The store would still need to download the whole reality tree from consul because the keys are indexed by node name primarily

rudle:
I'm not really sure what a good caching strategy would be here because daemon sets really need to have a consistent view of reality. I'm also not convinced it's necessary. We should see only a few calls to AllPods() per hour.

mpuncel:
I'm also not convinced it's necessary

mpuncel:
do daemon sets need a consistent view of reality? if the

pkg/ds/daemon_set.go
```diff
@@ -470,8 +474,8 @@ func (ds *daemonSet) PublishToReplication() error {
 	repl, err := replication.NewReplicator(
 		ds.DaemonSet.Manifest,
 		ds.logger,
-		nodes,
-		len(nodes)-ds.DaemonSet.MinHealth,
+		nodesToUpdate.Nodes(),
+		len(nodesToUpdate.Nodes())-ds.DaemonSet.MinHealth,
```
bwester: Is this the right computation for the number of active nodes?

pkg/ds/daemon_set.go
```diff
@@ -563,3 +567,32 @@ func (ds *daemonSet) CurrentPods() (types.PodLocations, error) {
 
 	return result, nil
 }
+
+func (ds *daemonSet) NodesToUpdate(superset types.PodLocations) (types.PodLocations, error) {
+	reality, _, err := ds.kpStore.AllPods(kp.REALITY_TREE)
```
bwester: I echo the concerns that this will fetch too much data. Perhaps it would be better to fetch reality for individual pods. It would mean more requests, but they could be spread out over time since we don't want to update all nodes at once anyway.

You're more familiar with the code right now than I am, but would it make for sense for this strategy to be implemented inside the replicator instead of being a filter that controls the nodes that the replicator sees?

rudle:
This is now a much simpler diff. UpdateOne will always read reality for the node it's about to update before doing a write and doing the healthCheck dance (`ensureInReality`). I expect this to save us a lot of writes in addition to meeting the original goal of this PR: we need to be able to kick off all daemon sets at once (as happens upon service deploy) without knocking consul over.

Notice also that the calls to reality are rate limited (the default used in daemon_set.go is 3, which is fine with me).

mpuncel:
👍

PR #633, opened 2016-09-01
Fix bugs related to rate limiting replications.

mpuncel:
Bugs fixed:
1) A 1 nanosecond ticker is not a safe default because that causes a lot
of CPU work. Now the ticker used for rate limiting will just be nil if
the specified wait interval is 0, and goroutines will only attempt to
read a value off the ticker's channel if the ticker is not nil.
2) Goroutines pulling values off the ticker's channel also read off of
the quit channel in a select, which will prevent workers from stalling
forever if the ticker is stopped before the workers exit.
3) A check for the ticker's channel being closed was removed because it
will never be closed according to time.Ticker's documentation.

### DIFF
--- a/pkg/replication/replication.go
+++ b/pkg/replication/replication.go
@@ -72,7 +72,7 @@ type replication struct {
 
 	// Used to rate limit node updates. A node will not be updated
 	// until a value can be read off of the channel.
-	rateLimiter time.Ticker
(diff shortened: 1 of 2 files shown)


rudle:
🚢

bwester:
✔️

PR #634, opened 2016-09-01
only log error if there is one to log

rudle:


mpuncel:
oh ha whoops i did this too

pkg/replication/replication.go
```diff
@@ -291,7 +291,7 @@ func (r *replication) Enact() {
 
 			} else if status.err == errCancelled {
 				r.logger.Errorf("The host '%v' was cancelled (probably due to an update) during replication for pod '%v'", status.node, r.manifest.ID())
-			} else {
+			} else if status.err != nil {
 				r.logger.Errorf("An unexpected error has occurred: %v", status.err)
 			}
 			r.logger.Infof("The host '%v' successfully replicated the pod '%v'", status.node, r.manifest.ID())
```
mpuncel: we don't want this message if there was an error

rudle:
#635 did it better

PR #635, opened 2016-09-01
Fixes a bug where error message is logged even without an error

mpuncel:
A msising case in an if statement when checking whether an error
occurred during a single node replication resulted in an error message
being logged even if there was no error. This commit fixes that.


rudle:
:ship:

PR #637, opened 2016-09-07
launch: Omit empty "location" and "version.tags"

bwester:
When serializing a launch.LaunchableStanza, omit the "location" and
"version.tags" fields when they are empty. This change makes a serialized
manifest a little prettier.


petertseng:
This seems OK.

PR #638, opened 2016-09-07
logging: Fix data race adding "Counter" to entries

bwester:
ProcessCounter.Fire() contained a data race modifying the entry.Data map
to add the "Counter" and "PID" fields. Logrus makes a copy of the Entry
object, but it is shallow, so multiple loggers could share the same map.

The bug is fixed by having Fire() update a copy of Data.

### DIFF
--- a/pkg/logging/logging_test.go
+++ b/pkg/logging/logging_test.go
@@ -159,6 +159,9 @@ func TestAddUnrecognizedHook(t *testing.T) {
 	Assert(t).IsNotNil(err, "Expected an error for adding an unrecognized hook output type")
 }
 
+// TestConsurrent checks that one logger can be used concurrently without panics (e.g.,
+// from concurrent map accesses) or data races (if the race detector is enabled), which
+// will fail the test instead of explicit t.Error() calls.
 func TestConcurrent(t *testing.T) {
 	logger := NewLogger(nil)
 	const n = 100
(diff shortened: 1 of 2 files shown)


pkg/logging/logging_test.go
```diff
@@ -157,3 +158,18 @@ func TestAddUnrecognizedHook(t *testing.T) {
 	err := logger.AddHook("unrecognized_type", "some_destination")
 	Assert(t).IsNotNil(err, "Expected an error for adding an unrecognized hook output type")
 }
+
+func TestConcurrent(t *testing.T) {
```
petertseng: since this test has no assertions, I assume the point is simply that without the attached code change, it will crash due to a concurrent write?

I suggest this be noted somewhere so that others don't get confused about "hmm, no assertions, what is this testing?"

petertseng:
👍  if some comment is added to the test.

pkg/logging/logging_test.go
bwester: > without the attached code change, it will crash due to a concurrent write?

That's the idea. I've added a comment to make that clearer.

pkg/logging/logging_test.go
```diff
@@ -157,3 +158,21 @@ func TestAddUnrecognizedHook(t *testing.T) {
 	err := logger.AddHook("unrecognized_type", "some_destination")
 	Assert(t).IsNotNil(err, "Expected an error for adding an unrecognized hook output type")
 }
+
+// TestConsurrent checks that one logger can be used concurrently without panics (e.g.,
```
petertseng: doh (consurrent)

bwester: 😢

PR #640, opened 2016-09-08
consulutil: Fix deadlock when SessionManager exits

bwester:
In SessionManager, once its "done" channel is closed, there is no
guarantee that any other goroutine will be reading its output channel.
The way to prevent blocking is to send inside a select statement. When a
session is acquired, SessionManager does this; when a session is lost or
released, that send wasn't guarded. As a consequence, the SessionManager
goroutine would never exit.

This problem is fixed by adding select around the second send.

### DIFF
--- a/pkg/kp/consulutil/session_controller.go
+++ b/pkg/kp/consulutil/session_controller.go
@@ -64,7 +64,10 @@ func SessionManager(
 			} else {
 				sessionLogger.NoFields().Info("session manager: released session")
 			}
-			output <- ""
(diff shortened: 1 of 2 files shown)


rudle:
Looks good to me. Thanks for the test!

bwester:
No thanks needed. I want to say that bug fixes should, whenever possible, include a unit test that demonstrates the bug being fixed. This will make sure we don't fix the same bug twice as our product evolves.

PR #642, opened 2016-09-08
User confirmation on daemon sets when minhealth is too high

drcapulet:


rudle:
This seems like a good idea given the current state of the world. Presently one must consult the logs to figure this sort of thing out.

PR #644, opened 2016-09-12
roll: CAS desired replicas

petertseng:
Sometimes, AddDesiredReplicas isn't the desired amount of checking that
we would like to have. AddDesiredReplicas retries its mutation so that
it is resistant to *unrelated* changes (such as the changing of the
disable flag), but if someone else changes the desired replica count, we
want to know about it. Thus, a CAS for desired replicas.

---

We are experiencing cases where session losses cause replica count
increases to happen multiple times. This attempts to stop that.

### DIFF
--- a/pkg/roll/run_update.go
+++ b/pkg/roll/run_update.go
@@ -252,14 +252,14 @@ func (u *update) rollLoop(podID types.PodID, hChecks <-chan map[types.NodeName]h
 					"nextAdd":    nextAdd,
 				}).Infof("Adding %d new nodes and removing %d old nodes", nextAdd, nextRemove)
 				if nextRemove > 0 {
-					err = u.rcs.AddDesiredReplicas(u.OldRC, -nextRemove)
+					err = u.rcs.CASDesiredReplicas(u.OldRC, oldNodes.Desired, oldNodes.Desired-nextRemove)
 					if err != nil {
 						u.logger.WithError(err).Errorln("Could not decrement old replica count")
 						break
 					}
 				}
 				if nextAdd > 0 {
-					err = u.rcs.AddDesiredReplicas(u.NewRC, nextAdd)
+					err = u.rcs.CASDesiredReplicas(u.NewRC, newNodes.Desired, newNodes.Desired+nextAdd)
 					if err != nil {
 						u.logger.WithError(err).Errorln("Could not increment new replica count")
 						break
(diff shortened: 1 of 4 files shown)


pkg/roll/run_update.go
```diff
@@ -252,14 +252,14 @@ func (u *update) rollLoop(podID types.PodID, hChecks <-chan map[types.NodeName]h
 					"nextAdd":    nextAdd,
 				}).Infof("Adding %d new nodes and removing %d old nodes", nextAdd, nextRemove)
 				if nextRemove > 0 {
-					err = u.rcs.AddDesiredReplicas(u.OldRC, -nextRemove)
+					err = u.rcs.CASDesiredReplicas(u.OldRC, oldNodes.Desired, oldNodes.Desired-nextRemove)
 					if err != nil {
 						u.logger.WithError(err).Errorln("Could not decrement old replica count")
 						break
```
petertseng: is break here still appropriate? This will cause it to continue the watch, I believe. Perhaps `return false` (acting as if the replication was cancelled) is more appropriate?

Maybe in normal circumstances we should `break`, but if we were pre-empted (can be signalled by making the error type different) we should `return false`?

rudle:
This will certainly address the symptom we're seeing! I have a lurking concern that we're mutating other fields in a laissez-faire way and encountering similar bugs. This might not show up on our radar as we have alerting on only the replicas_desired field.

Said another way, what if we CAS'd the whole rc.Fields struct? Would that be safer? What does it cost?

petertseng:
from what I see in the `rcstore.Store` interface, the possible mutations are:

```go
  SetDesiredReplicas(fields.ID, int) error
  AddDesiredReplicas(fields.ID, int) error
  CASDesiredReplicas(rcID fields.ID, expectedValue int, newValue int) error
  Enable(fields.ID) error
  Disable(fields.ID) error
  Delete(fields.ID, bool) error
```

We could try to abort our replica set if we've been disabled in the meantime, though I'm not all that convinced it will be that useful. What say you?

rudle:
aha. What a convenient interface. I don't think what's here is fine.

rudle:
:ship:

petertseng:
I have to ask for clarification because your 🚢  is at odds with "I don't think what's here is fine.", but I am inferring that you mixed up two things that you wanted to say.

For me personally, I still need to reason through the effects of `break` vs `return`, so please hold.

pkg/roll/run_update.go
bwester: It's fragile to abort the whole update if one operation has an error. What I'd like is for it to re-run its computation using the new RC state. With `break`, it will do that. Eventually. After the health update watch hits its wait time and re-sends all its data. That isn't ideal behavior, but I'll take it.

bwester:
I say :shipit:

PR #645, opened 2016-09-12
consulutil: Add exponential backoff to SessionManager

bwester:
Add an exponential backoff strategy to SessionManager to deal with
errors encountered while creating a session. Session loss is a
highly-correlated event, so jitter is added to mitigate a thundering
herd (mostly preparers) attempting to reestablish their sessions after a
loss occurrs.

Errors encountered while renewing a session are probably a symptom of an
overloaded Consul server, but it isn't clear how to use an exponential
backoff to mitigate that problem. This changes uses a heuristic of
sending the next create request after the maximum timeout from the last
successful create.

### DIFF
--- a/pkg/kp/consulutil/session_controller.go
+++ b/pkg/kp/consulutil/session_controller.go
@@ -8,11 +8,17 @@ import (
 
 	"github.com/square/p2/pkg/logging"
 	"github.com/square/p2/pkg/util/param"
+	"github.com/square/p2/pkg/util/randseed"
 )
 
-// SessionRetrySeconds specifies how long to wait between retries when establishing a
-// session to Consul.
-var SessionRetrySeconds = param.Int("session_retry_seconds", 5)
+// SessionRetrySeconds specifies the base time to wait between retries when establishing a
+// session. In the presence of errors, the effective time is derived from this base
+// following a strategy of exponential backoff with jitter.
+var SessionRetrySeconds = param.Int("session_retry_seconds", 10)
+
+// SessionMaxRetrySeconds is the maximum number of seconds to wait between failed
+// attempts to acquire a session.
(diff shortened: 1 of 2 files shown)


bwester:
This is an alternative to #629 (sorry @mpuncel, you're on vacation and this functionality is needed)

petertseng:
I think it looks good. 👍

pkg/kp/consulutil/session_controller.go
```diff
@@ -34,8 +40,29 @@ func SessionManager(
...
 	for {
+		if useDelay {
+			// Normalize timeout range
+			if delay < minDelay {
+				delay = minDelay
+			} else if delay > maxDelay {
+				delay = maxDelay
+			}
+			select {
+			case <-time.After(time.Duration(prng.Int63n(int64(delay)))):
```
anthonybishopric: Should this be `time.Duration(minDelay + prng.Int63n(int64(delay-minDelay)))`? Really, I'm poking at the meaning of "minDelay" when it could actually wait less time than this value implies.

pkg/kp/consulutil/session_controller.go
```diff
@@ -61,6 +91,12 @@ func SessionManager(
 			err = client.Session().RenewPeriodic(config.TTL, id, nil, done)
 			if err != nil {
 				sessionLogger.WithError(err).Error("session manager: lost session")
+				// Session loss is an indicator that Consul is very congested and we must
+				// back off. Howerver, it isn't clear how long to wait for. As a
```
anthonybishopric: sp "However"

anthonybishopric:
Two nits, otherwise :+1:

pkg/kp/consulutil/session_controller.go
bwester: I'll change the name of "minDelay" to match its actual use

PR #646, opened 2016-09-13
roll: Do not enable old RC on cleanup

petertseng:
Recall that the roll algorithm declares a deploy complete when the new
RC **desires** as many nodes as specified in the deploy, and at least
the minimum number of new nodes are healthy.

Imagine this situation:

We are in the middle of a deploy with 2 nodes desired, 1 node min
healthy on a 2 node environment.

* Old RC: 1 desired, 1 healthy, disabled.
* New RC: 1 desired, 1 healthy, enabled.

Roll algorithm decides (rightfully) that it can roll forward:

* Old RC: 0 desired, 1 healthy, disabled.
* New RC: 2 desired, 1 healthy, enabled.

At this point, the termination condition of the roll algorithm is met.
Upon termination, it sets the old RC's count to zero (not materially
useful; it's already at zero), **enables it** (dangerous!), and then
deletes it.

* Old RC: 0 desired, 1 healthy, enabled.
* New RC: 2 desired, 1 healthy, enabled.

Now the two RCs race:

* The old RC is attempting to remove the pod from the node that
  currently has it, since it desires to have zero but has one.
* The new RC is attempting to add the pod to the node that doesn't have
  it.

So in the following sequence of events:

* The old RC decides it needs to act (new RC has not yet relabeled the
  pod)
* The new RC acts and schedules the pod
* The old RC unschedules the pod

We have lost a pod.

If we do not enable the old RC, we avoid this situation.

### DIFF
--- a/pkg/roll/run_update.go
+++ b/pkg/roll/run_update.go
@@ -164,9 +164,6 @@ func (u *update) Run(quit <-chan struct{}) (ret bool) {
 		if !RetryOrQuit(func() error { return u.rcs.SetDesiredReplicas(u.OldRC, 0) }, quit, u.logger, "Could not zero old replica count") {
 			return
 		}
-		if !RetryOrQuit(func() error { return u.rcs.Enable(u.OldRC) }, quit, u.logger, "Could not enable old RC") {
-			return
-		}
 		if !RetryOrQuit(func() error { return u.rcs.Delete(u.OldRC, false) }, quit, u.logger, "Could not delete old RC") {
 			return
 		}


petertseng:
Open questions:

* This race has, I believe, always been present. Why do we not see nodes disappear more often? Have we just been getting lucky?
* Do feel free to double check that the sequence of events that I described is possible - I haven't checked whether there is some locking that prevents it.
* Is there a situation where we **do** want to enable the old RC before deleting it?

petertseng:
An alternative is to change the termination to check for *current* nodes rather than *desired* nodes. That change seems a bit more invasive, need to think about it a bit.

I'll also try making a test for the situation described.

anthonybishopric:
I think the only reason to re-enable an RC after an RU completed is to handle a canary deployment. If that case is addressed in another codepath (via LeaveOld) then this PR seems ok 👍

petertseng:
Ah, I see.

This PR, since it affects only *not* LeaveOld RUs, doesn't affect canarying behavior: In this code and the old code, the old RU stays disabled after a canary!

So then you ask, how do we roll back from a canary?

At the start of any rolling update, we always enable the new RC and disable the old.

PR #649, opened 2016-09-13
Only store 200 characters from health output

anthonybishopric:
Clients may unintentionally send enormous status strings in their status endpoint. While there is nothing inherently wrong with this, we do not want to store this full output in Consul. This sets an arbitrarily limit of 200 characters for status data before being truncated.

I verified that this worked using the integration test environment.

### DIFF
--- a/pkg/kp/healthmanager.go
+++ b/pkg/kp/healthmanager.go
@@ -41,6 +41,10 @@ var (
 	// health status to "unknown" with an error message, and further updates will be
 	// throttled until enough tokens have been accumulated.
 	HealthResumeLimit = param.Int64("health_resume_limit", 4)
+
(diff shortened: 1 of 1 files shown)


pkg/kp/healthmanager.go
```diff
@@ -342,6 +342,9 @@ func healthToKV(wr WatchResult, session string) (*api.KVPair, error) {
 	wr.Time = now
 	// This health check only expires when the key is removed
 	wr.Expires = now.Add(100 * 365 * 24 * time.Hour)
+	if len(wr.Output) > 200 {
```
bwester: Make "200" a configurable parameter, and this will be good to go.

PR #650, opened 2016-09-13
Stale watches

rudle:
Stale recursive consul list operations are 10x faster than consistent methods and nearly as safe.

### DIFF
--- a/pkg/kp/consulutil/watch.go
+++ b/pkg/kp/consulutil/watch.go
@@ -2,9 +2,12 @@ package consulutil
 
 import (
 	"bytes"
+	"fmt"
 	"time"
 
 	"github.com/hashicorp/consul/api"
+	"github.com/rcrowley/go-metrics"
+	p2metrics "github.com/square/p2/pkg/metrics"
 )
 
 // WatchPrefix watches a Consul prefix for changes to any keys that have the prefix. When
@@ -35,6 +38,13 @@ func WatchPrefix(
 	if pause < 250*time.Millisecond {
 		pause = 250 * time.Millisecond
 	}
+
+	listLatencyHistogram, consulLatencyHistogram, outputPairsBlocking, outputPairsHistogram := watchHistograms("prefix")
+
+	var (
+		safeListStart    time.Time
+		outputPairsStart time.Time
+	)
 	for {
 		select {
 		case <-done:
@@ -42,18 +52,24 @@ func WatchPrefix(
 		case <-timer.C:
 		}
 		timer.Reset(pause) // upper bound on request rate
+		safeListStart = time.Now()
 		pairs, queryMeta, err := SafeList(clientKV, done, prefix, &api.QueryOptions{
 			WaitIndex: currentIndex,
 		})
+		listLatencyHistogram.Update(int64(time.Since(safeListStart) / time.Millisecond))
 		switch err {
 		case CanceledError:
 			return
 		case nil:
 			currentIndex = queryMeta.LastIndex
+			consulLatencyHistogram.Update(int64(queryMeta.RequestTime))
+			outputPairsStart = time.Now()
 			select {
 			case <-done:
 			case outPairs <- pairs:
 			}
+			outputPairsBlocking.Update(int64(time.Since(outputPairsStart) / time.Millisecond))
+			outputPairsHistogram.Update(int64(sizeInBytes(pairs)))
 		default:
 			select {
 			case <-done:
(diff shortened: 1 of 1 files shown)


pkg/kp/consulutil/watch.go
```diff
@@ -35,35 +38,59 @@ func WatchPrefix(
...
 			case <-done:
 			case outErrors <- err:
+				timer.Reset(2*time.Second + pause)
 			}
-			timer.Reset(2*time.Second + pause) // backoff
 		}
 	}
 }
 
+func watchHistograms(watchType string) (listLatency metrics.Histogram, consulLatency metrics.Histogram, outputPairsWait metrics.Histogram, outputPairsBytes metrics.Histogram) {
```
petertseng: I don't see `watchType` being used in this function - should it be?

rudle: Good catch. A Sprintf appears!

pkg/kp/consulutil/watch.go
```diff
@@ -35,35 +38,59 @@ func WatchPrefix(
...
 			return
 		case nil:
 			currentIndex = queryMeta.LastIndex
+			outputPairsStart = time.Now()
 			select {
 			case <-done:
 			case outPairs <- pairs:
 			}
+			outputPairsBlocking.Update(int64(time.Since(outputPairsStart) / time.Millisecond))
+			outputPairsHistogram.Update(int64(unsafe.Sizeof(pairs)))
```
petertseng: The histogram name and commit comment imply this is an attempt to measure the byte size of `pairs`. I'm not 100% sure of that, but maybe you can convince me. I see that `SafeList` returns an `api.KVPairs` which is a `[]*KVPair`. Then https://golang.org/pkg/unsafe/#Sizeof tells me for a slice it returns "the size of the slice descriptor, not the size of the memory referenced by the slice." - can you confirm that is what we want?

If it's not, I wonder if we have to instead iterate through all the pairs and sum up the `len(pair.Value)`.

pkg/kp/consulutil/watch.go
```diff
@@ -80,25 +109,37 @@ func WatchSingle(
...
 			return
 		case <-timer.C:
 		}
+		safeGetStart = time.Now()
 		timer.Reset(250 * time.Millisecond) // upper bound on request rate
 		kvp, queryMeta, err := SafeGet(clientKV, done, key, &api.QueryOptions{
 			WaitIndex: currentIndex,
 		})
+		listLatencyHistogram.Update(int64(time.Since(safeGetStart) / time.Millisecond))
+		consulLatencyHistogram.Update(int64(queryMeta.RequestTime))
```
petertseng: above, you guarded with `queryMeta != nil`, but not here. Should you here?

pkg/kp/consulutil/watch.go
```diff
@@ -253,9 +300,14 @@ func WatchDiff(
...
 			timer.Reset(250 * time.Millisecond) // upper bound on request rate
 
+			safeListStart = time.Now()
 			pairs, queryMeta, err := SafeList(clientKV, quitCh, prefix, &api.QueryOptions{
-				WaitIndex: currentIndex,
+				WaitIndex:  currentIndex,
+				AllowStale: true,
 			})
+			listLatencyHistogram.Update(int64(time.Since(safeListStart) / time.Millisecond))
+			consulLatencyHistogram.Update(int64(queryMeta.RequestTime))
```
petertseng: same question with guarding `queryMeta` `nil`.

pkg/kp/consulutil/watch.go
```diff
@@ -309,10 +361,13 @@ func WatchDiff(
 				}
 			}
 
+			outputPairsStart = time.Now()
 			select {
 			case <-quitCh:
 			case outCh <- outgoingChanges:
 			}
+			outputPairsBlocking.Update(int64(time.Since(outputPairsStart) / time.Millisecond))
+			outputPairsHistogram.Update(int64(unsafe.Sizeof(pairs)))
```
petertseng: also Sizeof

pkg/kp/consulutil/watch.go
rudle: I'll move all of these inside the `switch err`, it's cleaner that way

pkg/kp/consulutil/watch.go
petertseng: https://play.golang.org/p/YveicAxnTr seems to tell us we'll get the same thing all the time

pkg/kp/consulutil/watch.go
petertseng: you missed this one!

petertseng:
one last unchecked queryMeta, then I think it's 👍

pkg/kp/consulutil/watch.go
```diff
@@ -35,35 +38,58 @@ func WatchPrefix(
...
 			case <-done:
 			case outPairs <- pairs:
 			}
+			outputPairsBlocking.Update(int64(time.Since(outputPairsStart) / time.Millisecond))
+			outputPairsHistogram.Update(int64(sizeInBytes(pairs)))
 		default:
 			select {
 			case <-done:
 			case outErrors <- err:
+				timer.Reset(2*time.Second + pause)
```
rudle: This is not a good change. Going to back it out.

petertseng: what happened?

rudle: This code went through a lot of iterations locally because it arrived here in review.

pkg/kp/consulutil/watch.go
```diff
@@ -35,35 +38,58 @@ func WatchPrefix(
...
 			case outPairs <- pairs:
 			}
+			outputPairsBlocking.Update(int64(time.Since(outputPairsStart) / time.Millisecond))
+			outputPairsHistogram.Update(int64(sizeInBytes(pairs)))
 		default:
 			select {
 			case <-done:
 			case outErrors <- err:
 			}
-			timer.Reset(2*time.Second + pause) // backoff
```
petertseng: For cleanest diffs, maybe keep the comment too

pkg/kp/consulutil/watch.go
```diff
@@ -35,25 +38,39 @@ func WatchPrefix(
...
+		outputPairsStart time.Time
+	)
 	for {
 		select {
 		case <-done:
 			return
 		case <-timer.C:
 		}
-		timer.Reset(pause) // upper bound on request rate
+		timer.Reset(pause)
```
petertseng: also keep comment, if going to keep this wher eit is, I guess

petertseng:
👍

rudle:
Thanks for the review!

PR #651, opened 2016-09-14
rc: abolish rcmetrics in favor of p2metrics

petertseng:
We would like to add metrics on the number of RCs, RUs, and DSes. It
presents an inconsistent interface to callers if one of these (RCs)
requires a SetMetricsRegistry function and the others do not.

To head off the inconsistency, now we use the global p2metrics in the RC
Farm.


pkg/rc/farm.go
```diff
@@ -13,10 +13,12 @@ import (
 	"github.com/square/p2/pkg/kp/rcstore"
 	"github.com/square/p2/pkg/labels"
 	"github.com/square/p2/pkg/logging"
+	p2metrics "github.com/square/p2/pkg/metrics"
```
rudle: How strongly do you feel about the name collision introduced by this package? I note that `p2/pkg/metrics` and `rcrowley/go-metrics` are almost always guaranteed to appear together. If we were to change the name of our package, now is the time to do it.

Of course, I can't think of a better name!

rudle:
😹

pkg/rc/farm.go
petertseng: We could name our package p2metrics, but I don't feel strongly enough to insist on the change here

PR #653, opened 2016-09-19
Add a Pod() function to HookEnv to initialize hooked pod.

mpuncel:
Sometimes a hook will need to initialize a pod struct based on the
hooked pod manifest. This is useful when the pod's launchables need to
be initialized, for example in order to place a special directory in
each launchable contained in the manifest.

Commit e8d2f903 changed the way pods are initialized such that instead
of initializing a pod by knowing its $POD_HOME (i.e. the concatenation
of the system pod root and the pod id), a caller must first instantiate
a pod factory using the system root (e.g. /data/pods) and then instruct
the pod factory to instantiate a pod of a specified ID. The goal of that
commit was to reduce the number of code sites that compute a pod's home
directory, so that logic can more easily be changed.

A side effect of this change is that a hook can no longer instantiate a
pod using the $HOOKED_POD_HOME var, because there is no longer a pod
constructor that takes the pod home as an argument. The system pod root
must be known.

As a result, this commit adds a $HOOKED_SYSTEM_POD_ROOT var with which a
pod factory may be instantiated, and a Pod() function on the HookEnv
struct conveniently instantiates a pod factory and subsequently a pod
based on $HOOKED_POD_ID.

For systems using a pod root other than "/data/pods", it is important to
update p2-preparer with this commit prior to deploying hooks that depend
on the Pod() function, so that the $HOOKED_SYSTEM_POD_ROOT variable is
exported to hooks prior to hooks depending on it. For systems that use
"/data/pods", the order of deployment does not matter because when the
pod root passed to pods.NewFactory is the empty string, DefaultPath is
used which is "/data/pods".

### DIFF
--- a/bin/p2-run-hooks/main.go
+++ b/bin/p2-run-hooks/main.go
@@ -19,6 +19,7 @@ var (
 	hookDir      = kingpin.Flag("hook-dir", "The root of the hooks").Default(hooks.DEFAULT_PATH).String()
 	manifestPath = kingpin.Flag("manifest", "The manifest to use (this is useful when we are in the before_install phase)").ExistingFile()
 	nodeName     = kingpin.Flag("node-name", "The name of this node (default: hostname)").String()
+	podRoot      = kingpin.Flag("pod-root", "The system root for pods").Default(pods.DefaultPath).String()
 )
 
 func main() {
@@ -33,7 +34,7 @@ func main() {
 		*nodeName = hostname
 	}
 
-	dir := hooks.Hooks(*hookDir, &logging.DefaultLogger)
+	dir := hooks.Hooks(*hookDir, *podRoot, &logging.DefaultLogger)
 
 	hookType, err := hooks.AsHookType(*hookType)
 	if err != nil {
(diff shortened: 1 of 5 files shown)


mpuncel:
after this change, I believe we could remove the `HOOKED_POD_HOME` env var by making `hookenv.PodHome()` call `hookenv.Pod()` and then call `pod.Home()`

bin/p2-run-hooks/main.go
```diff
@@ -19,6 +19,7 @@ var (
 	hookDir      = kingpin.Flag("hook-dir", "The root of the hooks").Default(hooks.DEFAULT_PATH).String()
 	manifestPath = kingpin.Flag("manifest", "The manifest to use (this is useful when we are in the before_install phase)").ExistingFile()
 	nodeName     = kingpin.Flag("node-name", "The name of this node (default: hostname)").String()
+	podRoot      = kingpin.Flag("pod-root", "The system root for pods (default: /data/pods").Default(pods.DefaultPath).String()
```
bwester: nit: When you use the `.Default()` option, its value is printed in the help message. You don't need to add the extra `"(default: ...)"` message.

mpuncel: fixed

PR #654, opened 2016-09-20
Add a LaunchableVersionID typedef backed by a string

mpuncel:
This adds greater type safety when working with launchable version IDs.
This is useful for systems that support upgrading or otherwise modifying
the launchable version ID within a pod manifest.

### DIFF
--- a/pkg/launch/launchable.go
+++ b/pkg/launch/launchable.go
@@ -17,9 +17,13 @@ type LaunchableID string
 
 func (l LaunchableID) String() string { return string(l) }
 
+type LaunchableVersionID string
+
+func (l LaunchableVersionID) String() string { return string(l) }
+
 type LaunchableVersion struct {
-	ID   string            `yaml:"id"`
-	Tags map[string]string `yaml:"tags,omitempty"`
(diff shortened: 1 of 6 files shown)


pkg/launch/launchable.go
```diff
@@ -17,9 +17,13 @@ type LaunchableID string
 
 func (l LaunchableID) String() string { return string(l) }
 
+type LaunchableVersionID string
```
bwester: I think it's fine to make version strings a distinct type. It seems wordy to call this a "version _identifier_". Typically, wouldn't developers just refer to the "version" of their code?

mpuncel: I agree that it's wordy, unfortunately `LaunchableVersion` already is a struct that includes not only the version identifier but also tags (e.g. os version)

bwester: Makes sense to me. But why is `LaunchableVersion` its own struct with an `ID` field? I have the same criticism of it: calling it a "version identifier" is strange. I really just want to call it the "version". But that's not relevant to this PR, which is fine to ship.

bwester:
🚢

PR #655, opened 2016-09-23
Use only one Store in the preparer

bwester:
When creating a Store for the preparer, cache a single instance and
always return that one. This allows consulStore's HTTP client to have a
global view of all requests to Consul.


mpuncel:
👍

PR #656, opened 2016-09-23
Allow HTTP/2.0 in p2-preparer

bwester:
Adds a new preparer configuration option: "http2". When enabled, it will
configure the preparer to create clients that support the HTTP/2.0
protocol. Starting with go1.6.1, the default client supports HTTP/2.0,
however specifying a client key/certificate or custom CA authority would
disable that default behavior. This new option allows HTTP/2.0 to be
switched back on.


mpuncel:
👍

anthonybishopric:
👍

PR #657, opened 2016-09-27
`cachedMatch` flag for `GetMatches`

anthonybishopric:
This change introduces a new parameter to the `Applicator.GetMatches`
function, `staleMatch`. It permits clients to specify whether or not
they tolerate cached results from the applicator or not.

The `ConsulApplicator` has been modified to use this flag to indicate
relying on underlying Consul aggreagator's cached label set, instead of
querying Consul for the full tree on each query.

The `HttpApplicator` has been altered to pass a `staleMatch` query parameter
to the receiving service.

The ds farm implementation has been given a struct field that can
be set to enable stale queries for all daemon sets. All other usages
of `GetMatches` keep `staleMatch` disabled for now.

### DIFF
--- a/bin/p2-ds-farm/main.go
+++ b/bin/p2-ds-farm/main.go
@@ -17,6 +17,12 @@ import (
 	"github.com/square/p2/pkg/logging"
 
 	ds_farm "github.com/square/p2/pkg/ds"
+
+	"gopkg.in/alecthomas/kingpin.v2"
+)
+
+var (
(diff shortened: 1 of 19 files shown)


mpuncel:
code looks good, a nit on using the word "stale" though since that means something different in consul terms, maybe `cachedMatch`" or `useCache`?

anthonybishopric:
@mpuncel renamed to `cachedMatch`, good call

PR #658, opened 2016-09-27
Use byte instead of SHA equivalence to determine if ds manifest changed.

mpuncel:
p2-dsctl will not update a daemon set if it perceives that no changes
have been made. In order to determine manifest equivalence, the SHA
of the existing and proposed manifests were being preparered. However,
manifest signature is not considered in SHA computation and as a result,
p2-dsctl did not allow updates that only change a manifest's signature,
which is sometimes an operational necessity.

Now, normalized byte equivalence of the manifests is used instead, so an
error will only be raised if the two manifests are exactly equivalent
including both contents and signature.

### DIFF
--- a/bin/p2-dsctl/main.go
+++ b/bin/p2-dsctl/main.go
@@ -1,6 +1,7 @@
 package main
 
 import (
+	"bytes"
 	"encoding/json"
 	"fmt"
 	"log"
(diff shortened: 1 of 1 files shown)


mpuncel:
This won't help because daemon sets already don't re-schedule a manifest with the same SHA as what's in /intent. It's better behavior to not allow the user to schedule SHA equivalent manifests even if there is a difference.

PR #659, opened 2016-09-28
Fix concurrent map access in FakePodStore

bwester:
Fix unit test error:

    fatal error: concurrent map read and map write

    goroutine 359 [running]:
    runtime.throw(...)
            runtime/panic.go:566
    runtime.mapaccess2(...)
            runtime/hashmap.go:340
    github.com/square/p2/pkg/kp/kptest.(*FakePodStore).Pod(...)
            github.com/square/p2/pkg/kp/kptest/fake_store.go:70

This commit acquires and releases locks around *all* access to the
podResults field. Commit 4462d61 appears to address the same problem.
However, it only serialized writes to the field, leaving reads to race.


petertseng:
I would have talked about rwmutex, but that's probably premature optimization

PR #662, opened 2016-09-29
Fix warnings found by analysis tools

bwester:
Simplifies the source code formatting via "gofmt" and fixes many questionable practices flagged by "go vet".

### DIFF
--- a/pkg/kp/consulutil/fake_kv.go
+++ b/pkg/kp/consulutil/fake_kv.go
@@ -31,34 +31,34 @@ func NewFakeClient() *FakeConsulClient {
 	}
 }
 
-func NewKVWithEntries(entries map[string]*api.KVPair) FakeKV {
+func NewKVWithEntries(entries map[string]*api.KVPair) *FakeKV {
 	if entries == nil {
 		entries = make(map[string]*api.KVPair)
 	}
 
-	return FakeKV{
+	return &FakeKV{
 		Entries: entries,
 	}
 }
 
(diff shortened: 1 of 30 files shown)


pkg/kp/consulutil/fake_kv.go
```diff
@@ -31,34 +31,34 @@ func NewFakeClient() *FakeConsulClient {
 	}
 }
 
-func NewKVWithEntries(entries map[string]*api.KVPair) FakeKV {
+func NewKVWithEntries(entries map[string]*api.KVPair) *FakeKV {
```
mpuncel: i'm curious why this change was necessary?

pkg/util/stream/publisher_test.go
```diff
@@ -59,7 +59,7 @@ RecvLoop:
 	for _, r := range received {
 		for i, s := range sent {
 			if r == s {
-				sent = sent[i+1 : len(sent)]
+				sent = sent[i+1:]
```
mpuncel: interesting thought Go didn't let you do this

pkg/kp/consulutil/fake_kv.go
bwester: `FakeKV` had a `sync.Mutex` in it, and mutexes can't be copied.

JackDanger:
This is great, @bwester - now when an editor is configured to run `go vet` it'll complain less. Thanks!

PR #664, opened 2016-10-04
Fix syntax in p2-dsctl test-selector example.

mpuncel:


bin/p2-dsctl/main.go
```diff
@@ -81,7 +81,7 @@ var (
 		This will output the hosts that match the selector,
 		The selector string uses same syntax as the kubernetes selectors without flags.
 		An example command is:
-		p2-dsctl test-selector "ice-cream in (the-fridge)"`,
+		p2-dsctl test-selector --selector "ice-cream in (the-fridge)"`,
```
rudle: while you're here, would you mind making this more realistic? `hardware in (metal,lxc)` would be my preference

rudle:
👍

bin/p2-dsctl/main.go
```diff
@@ -81,7 +81,7 @@ var (
 		This will output the hosts that match the selector,
 		The selector string uses same syntax as the kubernetes selectors without flags.
 		An example command is:
-		p2-dsctl test-selector "ice-cream in (the-fridge)"`,
+		p2-dsctl test-selector --selector "hardware in (lxc,metal)"`,
```
bwester: Shouldn't this be `sq/hardware`?

rudle: depends who the example is for :( I could see this one going either way

We have helpful internal docs that describe valid selector types.

mpuncel: i'll just put a placeholder instead

rudle:
🐑 it

PR #665, opened 2016-10-05
Improve clarity in help and logging in p2-dsctl.

mpuncel:
The help now states that the specified timeout is a per-node timeout.

Furthermore, the message printed when creating a daemon set indicates
that the daemon set will begin replicating without further action, with
the unfortunate edge case that this won't be true if a farm is not
running.


petertseng:
you say per-node in the commit message but per-pod in the help text -- intentional?

rudle:
LGTM

PR #666, opened 2016-10-05
Mpuncel/preparer uuid pods

mpuncel:


mpuncel:
closing until uninstalling is implemented because it will change some of this

PR #668, opened 2016-10-06
Mpuncel/write reality index

mpuncel:


mpuncel:
only the manifests need to match, and the code that does this analysis is tested in #675 in `result_set_test.go`. Gonna merge this one

PR #669, opened 2016-10-06
Configurable umask for p2-exec

rudle:
Default of 0022 matches our datacenter's config

### DIFF
--- a/bin/p2-exec/main.go
+++ b/bin/p2-exec/main.go
@@ -9,8 +9,11 @@ import (
 	"os"
 	"os/exec"
 	"path/filepath"
+	"strconv"
 	"strings"
 	"syscall"
+
+	"golang.org/x/sys/unix"
 
 	"github.com/square/p2/pkg/cgroups"
 	"github.com/square/p2/pkg/util"
@@ -29,6 +32,8 @@ var (
 	nolim          = kingpin.Flag("nolimit", "Remove rlimits.").Short('n').Bool()
 	clearEnv       = kingpin.Flag("clearenv", "Clear all environment variables before loading envDir(s).").Bool()
 	workDir        = kingpin.Flag("workdir", "Set working directory.").Short('w').String()
+	umask          = kingpin.Flag("umask", "Set the process umask. Use octal notation ex. 0022").Short('m').Default(umaskDefault).String()
+	umaskDefault   = ""
 
 	cmd = kingpin.Arg("command", "the command to execute").Required().Strings()
 )
@@ -36,6 +41,14 @@ var (
 func main() {
 	kingpin.Version(version.VERSION)
 	kingpin.Parse()
+
+	if *umask != umaskDefault {
+		effectiveUmask, err := strconv.ParseInt(*umask, 8, 0)
+		if err != nil {
+			log.Fatalf("umask not expressed in octal notation. %v\n", err)
+		}
+		unix.Umask(int(effectiveUmask))
+	}
 
 	if *clearEnv {
 		os.Clearenv()


bin/p2-exec/main.go
```diff
@@ -29,6 +29,7 @@ var (
 	nolim          = kingpin.Flag("nolimit", "Remove rlimits.").Short('n').Bool()
 	clearEnv       = kingpin.Flag("clearenv", "Clear all environment variables before loading envDir(s).").Bool()
 	workDir        = kingpin.Flag("workdir", "Set working directory.").Short('w').String()
+	umask          = kingpin.Flag("umask", "umask for this process and its subprocesses. Use golang octal notation ex. 022").Short('m').Default("0022").Int()
```
bwester: Octal notation is pretty standard these days; I wouldn't call it the "golang notation."

bin/p2-exec/main.go
```diff
@@ -37,6 +38,8 @@ func main() {
 	kingpin.Version(version.VERSION)
 	kingpin.Parse()
 
+	syscall.Umask(*umask)
```
bwester: I'd feel better about this being optional. If the flag is set, change the umask. Otherwise, don't do anything. That's how all the other flags in `p2-exec` work.

bwester: New code should use the ["golang.org/x/sys/unix"](https://godoc.org/golang.org/x/sys/unix) package instead of "syscall".

rudle:
I took a slightly different approach to support the desired behaviour for when this switch is not present. PTAL

bin/p2-exec/main.go
```diff
@@ -10,14 +10,19 @@ import (
...
 	"strings"
-	"syscall"
 
 	"github.com/square/p2/pkg/cgroups"
 	"github.com/square/p2/pkg/util"
 	"github.com/square/p2/pkg/version"
 	"gopkg.in/alecthomas/kingpin.v2"
 	"gopkg.in/yaml.v2"
 )
+import (
```
bwester: nit: Can you fold this back into the earlier import block?

bin/p2-exec/main.go
```diff
@@ -29,6 +34,8 @@ var (
 	nolim          = kingpin.Flag("nolimit", "Remove rlimits.").Short('n').Bool()
 	clearEnv       = kingpin.Flag("clearenv", "Clear all environment variables before loading envDir(s).").Bool()
 	workDir        = kingpin.Flag("workdir", "Set working directory.").Short('w').String()
+	umask          = kingpin.Flag("umask", "umask for this process and its subprocesses. Leave this off to use the system default. Use octal notation ex. 0022").Short('m').Default(umaskSentinal).String()
+	umaskSentinal  = "CannotParseAsNumeric"
```
bwester: nit: I would like to suggest that a _sentinel_ value of `""` would fill the same roll and be a little easier to work with.

bin/p2-exec/main.go
```diff
@@ -29,6 +34,8 @@ var (
 	nolim          = kingpin.Flag("nolimit", "Remove rlimits.").Short('n').Bool()
 	clearEnv       = kingpin.Flag("clearenv", "Clear all environment variables before loading envDir(s).").Bool()
 	workDir        = kingpin.Flag("workdir", "Set working directory.").Short('w').String()
+	umask          = kingpin.Flag("umask", "umask for this process and its subprocesses. Leave this off to use the system default. Use octal notation ex. 0022").Short('m').Default(umaskSentinal).String()
```
bwester: > umask for this process and its subprocesses. Leave this off to use the system default.

I would like to suggest "Set the process umask" as a simplification with equivalent meaning, especially considering the wording of all the other flags.

bin/p2-exec/main.go
```diff
@@ -37,6 +44,14 @@ func main() {
 	kingpin.Version(version.VERSION)
 	kingpin.Parse()
 
+	if *umask != umaskSentinal {
+		effectiveUmask, err := strconv.ParseInt(*umask, 8, 0)
+		if err != nil {
+			log.Fatalf("umask not expressed in octal notation. %v\n", err)
+		}
+		unix.Umask(int(effectiveUmask))
```
bwester: Check the error?

rudle: This function does not return an error. The return value is the old mask which is not interesting.

rudle:
Comments addressed

bin/p2-exec/main.go
```diff
@@ -29,6 +32,8 @@ var (
 	nolim          = kingpin.Flag("nolimit", "Remove rlimits.").Short('n').Bool()
 	clearEnv       = kingpin.Flag("clearenv", "Clear all environment variables before loading envDir(s).").Bool()
 	workDir        = kingpin.Flag("workdir", "Set working directory.").Short('w').String()
+	umask          = kingpin.Flag("umask", "Set the process umask. Use octal notation ex. 0022").Short('m').Default(umaskSentinal).String()
+	umaskSentinal  = ""
```
bwester: I was too subtle in my last comment: "Sentinal" is misspelled. Use "Sentinel" (or don't name this constant).

bin/p2-exec/main.go
bwester: Oops, sorry!

bin/p2-exec/main.go
rudle: I value the self documentation nature of this name and trust the compiler to inline it. I intend to rename this to umaskDefault to better describe the intention.

PR #670, opened 2016-10-07
[RFC] Perform a HEAD request instead of a GET

rudle:
This matches the convention of other backend systems and protects us from any extra bytes in the response.


rudle:
Any more comments from the team? I have high confidence in this change.

mpuncel:
lgtm

pkg/watch/health_test.go
```diff
@@ -117,7 +117,7 @@ output`))), nil)
```
petertseng: above, in the `[]byte`s, it seems that it would be more representative of a HEAD if there would be no body, but I suppose this tests that the body is never read...

I'm surprise the `val.Output` for the error cases below isn't tested

PR #671, opened 2016-10-10
Add exponential backoff to preparer installAndLaunchPod()

mpuncel:
By design, failures in the installAndLaunchPod() function are retried
constantly, so that transient failures do not leave a deploy unfinished.
The only way to recover from permanent errors is to schedule a new
version of the pod that succeeds.

However, trying to download a pod every second can have undesirable
consequences to the download server in times of high load. This commit
adds an exponential backoff to mitigate this issue.

The backoff time begins at 1 second as before, and every sequential
failure doubles the backoff time, with a ceiling at 1 minute.


rudle:
`time.After` behaviour has always seemed subtle to me so I tested this mutation strategy and it seems to work fine. The go docs remind me that `time.After` is syntax sugar for `time.NewTimer(duration).c`.

PR #672, opened 2016-10-10
Showing the correct error from AddKVPToMap

JackDanger:
Prior to this change `err` was either nil or contained a value from a
previous, unrelated error message.


bin/p2-inspect/main.go
```diff
@@ -105,7 +105,7 @@ func main() {
 		err = enc.Encode(statusMap)
 	case "list":
 		// "List" format is a flattened version of "tree"
-		output := make([]inspect.NodePodStatus, 0)
+		var output []inspect.NodePodStatus
```
JackDanger: `go vet` like every line in this file except this one.

mpuncel:
👍

PR #673, opened 2016-10-11
Remove references to deprecated health Output field

rudle:
This will break dependencies of pkg/, please merge with care


anthonybishopric:
🎸

bwester:
You'll also need to do this for `kp.WatchResult`, because we implemented all this twice.

rudle:
PTAL another look, I removed references to Output from kp.WatchResult

PR #674, opened 2016-10-11
Reset pod installation backoff time if new manifest is received.

mpuncel:
Commit f98b0a1116cc82caab27a6202d91aea4d0b1019f introduced an
exponential backoff for pod installation and launching when an error
occurs. The backoff starts at 1 second and doubles on each error. The
backoff time is only reset when a pod launch is successful.

This commit also resets the backoff time if a new manifest is received,
which will reduce the amount of time it requires for a "fixed" pod
manifest update to be processed. For example, if a manifest is deployed
with a bad launchable URL and the backoff time rises to a minute, it is
desirable that a corrected manifest can be scheduled and processed
immediately, instead of after a large backoff time.

### DIFF
--- a/pkg/preparer/orchestrate.go
+++ b/pkg/preparer/orchestrate.go
@@ -17,7 +17,10 @@ import (
 
 // The Pod ID of the preparer.
 // Used because the preparer special-cases itself in a few places.
-const POD_ID = types.PodID("p2-preparer")
+const (
+	POD_ID             = types.PodID("p2-preparer")
+	minimumBackoffTime = 1 * time.Second
+)
 
 // slice literals are not const
 var svlogdExec = []string{"svlogd", "-tt", "./main"}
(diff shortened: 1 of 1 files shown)


pkg/preparer/orchestrate.go
```diff
@@ -186,6 +186,7 @@ func (p *Preparer) handlePods(podChan <-chan ManifestPair, quit <-chan struct{})
 		case <-quit:
 			return
 		case nextLaunch = <-podChan:
+			backoffTime = 1 * time.Second
```
mpuncel: no, does it need to be?

mpuncel: historically this was hardcoded inside a `time.After()`

petertseng: maybe at least name something `minimumBackoffTime := 1 * time.Second`?

petertseng:
Is this important enough that I should not roll preparers to production without this?

mpuncel:
@petertseng I think it's okay to proceed, this just means updates will be delayed by a minute iff the previous version never installed or launched correctly

PR #675, opened 2016-10-12
Mpuncel/preparer uuid pods

mpuncel:


mpuncel:
the first two commits of this PR are from https://github.com/square/p2/pull/668 so review that one first

bin/p2-launch/main.go
```diff
@@ -56,7 +56,7 @@ func main() {
 	}
 
 	podFactory := pods.NewFactory(*podRoot, types.NodeName(*nodeName))
-	pod := podFactory.NewPod(manifest.ID())
+	pod := podFactory.NewPod(manifest.ID(), nil)
```
mpuncel: in the interest of keeping the PR small some tools are changed to just have the old behavior. Later on there may be a need to `p2-launch` a pod with a UUID, although there are questions that are raised there (like `p2-launch` needing to talk to consul)

pkg/hooks/hook_env.go
```diff
@@ -85,7 +85,8 @@ func (h *HookEnv) Pod() (*pods.Pod, error) {
 		return nil, err
 	}
 
-	return factory.NewPod(podID), nil
+	// TODO: pass the pod uuid to hooks when applicable.
+	return factory.NewPod(podID, nil), nil
```
mpuncel: also saving this for a future PR

pkg/watch/health.go
```diff
@@ -131,6 +132,10 @@ func updatePods(
 	for _, pod := range current {
 		inReality := false
 		for _, man := range reality {
+			if man.PodUniqueKey != nil {
+				// We don't health check uuid pods
```
rudle: :eyes:

How hard would it be to separate the components of the decision "Should I health check this pod?" from "Is this pod referenced by a UUID"? At the very least, I'd like to see this conditional encapsulated so we can do something more nuanced in the future.

bin/p2-launch/main.go
mpuncel: I actually made that change in #677 but i can bring it to this commit

pkg/watch/health.go
mpuncel: I'd worry that separating the decision could result in a situation where health checking is on for a uuid pod, in which case it will override the health of a legacy pod (because we just a health namespace by node/pod id)

bin/p2-launch/main.go
mpuncel: actually I'm going to bring that in later

mpuncel:
rebased on master, going to do a final test

mpuncel:
Gave this PR one more test drive and i'm happy with it

integration/hello/README.md
```diff
@@ -0,0 +1,20 @@
+# Hello app
+
+This directory contains the source for the "hello" pod which is used in the
+integration tests. All it does is read the port it should run on from its
+config and responds to HTTP requests to '/'. The binary is compiled and built
+as a hoist artifact in hoisted-hello_def456.tar.gz so that the integration test
+does not need to actually compile anyand built as a hoist artifact in
```
petertseng: repetition here

PR #676, opened 2016-10-12
preparer: disable HTTP/2 by setting TLSNextProto to empty map

petertseng:
As noted in https://golang.org/doc/go1.6#http2.

It would appear from https://golang.org/pkg/net/http/#Transport that we
were inadvertently enabling HTTP/2 when we did not want to.

### DIFF
--- a/pkg/preparer/setup.go
+++ b/pkg/preparer/setup.go
@@ -1,6 +1,7 @@
 package preparer
 
 import (
+	"crypto/tls"
 	"io/ioutil"
 	"net"
 	"net/http"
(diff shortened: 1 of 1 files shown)


mpuncel:
🚢

mpuncel:
I wonder how we could verify that it's working correctly

petertseng:
I don't have super good ideas other than "build, deploy, and observe"

bwester:
FYI, for the purposes of experimentation, this could be accomplished by [setting an environment variable](https://golang.org/pkg/net/http/#pkg-overview).

    GODEBUG=http2client=0

bwester:
While this change does add the documented way to disable HTTP/2, I don't think the preparer was using it due to the [undocumented ways](https://go.googlesource.com/go/+/go1.7/src/net/http/transport.go#191) in which it would be disabled, namely by setting a custom `Dial` and `TLSClientConfig`.

PR #677, opened 2016-10-13
Mpuncel/unique key string

mpuncel:
Note to reviewers, this contains commits from #675, so review that first


rudle:
It looks like the integration test failed for this PR. Possible there's some flakiness somewhere?

mpuncel:
I've determined that the integration test always fails on this PR, so something must be wrong. #675 seems to be okay

anthonybishopric:
👍

PR #679, opened 2016-10-14
List labels by type

anthonybishopric:
This is useful for secondary caches that can retrieve the full list of labeled entries


pkg/labels/consul_applicator.go
```diff
@@ -124,6 +117,22 @@ func (c *consulApplicator) GetMatches(selector labels.Selector, labelType Type,
 	return res, nil
 }
 
+func (c *consulApplicator) ListLabels(labelType Type) ([]Labeled, error) {
```
rudle: I'd like to see (at least in-memory) rate limiting in here someday. This is a heavy function to call.

PR #680, opened 2016-10-17
Disable each launchable before stopping any of them.

mpuncel:
The previous behavior was to shut down with the following sequence:
1) disable 1st launchable
2) stop 1st launchable
3) disable 2nd launchable
...

The new behavior is to disable all launchables before stopping
launchables. This allows a launchable that depends on others for
successful operation to do shutdown handling prior to the other
launchables receiving term or kill signals.

### DIFF
--- a/pkg/opencontainer/containers.go
+++ b/pkg/opencontainer/containers.go
@@ -285,9 +285,13 @@ func (l *Launchable) stop(serviceBuilder *runit.ServiceBuilder, sv runit.SV) err
 	return nil
 }
 
-// Halt causes the launchable to halt execution if it is running.
-func (l *Launchable) Halt(serviceBuilder *runit.ServiceBuilder, sv runit.SV) error {
+func (l *Launchable) Disable() error {
 	// "disable" script not supported for containers
+	return nil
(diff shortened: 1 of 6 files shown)


anthonybishopric:
👍

mpuncel:
fixed a NPE in a test, (`TestUninstall` needs a logger now)

pkg/opencontainer/containers.go
```diff
@@ -285,8 +285,12 @@ func (l *Launchable) stop(serviceBuilder *runit.ServiceBuilder, sv runit.SV) err
...
 }
 
+func (l *Launchable) Disable() error {
+	return nil
+}
+
 // Halt causes the launchable to halt execution if it is running.
-func (l *Launchable) Halt(serviceBuilder *runit.ServiceBuilder, sv runit.SV) error {
+func (l *Launchable) Stop(serviceBuilder *runit.ServiceBuilder, sv runit.SV) error {
 	// "disable" script not supported for containers
```
bwester: nit: This comment should be moved up to the new `Disable()` method.

PR #681, opened 2016-10-18
Implement HTTP applicator endpoints and simple server.

anthonybishopric:
This change introduces many more endpoints implemented in the
HTTP label applicator as well as a simple server that clients
can use.

This change also adds the Gorilla multiplexer.


pkg/labels/batcher.go
```diff
@@ -0,0 +1,88 @@
...
+		}
+	}
+	b.pending = []chan []Labeled{}
+	b.batchInProgress = false
+}
+
+func (b *TypeBatcher) Retrieve() ([]Labeled, error) {
+	b.createMux.Lock()
+	if !b.batchInProgress {
+		b.batchInProgress = true
```
mpuncel: might be neater to set this inside of `handleBatch()`

pkg/labels/batcher.go
```diff
@@ -0,0 +1,88 @@
...
+	allLabels, err := b.applicator.ListLabels(b.labelType)
+	<-after
+	b.createMux.Lock()
+	defer b.createMux.Unlock()
+	b.queriesPerBatch.Update(int64(len(b.pending)))
+	for _, ch := range b.pending {
+		if err != nil {
+			close(ch)
+		} else {
+			ch <- allLabels
```
mpuncel: you should `close(ch)` here as well

pkg/labels/batcher.go
```diff
@@ -0,0 +1,88 @@
...
+	if !b.batchInProgress {
+		b.batchInProgress = true
+		go b.handleBatch()
+	}
+	respCh := make(chan []Labeled)
+	b.pending = append(b.pending, respCh)
+	b.createMux.Unlock()
+	res, ok := <-respCh
+	if !ok {
+		return nil, fmt.Errorf("Could not retrieve results")
```
mpuncel: if we put this into use we'll probably want a better error message here

pkg/labels/http_applicator.go
```diff
@@ -13,18 +15,25 @@ import (
...
 var _ Applicator = &httpApplicator{}
 
-func NewHttpApplicator(client *http.Client, matchesEndpoint *url.URL) (*httpApplicator, error) {
+func NewHTTPApplicator(client *http.Client, matchesEndpoint *url.URL) (*httpApplicator, error) {
+	// take only the base
+	base, err := url.Parse("/")
+	if err != nil {
+		return nil, err
+	}
+	matchesEndpoint = matchesEndpoint.ResolveReference(base)
```
mpuncel: this should happen after the `nil` check

pkg/labels/http_applicator.go
```diff
@@ -41,48 +50,196 @@ func NewHttpApplicator(client *http.Client, matchesEndpoint *url.URL) (*httpAppl
...
+		return nil, err
 	}
 	req.Header.Add("Accept", "application/json")
 
 	resp, err := h.client.Do(req)
 	if err != nil {
+		return nil, err
+	}
+
+	if err != nil {
```
mpuncel: delete

pkg/labels/label_http_server.go
```diff
@@ -0,0 +1,233 @@
...
+	return &labelHTTPServer{applicator, NewBatcher(applicator, batchTime), batchTime > 0, logger}
+}
+
+func (l *labelHTTPServer) Handler() *mux.Router {
+	r := mux.NewRouter()
+	r.Methods("GET").Path("/select").HandlerFunc(l.Select)
+	r.Methods("GET").Path("/labels/{type}/{id}").HandlerFunc(l.GetLabels)
+	r.Methods("GET").Path("/labels/{type}").HandlerFunc(l.ListLabels)
+	r.Methods("POST").Path("/labels/{type}/{id}/{name}").HandlerFunc(l.SetLabel)
+	r.Methods("POST").Path("/labels/{type}/{id}").HandlerFunc(l.SetLabels)
```
mpuncel: nit maybe this should be `PUT`?

pkg/labels/batcher.go
anthonybishopric: You can't safely set this flag in a separate goroutine

pkg/labels/batcher.go
anthonybishopric: Yeah, there needs to be error logging when we get an error up a few lines.

pkg/labels/label_http_server.go
anthonybishopric: I was being lazy because of `client.Post`.

anthonybishopric: I'll change it, no broken windows!

anthonybishopric:
@bwester that's a good point. Serializability is lost when using this service to query label data as written. It would definitely be solved by using the hold time option as a way to collect requests prior to actually executing the query. 

I see a fairly significant downside to doing this. The total time that requests will be batched for goes from `max(holdTime, labelRetrievalTime)` to just `holdTime`. I think in practice this will somewhat minimize the batch count per query and increase the overall latency of match results, since you now need to wait `holdTime + labelRetrievalTime`.

However, I think your suggested solution still lets us perform batching and doesn't compromise data quality, which is critical. To your point, replication controllers do not act safely if the labels they've written are not immediately readable:

* RC abc has a pod marked with its ID
* RC def needs to `addOne`, it marks that pod with its ID
* RC def then checks again which nodes it manages and decides it needs to `addOne` again.

I will alter the batch handling to match your description and we can tune hold times as needed. We don't need another Faustian deal for performance gains.

PR #682, opened 2016-10-18
Fix health path and add feature to p2-label

anthonybishopric:
Two unrelated commits in this PR. 

First, there is a bugfix for health watches. You can accidentally get the health for any service matching the string prefix of the name.

The other is a small feature for `p2-label` to allow setting labels on specific entities.

### DIFF
--- a/bin/p2-label/main.go
+++ b/bin/p2-label/main.go
@@ -16,7 +16,8 @@ var (
 	cmdApply              = kingpin.Command(CmdApply, "Apply label changes to all objects matching a selector")
 	applyAutoConfirm      = cmdApply.Flag("yes", "Autoconfirm label applications. Use with caution!").Short('y').Bool()
 	applyLabelType        = cmdApply.Flag("labelType", "The type of label to adjust. Sometimes called the \"label tree\". Supported types can be found here:\n\thttps://godoc.org/github.com/square/p2/pkg/labels#pkg-constants").Short('t').Required().String()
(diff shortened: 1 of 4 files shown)


bin/p2-label/main.go
```diff
@@ -74,21 +80,26 @@ func main() {
...
-		if err != nil {
-			fmt.Fprintf(os.Stderr, "Error while finding label matches. Check the syntax.\n%v\n", err)
-			exitCode = 1
-			break
+		var matches []labels.Labeled
+		if *applySubjectSelector != "" {
+			subject, err := klabels.Parse(*applySubjectSelector)
+			if err != nil {
+				fmt.Fprintf(os.Stderr, "Error while parsing subject label. Check the syntax.\n%v\n", err)
+				break
```
mpuncel: not your fault but there should be an `exitCode = 1` here probably

bin/p2-label/main.go
```diff
@@ -65,6 +66,11 @@ func main() {
 		fmt.Printf("%s/%s: %s\n", labelType, *showID, labelsForEntity.Labels.String())
 		return
 	case CmdApply:
+		if *applySubjectSelector == "" && *applySubjectID == "" {
+			fmt.Fprint(os.Stderr, "Must pass either an ID or a selector for objects to apply the given label to")
+			exitCode = 1
+			break
+		}
```
mpuncel: should you check and err if they're both set? looks like the code will use the selector and not the ID in that case which could be surprising

PR #685, opened 2016-10-19
Break up kp.Store interface into subsets

mpuncel:
kp.Store was a massive interface which makes it hard to implement with
new stores. This commit breaks up the interface into package-defined
interfaces where they are needed that only contain the functions that
are used.

This will enable implementing individual store interfaces with a GRPC
storage server.

### DIFF
--- a/pkg/kp/kv.go
+++ b/pkg/kp/kv.go
@@ -39,25 +39,6 @@ type ManifestResult struct {
 	PodUniqueKey *types.PodUniqueKey
 }
 
-type Store interface {
-	SetPod(podPrefix PodPrefix, nodename types.NodeName, manifest manifest.Manifest) (time.Duration, error)
-	Pod(podPrefix PodPrefix, nodename types.NodeName, podId types.PodID) (manifest.Manifest, time.Duration, error)
-	DeletePod(podPrefix PodPrefix, nodename types.NodeName, podId types.PodID) (time.Duration, error)
-	PutHealth(res WatchResult) (time.Time, time.Duration, error)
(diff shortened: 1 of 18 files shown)


pkg/kp/kv.go
```diff
@@ -110,7 +91,7 @@ type consulStore struct {
 	podStore podstore.Store
 }
 
-func NewConsulStore(client consulutil.ConsulClient) Store {
+func NewConsulStore(client consulutil.ConsulClient) *consulStore {
```
rudle: Is there any value in using an interface vs concrete type here? I feel that the philosophical answer is "yes, of course, of course. hmm." What was your experience in interacting with the callsites of this method? Based on what I've seen so far, it seems like we were leaning of type inference anyway so it's probably a moot point.

… time passes …

To wit, golint decrees: `    94  53 warning         exported func NewConsulStore returns unexported type *kp.consulStore, which can be annoying to use (go-golint)

… still more time passes … 

I'm now concluding that this needs to be made public or into an interface. WDYT?

rudle: To wit, golint decrees: `    94  53 warning         exported func NewConsulStore returns unexported type *kp.consulStore, which can be annoying to use (go-golint)
`

anthonybishopric: Keeping it private has a nice benefit in that it forces the client, if they ever pass it around, to define the subset of the interface that they need. Accepting a `*consulStore` as a parameter would be way worse, for obvious reasons.

rudle: I guess I don't have a constructive alternative to suggest.

PR #687, opened 2016-10-19
Further constrain the interface required for the HTTP label server.

anthonybishopric:
The HTTP label server does not need to have access to the Watch functions yet.


mpuncel:
👍

PR #688, opened 2016-10-21
Mpuncel/grpc kp store

mpuncel:
This PR vendors grpc dependencies, adds a proof of concept implementation of `Pod()` and `SetPod()`, and a binary that can be used to read a manifest from intent or reality.


mpuncel:
gonna decline since the utility of this is low, it was mainly my initial experiment into grpc

PR #689, opened 2016-10-21
Inject labelers whenever possible

anthonybishopric:
This reduces the number of hardcoded instantiations of
`labels.NewConsulApplicator` in favor of injected labeler
interfaces that are constrained to the needs of the given client.

It also allows all CLIs to use HTTP applicators as a default option
in `kv.ParseWithConsulOptions`.

### DIFF
--- a/pkg/kp/flags/kingpin.go
+++ b/pkg/kp/flags/kingpin.go
@@ -9,12 +9,14 @@ import (
 	"net/http"
 
 	"github.com/square/p2/pkg/kp"
+	"github.com/square/p2/pkg/labels"
 	netutil "github.com/square/p2/pkg/util/net"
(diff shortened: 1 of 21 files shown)


pkg/kp/flags/kingpin.go
```diff
@@ -51,11 +53,25 @@ func ParseWithConsulOptions() (string, kp.Options) {
...
+		Address:  *consulURL,
 		Token:    *token,
-		Client:   netutil.NewHeaderClient(*headers, transport),
+		Client:   httpClient,
 		HTTPS:    *https,
 		WaitTime: *wait,
 	}
+
+	var applicator labels.ApplicatorWithoutWatches
+	var err error
```
mpuncel: don't think you need this

PR #691, opened 2016-10-24
kp: HealthManager waits for session termination

bwester:
When shutting down a consulHealthManager, its Close method will now wait for
its Consul session to be destroyed. This new behavior makes it more likely that
a clean restart of p2-preparer will be able to immediately write health values.
Otherwise, the new p2-preparer will have to wait for its previous session to
expire.


mpuncel:
👍

PR #692, opened 2016-10-24
p2-preparer: Stop the health checker last

bwester:
When exiting, stop the health checker as the last action of the process.

The preparer is slow to exit, especially if there is a pod update in
progress. There could be a substantial amount of time between the start
of a shutdown (when a SIGTERM arrives) and when the preparer finally
exits. This change keeps the health checker alive during this time to
minimize the amount of time when pod health is not published.


mpuncel:
👍

PR #693, opened 2016-10-25
Mpuncel/grpc label store

mpuncel:


mpuncel:
had some struggles with the integration test, turns out `grpc.Dial()` does not block until a connection is made unless you pass it an option! in any case, i expect the test to succeed reliably now and this is ready for review

pkg/grpc/label_store/client/client.go
```diff
@@ -0,0 +1,126 @@
...
+	klabels "k8s.io/kubernetes/pkg/labels"
+)
+
+type Client struct {
+	conn             *grpc.ClientConn
+	labelStoreClient label_protos.LabelStoreClient
+	logger           logging.Logger
+}
+
+func NewClient(grpcAddress string, tlsConfig *tls.Config, logger logging.Logger) (Client, error) {
```
mpuncel: I think I should change this to take the `conn` as the argument to support things like `grpc.WithBalancer()`

pkg/grpc/label_store/protos/label_store.proto
```diff
@@ -0,0 +1,31 @@
...
+
+message WatchMatchesRequest {
+  string selector = 1;
+  LabelType label_type = 2;
+}
+
+message Labeled {
+  LabelType label_type = 1;
+  string id = 2;
+  map<string,string> labels = 3;
```
mpuncel: protobuf maps can only use "integral types" for keys. values can be whatever we want, although you cannot do a typedef of a primitive type e.g. `message Value string`.

Then we sort of lose the enforcement that values have to be strings, and if we ever json marshal the type it'll be messed up. I don't anticipate us doing that though. Probably using <string, string> is acceptable and won't kill us, i do wish we could do typedefs though

bin/grpc/clients/label-store-client/main.go
```diff
@@ -0,0 +1,83 @@
...
+	cmdWatchMatches = kingpin.Command(cmdWatchMatchesText, "Watch the matches for a label selector")
+	labelType       = cmdWatchMatches.Flag("label-type", "The type of label watch to do, e.g. pod.").Required().String()
+	selector        = cmdWatchMatches.Flag("selector", "A kubernetes style label selector").Required().String()
+	numFetches      = cmdWatchMatches.Flag("num-fetches", "The number of watch iterations to display before exiting").Short('n').Default("1").Int()
+)
+
+func main() {
+	cmd := kingpin.Parse()
+
+	logger := log.New(os.Stderr, "", 0)
```
mpuncel: sounds good, once i have a real laptop i can do that

mpuncel:
the underscores will bother me too, i'll fix

PR #694, opened 2016-10-25
Verify pod authenticity after SHA comparison

anthonybishopric:
This change reduces the number of error messages
emitted by the preparer when a manifest is signed by
a GPG key that has expired, but was not expired when
the installation took place.


mpuncel:
brilliant

PR #696, opened 2016-10-26
Daemon set bugfixes & cleanup

rudle:
I still need to verify this code works via integration test. It's a fairly big change, so I thought I'd solicit opinions early. Commits are fairly isolated, please review commit by commit.

### DIFF
--- a/pkg/replication/replication.go
+++ b/pkg/replication/replication.go
@@ -265,7 +265,6 @@ func (r *replication) Enact() {
 						r.logger.Errorf("The host '%v' timed out during replication for pod '%v'", node, r.manifest.ID())
 					case errCancelled:
 						r.logger.Errorf("The host '%v' was cancelled (probably due to an update) during replication for pod '%v'", node, r.manifest.ID())
-					case nil:
 					default:
 						r.logger.Errorf("An unexpected error has occurred: %v", err)
 					}
@@ -424,12 +423,9 @@ func (r *replication) updateOne(
 	timer := time.NewTimer(exponentialBackoff)
 	for err != nil {
 		nodeLogger.WithError(err).Errorln("Could not write intent store")
-		select {
-		case r.errCh <- err:
-		default:
-		}
 
 		select {
+		case r.errCh <- err:
 		case <-r.quitCh:
 			r.logger.Infoln("Caught quit signal during updateOne")
 			return errQuit
--- a/pkg/replication/replication.go
+++ b/pkg/replication/replication.go
@@ -237,19 +237,11 @@ func (r *replication) Enact() {
 	// this loop multiplexes the node queue across some goroutines
 	for i := 0; i < r.active; i++ {
 		go func() {
+			// nodeQueue is managed below to throttle these goroutines
 			for node := range nodeQueue {
 				exitCh := make(chan struct{})
 				timeoutCh := make(chan struct{})
 
-				if r.rateLimiter != nil {
-					// Wait until we can read off the rate limit channel before
-					// updating the node
-					select {
-					case <-r.rateLimiter.C:
-					case <-r.quitCh:
-						return
-					}
-				}
 				go func() {
 					defer close(exitCh)
 					err = r.updateOne(node, done, timeoutCh, aggregateHealth)
@@ -284,11 +276,21 @@ func (r *replication) Enact() {
 	go func() {
 		defer close(nodeQueue)
 		for _, node := range r.nodes {
+			if r.rateLimiter != nil {
+				select {
+				case <-r.replicationCancelledCh:
+					return
+				case <-r.quitCh:
+					return
+				case <-r.rateLimiter.C:
+				}
+			}
 			select {
-			case nodeQueue <- node:
-				// a worker will consume it
+			case <-r.replicationCancelledCh:
+				return
 			case <-r.quitCh:
 				return
+			case nodeQueue <- node:
 			}
 		}
 	}()
--- a/pkg/replication/replication.go
+++ b/pkg/replication/replication.go
@@ -18,6 +18,7 @@ import (
 	"github.com/square/p2/pkg/rc"
 	"github.com/square/p2/pkg/types"
 	"github.com/square/p2/pkg/util"
+	"github.com/square/p2/pkg/util/param"
 
 	"github.com/Sirupsen/logrus"
 )
@@ -25,6 +26,11 @@ import (
 type Labeler interface {
 	GetLabels(labelType labels.Type, id string) (labels.Labeled, error)
 }
+
+var (
+	ensureRealityPeriodMillis = param.Int("ensure_in_reality_millis", 5000)
+	ensureHealthyPeriodMillis = param.Int("ensure_healthy_millis", 1000)
+)
 
 type nodeUpdated struct {
 	node types.NodeName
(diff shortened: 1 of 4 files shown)


pkg/replication/replication.go
```diff
@@ -363,39 +363,50 @@ func (r *replication) handleReplicationEnd(session kp.Session, renewalErrCh chan
...
-) error {
-	targetSHA, _ := r.manifest.SHA()
-	nodeLogger := r.logger.SubLogger(logrus.Fields{"node": node})
-	nodeLogger.WithField("sha", targetSHA).Infoln("Updating node")
-
+func (r *replication) shouldScheduleForNode(node types.NodeName, logger logging.Logger) bool {
 	nodeReality, err := r.queryReality(node)
 	if err != nil || nodeReality == nil {
-		nodeLogger.WithError(err).Errorln("Could not read Reality for this node. Will proceed to schedule onto it.")
+		logger.WithError(err).Errorln("Could not read Reality for this node. Will proceed to schedule onto it.")
```
mpuncel: not related to your code, but I think it's better to check `err == pods.NoCurrentManifest` rather than `nodeReality == nil` and don't log the error, (although an `Infoln()` would probably be useful)

pkg/replication/replication.go
```diff
@@ -443,7 +446,15 @@ func (r *replication) updateOne(
 }
 
 func (r *replication) queryReality(node types.NodeName) (manifest.Manifest, error) {
-	r.concurrentRealityRequests <- struct{}{}
+	for {
+		select {
+		case r.concurrentRealityRequests <- struct{}{}:
+			break
+		case <-time.After(1 * time.Second):
+			r.logger.Infof("Waiting on concurrentRealityRequests for pod: %s/%s", node.String(), r.manifest.ID())
```
mpuncel: that's gonna be a lot of logging, is this useful?

pkg/replication/replication_test.go
```diff
@@ -1 +1,117 @@
...
+
+	intent, _, err := r.store.AllPods(kp.INTENT_TREE)
+	if err != nil {
+		t.Fatalf("Encountered error while fetching intent: %v\n", err)
+	}
+
+	if len(intent) != 2 {
+		t.Errorf("Expected to have 2 intent records scheduled but got %d.\n%v", len(intent), intent)
+	}
+	reality, _, err := r.store.AllPods(kp.REALITY_TREE)
```
rudle: darn, this is a compile error now

pkg/replication/replication.go
```diff
@@ -416,7 +416,10 @@ func (r *replication) updateOne(
 	timer := time.NewTimer(exponentialBackoff)
 	for err != nil {
 		nodeLogger.WithError(err).Errorln("Could not write intent store")
-		r.errCh <- err
+		select {
+		case r.errCh <- err:
```
mpuncel: any reason not to move this into the lower select? that way we only drop the error if this has to take another action

pkg/replication/replication.go
```diff
@@ -231,23 +232,42 @@ func (r *replication) Enact() {
...
+
+					switch err {
+					case errTimeout:
+						r.timedOutReplicationsMutex.Lock()
+						r.timedOutReplications = append(r.timedOutReplications, node)
+						r.timedOutReplicationsMutex.Unlock()
+						r.logger.Errorf("The host '%v' timed out during replication for pod '%v'", node, r.manifest.ID())
+					case errCancelled:
+						r.logger.Errorf("The host '%v' was cancelled (probably due to an update) during replication for pod '%v'", node, r.manifest.ID())
+					case nil:
```
mpuncel: nit: this should be impossible since you `return`'d above if `err == nil`

pkg/replication/replication_test.go
```diff
@@ -1 +1,117 @@
...
+
+	}
+}
+
+func TestEnactCancellation(t *testing.T) {
+	errCh := make(chan error)
+	defer close(errCh)
+	r := newTestReplication(errCh)
+	r.active = 1
+	r.rateLimiter = time.NewTicker(2 * time.Second) // we need time to cancel
```
mpuncel: can you call `r.Cancel()` before `r.Enact()` to avoid needing to do this?

pkg/replication/replication_test.go
```diff
@@ -1 +1,117 @@
...
+
+	var wg sync.WaitGroup
+	wg.Add(1)
+	go func() {
+		r.Enact()
+		wg.Done()
+	}()
+	r.Cancel()
+	wg.Wait()
```
mpuncel: if this test was failing this would probably stall forever right? maybe instead of using a waitgroup do a send on a channel inside a `select` with a `time.After()`?

pkg/replication/replication_test.go
```diff
@@ -1 +1,117 @@
...
+	}
+}
+
+// Manages the errCh and fails the test if an error encountered
+func newTestReplication(errCh chan error) *replication {
+	podID := types.PodID("testPod")
+	mb := manifest.NewBuilder()
+	mb.SetID(podID)
+	nodes := []types.NodeName{"abc123.example.com", "def456.example.com"}
+	podStore := kptest.NewFakePodStore(nil, nil) // TODO
```
mpuncel: what's the TODO?

pkg/replication/replication_test.go
```diff
@@ -1 +1,117 @@
...
+	logger := logging.TestLogger()
+	preparer := kptest.NewFakePreparer(podStore, logger)
+	preparer.Enable()
+
+	replicationErrChan := make(chan error)
+	replicationCancelledChan := make(chan struct{})
+	replicationDoneCh := make(chan struct{})
+	enactedCh := make(chan struct{})
+	quitCh := make(chan struct{})
+	concurrentRealityRequests := make(chan struct{}, 100)
```
mpuncel: so many channels!

mpuncel:
these changes all look very solid, really glad to see some tests too! thanks for doing this

pkg/replication/replication.go
```diff
@@ -363,39 +363,50 @@ func (r *replication) handleReplicationEnd(session kp.Session, renewalErrCh chan
...
+			logger.Info("Reality for this node matches this DS. No action required.")
+			return false
 		}
 	}
 
-	_, err = r.store.SetPod(
+	return true
+}
+
+// note: logging should be delegated somehow
```
rudle: It's been moved around a few times. I'm going to remove it? @anthonybishopric

pkg/replication/replication.go
```diff
@@ -363,39 +363,50 @@ func (r *replication) handleReplicationEnd(session kp.Session, renewalErrCh chan
...
+func (r *replication) updateOne(
+	node types.NodeName,
+	done chan<- nodeUpdated,
+	timeoutCh <-chan struct{},
+	aggregateHealth *podHealth,
+) error {
+	targetSHA, _ := r.manifest.SHA()
+	nodeLogger := r.logger.SubLogger(logrus.Fields{"node": node})
+	nodeLogger.WithField("sha", targetSHA).Infoln("Updating node")
```
rudle: This is going to make the logs so pretty.

rudle:
Thanks for the review @mpuncel. I've addressed all the comments so I'll merge when this is green.

PR #697, opened 2016-10-27
Mpuncel/pod process reporter

mpuncel:


pkg/preparer/pod_process_reporter/reporter.go
```diff
@@ -0,0 +1,180 @@
+/*
+This package provides a server listening on a unix socket for the exit status
+of pods. The status of each pod is written back to the consul datastore. This
+is useful for allowing external systems to examine the success or failure of a
+pod.
+*/
+package pod_process_reporter
```
rudle: snake case isn't conventional go package name style: https://blog.golang.org/package-names

May I suggest `podprocess`?

pkg/preparer/pod_process_reporter/reporter.go
```diff
@@ -0,0 +1,180 @@
...
+}
+
+func New(config ReporterConfig, logger logging.Logger, podStatusStore podstatus.Store) *Reporter {
+	if config.ProcessResultSocketPath == "" {
+		logger.Infoln("process_result_socket_path not configured, process exit status will not be captured")
+		return nil
+	}
+
+	// TODO: do more validation here?
+	if config.EnvironmentExtractorPath == "" {
```
rudle: An `os.File.Stat` for existence and executability would be a good call

pkg/preparer/pod_process_reporter/reporter.go
```diff
@@ -0,0 +1,180 @@
...
+		conn, err := server.Accept()
+		if err != nil {
+			select {
+			case <-quitCh:
+				// We expect an error when server.Close() is called, it's time to shut down
+				r.logger.Infoln("Process event reporter shutting down")
+				return nil
+			default:
+			}
+			r.logger.Errorf("Error during Accept(): %v", err)
```
rudle: Move this up so we always log the error

pkg/preparer/setup.go
```diff
@@ -376,11 +384,15 @@ func New(preparerConfig *PreparerConfig, logger logging.Logger) (*Preparer, erro
...
-	if len(preparerConfig.FinishExec) > 0 {
-		finishExec = preparerConfig.FinishExec
-	} else {
-		finishExec = pods.DefaultFinishExec
+	finishExec := pods.NopFinishExec
+	podProcessReporterLogger := logger.SubLogger(logrus.Fields{
+		"component": "PodProcessReporter",
+	})
+
+	// Returns nil if config is unset
```
rudle: I feel not at all strongly about this, but: did you consider using a second return value that signals whether or not this should be enabled? `nil` as a sentinel value is fine in most cases but I thought it worth mentioning.

pkg/preparer/pod_process_reporter/reporter.go
mpuncel: It's here intentionally because we expect an error under normal shutdown (`server.Close()` causes an error in `Accept()`)

mpuncel:
declining because we've decided to use a sqlite based solution instead of using a unix socket

PR #698, opened 2016-10-31
Mpuncel/sqlite pod process reporter

mpuncel:


pkg/kp/statusstore/podstatus/status.go
```diff
@@ -5,53 +5,39 @@ import (
...
-// the current status of the service are useful to expose
-type ServiceStatus struct {
-	Name     string       `json:"name"`
-	State    ServiceState `json:"state"`
-	LastExit *ExitStatus  `json:"last_exit"`
+// Encapsulates information regarding the state of a process. Currently only
+// information about the last exit is exposed.
+type ProcessStatus struct {
+	LaunchableID launch.LaunchableID `json:"launchable_id"`
+	EntryPoint   string              `json:"name"`
```
anthonybishopric: Did you mean to change the JSON tag too?

pkg/preparer/podprocess/finish_exec.go
```diff
@@ -0,0 +1,13 @@
+package podprocess
+
+func (r *Reporter) FinishExec() []string {
+	return []string{
+		"/usr/bin/timeout",
```
anthonybishopric: This is not available on all systems. Can it be made configurable / opt-in?

pkg/preparer/podprocess/finish_exec.go
```diff
@@ -0,0 +1,13 @@
+package podprocess
```
anthonybishopric: Do you envision this file growing much?

pkg/preparer/podprocess/finish_output.go
```diff
@@ -0,0 +1,48 @@
...
+	EntryPoint   string              `json:"entry_point"`
+	PodUniqueKey types.PodUniqueKey  `json:"pod_unique_key"`
+
+	// These two model the arguments given to the ./finish script under runit:
+	// (http://smarden.org/runit/runsv.8.html)
+	ExitCode   int `json:"exit_code"`
+	ExitStatus int `json:"exit_status"`
+}
+
+func (f FinishOutput) Insert(db *sql.DB) error {
```
anthonybishopric: SQL DB style recommendation - rather than putting this function on the data type, put it on a service type that has a function `InsertFinish(finishOutput, db)`. Constraining DB operations to a single interface will make it easier to stub later. 

It also results in the various queries / statements and sqlite implementation being consolidated to a single location.

pkg/preparer/podprocess/reporter.go
```diff
@@ -0,0 +1,313 @@
...
+	var exitCode, exitStatus int
+
+	defer func() {
+		err = r.writeLastID(lastID)
+		if err != nil {
+			r.logger.WithError(err).Errorln("Could not write last ID to workspace file, duplicate messages may be reported")
+		}
+	}()
+	for rows.Next() {
+		err = rows.Scan(&id, &date, &podID, &podUniqueKey, &launchableID, &entryPoint, &exitCode, &exitStatus)
```
anthonybishopric: Move the query + scan to a `readFinishes` helper that returns a `[]finishRow` with these fields.

pkg/preparer/podprocess/reporter.go
```diff
@@ -0,0 +1,313 @@
...
+	"github.com/square/p2/pkg/logging"
+	"github.com/square/p2/pkg/types"
+	"github.com/square/p2/pkg/util"
+
+	"github.com/Sirupsen/logrus"
+	_ "github.com/mattn/go-sqlite3"
+)
+
+const (
+	workspaceFileName     = "last_sql_id"
```
anthonybishopric: `"last_synced_finish_id"`

pkg/preparer/podprocess/reporter.go
```diff
@@ -0,0 +1,313 @@
...
+	return nil
+}
+
+func (r *Reporter) openAndMaybeMigrateDatabase() (*sql.DB, error) {
+	db, err := sql.Open("sqlite3", r.sqliteDBPath)
+	if err != nil {
+		return nil, util.Errorf("Could not open sqlite database: %s", err)
+	}
+
+	sqlStmt := `
```
anthonybishopric: is it worth having a lightweight schema version number here? A follow up migration as trivial as adding a column will make this complicated real fast

pkg/preparer/podprocess/reporter.go
```diff
@@ -0,0 +1,313 @@
...
+	create table if not exists finishes (
+	    id integer not null primary key,
```
anthonybishopric: do you need an `autoincrement` here too?

pkg/kp/statusstore/podstatus/status.go
mpuncel: yes 🙏

pkg/preparer/podprocess/finish_exec.go
mpuncel: will make it configurable

pkg/preparer/podprocess/reporter.go
mpuncel: no: https://sqlite.org/faq.html#q1

mpuncel: although maybe i'll add it for clarity assuming it works

mpuncel: oh man i think I do need it for when we delete values: ```The new key will be unique over all keys currently in the table, but it might overlap with keys that have been previously deleted from the table. To create keys that are unique over the lifetime of the table, add the AUTOINCREMENT keyword to the INTEGER PRIMARY KEY declaration. ```

pkg/preparer/podprocess/finish_exec.go
mpuncel: I don't think so, are you suggesting I merge it into `reporter.go`? The reason I had it separate is because the preparer calls this function directly to determine the finish exec script to write, so it seemed separate from `reporter.go` in terms of purpose

pkg/preparer/podprocess/reporter.go
mpuncel: I'm curious for the motivation here? the function doesn't do much else (the only other thing is `r.getLastID()`) and I don't anticipate other places to need to do the scan

mpuncel: actually the submission is separate too, i'll move it to another function

mpuncel:
btw I plan on adding an integration test for this once the env extractor PR is also merged, or perhaps I can add it to that one once this one is merged

anthonybishopric:
Nice changes, I think this is good to go 👍

mpuncel:
squashed the fixup commit, and fixed a bug in the migration code (have to insert a row when you create the table or the update won't do anything).

will merge on green

mpuncel:
fixed some errcheck complaints

PR #700, opened 2016-10-31
Emit warning messages if a launchable script is running for minutes.

mpuncel:
Disable scripts that hang forever or run for a very long time are a
common source of confusion during p2 deploys, because they result in
p2-preparer appearing slow to accomplish the update. Emitting warning
messages to the logs makes it easier to identify when a slow launchable
lifecycle script is to blame.

### DIFF
--- a/pkg/pods/pod.go
+++ b/pkg/pods/pod.go
@@ -114,7 +114,11 @@ func (pod *Pod) Halt(manifest manifest.Manifest) (bool, error) {
 
 	success := true
 	for _, launchable := range launchables {
-		err = launchable.Disable()
+		var err error
+		disableFunc := func() {
+			err = launchable.Disable()
+		}
+		pod.withTimeWarnings("disable", launchable.ServiceID(), disableFunc)
 		if err != nil {
 			// do not set success to false on a disable error
 			pod.logLaunchableWarning(launchable.ServiceID(), err, "Could not disable launchable")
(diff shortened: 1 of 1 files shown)


pkg/pods/pod.go
```diff
@@ -738,3 +749,42 @@ func (p *Pod) logLaunchableWarning(serviceID string, err error, message string)
 func (p *Pod) logInfo(message string) {
 	p.logger.WithFields(logrus.Fields{}).Info(message)
 }
+
+// Runs function f and emits warnings if it hasn't completed within 1 minute, 2
+// minutes, 5 minutes, 10 minutes, and each 5 minute increment afterward. This
+// is useful for identifying the cause of a very long install time, for
+// instance a disable or post-activate script hanging forever.
+func (p *Pod) withTimeWarnings(scriptType string, serviceID string, f func()) {
```
anthonybishopric: Why not just embed this directly into the lifecycle script execution method? Or, alternatively, make this a `timeutil` thing.

rudle: I agree, but weakly. The pod state is only interrogated to populate the logger. This would be a useful timeutil helper if it took a logger as an argument.

rudle:
This LGTM. I agree with Anthony, but I also understand that we might be too concerned about generality. I recall that functions that take functions as arguments tend not to generalize so well.

mpuncel:
famous last words but I do think that post activate and disable are the only places where we're going to be running a user supplied script and logging in this way, so it's not super useful as a generic function. We have a hard timeout on system hooks so those are different

PR #702, opened 2016-11-03
Export pod.uniqueName() function.

mpuncel:
This function is useful as a library function for generating a
deterministic unique string for a pod. For instance a hook may wish to
manage per-pod files using this string in the filename to avoid filename
collisions while programmatically being able to clean up or otherwise
locate them.


anthonybishopric:
🚢

PR #704, opened 2016-11-18
Make restart policy a per-launchable setting instead of per-pod.

mpuncel:
This will allow greater flexibility in the way pods are composed and
run. For example, the scheduled pod may have an entry point that should
only run once, such as a job like a database backfill. There may be
supporting launchables that need to be up to support the job, e.g. an
instance of stunnel which should be restarted if it crashes. However,
the job should not be restarted.

### DIFF
--- a/pkg/runit/servicebuilder.go
+++ b/pkg/runit/servicebuilder.go
@@ -63,6 +63,14 @@ type ServiceTemplate struct {
 	Finish   []string `yaml:"finish,omitempty"`
 	Sleep    *int     `yaml:"sleep,omitempty"`
 	LogSleep *int     `yaml:"logsleep,omitempty"`
+
+	// Determines whether the service should be restarted by runit after it
+	// exits. This isn't actually written to the servicebuilder file but
+	// determines how the stage directory is written and the service
+	// invoked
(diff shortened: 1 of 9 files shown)


pkg/runit/servicebuilder.go
```diff
@@ -63,6 +63,12 @@ type ServiceTemplate struct {
 	Finish   []string `yaml:"finish,omitempty"`
 	Sleep    *int     `yaml:"sleep,omitempty"`
 	LogSleep *int     `yaml:"logsleep,omitempty"`
+
+	// Determines whether the service should be restarted by runit after it
+	// exits. This isn't actually written to the servicbuilder file but
```
rudle: typo: servicbuilder

mpuncel:
I also had my qualms about adding a field to the template, but i also didn't want to pass a map[serviceName]RestartPolicy or something like that. Now i just thought of a way that might be better though, can make a `ServiceConfig` struct that has both a template and a restart policy. i'll see how that looks

mpuncel:
hmm we actually just straight up marshal the `map[string]ServiceTemplate` being passed around, so it gets a little ugly

PR #705, opened 2016-11-28
Mpuncel/uuid p2 rm

mpuncel:


bin/p2-rm/main.go
```diff
@@ -24,183 +20,90 @@ import (
...
+	err := handleLegacyPodRemoval(consulClient, labeler)
+	if err != nil {
+		fmt.Fprintf(os.Stderr, err.Error())
+		os.Exit(1)
+	}
+}
+
+func handleLegacyPodRemoval(consulClient consulutil.ConsulClient, labeler labels.ApplicatorWithoutWatches) error {
 	if *nodeName == "" {
 		hostname, err := os.Hostname()
```
rudle: Is this behaviour all that helpful? I didn't find `p2-rm` installed in the datacenter so I can't imagine a use case for it.

bin/p2-rm/main.go
```diff
@@ -24,183 +20,90 @@ import (
...
+		os.Exit(1)
 	}
 
-	if podLabels.Labels.Has(rc.RCIDLabel) {
-		return true, fields.ID(podLabels.Labels.Get(rc.RCIDLabel)), nil
-	}
+	rm := NewUUIDP2RM(consulClient, labeler)
 
-	return false, "", nil
+	// UUID pods are never managed by replication controllers
```
rudle: This looks like a bug waiting to bite us next year sometime. Does it make sense to pull `checkForManagingReplicationController` up into some shared code?

bin/p2-rm/main.go
mpuncel: i'm okay with getting rid of it

bin/p2-rm/main.go
mpuncel: good call

mpuncel:
the suggestion to check for managing RCs led me to refactor this completely, there's now only one P2RM object that is constructed differently and just makes difference decisions about how to delete the pod

bin/p2-rm/rm.go
```diff
@@ -0,0 +1,167 @@
...
+
+	if podLabels.Labels.Has(rc.RCIDLabel) {
+		return true, fields.ID(podLabels.Labels.Get(rc.RCIDLabel)), nil
+	}
+
+	return false, "", nil
+}
+
+func (rm *P2RM) decrementDesiredCount(id fields.ID) error {
+	session, _, err := rm.Store.NewSession(sessionName(id), nil)
```
petertseng: showing renewal errors deemed no longer useful? (Not like we were taking any action on it anyway...)

bin/p2-rm/rm.go
```diff
@@ -0,0 +1,167 @@
...
+	}
+
+	err = rm.PodStore.Unschedule(podUniqueKey)
+	if err != nil {
+		return err
+	}
+
+	err = rm.Labeler.RemoveAllLabels(labels.POD, podUniqueKey.String())
+	if err != nil {
+		return err
```
petertseng: any desire to show the same "partially deleted, re-run to delete the labels" here?

bin/p2-rm/rm.go
```diff
@@ -0,0 +1,174 @@
...
+		return fmt.Errorf("Unable to unschedule pod: %s", err)
+	}
+	return nil
+}
+
+func (rm *P2RM) removePodLabels() error {
+	err := rm.Labeler.RemoveAllLabels(labels.POD, rm.LabelID)
+	if err != nil {
+		return fmt.Errorf(
+			"node is partially deleted. re-run command to finish deleting\n"+
```
petertseng: would "node" make sense for a UUID pod?

bin/p2-rm/rm.go
```diff
@@ -0,0 +1,174 @@
...
+
+	err = rm.PodStore.Unschedule(podUniqueKey)
+	if err != nil {
+		return fmt.Errorf("Unable to unschedule pod: %s", err)
+	}
+	return nil
+}
+
+func (rm *P2RM) removePodLabels() error {
+	err := rm.Labeler.RemoveAllLabels(labels.POD, rm.LabelID)
```
petertseng: would `rm.LabelID` make sense for a UUID pod? previously, it had been `podUniqueKey.String()`

petertseng: answering own question: yes. `rm.LabelID` gets set to `podUniqueKey.String()` for a UUID pod.

bin/p2-rm/rm.go
petertseng: Previously, there was code that was doing:

```go
	go func(errChan chan error) {
 		select {
 		case <-errChan:
 			fmt.Fprintf(os.Stderr, "Got renewal error, proceeding without a lock\n%v\n", err)
 		default:
 		}
 	}(renewalErrChan)
```

I was just confirming that its removal was intentional, which it very well could be

mpuncel:
made one last change that allows the command to succeed if the uuid pod doesn't exist, which is necessary to be able to re-run it to do label deletions

bin/p2-rm/rm.go
mpuncel: ha i have no memory of doing that, but i stand by the decision to remove it since it's confusing and doesn't change behavior.

bin/p2-rm/rm.go
```diff
@@ -0,0 +1,178 @@
...
+}
+
+func (rm *P2RM) deletePod() error {
+	if rm.PodUniqueKey == "" {
+		err := rm.deleteLegacyPod()
+		if err != nil {
+			return err
+		}
+	} else {
+		err := rm.deleteUUIDPod(rm.PodUniqueKey, rm.PodName)
```
mpuncel: just realized i don't need to pass args here

mpuncel:
rebased

mpuncel:
this PR is build failure central

PR #706, opened 2016-11-28
Correctly set pod status field for uuid pods.

mpuncel:
When installing a pod, the PodStatus field will be set to
"launched".

When uninstalling it will be set to "removed".


rudle:
:ship:

PR #707, opened 2016-11-29
Remove launchable_id key from launchable stanza

mpuncel:
Currently P2 has two notions of launchable_id, one being the key in the
launchable stanza map, and the other being the value under
launchable_id: in the launchable stanza itself. Typically these are the
same and they cause confusion to our users.

Since maps always need keys, it makes sense to remove the inner
launchable_id key from launchable stanza and use the map key instead.

### DIFF
--- a/pkg/manifest/manifest.go
+++ b/pkg/manifest/manifest.go
@@ -331,8 +331,8 @@ func (manifest *manifest) WriteConfig(out io.Writer) error {
 
 func (manifest *manifest) WritePlatformConfig(out io.Writer) error {
 	platConf := make(map[launch.LaunchableID]interface{})
-	for _, stanza := range manifest.LaunchableStanzas {
-		platConf[stanza.LaunchableId] = map[launch.LaunchableID]interface{}{
+	for launchableID, stanza := range manifest.LaunchableStanzas {
+		platConf[launchableID] = map[launch.LaunchableID]interface{}{
 			"cgroup": stanza.CgroupConfig,
 		}
 	}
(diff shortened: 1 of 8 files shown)


mpuncel:
I could use some help thinking through any bad consequences that of a the two launchable IDs mismatching. I should probably also search through all of our manifests to see if this is a common occurrence

pkg/manifest/manifest.go
```diff
@@ -398,16 +398,14 @@ func ValidManifest(m Manifest) error {
 	if m.ID() == "" {
 		return fmt.Errorf("manifest must contain an 'id'")
 	}
-	for key, stanza := range m.GetLaunchableStanzas() {
+	for launchableID, stanza := range m.GetLaunchableStanzas() {
 		switch {
 		case stanza.LaunchableType == "":
-			return fmt.Errorf("'%s': launchable must contain a 'launchable_type'", key)
-		case stanza.LaunchableId == "":
```
mpuncel: If we desire a less risky change, we could probably make this validation fail if the two notions of launchable ID differ and address any violations case-by-case

pkg/manifest/manifest.go
```diff
@@ -331,8 +331,8 @@ func (manifest *manifest) WriteConfig(out io.Writer) error {
 
 func (manifest *manifest) WritePlatformConfig(out io.Writer) error {
 	platConf := make(map[launch.LaunchableID]interface{})
-	for _, stanza := range manifest.LaunchableStanzas {
-		platConf[stanza.LaunchableId] = map[launch.LaunchableID]interface{}{
+	for launchableID, stanza := range manifest.LaunchableStanzas {
+		platConf[launchableID] = map[launch.LaunchableID]interface{}{
```
mpuncel: I don't have great knowledge about how `platConf` gets used, so would appreciate comment

petertseng:
I feel good about this to remove duplication and potential confusion. We can probably easily check whether any manifest has a different launchable_id than key, and if none does then full speed ahead.

petertseng:
also checked, same result.

PR #710, opened 2016-11-30
Chmod 0777 sqlite database directory after creating it

mpuncel:
This gets around the umask for the process creating the directory with
permissions other than 0777.

Additionally, fix a bug where if the directory already existed and was
not executable, it would not be made executable (because chown 0666 was
used instead of 0777).


mpuncel:
thanks @rudle for helping me tracked this down you saved a ton of time

rudle:
:ship:

PR #712, opened 2016-11-30
Mpuncel/sqlite pruning

mpuncel:


pkg/preparer/podprocess/reporter.go
```diff
@@ -241,8 +271,46 @@ func (r *Reporter) initWorkspaceDir() error {
...
+	wg.Add(1)
+	go func() {
+		defer wg.Done()
+		r.publishProcessExits(quitCh)
+	}()
+
+	wg.Add(1)
+	go func() {
+		defer wg.Done()
+		r.pruneRows(quitCh)
```
rudle: What happens if this doesn't return? Can we make pruning best effort? I'd feel more comfortable with this approach if we had an error reporter wired up.

pkg/preparer/podprocess/finish_service.go
```diff
@@ -23,6 +23,8 @@ type FinishService interface {
 	GetLatestFinishes(lastID int64) ([]FinishOutput, error)
 	// Gets the last finish result for a given PodUniqueKey
 	LastFinishForPodUniqueKey(podUniqueKey types.PodUniqueKey) (FinishOutput, error)
+	// Deletes any rows with dates before the specified time
+	PruneRowsBefore(time.Time) error
```
rudle: testing single comment on line

mpuncel:
tested this on a host and it worked as expected (although the file size went up probably because of the index)

PR #714, opened 2016-12-02
Mpuncel/remove reality index uninstall

mpuncel:


mpuncel:
LMK if you guys think the `current_manifest.yaml` change is stupid. That shouldn't be a state we encounter but i guess uninstall loops because of that file missing seem wrong

pkg/pods/pod.go
```diff
@@ -352,30 +352,33 @@ func (pod *Pod) EnvDir() string {
...
 		return err
 	}
 
-	// halt launchables
-	for _, launchable := range launchables {
-		disableFunc := func() {
-			err = launchable.Disable()
-		}
-		pod.withTimeWarnings("disable", launchable.ServiceID(), disableFunc)
+	if err == nil {
```
anthonybishopric:
```
if err == NoCurrentManifest { return }
```

anthonybishopric:
👍  to eliminating an endless looping behavior

pkg/pods/pod.go
mpuncel: i'll move the inner stuff to a function so it's easier to read

mpuncel:
might want to give this a once over again, the first commit is rewritten for clarity (but behavior is the same)

PR #715, opened 2016-12-05
Add configurable pause time between RC watches in RC farm.

mpuncel:
The RC tree size can grow up to a few Mb in large deployments, which
means that running many instances of the RC farm results in a sizable
amount of bandwidth consumption.

This commit adds a configurable time.Duration to the RC farm which
causes it to pause in between watches on the entire RC tree. This trades
responsiveness to new replication controllers for a decrease in request
rate to the datastore (and thus a bandwidth reduction).

Individual RC watches are unaffected, so once an RC is noticed and
adopted by a farm, changes to replicas_desired will still be processed
very quickly.

### DIFF
--- a/bin/p2-rctl-server/main.go
+++ b/bin/p2-rctl-server/main.go
@@ -107,6 +107,7 @@ func main() {
 		logger,
 		klabels.Everything(),
 		alerter,
+		1*time.Second,
 	).Start(nil)
 	roll.NewFarm(
 		roll.UpdateFactory{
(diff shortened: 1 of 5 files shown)


rudle:
:ship:
Which metrics will this affect? Perhaps bandwidth?

mpuncel:
Yeah it should allow us to tune responsiveness of the RC farm to new and deleted RCs and trade off with bandwidth

PR #716, opened 2016-12-05
Inline hook manifest into preparer config.

mpuncel:
Previously hook manifests were fetched from consul, where they were not
namespaced by hostname. This is a difficult operational strategy because
updating a hook manifest in consul will result in simultaneous updates
for every preparer, making partial or slow rollouts impossible.

With the manifest inlined in the preparer's config, a hook rollout means
a preparer rollout.


pkg/preparer/setup.go
```diff
@@ -399,13 +398,22 @@ func New(preparerConfig *PreparerConfig, logger logging.Logger) (*Preparer, erro
...
+	var hooksManifest manifest.Manifest
+	var hooksPod *pods.Pod
+	if preparerConfig.HooksManifest != "" {
+		hooksManifest, err = manifest.FromBytes([]byte(preparerConfig.HooksManifest))
+		if err != nil {
+			return nil, util.Errorf("Could not parse configured hooks manifest: %s", err)
+		}
+		hooksPodFactory := pods.NewHookFactory(filepath.Join(preparerConfig.PodRoot, "hooks"), preparerConfig.NodeName)
+		hooksPod = hooksPodFactory.NewHookPod(hooksManifest.ID())
+	}
```
rudle: Does the lack of an else clause here preserve some "hookless" behaviour we were supporting before? I'm asking because I'd like to see an error check (& crash) here instead of proceeding.

mpuncel: yeah in theory it should be possible to run the preparer without any hooks, although it's not a big deal if we want to require it

mpuncel:
now you must provide a sentinel value for the hooks manifest, or provide a valid manifest

PR #717, opened 2016-12-06
Mpuncel/grpc pod store

mpuncel:


mpuncel:
@anthonybishopric and @rudle  approved on slack

PR #719, opened 2016-12-13
use a serializable Nothing()

anthonybishopric:
The Kubernetes labels package ships with helper `labels.Nothing()` that returns a selector that cannot match anything. Unfortunately, this helper selector is not serializable to and from strings - attempts to do `labels.Parse(labels.Nothing().String())` result in parse errors. 

This adds a helper to P2's labels package to create a selector that is serializable.


rudle:
Here's the similar upstream change. We can revisit this patch once it's merged.

https://github.com/kubernetes/kubernetes/pull/38719

PR #720, opened 2016-12-13
artifact: Pass both pod ID and artifact name to discover endpoint

petertseng:
Instead of always passing only the launchable ID, we need to allow the
manifest to specify an override for the launchable.
An artifact_name field is added to the version stanza of a launchable
stana for this purpose.

At the same time, we need to also pass the pod ID to the discover
endpoint (and for this there is no override), so that we can check
whether the artifact<->pod pair makes sense.

So now the pod ID is passed as the parameter in the URL, while the
launchable ID is passed as a query parameter.

### DIFF
--- a/pkg/artifact/registry.go
+++ b/pkg/artifact/registry.go
@@ -9,12 +9,14 @@ import (
 	"github.com/square/p2/pkg/auth"
 	"github.com/square/p2/pkg/launch"
 	"github.com/square/p2/pkg/osversion"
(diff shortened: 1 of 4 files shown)


petertseng:
Considering adding a quick test to make sure that the override business does as I expect.

petertseng:
Tests have been added.

petertseng:
There has been no change in non-test code since the last approval (only test code changed), so I'll take the last approval as authoritative for this change as well.

PR #721, opened 2016-12-16
Make p2-inspect able to understand launchable versions.

mpuncel:
Launchables are not required to specify a Location anymore, which
changes the way we want p2-inspect to show version information.

### DIFF
--- a/pkg/inspect/inspect.go
+++ b/pkg/inspect/inspect.go
@@ -5,6 +5,7 @@ import (
 	"sort"
 
 	"github.com/square/p2/pkg/health"
+	"github.com/square/p2/pkg/launch"
 	"github.com/square/p2/pkg/store/consul"
 	"github.com/square/p2/pkg/types"
 )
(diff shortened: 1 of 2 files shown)


mpuncel:
```~/Development/go/src/github.com/square/p2/bin/p2-inspect master * ./p2-inspect `p2-headers` --pod p2-preparer --node aam12.sjc1.square | jq ''
{
  "p2-preparer": {
    "aam12.sjc1.square": {
      "intent_manifest_sha": "74c43f5a6cfe1b5b36868b769502e8b2aeae173a98dc87b9878b54901fead40a",
      "reality_manifest_sha": "74c43f5a6cfe1b5b36868b769502e8b2aeae173a98dc87b9878b54901fead40a",
      "reality_versions": {
        "p2-health-checker": {
          "location": "https://hoistrepo-api.vip/download/deployable/p2-health-checker/p2-health-checker_3fbb3dc963841aa473b10b09acfcde46bfef29dd.tar.gz"
        },
        "p2-log-bridge": {
          "location": "https://hoistrepo-api.vip/download/deployable/p2-log-bridge/p2-log-bridge_3fbb3dc963841aa473b10b09acfcde46bfef29dd.tar.gz"
        },
        "p2logconsumer": {
          "location": "https://hoistrepo-api.vip/download/deployable/p2logconsumer/p2logconsumer_3fbb3dc963841aa473b10b09acfcde46bfef29dd.tar.gz"
        },
        "preparer": {
          "location": "https://hoistrepo-api.vip/download/deployable/p2-preparer/p2-preparer_4ab070dbfc6af9ba21eadb646e75e68755fbc008.tar.gz"
        }
      },
      "health": "passing"
    }
  }
}
~/Development/go/src/github.com/square/p2/bin/p2-inspect master * ./p2-inspect `p2-headers` --pod slug --node aam12.sjc1.square | jq ''
{
  "slug": {
    "aam12.sjc1.square": {
      "intent_manifest_sha": "3fe0dfbf81c5533a2ba40efa7701af173a80ae3cca205ec4d977181c386d58ab",
      "reality_manifest_sha": "3fe0dfbf81c5533a2ba40efa7701af173a80ae3cca205ec4d977181c386d58ab",
      "reality_versions": {
        "slug": {
          "version": {
            "id": "9b81807da207ff21516f898c4e9ea726b8ec465b"
          }
        }
      },
      "health": "passing"
    }
  }
}```

File: pkg/inspect/inspect.go [hunk 1]
```diff
@@ -14,14 +15,25 @@ const (
 	REALITY_SOURCE
 )
 
+type LaunchableVersion struct {
+	Location string                    `json:"location,omitempty"`
```
rudle: To make sure I understand: this field, and thus this whole type will is only here to support a diversity of manifest versions. If that's the case, I'm surprised that this type is only useful out here at the edge. It might be better to encapsulate this sort of thing in either: `launch.LaunchableVersion` or, if we want that type to stay simple, an adapter in the `manifests` package.

mpuncel: typically `LaunchableStanza` would be the structure you'd see this presented in, but that's more information than `p2-inspect` should provide. You're right that this struct might be useful elsewhere, but I don't think it is today

mpuncel: maybe there is a case for adding a `location:` option to `LaunchableVersion` though, so that there's only one place to specify version

rudle: Based on my grep skills, this type extension thing is done in exactly one place. If we expect LaunchableVersion to replace all need to reference Location (we do) then I'm ok with this sort of code debt.

mpuncel:
addressed conflict with #725

PR #722, opened 2016-12-20
Fix bug in recovery from throttling of health publishing.

mpuncel:
p2-preparer executes health checks on the pods it has deployed, and
writes the health value to consul if and only if that health value has
changed since the last time it was written. This is done to reduce QPS
of health checks on Consul.

This means that the preparer must keep a copy of its view of the health
value that is stored remotely in order to make the decision of whether
to write or not.

Additionally, services with flapping health will have their health
updates throttled. This is implemented by writing an "unknown" health
value to Consul, and ceasing all health writes until the service stops
flapping.

This commit fixes a bug where a throttle would cause "unknown" to be
written to Consul, however the local view of the remote health value
would be set to the actual service health (e.g. "passing" or
"critical"). As a result, when the service recovers from flapping, no
health will be written because the health value matches what the
preparer thinks is stored remotely.

### DIFF
--- a/pkg/store/consul/healthmanager.go
+++ b/pkg/store/consul/healthmanager.go
@@ -128,15 +128,19 @@ func (m *consulHealthManager) NewUpdater(pod types.PodID, service string) Health
 	go func() {
 		sub := m.sessionPub.Subscribe()
 		defer sub.Unsubscribe()
+
+		subLogger := m.logger.SubLogger(logrus.Fields{
+			"service": service,
+			"pod":     pod,
+			"node":    m.node,
+		})
+		throttledCheckStream := throttleChecks(checksStream, *HealthMaxBucketSize, subLogger)
+
 		processHealthUpdater(
 			m.client,
-			checksStream,
+			throttledCheckStream,
 			sub.Chan(),
-			m.logger.SubLogger(logrus.Fields{
-				"service": service,
-				"pod":     pod,
-				"node":    m.node,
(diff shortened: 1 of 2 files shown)


rudle:
This change LGTM (I think), but the throttling behaviour deserves to be tested.

mpuncel:
`throttle` is actually instantiated and used all in one function with a single goroutine, so I think it wouldn't trigger the race detector anyway.

I agree that I wish there was a test around the throttling code, but it would require a lot of work to refactor all of this. Another day.

bwester:
Now that you point this out, it's obvious the old code is wrong. Amazing find!

I agree that this code is very complex. Perhaps it might be easier if throttling and flap detection were implemented as a self-contained pipeline unit ahead of the main `processHealthUpdater()` loop?

mpuncel:
I added a commit to implement benji's suggestion. There is now an intermediary in the pipeline ahead of `processHealthUpdater()` that converts all values to "unknown" during a health flapping event. `processHealthUpdater()` is only responsible for not writing the same health value more than once in a row

mpuncel:
Notably missing is a test for recovering from a flapping state (which is the original bug fixed here). I'm not sure how to add a test for that without rearchitecting `throttleChecks()` to allow clearing a throttle externally or making the test sleep for a significant amount of time.

mpuncel:
addressed conflicts and squashed fixup commit

PR #723, opened 2016-12-21
Mpuncel/unified storage

mpuncel:
This is the beginning of what will be a long effort to decouple P2 from the storage semantics of Consul. The idea is resurrected from https://github.com/square/p2/pull/399

This PR moves 3 types into a new package `pkg/store`: `RollingUpdate`, `ReplicationController` and `Manifest`.

Code behavior should be unchanged, only import paths, type names and in some cases variable names where there were conflicts were changed.

I will add commits to this PR but each commit passes all tests and is incremental, so if a reviewer jumps in and approves it i'll merge it (if others think this is acceptable of course).


rudle:
With what's presented here, I can't see why the `store` package is different than the extant `types` package. The store package, as written, has some dependencies: on `uri`, `github.com/square/p2/pkg/util` for example, whereas the `types` package does not. I predict that as we add more behaviour to `store`, it will grow in its dependency count. I can share a rendering of the dependency graph privately, if you'd like; davecheney/graphpkg was quite useful.

Given the risk of a dependency cycle causing us great sadness in the future, what do you think about leaving the types in a dependency-free package (`github.com/square/p2/pkg/types` or a sub package of this new one) keeping the constructors and RPC stubs in `store`.

It's possible you've already considered all this and I'm being a bit dense. Please enlighten me!

mpuncel:
I guess I'm more optimistic about being able to avoid circular imports. `pkg/uri` for example should never depend on a logical P2 type.

mpuncel:
What do you think about trying to proceed with just `pkg/store` (and i'd probably move the current `types` package into it too). I think it will be easier to fix a circular import down the line by creating a second package rather than the other way around

rudle:
What about a compromise? When you move `types.go` into this package, can you also move the type definitions from the rest of the files into that one? That will make it easy to extract if we need to in the future.

mpuncel:
sounds good 👍

mpuncel:
now i'm thinking my logic is flawed, it's probably easier to start with two packages now and merge them later rather than the other way around

mpuncel:
also I think putting all the types into `types.go` will make for a gigantic file, maybe that's okay though

mpuncel:
i'm starting to lose sight of the goal here. I took a look at the dependency graph and `pkg/store` mostly depends on standard library things but also things like `pkg/launch` which depends on some types from `pkg/runit` and `pkg/cgroups` etc. I could go super down the rabbit hole and put all of these things into `pkg/store` and have those things depend on `pkg/store` for their types. Ultimately that makes sense from the perspective of "everything we store in consul or any other datastore should be in `pkg/store`.

However it means a LOT of stuff will be in `pkg/store`

mpuncel:
it looks like kubernetes has a `pkg/api/types.go` that has ~ 2500 lines just of type definitions, so that gives me a little more confidence

rudle:
> "everything we store in consul or any other datastore should be in pkg/store."

Is that the stated goal? That does sound like a rabbit hole, and a unwieldy package.

A more refined set of packages would probably try to do something like _cohesively implement all of  the API (ie. CRUD + Watch)_ on resources defined in `types`.

mpuncel:
declining this, I think #725 accomplished most of the goal here, to make it obvious when a coupling to consul is being used

PR #724, opened 2016-12-22
Add stack to recovered panic output in farms.

mpuncel:
The replication controller and rolling update farms run many workers in
separate goroutines. Since panics such as nil pointer dereferences
happen from time to time, each worker is wrapped in a recover() in order
to catch panics and log them. However, only the panic message e.g.
"runtime error: invalid memory address or nil pointer dereference" was
being shown, with no line number or stack information.

This commit adds stack information to the output of the log message,
making it trivial to find where the original panic occurred.

### DIFF
--- a/pkg/rc/farm.go
+++ b/pkg/rc/farm.go
@@ -1,6 +1,7 @@
 package rc
 
 import (
+	"fmt"
 	"sync"
 	"time"
 
@@ -229,9 +230,15 @@ START_LOOP:
 					defer func() {
 						if r := recover(); r != nil {
 							err := util.Errorf("Caught panic in rc farm: %s", r)
+
+							stackErr, ok := err.(util.StackError)
+							msg := "Caught panic in rc farm"
+							if ok {
+								msg = fmt.Sprintf("%s:\n%s", msg, stackErr.Stack())
+							}
 							rcLogger.WithError(err).
 								WithField("rc_id", id).
-								Errorln("Caught panic in rc farm")
+								Errorln(msg)
 						}
 					}()
 					// disabled-ness is handled in watchdesires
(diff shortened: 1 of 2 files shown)


pkg/rc/farm.go
```diff
@@ -228,7 +229,7 @@ START_LOOP:
 				go func(id fields.ID) {
 					defer func() {
 						if r := recover(); r != nil {
-							err := util.Errorf("Caught panic in rc farm: %s", r)
+							err := util.Errorf("Caught panic in rc farm: %s\n%s", r, debug.Stack())
```
bwester: This is interesting: `util.Errorf()` already captures a full stack trace—it just isn't used anywhere. Could there be a more general way to report the stack? Perhaps not, because that method is used everywhere as a generic `error` wrapper, but a full stack trace would only be useful in specific places like this.

pkg/rc/farm.go
```diff
@@ -228,7 +229,7 @@ START_LOOP:
...
 							rcLogger.WithError(err).
```
bwester: `util.WithError()` optionally tags the error with a filename and line number where the error was created (if the error implements `util.CallsiteError`). As written, this log line's tags will point to the "Caught panic..." message above. For a panic handler, would it make sense to log the location of the panic instead? I assume is what is actually interesting.

pkg/rc/farm.go
mpuncel: maybe I should just be logging `err.Stack()` in the log message below, which is what I really care about

pkg/rc/farm.go
mpuncel: I do like the "caught panic" message because it shows me that the panic was swallowed and didn't crash the server. It is slightly unfortunate that the line number doesn't point to the _actual_ problem, but since that information is present in the stack trace I think it's probably okay?

bwester:
This version too looks all right. The main test is: Does this give you enough information to debug a panic? If so (and you might want to verify that), then 🚢 .

mpuncel:
I used it to debug a panic yesterday so I'll say yes :)

PR #725, opened 2016-12-22
Move pkg/kp underneath pkg/store and rename to consul.

mpuncel:
pkg/kp is inextricably tied to Consul semantics. A lot of the operations
are specific to consul, e.g. session creation, locking, and accepting a
PodPrefix parameter in GetPod() and SetPod().

This PR organizationally paves the way to support multiple backends by
naming kp what it is: consul-specific.

The idea is that pkg/store will have constructors for separate backend
objects that provide the same functionality. For example, any package
that needs to operate on replication controllers might define an
interface like

type RCGetter interface {
    GetRC(rc_fields.ID) rc_fields.ReplicationController
}

which it will take as an argument to its constructor.

pkg/store can offer multiple constructurs e.g. for ConsulStore and
ETCDStore, both of which implement the GetRC interface.


mpuncel:
this PR is an alternative to https://github.com/square/p2/pull/399. It aims to set up the repo to better be datastore agnostic in the future by calling `pkg/kp` what it is: consul-specific.

I'm not of the opinion that #399  is the wrong way of going about things. We don't really care if the concretely stored types are in separate packages, at least not for being backend agnostic (although maybe it makes the repo harder to read). Really all we care about is some unified interface defined in terms of the concrete types and implemented by multiple datastore backends.

pkg/rc/farm.go
```diff
@@ -53,7 +53,7 @@ type Farm struct {
 
 	children map[fields.ID]childRC
 	childMu  sync.Mutex
-	session  kp.Session
+	session  consul.Session
```
rudle: I like this change because this sort of thing is now a very visible sign of tight coupling.

pkg/replication/replicator.go
```diff
@@ -134,14 +134,14 @@ func (r replicator) initializeReplicationWithCheck(
 	overrideLock bool,
 	ignoreControllers bool,
 	concurrentRealityRequests int,
-	checkPreparers bool,
+	checconsulreparers bool,
```
rudle: haha, I think this is a `sed` error.

pkg/store/consul/rcstore/consul_store.go
```diff
@@ -639,13 +639,13 @@ func (s *consulStore) ownershipLockPath(rcID fields.ID) (string, error) {
 
 // Acquires a lock on the RC that should be used by RC farm goroutines, whose
 // job it is to carry out the intent of the RC
-func (s *consulStore) LockForOwnership(rcID fields.ID, session kp.Session) (consulutil.Unlocker, error) {
-	lockPath, err := s.rcLockPath(rcID)
+func (s *consulStore) LockForOwnership(rcID fields.ID, session consul.Session) (consulutil.Unlocker, error) {
+	LockPath, err := s.rcLockPath(rcID)
```
rudle: hm? I don't see why this happened.

pkg/replication/replicator.go
mpuncel: oh gosh lol it matched `kp`

mpuncel:
Added a package comment describing the rationale and future purpose of `pkg/store`, going to proofread for any more sed errors now

bin/grpc/servers/podstore/server.go
```diff
@@ -29,9 +29,9 @@ func main() {
 	// Parse custom flags + standard Consul routing options
 	_, opts, _ := flags.ParseWithConsulOptions()
 
-	client := kp.NewConsulClient(opts)
-	podStore := kp_podstore.NewConsul(client.KV())
-	podStatusStore := podstatus.NewConsul(statusstore.NewConsul(client), kp.PreparerPodStatusNamespace)
+	client := consul.NewConsulClient(opts)
+	podStore := consul_podstore.NewConsul(client.KV())
+	podStatusStore := podstatus.NewConsul(statusstore.NewConsul(client), consul.PreparerPodStatusNamespace)
```
mpuncel: pod status namespaces will likely carry over to other datastores as well, but I don't know where else to put it at the moment

bin/p2-ds-farm/main.go
```diff
@@ -40,10 +40,10 @@ func main() {
 	quitCh := make(chan struct{})
 
 	_, consulOpts, labeler := flags.ParseWithConsulOptions()
-	client := kp.NewConsulClient(consulOpts)
+	client := consul.NewConsulClient(consulOpts)
 	logger := logging.NewLogger(logrus.Fields{})
 	dsStore := dsstore.NewConsul(client, 3, &logger)
-	kpStore := kp.NewConsulStore(client)
+	consulStore := consul.NewConsulStore(client)
```
mpuncel: I want to rename this to `NewStore` since `Consul` is redundant now but i'll save that for later because `gorename` is fighting me

mpuncel:
actually i'm going to do the proofreading later tonight

pkg/preparer/podprocess/reporter.go
```diff
@@ -234,7 +234,7 @@ func (r *Reporter) initWorkspaceDir() error {
 		return util.Errorf("Could not stat workspace directory: %s", err)
 	} else {
 		if !dirInfo.IsDir() {
-			return util.Errorf("Configured worskpace directory %s already exists but is not a directory", r.workspaceDirPath)
```
mpuncel: will fix

pkg/replication/replicator.go
```diff
@@ -134,14 +134,14 @@ func (r replicator) initializeReplicationWithCheck(
 	overrideLock bool,
 	ignoreControllers bool,
 	concurrentRealityRequests int,
-	checkPreparers bool,
+	chekPreparers bool,
```
mpuncel: will fix all these

mpuncel: this is when I tried to do a sed to fix my previous sed :D i've learned a lesson now

pkg/watch/health.go
```diff
@@ -239,7 +239,7 @@ func (p *PodWatch) checkHealth() {
 		return
 	}
 
-	if err = p.updater.PutHealth(resToKPRes(health)); err != nil {
```
mpuncel: will fix caps

pkg/preparer/podprocess/reporter.go
mpuncel: if i had spelled this right initially it wouldn't have matched `kp`

mpuncel:
I really like this change, also now the word "consul" looks really strange to me

PR #726, opened 2017-01-09
Update dependencies

bwester:
Upgrade a bunch of dependencies, just to stay current and bug-free. Tagged versions are preferred to head revision, when applicable.


bwester:
Haha, thanks for reading through the commit messages. I'll fix them. For reference, it was x/net that caused x/text to be pulled in.

PR #730, opened 2017-01-19
Mpuncel/entry point handle basename

mpuncel:


pkg/hoist/hoist_launchable_test.go
```diff
@@ -348,7 +298,7 @@ func TestLaunchWithFailingEnable(t *testing.T) {
 	err := hl.Launch(sb, sv)
 	Assert(t).IsNotNil(err, "Expected error while launching")
 	_, ok := err.(launch.EnableError)
-	Assert(t).IsTrue(ok, "Expected enable error to be returned")
+	Assert(t).IsTrue(ok, fmt.Sprintf("Expected enable error to be returned, was %s", err))
```
petertseng: a nitpicker might suggest that this be made separate

mpuncel:
@petertseng I rewrote the implementation in hoist_launchable.go because the integration test caught a problem. tests are the same since you last looked

pkg/hoist/failing_scripts_test_hoist_launchable/installs/testLaunchable_abc123/bin/launch
```diff
@@ -0,0 +1,2 @@
+#!/bin/bash
+exit 0
```
petertseng: just so I'm clear... I'm guessing this now needs to be added because the `failing_scripts_test_hoist_launchable` test would otherwise have a missing `bin/launch` cahsing an error. let me know if I'm wrong.

integration/single-node-slug-deploy/Vagrantfile
```diff
@@ -8,8 +8,8 @@ Vagrant.configure(VAGRANTFILE_API_VERSION) do |config|
   config.vm.provider "virtualbox" do |vb|
      vb.customize ["modifyvm", :id, "--accelerate3d", "off"]
   end
-  config.vm.box_url = "https://f0fff3908f081cb6461b407be80daf97f07ac418.googledrive.com/host/0BwtuV7VyVTSkUG1PM3pCeDJ4dVE/centos7.box"
```
petertseng: whoa (this was such a strange URL to be getting a box from!)

pkg/hoist/hoist_launchable.go
```diff
@@ -228,27 +229,31 @@ func (hl *Launchable) Executables(
 		absEntryPointPath := filepath.Join(hl.InstallDir(), relativeEntryPoint)
 
 		entryPointInfo, err := os.Stat(absEntryPointPath)
-		if err != nil {
+		if os.IsNotExist(err) {
```
petertseng: sorry, did you just revert the previous commit with this?

pkg/hoist/hoist_launchable.go
```diff
@@ -259,18 +264,18 @@ func (hl *Launchable) Executables(
 			}
 
 			p2ExecArgs := p2exec.P2ExecArgs{
-				Command:          []string{filepath.Join(serviceDir, service.Name())},
+				Command:          []string{filepath.Join(hl.InstallDir(), relativePath)},
 				User:             hl.RunAs,
 				EnvDirs:          []string{hl.PodEnvDir, hl.EnvDir()},
-				ExtraEnv:         map[string]string{launch.EntryPointEnvVar: service.Name()},
+				ExtraEnv:         map[string]string{launch.EntryPointEnvVar: entryPointName},
```
petertseng: confirm it is more desirable to give them `bin__launch` rather than just `bin/launch`? (I have no strong opinion, just want to make sure)

pkg/hoist/hoist_launchable_test.go
```diff
@@ -120,38 +97,11 @@ func TestNonStandardEntryPoint(t *testing.T) {
...
-			t.Errorf("Did not find %s among executable paths", expected)
-		}
-	}
-}
-
-// Tests that the Executables() function fails if two runit services with the same
-// service name will be created. We specify bin/start (a normal file) and bin/start2
-// (a directory containing one file called start) which will generate a naming
-// collision and thus (we hope) an error
-func TestErrorIfConflictingServiceNames(t *testing.T) {
```
petertseng: I suppose if you really wanted you could keep this best but instead now it should ensure it accepts conflicting service names?

pkg/hoist/hoist_launchable_test.go
```diff
@@ -120,38 +97,11 @@ func TestNonStandardEntryPoint(t *testing.T) {
...
-	fakeLaunchable, sb := FakeHoistLaunchableForDir("multiple_script_test_hoist_launchable")
```
petertseng: but otherwise, if you will delete this test, maybe you should delete the multiple_script_test_hoist_launchable dir too?

pkg/hoist/failing_scripts_test_hoist_launchable/installs/testLaunchable_abc123/bin/launch
mpuncel: 👍

pkg/hoist/hoist_launchable.go
mpuncel: hmmmm that's interesting, it's probably better to keep the slashes. then it's just the runit service name where the double underscores happen. good suggestion will fix

pkg/hoist/hoist_launchable_test.go
mpuncel: it's used for other tests

petertseng:
> I am slightly confused because I think you reverted your first commit in your second

Slightly sad that Github deems the comment to be on an outdated diff https://github.com/square/p2/pull/730#discussion_r96950405

mpuncel:
@petertseng I undid the first commit revert so I think it should be outdated?

mpuncel:
going to squash that last commit in and merge

PR #731, opened 2017-01-23
Fix p2-rm output when removing a uuid pod.

mpuncel:
When removing a uuid pod, the uuid is passed on the command line instead
of the node name since a node name is not necessarily enough information
to uniquely identify the pod to be removed.

However the output at the end of the script was always depending on the
node name, resulting in weird output like ": successfully removed
<app>". The output is now dependent on whether the pod removed was a
UUID pod or not.

### DIFF
--- a/bin/p2-rm/main.go
+++ b/bin/p2-rm/main.go
@@ -80,7 +80,11 @@ func handlePodRemoval(consulClient consulutil.ConsulClient, labeler labels.Appli
 		}
 	}
 
-	fmt.Printf("%s: successfully removed %s\n", rm.NodeName, rm.PodID)
+	if rm.NodeName != "" {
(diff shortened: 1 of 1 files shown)


bin/p2-rm/main.go
```diff
@@ -80,7 +80,11 @@ func handlePodRemoval(consulClient consulutil.ConsulClient, labeler labels.Appli
 		}
 	}
 
-	fmt.Printf("%s: successfully removed %s\n", rm.NodeName, rm.PodID)
+	if rm.NodeName != "" {
```
petertseng: at first, I was going to leave a comment of the form "But the condition at the top says `if *podUniqueKey != ""`, and now it is not clear how the two are related", but then I realized that even if that were changed, the condition here would be something like `if rm.PodUniqueKey != ""`, which isn't necessarily better. So I guess I have no change to suggest

PR #733, opened 2017-01-23
Mpuncel/upgrade consul library

mpuncel:


mpuncel:
note to reviewers: I made a commit for every godep operation that ran successfully. This means that the commit for `godep update github.com/hashicorp/consul/...` has subvendored packages `github.com/square/p2/vendor/github.com/hashicorp/consul/vendor/*` checked in.

The final godep-save operation kept failing with missing packages, so I ran `go get` for each package and checked them out at the version that the consul library had vendored.

A final `godep save ./...` removed all of the inner vendor directories.

It's possible that some packages used by the consul client might still be out of date, unfortunately they don't use godep so it's not as simple as a `godep restore` in the consul repo.

mpuncel:
it appears this version doesn't build, I probably have more packages to update

mpuncel:
I iteratively updated dependencies to match those of the consul library until compilation worked.

mpuncel:
I wrote a rake task that iterates over the packages in consul's `vendor.json` and checks out everything in $GOPATH with the same version and then runs `godep update` on it. Looks like it worked!

mpuncel:
if you're concerned that 140k lines were deleted and only 33k were added, it seems that `github.com/fsouza/go-dockerclient/` was removed and accounted for 95kloc: 
```
~/Development/go/src/github.com/square/p2 master git diff mpuncel/upgrade-consul-library  --name-status | grep ^A | grep fsouza\/go-dockerclient | awk '{print $2}' | xargs cat | wc -l
   95768
```

bwester:
I replicated your work through other means and got practically the same result as you, so I have confidence that this is a good update. This PR currently has a few loose ends: packages listed in Godeps.json that aren't in vendor/ and packages that are vendored that aren't used. Re-vendoring the repo should fix all of that up.

I have one nit: setting all external packages to the version specified in Consul's repo ends up _downgrading_ a few of the packages in this library (see #726). Can you exclude those packages' changes from this PR?

Rakefile
```diff
@@ -1,14 +1,16 @@
+require "json"
+
 def e(cmd)
   puts cmd
-  system(cmd) || abort("Error running `#{cmd}`")
+  system(cmd) || raise("Error running `#{cmd}`")
```
bwester: Why the difference?

Rakefile
```diff
@@ -1,14 +1,16 @@
...
 end
 
 def target(expand="")
   root = File.dirname(__FILE__)
   File.expand_path(expand, File.join(root, 'target'))
 end
 
-task :godep_check do
+task :odep_check do
```
bwester: Oops

Rakefile
```diff
@@ -81,3 +83,46 @@ end
...
+
+  # now iterate over our Godeps.json and find any versions that mismatch and print them
+  p2_godep_contents["Deps"].each do |package|
+    package_name = package["ImportPath"]
+    consul_rev = consul_deps[package_name]
+    next unless consul_rev
+
+    our_rev = package["Rev"]
+    if consul_rev != our_rev
+      # grab just the first three items e.g. github.com/hashicorp/consul not github.com/hashicor/consul/api/inner/package/thing
```
bwester: nit: hashicor**p**

Rakefile
mpuncel: so it can be caught instead of exiting the program. I also think if any of the other commands were to encounter an error here the stack trace information would be useful

Rakefile
```diff
@@ -81,3 +83,46 @@ end
...
+      first_three = package_name.split("/")[0..2]
+
+      begin
+        Dir.chdir(File.join(ENV["GOPATH"], "src", first_three[0], first_three[1], first_three[2])) do
```
bwester: nit: `File.join(foo, *first_three)`

Rakefile
```diff
@@ -81,3 +83,46 @@ end
...
+  # build a map of package names to revisions from consul's vendor.json
+  consul_deps = {}
+  consul_vendor_contents["package"].each do |package|
+    consul_deps[package["path"]] = package["revision"]
+  end
+
+  # now iterate over our Godeps.json and find any versions that mismatch and print them
+  p2_godep_contents["Deps"].each do |package|
+    package_name = package["ImportPath"]
+    consul_rev = consul_deps[package_name]
```
bwester: What behavior will this script have when a new dependency is encountered in Consul?

mpuncel: it will fail. I'll make it do a `go get` in that case

mpuncel:
@bwester I think i'm going to change my script to only do a `godep update` if the revision in consul's deps is newer than ours. then i'll probably scrap all the commits and run the script

mpuncel:
merged #734 instead, closing this since it's a duplicate

PR #735, opened 2017-01-25
Restore old naming scheme for runit service directories for legacy pods.

mpuncel:
Commit b22751cc08446367b6181ee5b602fc0d3a469c0e had modified the naming
scheme for runit service directories (i.e. /var/service/*).

Prior to that commit, the directory names were built like the following:
"/var/service/{podID}__{launchableID}__{filepath.Base(entryPointPath)}
yielding "/var/service/some-pod__some-launchable__launch" for a pod
named "some-pod", launchable named "some-launchable" with the bin/launch
entry point used.

After that commit, the naming scheme was modified to include information
about the directory hierarchy to the entry point relative to the root of
the artifact: e.g. bin__launch is used for the entry point portion
instead of just the file basename. This allows artifacts to have
multiple entry points with the same filename (in different directories).

This commit reverts this naming scheme change for "legacy" pods which do
not have a UUID, and keeps the new naming scheme if the pod has a UUID.

### DIFF
--- a/pkg/hoist/hoist_launchable.go
+++ b/pkg/hoist/hoist_launchable.go
@@ -40,6 +40,18 @@ type Launchable struct {
 	Location         *url.URL                   // URL to download the artifact from
 	VerificationData auth.VerificationData      // Paths to files used to verify the artifact
 	EntryPoints      []string                   // paths to entry points to launch under runit
+
+	// IsUUIDPod indicates whether the launchable is part of a "uuid pod"
+	// vs a "legacy pod". Currently this information is used for determining the name of the runit service directories to use
+	// for each of a pod's processes. UUID pods use the "new" naming scheme and legacy pods use the "old" naming scheme.
+	//
+	// Example old naming scheme:
+	// /var/service/some-pod__some-launchable__launch/
+	//
+	// Example new naming scheme (includes full relative path to entry point with
+	// slashes exchanged for double underscores):
+	// /var/service/some-pod-<uuid>__some-launchable__bin__launch/
+	IsUUIDPod bool
 }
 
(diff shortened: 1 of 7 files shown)


pkg/hoist/hoist_launchable_test.go
```diff
@@ -141,7 +230,8 @@ func TestDisable(t *testing.T) {
 }
 
 func TestFailingDisable(t *testing.T) {
-	hl, sb := FakeHoistLaunchableForDir("failing_scripts_test_hoist_launchable")
+	// IsUUIDPod has no bearing on the behavior of this test
```
rudle: What does this comment really _mean_? As usual, I recommend writing a specifically named wrapper for the default behaviour of FakeHoistLaunchableForDir (can give it the same name) and a specific one with more args, maybe called `FakeHoistLaunchableForDirWithUUIDPod`

pkg/hoist/hoist_launchable.go
```diff
@@ -250,7 +273,12 @@ func (hl *Launchable) Executables(
...
 
 		for _, relativePath := range relativeExecutablePaths {
-			entryPointName := strings.Replace(relativePath, "/", "__", -1)
+			var entryPointName string
+			if hl.IsUUIDPod {
+				entryPointName = strings.Replace(relativePath, "/", "__", -1)
+			} else {
+				entryPointName = filepath.Base(relativePath)
+			}
 			serviceName := fmt.Sprintf("%s__%s", hl.ServiceId, entryPointName)
```
rudle: o_O

mpuncel:
implemented your suggestion for the test helpers, squashed it in

rudle:
gogogogogo

PR #738, opened 2017-01-27
ds: report eligible and healthy pods per daemon set

petertseng:
We'd like to get more insight into how our daemon sets are doing
health-wise. Adding this reporting will help us do that.

### DIFF
--- a/pkg/ds/daemon_set.go
+++ b/pkg/ds/daemon_set.go
@@ -11,6 +11,7 @@ import (
 	"github.com/square/p2/pkg/health/checker"
 	"github.com/square/p2/pkg/labels"
 	"github.com/square/p2/pkg/logging"
+	p2metrics "github.com/square/p2/pkg/metrics"
 	"github.com/square/p2/pkg/replication"
 	"github.com/square/p2/pkg/scheduler"
 	"github.com/square/p2/pkg/store/consul"
@@ -18,6 +19,7 @@ import (
 	"github.com/square/p2/pkg/types"
 	"github.com/square/p2/pkg/util"
 
+	"github.com/rcrowley/go-metrics"
 	klabels "k8s.io/kubernetes/pkg/labels"
 )
 
@@ -44,6 +46,8 @@ type DaemonSet interface {
 
 	// Returns a list of all nodes that are selected by this daemon set's selector
 	EligibleNodes() ([]types.NodeName, error)
+
+	MetricNames(suffix string) []string
 
 	// WatchDesires watches for changes to its daemon set, then schedule/unschedule
 	// pods to to the nodes that it is responsible for
@@ -170,6 +174,19 @@ func (ds *daemonSet) EligibleNodes() ([]types.NodeName, error) {
 	return ds.scheduler.EligibleNodes(ds.Manifest, ds.NodeSelector)
 }
 
+func (ds *daemonSet) MetricNames(suffix string) []string {
+	prefix := "daemonset"
+	middles := []string{
+		ds.PodID().String() + "." + ds.ClusterName().String(),
+		ds.ID().String(),
+	}
+	var ret []string
+	for _, middle := range middles {
+		ret = append(ret, fmt.Sprintf("%s.%s.%s", prefix, middle, suffix))
+	}
+	return ret
+}
+
 func (ds *daemonSet) WatchDesires(
 	quitCh <-chan struct{},
 	updatedCh <-chan *fields.DaemonSet,
@@ -235,6 +252,12 @@ func (ds *daemonSet) WatchDesires(
 				}
 				ds.DaemonSet = *newDS
 
+				if reportErr := ds.reportEligible(); reportErr != nil {
+					// An error in sending the metrics shouldn't stop us from doing updates.
+					// Report it, and move on.
+					ds.logger.WithError(reportErr).Warnf("Error reporting number of eligible nodes")
+				}
+
 				if ds.Disabled {
 					ds.cancelReplication()
 					continue
@@ -280,6 +303,11 @@ func (ds *daemonSet) WatchDesires(
 					// channel closed
 					return
 				}
+				if reportErr := ds.reportEligible(); reportErr != nil {
+					// An error in sending the metrics shouldn't stop us from doing updates.
+					// Report it, and move on.
+					ds.logger.WithError(reportErr).Warnf("Error reporting number of eligible nodes")
+				}
 				if ds.Disabled {
 					continue
 				}
@@ -308,6 +336,18 @@ func (ds *daemonSet) WatchDesires(
 	}()
 
 	return errCh
+}
+
+func (ds *daemonSet) reportEligible() error {
+	eligible, err := ds.EligibleNodes()
+	if err != nil {
+		return util.Errorf("Error retrieving eligible nodes for daemon set: %v", err)
+	}
+	for _, name := range ds.MetricNames("total") {
+		gauge := metrics.GetOrRegisterGauge(name, p2metrics.Registry)
+		gauge.Update(int64(len(eligible)))
+	}
+	return nil
 }
 
 // Watch for changes to nodes and sends update and delete signals
--- a/pkg/ds/farm.go
+++ b/pkg/ds/farm.go
@@ -15,6 +15,7 @@ import (
 	"github.com/square/p2/pkg/labels"
 	"github.com/square/p2/pkg/logging"
 	p2metrics "github.com/square/p2/pkg/metrics"
+	"github.com/square/p2/pkg/replication"
 	"github.com/square/p2/pkg/scheduler"
 	"github.com/square/p2/pkg/store/consul"
 	"github.com/square/p2/pkg/store/consul/consulutil"
@@ -47,6 +48,7 @@ type Farm struct {
 
 	healthChecker *checker.ConsulHealthChecker
 
+	monitorHealth  bool
 	cachedPodMatch bool
 }
 
@@ -69,6 +71,7 @@ func NewFarm(
 	alerter alerting.Alerter,
 	healthChecker *checker.ConsulHealthChecker,
 	rateLimitInterval time.Duration,
+	monitorHealth bool,
 	cachedPodMatch bool,
 ) *Farm {
 	if alerter == nil {
@@ -87,6 +90,7 @@ func NewFarm(
 		alerter:           alerter,
 		healthChecker:     healthChecker,
 		rateLimitInterval: rateLimitInterval,
+		monitorHealth:     monitorHealth,
 		cachedPodMatch:    cachedPodMatch,
 	}
 }
@@ -563,6 +567,34 @@ func (dsf *Farm) spawnDaemonSet(
 	deletedCh := make(chan *ds_fields.DaemonSet)
 
 	desiresCh := ds.WatchDesires(quitSpawnCh, updatedCh, deletedCh)
+
+	if dsf.monitorHealth {
+		go func() {
+			aggregateHealth := replication.AggregateHealth(ds.PodID(), *dsf.healthChecker)
+			ticks := time.NewTicker(1 * time.Minute)
+
+			defer aggregateHealth.Stop()
+			defer ticks.Stop()
+
+			for {
+				select {
+				case <-quitSpawnCh:
+					return
+				case <-ticks.C:
+					eligible, err := ds.EligibleNodes()
+					if err != nil {
+						dsLogger.WithError(err).Warnf("Error finding eligible nodes; can't count healthy nodes")
+						continue
+					}
+					numHealthy := aggregateHealth.NumHealthyOf(eligible)
+					for _, name := range ds.MetricNames("healthy") {
+						gauge := metrics.GetOrRegisterGauge(name, p2metrics.Registry)
+						gauge.Update(int64(numHealthy))
+					}
+				}
+			}
+		}()
+	}
 
 	return &childDS{
 		ds:        ds,
--- a/pkg/replication/health_aggregator.go
+++ b/pkg/replication/health_aggregator.go
@@ -80,6 +80,19 @@ func (p *podHealth) GetHealth(host types.NodeName) (health.Result, bool) {
 	return h, ok
 }
 
+func (p *podHealth) NumHealthyOf(hosts []types.NodeName) int {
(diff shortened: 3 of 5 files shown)


pkg/ds/daemon_set.go
```diff
@@ -304,12 +313,26 @@ func (ds *daemonSet) WatchDesires(
 			case <-quitCh:
 				return
 			}
+
+			if err == nil {
+				err = reportErr
```
petertseng: argh no, this doesn't happen if we `continue`. This is why this is hard. I have to move this to the top of the function now. I don't want to put it after every `continue` since there are too many of them!

pkg/ds/farm.go
```diff
@@ -564,6 +565,25 @@ func (dsf *Farm) spawnDaemonSet(
 
 	desiresCh := ds.WatchDesires(quitSpawnCh, updatedCh, deletedCh)
 
+	go func() {
+		aggregateHealth := replication.AggregateHealth(ds.PodID(), *dsf.healthChecker)
```
petertseng: `dsf.healthChecker` is a pointer to an interface. I have a sneaking suspicion it should get changed.

pkg/ds/farm.go
```diff
@@ -564,6 +565,25 @@ func (dsf *Farm) spawnDaemonSet(
...
+		ticks := time.NewTicker(1 * time.Minute)
+
+		defer aggregateHealth.Stop()
```
petertseng: `defer`ring these two operations is gratuitous. Could put it under `case <-quitSpawnCh:`; this is proofing it against the probability (whose value is unknown) that this goroutine gains more exit points.

pkg/ds/farm.go
petertseng: Consider this possibility - we have this `AggregateHealth` call that is used to *always* monitor the health (to emit the metric).

When we go enact a replication, we make **another** `AggregateHealth` call that creates a separate watch.

**Why not just pass this aggregator to the replication**?

It is entirely possible. ds is the only package using replication at the moment, so I feel we are relatively free to change the interface.

petertseng: Finally, a third consideration: Previously, the number of active watches accounted for by daemon sets is equal to the number of daemon sets currently doing replications.

With this, the number of active watches increases to the number of daemon sets, at all times, whether they be replicating or not. Does this hurt performance? If so, what do we do about it?

pkg/ds/farm.go
petertseng: Excuse me, what? This line panics on closing a nil channel... the `quit` channel of `replication.podHealth` is NEVER set, which means the health aggregator is unstoppable. This can't be right.

This implies we were never stopping any of our health aggregators??? We start one for every replication, and never clean them up, so the number of watches increases monotonically over time???

petertseng: After a replication ends, there is nobody reading off of the aggregator's channel anymore, so it blocks itself sending on the channel. Therefore, it no longer does any more watches. Therefore, the only impact is we accumulate goroutines that are blocked on channels, but we don't kill the network/consul, which is at least not as bad as I thought it would be.

pkg/ds/farm.go
petertseng: I'll not change it now, but a possibility for later.

pkg/artifact/downloader.go
```diff
@@ -65,13 +65,7 @@ func (l *downloader) Download(location *url.URL, verificationData auth.Verificat
...
 	}
 
-	// rewind a second time to allow the archive to be unpacked
-	_, err = artifactFile.Seek(0, os.SEEK_SET)
-	if err != nil {
-		return util.Errorf("Could not reset artifact file position after verification: %v", err)
-	}
-
-	err = gzip.ExtractTarGz(owner, artifactFile, dst)
+	err = gzip.ExtractTarGz(owner, artifactFile.Name(), dst)
```
petertseng: sorry, what? Did I accidentally throw in a change I didn't want

petertseng: You never saw this.

pkg/ds/daemon_set.go
```diff
@@ -310,6 +323,17 @@ func (ds *daemonSet) WatchDesires(
 	return errCh
 }
 
+func (ds *daemonSet) reportEligible() error {
+	eligible, err := ds.EligibleNodes()
+	if err != nil {
+		return util.Errorf("Error retrieving eligible nodes for daemon set: %v", err)
+	}
+	name := "total_" + ds.ID().String()
```
petertseng: _ as separator?

pkg/ds/farm.go
```diff
@@ -564,6 +565,26 @@ func (dsf *Farm) spawnDaemonSet(
...
+		defer ticks.Stop()
+
+		for {
+			select {
+			case <-quitSpawnCh:
+				return
+			case <-ticks.C:
+				numHealthy := aggregateHealth.NumHealthy()
+				name := "healthy_" + ds.ID().String()
```
petertseng: once again, _ as separator?

pkg/ds/daemon_set.go
mpuncel: hyphen somehow seems like a better choice to me but not sure i really know why i think that

petertseng: is it because there are already hyphens in the daemon set ID (given that it is a UUID_?

I don't feel strongly either way.

pkg/ds/daemon_set.go
```diff
@@ -235,6 +235,12 @@ func (ds *daemonSet) WatchDesires(
 				}
 				ds.DaemonSet = *newDS
 
+				if reportErr := ds.reportEligible(); reportErr != nil {
+					// An error in sending the metrics shouldn't stop us from doing updates.
+					// Report it, and move on.
+					ds.logger.Warnf("Error reporting number of eligible nodes: %v", reportErr)
```
petertseng: Does "too" imply I should still keep `reportErr` in the `Warnf` as well? Added.

mpuncel: probably get rid of the `reportErr` in the message, not a huge deal though

pkg/ds/farm.go
petertseng: I may have to do this one at a a time. First, a config option to turn on health monitoring. Then, if everything is good, we optimize by passing the health monitors to the replicators.

This is because if we do both, we have to do this weird thing where if health monitoring is enabled we pass the existing health monitor to the replicator, otherwise we construct a new one when the replicator is constructed.

petertseng:
A test deploy shows that the metrics get reported, and I didn't notice a huge performance impact.

I've switched to hyphens as separators.

Is it legal to use dots? What if it were something like dsid.healthy and dsid.total? Grouping them together under the DS ID seems useful.

petertseng:
I've made an executive decision to use some dots. going to test a deploy again.

pkg/replication/health_aggregator.go
```diff
@@ -79,6 +80,18 @@ func (p *podHealth) GetHealth(host types.NodeName) (health.Result, bool) {
 	return h, ok
 }
 
+func (p *podHealth) NumHealthy() int {
```
rudle: I can't figure out where this scopes its node list to the nodes targeted by the daemon set. I predict that you'll find that this overcounts the number of healthy hosts.

petertseng: I think you're right.

petertseng: Now we have to count the eligible nodes, them pass them into `NumHealthyOf`.

pkg/ds/daemon_set.go
```diff
@@ -170,6 +175,15 @@ func (ds *daemonSet) EligibleNodes() ([]types.NodeName, error) {
 	return ds.scheduler.EligibleNodes(ds.Manifest, ds.NodeSelector)
 }
 
+func (ds *daemonSet) metricName(name string) string {
+	return strings.Join([]string{
+		"daemonset",
+		ds.PodID().String(),
```
petertseng: for convenience, show pod ID. Maybe could also show ClusterName? I'd have to add a String() to ClusterName.

If we show both pod ID and cluster name, do we need to show the DS ID? Only if people make daemon sets that clash on both these properties...

pkg/ds/daemon_set.go
```diff
@@ -45,6 +48,8 @@ type DaemonSet interface {
 	// Returns a list of all nodes that are selected by this daemon set's selector
 	EligibleNodes() ([]types.NodeName, error)
 
+	metricName(name string) string
```
petertseng: warning: unexported function in interface prevents anyone outside this package from implementing the interface.

$GOPATH/pkgb/hi.go

```go
package pkgb

// Greeter can greet things.
type Greeter interface {
  greet(string) string
}
```

$GOPATH/pkga/hi.go

```go
package main

import "fmt"
import "pkgb"

type helloer struct{}

func (h *helloer) greet(name string) string {
  return "Hello, " + name + "!"
}

var _ pkgb.Greeter = &helloer{}

func main() {
  h := &helloer{}
  fmt.Println(h.greet("World"))
}
```

build output:

```
./hi.go:12: cannot use helloer literal (type *helloer) as type pkgb.Greeter in assignment:
	*helloer does not implement pkgb.Greeter (missing pkgb.greet method)
		have greet(string) string
		want pkgb.greet(string) string
```

I don't know if we care enough to change that.

mpuncel: can we lowercase the interface name to make that clearer? or is it referred to anywhere

petertseng: I actually don't think it is, according to `git grep \\bds\\.DaemonSet`, though it's less work to export the function. I don't really have a reason to keep it unexported... unless less work if anyone wants to implement it...?

petertseng:
I would really also like to have "number of up-to-date nodes", but that may have to come later.

petertseng:
I think I should put the cluster name in the metric name, because otherwise it gets too hard to find.

I can add the daemon set ID as a separate metric, I suppose, if people really want to just use daemon sets.

petertseng:
> just use daemon sets

I mean "just use daemon set IDs"

petertseng:
I was thinking "I would like these metrics to be discoverable **either** by DS ID **or** by pod ID + cluster name". Now, unless there's a way to alias one metric to another, now I'm emitting each metric twice. I'll test this out again.

petertseng:
Sorry, but going to ask for a re-review on the changes made since last review, since they are non-trivial

```diff
diff --git a/pkg/ds/daemon_set.go b/pkg/ds/daemon_set.go
index cab35a99..8f37545a 100644
--- a/pkg/ds/daemon_set.go
+++ b/pkg/ds/daemon_set.go
@@ -4,7 +4,6 @@ import (
 	"fmt"
 	"os"
 	"os/user"
-	"strings"
 	"time"
 
 	"github.com/square/p2/pkg/ds/fields"
@@ -48,7 +47,7 @@ type DaemonSet interface {
 	// Returns a list of all nodes that are selected by this daemon set's selector
 	EligibleNodes() ([]types.NodeName, error)
 
-	metricName(name string) string
+	MetricNames(suffix string) []string
 
 	// WatchDesires watches for changes to its daemon set, then schedule/unschedule
 	// pods to to the nodes that it is responsible for
@@ -175,13 +174,17 @@ func (ds *daemonSet) EligibleNodes() ([]types.NodeName, error) {
 	return ds.scheduler.EligibleNodes(ds.Manifest, ds.NodeSelector)
 }
 
-func (ds *daemonSet) metricName(name string) string {
-	return strings.Join([]string{
-		"daemonset",
-		ds.PodID().String(),
+func (ds *daemonSet) MetricNames(suffix string) []string {
+	prefix := "daemonset"
+	middles := []string{
+		ds.PodID().String() + "." + ds.ClusterName().String(),
 		ds.ID().String(),
-		name,
-	}, ".")
+	}
+	var ret []string
+	for _, middle := range middles {
+		ret = append(ret, fmt.Sprintf("%s.%s.%s", prefix, middle, suffix))
+	}
+	return ret
 }
 
 func (ds *daemonSet) WatchDesires(
@@ -340,9 +343,10 @@ func (ds *daemonSet) reportEligible() error {
 	if err != nil {
 		return util.Errorf("Error retrieving eligible nodes for daemon set: %v", err)
 	}
-	name := ds.metricName("total")
-	gauge := metrics.GetOrRegisterGauge(name, p2metrics.Registry)
-	gauge.Update(int64(len(eligible)))
+	for _, name := range ds.MetricNames("total") {
+		gauge := metrics.GetOrRegisterGauge(name, p2metrics.Registry)
+		gauge.Update(int64(len(eligible)))
+	}
 	return nil
 }
 
diff --git a/pkg/ds/farm.go b/pkg/ds/farm.go
index d52c730f..af6a8b22 100644
--- a/pkg/ds/farm.go
+++ b/pkg/ds/farm.go
@@ -587,9 +587,10 @@ func (dsf *Farm) spawnDaemonSet(
 						continue
 					}
 					numHealthy := aggregateHealth.NumHealthyOf(eligible)
-					name := ds.metricName("healthy")
-					gauge := metrics.GetOrRegisterGauge(name, p2metrics.Registry)
-					gauge.Update(int64(numHealthy))
+					for _, name := range ds.MetricNames("healthy") {
+						gauge := metrics.GetOrRegisterGauge(name, p2metrics.Registry)
+						gauge.Update(int64(numHealthy))
+					}
 				}
 			}
 		}()
diff --git a/pkg/ds/fields/fields.go b/pkg/ds/fields/fields.go
index d848c231..8d2b5266 100644
--- a/pkg/ds/fields/fields.go
+++ b/pkg/ds/fields/fields.go
@@ -20,6 +20,10 @@ func (id ID) String() string {
 // Cluster name is where the DaemonSet lives on
 type ClusterName string
 
+func (cn ClusterName) String() string {
+	return string(cn)
+}
+
 // DaemonSet holds the runtime state of a Daemon Set as saved in Consul.
 type DaemonSet struct {
 	// UUID for this DaemonSet
```

mpuncel:
so we're publishing two metrics per daemon set? if that is indeed what's happening then 👍

petertseng:
Yeah, one for the human-readable names (to make it easier to find when browsing all metrics if I don't have a ID -> name mapping on hand), and one for the DS ID (in case any tool just wants to use that directly, and/or in case there's a conflict in name)

PR #739, opened 2017-01-27
gzip: replace custom code with call to tar

petertseng:
As we've seen time and time again, it's very difficult to get our custom
tar code right (see #116, #373, #374, #376, #623). We have decided
simply to call `tar`, dropping privileges to the RunAs user of the pod.

Using `tar` insulates us from any more edge cases we missed as long as
`tar` was the tool used to create the tarball, and dropping privileges
lets Unix permissions prevent any unintended system file modifications.

### DIFF
--- a/pkg/artifact/downloader.go
+++ b/pkg/artifact/downloader.go
@@ -65,18 +65,12 @@ func (l *downloader) Download(location *url.URL, verificationData auth.Verificat
 		return err
 	}
 
-	// rewind a second time to allow the archive to be unpacked
-	_, err = artifactFile.Seek(0, os.SEEK_SET)
-	if err != nil {
-		return util.Errorf("Could not reset artifact file position after verification: %v", err)
-	}
-
 	err = artifactFile.Chmod(0644)
 	if err != nil {
 		return err
 	}
 
-	err = gzip.ExtractTarGz(owner, artifactFile, dst)
+	err = gzip.ExtractTarGz(owner, artifactFile.Name(), dst)
 	if err != nil {
 		_ = os.RemoveAll(dst)
 		return util.Errorf("error while extracting artifact: %s", err)
--- a/pkg/artifact/downloader.go
+++ b/pkg/artifact/downloader.go
@@ -71,6 +71,11 @@ func (l *downloader) Download(location *url.URL, verificationData auth.Verificat
 		return util.Errorf("Could not reset artifact file position after verification: %v", err)
 	}
 
+	err = artifactFile.Chmod(0644)
+	if err != nil {
+		return err
+	}
+
 	err = gzip.ExtractTarGz(owner, artifactFile, dst)
 	if err != nil {
 		_ = os.RemoveAll(dst)
--- a/pkg/gzip/gzip.go
+++ b/pkg/gzip/gzip.go
@@ -1,37 +1,26 @@
 package gzip
 
 import (
-	"archive/tar"
-	"compress/gzip"
-	"io"
 	"os"
-	"path/filepath"
-	"strings"
+	"os/exec"
+	"os/user"
+	"syscall"
 
-	"github.com/square/p2/pkg/user"
+	p2user "github.com/square/p2/pkg/user"
 	"github.com/square/p2/pkg/util"
 )
 
-// ExtractTarGz reads a gzipped tar stream and extracts all files to the destination
-// directory. If an owner name is specified, all files will be created to be owned by that
-// user; otherwise, the tar specifies ownership.
-//
-// If any file would be extracted outside of the destination directory due to relative paths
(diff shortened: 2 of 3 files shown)


pkg/gzip/gzip_test.go
```diff
@@ -47,6 +45,10 @@ func TestFileWithoutDir(t *testing.T) {
 }
 
 func TestSelfHardlink(t *testing.T) {
+	if runtime.GOOS == "darwin" {
+		t.Skip("bsdtar doesn't handle a selflink tarball")
```
petertseng: a little suspicious at this.

The self-hardlink tarballs are only creatable by GNU tars, and only extractable by GNU tars. We have a self-hardlink tarball checked in to `testdata`, and obviously this test will fail if `tar` is a BSD tar on the system running the test.

This test is effectively asking "is a file that is only creatable by GNU tars extractable via the `tar` on the system the test happens to be run on?"

Testing that the tarballs are, in fact, extractable by GNU tar... doesn't really seem to be our job. Maybe we can just delete this test.

pkg/gzip/gzip.go
```diff
@@ -43,132 +32,11 @@ func ExtractTarGz(owner string, fp io.Reader, dest string) (err error) {
...
-				}
-				return nil
-			}()
-			if err != nil {
-				return err
-			}
-		default:
-			return util.Errorf("unhandled type flag %q (header %v)", hdr.Typeflag, hdr)
+	cmd := exec.Command("tar", "xzf", filename, "-C", dest)
+	if currentUser.Username != owner {
```
rudle: Is it harmful to remove this if? It would slightly reduce the complexity of this method!

rudle: Wasn't it the case that different versions of GNU's tar implementation had different behaviour wrt self-hardlink? In that case, this test captures a regression in an integration style. I support that.

pkg/gzip/gzip_test.go
rudle: Wasn't it the case that different versions of GNU's tar implementation had different behaviour wrt self-hardlink? In that case, this test captures a regression in an integration style. I support that.

pkg/gzip/gzip.go
petertseng: I'm sad to say that removing it... causes tests to fail, since we don't run the tests as root, and non-root users understandably cannot specify a user to switch to. Not even themselves.

rudle: aha

mpuncel: might be a useful comment

petertseng: comment is in, though I still need to track down integration test failures. I wonder why it would.

petertseng:
With a chmod, the test now passes locally for me. Take a look to see if the chmod is reasonable.

pkg/artifact/downloader.go
```diff
@@ -65,13 +65,12 @@ func (l *downloader) Download(location *url.URL, verificationData auth.Verificat
 		return err
 	}
 
-	// rewind a second time to allow the archive to be unpacked
-	_, err = artifactFile.Seek(0, os.SEEK_SET)
+	err = artifactFile.Chmod(0644)
```
petertseng: one could also consider a chmod 0400 followed by a chown to the app user. or chmod 0640 and chgrp to the app user.

pkg/gzip/gzip.go
```diff
@@ -43,132 +32,18 @@ func ExtractTarGz(owner string, fp io.Reader, dest string) (err error) {
...
+	cmd := exec.Command("tar", "xpzf", filename, "-C", dest)
+	if currentUser.Username != owner {
+		// If we are running as a non-root user (e.g. in tests), don't change user.
+		// Non-root users are understandably not allowed to change to other users...
+		// not even themselves.
+		cmd.SysProcAttr = &syscall.SysProcAttr{
+			Credential: &syscall.Credential{Uid: uint32(ownerUID), Gid: uint32(ownerGID)},
 		}
 	}
+	output, err := cmd.CombinedOutput()
```
mpuncel: love all the red

pkg/gzip/gzip.go
```diff
@@ -43,132 +32,18 @@ func ExtractTarGz(owner string, fp io.Reader, dest string) (err error) {
...
-				return nil
-			}()
-			if err != nil {
-				return err
-			}
-		default:
-			return util.Errorf("unhandled type flag %q (header %v)", hdr.Typeflag, hdr)
+	cmd := exec.Command("tar", "xpzf", filename, "-C", dest)
+	if currentUser.Username != owner {
+		// If we are running as a non-root user (e.g. in tests), don't change user.
```
petertseng: I don't think the conditional here matches the comment??

PR #741, opened 2017-01-30
Make size.ByteCount type serialize to yaml in human-readable form.

mpuncel:
In particular this makes p2-norm more user friendly. Prior to this
commit, passing a launchable with "memory: 1.0G" through p2-norm would
result in the memory value becoming an integer byte count instead of the
human readable notation.

### DIFF
--- a/pkg/util/size/bytes.go
+++ b/pkg/util/size/bytes.go
@@ -69,11 +69,10 @@ func Parse(sizeStr string) (ByteCount, error) {
 	return ByteCount(size), nil
 }
 
-// MarshalYAML() serializes a ByteCount into YAML. To maintain a canonical representation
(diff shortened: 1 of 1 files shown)


mpuncel:
note: this will cause diff churn on already-normalized manifests, but I think it's a good move in the long run

bwester:
This has been attempted before: #610. The one comment left there still applies: "You may want to double check that the code used to emit `PLATFORM_CONFIG_PATH` still uses a raw byte count, as clients read that value and do not have knowledge of the human-friendly shorthand."

Or more generally, note that `ByteCount` is serialized to more places that just a manifest. Either (1) ensure that the human-readable format is also transitively readable or (2) only human-encode the values when written to a manifest.

PR #743, opened 2017-01-31
Mpuncel/remove consul dep

mpuncel:


mpuncel:
This PR first removes our dependency on github.com/consul/consul and then performs the upgrade. this was much easier as there are many fewer dependencies

mpuncel:
this may have made the build more flakey 😞

mpuncel:
declining this because our tests are heavily reliant on the consul dependency and it's not easy to replace

PR #744, opened 2017-02-02
Introduce p2-fetch-manifest

rudle:
A tool to conveniently write the contents of /intent for the given pod on the
given host out to the file system. This will promote self-serve preparer (and
other pod) repair.

### DIFF
--- a/bin/p2-fetch-manifest/main.go
+++ b/bin/p2-fetch-manifest/main.go
@@ -1,0 +1,97 @@
+package main
+
+import (
+	"errors"
+	"fmt"
+	"os"
+	"time"
+
+	"github.com/Sirupsen/logrus"
+	"github.com/square/p2/pkg/logging"
+	"github.com/square/p2/pkg/manifest"
+	"github.com/square/p2/pkg/store/consul"
+	"github.com/square/p2/pkg/store/consul/consulutil"
+	"github.com/square/p2/pkg/store/consul/flags"
+	"github.com/square/p2/pkg/version"
+	kingpin "gopkg.in/alecthomas/kingpin.v2"
+)
+
+const (
+	cmdFetchText = "fetch"
+)
+
+var (
+	cmdFetch     = kingpin.Command(cmdFetchText, "Fetch the manifest for a given pod from the datastore and write it to a file on the disk. Combines well with p2-launch.")
+	fetchTree    = cmdFetch.Flag("tree", "Intent or Reality").Default("intent").Short('t').Enum("intent", "reality")
+	fetchPod     = cmdFetch.Flag("podID", "The pod ID for which to fetch its manifest").Short('i').Required().String()
+	fetchEntropy = cmdFetch.Flag("force-restart", "Introduce entropy into the manifest before outputting it. Scheduling the resulting manifest will definitely cause a restart of the pod.").Short('p').Bool()
+)
+
+func main() {
+	kingpin.Version(version.VERSION)
+	logger := logging.NewLogger(logrus.Fields{})
+	cmd, consulOpts, labeler := flags.ParseWithConsulOptions()
+	_ = labeler
+	client := consul.NewConsulClient(consulOpts)
+
+	switch cmd {
+	case cmdFetchText:
+		hostname, err := os.Hostname()
+		if err != nil {
+			logger.Fatal(err)
+		}
+		man, err := fetch(client, *fetchTree, hostname, *fetchPod, *fetchEntropy)
+		if err != nil {
+			logger.Fatal(err)
+		}
+		filename := fmt.Sprintf("%s.yaml.signed", *fetchPod)
+		file, err := os.OpenFile(filename, os.O_WRONLY|os.O_CREATE, 0600)
+		if err != nil {
+			logger.Fatalf("Could not open file at %s. %v", filename, err)
+		}
+
+		err = man.Write(file)
+		if err != nil {
+			logger.Fatalf("Unable to write manifest: %v", err)
+		}
+		fmt.Printf("%s", file.Name())
+	default:
+		logger.Fatalln("Did not receive a command. Try --help.")
+	}
+}
+
+func fetch(client consulutil.ConsulClient, tree, hostname, pod string, entropy bool) (manifest.Manifest, error) {
+	key := fmt.Sprintf("%s/%s/%s", tree, hostname, pod)
+	resp, _, err := client.KV().Get(key, nil)
+	if err != nil {
+		return nil, err
+	}
+	if resp == nil {
+		return nil, errors.New("Got nil response from the datastore")
+	}
+
+	man, err := manifest.FromBytes(resp.Value)
+	if err != nil {
+		return nil, err
+	}
+
+	if entropy {
+		man, err = introduceEntropy(man)
+		if err != nil {
+			return nil, err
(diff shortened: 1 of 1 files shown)


bin/p2-fetch-manifest/main.go
```diff
@@ -0,0 +1,97 @@
...
+	cmdFetch     = kingpin.Command(cmdFetchText, "Fetch the manifest for a given pod from the datastore and write it to a file on the disk. Combines well with p2-launch.")
+	fetchTree    = cmdFetch.Flag("tree", "Intent or Reality").Default("intent").Short('t').Enum("intent", "reality")
+	fetchPod     = cmdFetch.Flag("podID", "The pod ID for which to fetch its manifest").Short('i').Required().String()
+	fetchEntropy = cmdFetch.Flag("force-restart", "Introduce entropy into the manifest before outputting it. Scheduling the resulting manifest will definitely cause a restart of the pod.").Short('p').Bool()
+)
+
+func main() {
+	kingpin.Version(version.VERSION)
+	logger := logging.NewLogger(logrus.Fields{})
+	cmd, consulOpts, labeler := flags.ParseWithConsulOptions()
```
petertseng: labeler never used, except for `_ = labeler` below. Is naming it (rather than `_`) intentional?

bin/p2-fetch-manifest/main.go
```diff
@@ -0,0 +1,97 @@
...
+)
+
+const (
+	cmdFetchText = "fetch"
+)
+
+var (
+	cmdFetch     = kingpin.Command(cmdFetchText, "Fetch the manifest for a given pod from the datastore and write it to a file on the disk. Combines well with p2-launch.")
+	fetchTree    = cmdFetch.Flag("tree", "Intent or Reality").Default("intent").Short('t').Enum("intent", "reality")
+	fetchPod     = cmdFetch.Flag("podID", "The pod ID for which to fetch its manifest").Short('i').Required().String()
```
petertseng: I compared this to other in this directory buy using `git grep "Flag.*pod"`. There are many instances of flag `pod`, and I might like to extoll the virtues of consistency.

On the subject of consistency, if this is to remain, perhaps for consistency with `force-restart` it should be `pod-id`...

bin/p2-fetch-manifest/main.go
```diff
@@ -0,0 +1,97 @@
...
+		}
+	}
+
+	return man, nil
+}
+
+func introduceEntropy(man manifest.Manifest) (manifest.Manifest, error) {
+	mb := man.GetBuilder()
+	config := man.GetConfig()
+	config["_restart"] = fmt.Sprintf("restarted at: %s", time.Now().Format(time.UnixDate))
```
petertseng: again on consistency with other tools that may add `_restart` - any chance of showing the username of the person who has done so?

Then again, that might not be so useful information.

bin/p2-fetch-manifest/main.go
```diff
@@ -0,0 +1,97 @@
...
+}
+
+func fetch(client consulutil.ConsulClient, tree, hostname, pod string, entropy bool) (manifest.Manifest, error) {
+	key := fmt.Sprintf("%s/%s/%s", tree, hostname, pod)
+	resp, _, err := client.KV().Get(key, nil)
+	if err != nil {
+		return nil, err
+	}
+	if resp == nil {
+		return nil, errors.New("Got nil response from the datastore")
```
petertseng: a person who doesn't already know might not be able to deduce that this means there's no manifest for the given pod at the given host.

bin/p2-fetch-manifest/main.go
```diff
@@ -0,0 +1,97 @@
...
+		if err != nil {
+			logger.Fatal(err)
+		}
+		filename := fmt.Sprintf("%s.yaml.signed", *fetchPod)
+		file, err := os.OpenFile(filename, os.O_WRONLY|os.O_CREATE, 0600)
+		if err != nil {
+			logger.Fatalf("Could not open file at %s. %v", filename, err)
+		}
+
+		err = man.Write(file)
```
petertseng: the filename implies that we wish this to be signed. I believe this works fine if we do not modify the manifest, as the manifest comes out of the datastore signed.

What happens if we wish to introduce entropy? How do we sign the modified manifest?

bin/p2-fetch-manifest/main.go
```diff
@@ -0,0 +1,97 @@
...
+	"github.com/square/p2/pkg/version"
+	kingpin "gopkg.in/alecthomas/kingpin.v2"
+)
+
+const (
+	cmdFetchText = "fetch"
+)
+
+var (
+	cmdFetch     = kingpin.Command(cmdFetchText, "Fetch the manifest for a given pod from the datastore and write it to a file on the disk. Combines well with p2-launch.")
```
petertseng: can I say I feel weird asking for a command when there's only one possible choice? Are we leaving room for future expansion, if there will be future commands? If not, what of all the other commands in this directory that ignore the command? p2-schedule, p2-watch, p2-wipe-reality ...

bin/p2-fetch-manifest/main.go
bwester: Is this the first time (outside our internal extensions to P2) that we've promoted the view that you have to change the manifest and redeploy an app to simply "restart" a service?

While this method of triggering a restart works fine in Square's infrastructure, it's because we have full control over the services we deploy, and we can guarantee that a stray `_restart` key in this configuration is harmless. I don't think we can make any such assumptions in an open-source context. P2 itself should not ascribe any semantic meaning to the `config` data, nor should it modify it in any way. It's for the services.

As long as the manifest must change to trigger a redeploy/restart, a natural way around this new restraint is to introduce a non-config section to the manifest whose sole purpose is to allow for restarts. A simple integer counter would suffice. A field with a counter could allow other metadata to be stored, like the user or time at which the restart was requested. (Though really, I'm not convinced it makes sense to encode metadata about a deployment into a reusable description of what to deploy.)

bin/p2-fetch-manifest/main.go
bwester: 👍

bin/p2-fetch-manifest/main.go
```diff
@@ -0,0 +1,97 @@
...
+	fetchTree    = cmdFetch.Flag("tree", "Intent or Reality").Default("intent").Short('t').Enum("intent", "reality")
+	fetchPod     = cmdFetch.Flag("podID", "The pod ID for which to fetch its manifest").Short('i').Required().String()
+	fetchEntropy = cmdFetch.Flag("force-restart", "Introduce entropy into the manifest before outputting it. Scheduling the resulting manifest will definitely cause a restart of the pod.").Short('p').Bool()
```
bwester: From a purely functional view, changing a manifest to get it to restart is an orthogonal and unrelated task to fetching a single manifest. They are independently useful on their own. It doesn't make sense to me to couple them together into the same binary, much less exposing one as a mere flag.

Could you make "force restart" into its own binary? If that strikes you as too heavyweight, is there another existing P2 command that locally modifies a manifest? Putting the "force restart" modification into that command would at least group similar kinds of functionality.

nit: Your use of "entropy" here does not fit its common definition in computer science. I also don't think it's relevant _how_ a restart is forced, merely that it is.

PR #745, opened 2017-02-06
Add --json flag to the "list" subcommand on p2-dsctl

mpuncel:
This makes it easier to take a high level view of all daemon sets, for
example "show all of the node selectors currently in use".


rudle:
heck yes

PR #748, opened 2017-02-08
ds: failsafe: If no nodes are eligible, do nothing (and supporting commits - test changes)

petertseng:
It is expected that all daemon sets have non-empty target node sets at
all times.

If our label store ever indicates otherwise, we assume the label store
has erroneously returned empty results, and do nothing, to be safe.

This is similar to the p2-preparer's failsafe where it refuses to act if
it does not see its own key in intent.

The proper way to unschedule a daemon set on all its current nodes is to
delete it.

### DIFF
--- a/pkg/ds/farm_test.go
+++ b/pkg/ds/farm_test.go
@@ -488,6 +488,23 @@ func TestFarmSchedule(t *testing.T) {
 	Assert(t).IsNil(err, "Expected pod to have a dsID label")
 	dsID = labeled.Labels.Get(DSIDLabel)
 	Assert(t).AreEqual(anotherDSData.ID.String(), dsID, "Unexpected dsID labeled")
+
+	// If no nodes are eligible, DS should take no action.
+	mutator = func(dsToUpdate ds_fields.DaemonSet) (ds_fields.DaemonSet, error) {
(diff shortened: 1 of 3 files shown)


pkg/ds/farm_test.go
```diff
@@ -457,6 +462,10 @@ func TestFarmSchedule(t *testing.T) {
 	dsID = labeled.Labels.Get(DSIDLabel)
 	Assert(t).AreEqual(anotherDSData.ID.String(), dsID, "Unexpected dsID labeled")
 
+	// Verify node3 is scheduled
```
petertseng: unscheduled, please

rudle:
:ship:

petertseng:
I accuse the last commit of this PR ("ds: Verify delete in farm test") of making the tests flaky.

It's not the fault of that commit per se - since it only changed tests, that means it's revealing a race in our existing code, but it is still troubling.

I suspect a race between the closeChild of a session expiry and the close-on-delete of handleDSChanges. There is a nil deref against the line `err := <-child.errCh`, and the only way this happens on this line is if `child` is nil (receiving on a nil channel is safe). Need to investigate why `child` would be nil at that point in the code.

PR #750, opened 2017-02-08
ds: failsafe: no-op if no daemon sets (and supporting commits)

petertseng:


pkg/ds/farm.go
```diff
@@ -262,8 +262,13 @@ func (dsf *Farm) handleDSChanges(changes dsstore.WatchedDaemonSets, quitCh <-cha
 		return
 	}
 
+	if changes.Total == 0 {
+		dsf.logger.Errorf("No daemon sets present, asked to delete %d; farm refuses to delete its %d daemon sets", len(changes.Deleted), len(dsf.children))
```
petertseng: actually, this troubles me. This causes the farm's state to become out of sync with the watch's state. By this I mean: The watch will assume that the daemon set farm has processed the deletion of all its daemon sets, but the daemon set farm has, in fact, not.

This means that if the same watch perceives an update to an existing DS, the watch will say it's a new DS (assuming the farm has deleted it). this may have undesirable effects. But does that mean we should terminate the watch instead? Or somehow inform the watch that the changes were rejected, so it keeps its old state?

pkg/store/consul/consulutil/watch.go
```diff
@@ -327,7 +328,7 @@ func WatchDiff(
 				mapCopy[key] = val
 			}
 
-			outgoingChanges := &WatchedChanges{}
+			outgoingChanges := &WatchedChanges{Total: len(pairs)}
```
petertseng: Maybe the thing to do is, instead, to make `WatchDiff` not emit an output (and not update its state) if the total is 0. This will affect all users of `WatchDiff` though. Which coincidentally is only `dsstore` today.

pkg/ds/farm.go
mpuncel: hmmmm that's a good observation.

Maybe we should treat it as a soft-panic and abort the whole farm and start over?

petertseng: Would you prefer (as I am suspecting) simply having the watch not report such a situation, as #751 does?

Aborting the entire farm could be a choice. Can I be sure I'm not going to hammer Consul by repeatedly asking for more watches if this happens repeatedly for a period of time? Maybe have to back off for a bit.

pkg/store/consul/consulutil/watch.go
mpuncel: saw your other PR that does this.

If we go that route we need to be sure that the caller never expects that it might be a 404. For example listening rolling updates I expect that there will be 0 during steady state. This method means EVERY caller has to protect against an empty result set but they can at least customize their behavior. will need to think about it but i think for now i prefer this approach

petertseng: Well... it's a bit weird because the other PR only does this for `WatchDiff`, which means that the only affected farm is the DS farm. the RU farm doesn't use WatchDiff  - it uses a plain Watch (and is probably internally doing its own diff...)

mpuncel:
are we keeping this PR now that we removed daemon set deletion behavior?

petertseng:
Let's revisit when we want to re-enable, and choose between this and #751

PR #751, opened 2017-02-08
consulutil: Don't propagate a 0-length result

petertseng:
We assume that there will never be a case where there are no results, so
we shield our clients from potentially deleting all of their stored
results.

Test included, but is bogus.

~~Also, tests are failing for maybe-related reasons probably because there are other tests in which we delete the only daemon set. That is unfortunate.~~ Dealt with that.

### DIFF
--- a/pkg/ds/farm_test.go
+++ b/pkg/ds/farm_test.go
@@ -911,6 +911,92 @@ func TestMultipleFarms(t *testing.T) {
 	Assert(t).IsNil(err, "Expected pod not to have a dsID label")
 }
 
+// Honestly, this test is a bit bogus.
+// dsstoretest.NewFake doesn't use consulutil's WatchDiff,
+// which is what would protect us in production.
+// All this tests is that the fake DS store's WatchDiff behavior is the same as that of consulutil's WatchDiff.
+func TestFailsafe(t *testing.T) {
+	retryInterval = testFarmRetryInterval
+
+	//
+	// Instantiate farm
+	//
+	dsStore := dsstoretest.NewFake()
+	consulStore := consultest.NewFakePodStore(make(map[consultest.FakePodStoreKey]manifest.Manifest), make(map[string]consul.WatchResult))
+	applicator := labels.NewFakeApplicator()
+	logger := logging.DefaultLogger.SubLogger(logrus.Fields{
+		"farm": "farmFailsafe",
+	})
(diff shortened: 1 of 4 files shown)


pkg/ds/farm_test.go
```diff
@@ -911,6 +911,92 @@ func TestMultipleFarms(t *testing.T) {
 	Assert(t).IsNil(err, "Expected pod not to have a dsID label")
 }
 
+// Honestly, this test is a bit bogus.
+// dsstoretest.NewFake doesn't use consulutil's WatchDiff,
```
petertseng: my confidence in this test would be much increased if dsstoretest's `WatchDiff` and consulutil's `WatchDiff` would both use a common function, under which condition I would assume that they behave the same.

petertseng:
Given the concern I raised in https://github.com/square/p2/pull/750#discussion_r100179331 , this is my alternative.

pkg/ds/farm.go
```diff
@@ -263,7 +263,7 @@ func (dsf *Farm) handleDSChanges(changes dsstore.WatchedDaemonSets, quitCh <-cha
 	}
 
 	if len(changes.Created) > 0 {
-		dsf.logger.Infof("The following daemon sets have been created:")
+		dsf.logger.Infof("The following %d daemon sets have been created:", len(changes.Created))
```
mpuncel: it might be nice to just list their IDs in this message instead of it getting interspersed with everything else. You could just `%s` a slice of ids

petertseng: I should move that commit to a separate PR, since it's gotten into both #750 and this one now.

pkg/store/consul/consulutil/watch.go
```diff
@@ -317,6 +317,12 @@ func WatchDiff(
 				}
 				timer.Reset(2 * time.Second) // backoff
 				continue
+			} else if len(pairs) == 0 {
+				// Let's be safe and assume that we will never have an empty result.
+				// We will not be sending this to the watcher.
+				// Assume it means something happened with Consul, and try again.
+				timer.Reset(2 * time.Second) // backoff
+				continue
 			}
```
mpuncel: while you're in here, maybe we should do a sanity check that `queryMeta.LastIndex` isn't less than `currentIndex`. Can save for a later PR if you desire

petertseng:
Let's revisit when we want to re-enable, and choose between this and #750

PR #752, opened 2017-02-08
ds: Show created, updated, deleted daemon sets more compactly

petertseng:
Instead of showing the full details of all daemon sets, one per log
line, we show the count, then all IDs, in one log line.

### DIFF
--- a/pkg/ds/farm.go
+++ b/pkg/ds/farm.go
@@ -263,10 +263,8 @@ func (dsf *Farm) handleDSChanges(changes dsstore.WatchedDaemonSets, quitCh <-cha
 	}
 
 	if len(changes.Created) > 0 {
-		dsf.logger.Infof("The following daemon sets have been created:")
+		dsf.logger.Infof("The following %d daemon sets have been created: %s", len(changes.Created), dsIDs(changes.Created))
 		for _, dsFields := range changes.Created {
-			dsf.logger.Infof("%v", *dsFields)
-
 			var dsUnlocker consulutil.Unlocker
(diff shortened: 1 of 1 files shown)


pkg/ds/farm.go
```diff
@@ -263,7 +263,7 @@ func (dsf *Farm) handleDSChanges(changes dsstore.WatchedDaemonSets, quitCh <-cha
 	}
 
 	if len(changes.Created) > 0 {
-		dsf.logger.Infof("The following daemon sets have been created:")
+		dsf.logger.Infof("The following %d daemon sets have been created:", len(changes.Created))
```
petertseng: Hmm, interesting idea to just give a list of the daemon set IDs in this message, just a second.

pkg/ds/farm.go
```diff
@@ -263,10 +263,8 @@ func (dsf *Farm) handleDSChanges(changes dsstore.WatchedDaemonSets, quitCh <-cha
 	}
 
 	if len(changes.Created) > 0 {
-		dsf.logger.Infof("The following daemon sets have been created:")
+		dsf.logger.Infof("The following %d daemon sets have been created: %s", len(changes.Created), dsIDs(changes.Created))
 		for _, dsFields := range changes.Created {
-			dsf.logger.Infof("%v", *dsFields)
```
petertseng: well, note that we're now showing just the IDs, not all the fields. If you think you'd rather like to see all the fields instead, let's talk.

Or maybe I can just keep this line. Prepend "created"?

PR #753, opened 2017-02-09
ds: spawn a daemon set before deleting it

petertseng:
In the case where a daemon set is deleted, but nobody has a lock on it,
some daemon set farm must grab a lock on it and spawn it so that it can
react to its deletion signal.

Previously, any daemon set farm attempting to do so would not spawn the
daemon set, which has two bad effects:

1. No deletion actually occurs for the daemon set.
2. The farm dereferences nil when checking whether the nonexistent child
   has errored or reacted to its delete signal.

The second effect has the effect of making our tests sometimes fail.


rudle:
LGTM with green build

PR #754, opened 2017-02-09
Disable daemon set deletion behaviour

rudle:
The implementation that unscheduled all pods was hard to build a failsafe
around, so we prefer to remove it altogether until we can get a proper failsafe
built.


pkg/ds/daemon_set_test.go
```diff
@@ -306,13 +307,14 @@ func TestSchedule(t *testing.T) {
...
+	// Behaivour change: Deleting a daemon set will no longer unschedule its pods (for now)
+	numNodes = waitForNodes(t, ds, expectedNodes, desiresErrCh, dsChangesErrCh)
+	Assert(t).AreEqual(numNodes, expectedNodes, "Unexpected number of nodes scheduled")
 
 	scheduled = scheduledPods(t, ds)
-	Assert(t).AreEqual(len(scheduled), 0, "expected all nodes to have been unlabeled")
+	Assert(t).AreEqual(len(scheduled), expectedNodes, "Expected no nodes to be unscheduled")
 
-	err = waitForPodsInIntent(consulStore, 0)
+	err = waitForPodsInIntent(consulStore, 9)
```
petertseng: tell me about this 9... it's certainly not 11, which makes me wonder about it

pkg/ds/farm_test.go
```diff
@@ -512,7 +512,8 @@ func TestFarmSchedule(t *testing.T) {
 	Assert(t).IsNil(err, "Expected daemon set to be deleted in farm")
 
 	// Verify node3 is unscheduled
-	labeled, err = waitForPodLabel(applicator, false, "node3/testPod")
+	// Behaviour change! Daemon Set deletions do not delete their pods (for now)
+	labeled, err = waitForPodLabel(applicator, true, "node3/testPod")
 	Assert(t).IsNil(err, "Expected pod not to have a dsID label")
```
petertseng: change expect text

pkg/ds/daemon_set_test.go
rudle: I think it should be 23

rudle:
tests pass!

pkg/ds/daemon_set_test.go
```diff
@@ -306,13 +307,14 @@ func TestSchedule(t *testing.T) {
 	if err != nil {
 		t.Fatalf("Unable to delete daemon set: %v", err)
 	}
-	numNodes = waitForNodes(t, ds, 0, desiresErrCh, dsChangesErrCh)
-	Assert(t).AreEqual(numNodes, 0, "took too long to unschedule")
+	// Behaivour change: Deleting a daemon set will no longer unschedule its pods (for now)
```
petertseng: you spelled it wrong, and I'm not talking about the ou, I'm talking about the iv

pkg/ds/daemon_set_test.go
```diff
@@ -302,20 +302,22 @@ func TestSchedule(t *testing.T) {
 	//
 	// Deleting the daemon set should unschedule all of its nodes
 	//
+	// beforeDeletePods := labeledPods(t, ds)
```
petertseng: why leave this commented line in?

edit: oh hmm it gets made into a real line in next commit. oooook

pkg/ds/daemon_set_test.go
```diff
@@ -302,20 +302,22 @@ func TestSchedule(t *testing.T) {
...
+	Assert(t).IsNil(err, "Unexpected number of pods labeled")
 	ds.logger.NoFields().Info("Deleting daemon set...")
 	err = dsStore.Delete(ds.ID())
 	if err != nil {
 		t.Fatalf("Unable to delete daemon set: %v", err)
 	}
 	// Behaivour change: Deleting a daemon set will no longer unschedule its pods (for now)
-	numNodes = waitForNodes(t, ds, expectedNodes, desiresErrCh, dsChangesErrCh)
-	Assert(t).AreEqual(numNodes, expectedNodes, "Unexpected number of nodes scheduled")
+	numNodes = waitForNodes(t, ds, 12, desiresErrCh, dsChangesErrCh)
```
petertseng: expected->12 in previous commit might be good, cuz this ain't schedule -> label

pkg/ds/farm_test.go
```diff
@@ -506,7 +506,7 @@ func TestFarmSchedule(t *testing.T) {
 	dsID = labeled.Labels.Get(DSIDLabel)
 	Assert(t).AreEqual(anotherDSData.ID.String(), dsID, "Unexpected dsID labeled")
 
-	dsStore.Delete(anotherDSData.ID)
+	err = dsStore.Delete(anotherDSData.ID)
```
petertseng: that was a line I added? 

yeah it was.

I am sorry. The errors are completely bogus because I didn't assign too...

pkg/ds/daemon_set_test.go
rudle: yeah, this number is nonsense - this test will fail. 🤷‍♂️

rudle:
Merging on green

pkg/ds/daemon_set_test.go
```diff
@@ -295,25 +288,29 @@ func TestSchedule(t *testing.T) {
...
+	Assert(t).AreEqual(numNodes, expectedNodes, "Unexpected number of nodes labeled")
 
-	scheduled = scheduledPods(t, ds)
-	Assert(t).AreEqual(len(scheduled), 0, "expected all nodes to have been unlabeled")
+	labeled = labeledPods(t, ds)
+	Assert(t).AreEqual(len(labeled), expectedNodes, "Expected no nodes to be unlabeled")
 
-	err = waitForPodsInIntent(consulStore, 0)
-	Assert(t).IsNil(err, "Unexpected number of pods scheduled")
+	err = waitForPodsInIntent(consulStore, len(beforeDeletePods))
```
petertseng: maybe this is what's causing test flakes - we'd like it to be 12 (Right?), but sometimes tests expect it to be 1, presumably because the `scheduledPods` call doesn't get all of them.

mpuncel: that's probably it yep

pkg/ds/daemon_set_test.go
```diff
@@ -295,25 +288,29 @@ func TestSchedule(t *testing.T) {
...
+	Assert(t).IsNil(err, "Unable to list intent/")
 	ds.logger.NoFields().Info("Deleting daemon set...")
 	err = dsStore.Delete(ds.ID())
 	if err != nil {
 		t.Fatalf("Unable to delete daemon set: %v", err)
 	}
-	numNodes = waitForNodes(t, ds, 0, desiresErrCh, dsChangesErrCh)
-	Assert(t).AreEqual(numNodes, 0, "took too long to unschedule")
+	// behavior change: Deleting a daemon set will no longer unschedule its pods (for now)
+	numNodes = waitForNodes(t, ds, 12, desiresErrCh, dsChangesErrCh)
```
petertseng: hey hold on a second. This line waits for the number of nodes to decrease to 12 (spoiler, it never will)

pkg/ds/daemon_set_test.go
```diff
@@ -295,25 +288,29 @@ func TestSchedule(t *testing.T) {
...
+	Assert(t).AreEqual(numNodes, expectedNodes, "Unexpected number of nodes labeled")
```
petertseng: and then this line checks that the number of nodes is 23, not 12

pkg/ds/daemon_set_test.go
petertseng: This doesn't cause any test failures of course, but it adds 10 seconds to each test.

Then again, you may say that's a feature: We make sure that no pods are deleted for ten seconds.

PR #755, opened 2017-02-09
Add fail safe in pcstore.WatchAndSync to protect missing pod clusters.

mpuncel:
To protect against data loss or misfiring consul queries, this commit
adds a failsafe to the pcstore.WatchAndSync function in which it will
not call "SyncCluster" or "DeleteCluster" when the pod cluster watch
returns no data.

This is because a syncer may wish to perform destructive action at the
time of a pod cluster deletion, and if the entire set goes away it may
be disastrous.

### DIFF
--- a/pkg/store/consul/pcstore/consul_store_test.go
+++ b/pkg/store/consul/pcstore/consul_store_test.go
@@ -921,6 +921,8 @@ type RecordingSyncer struct {
 	// Used to signal when SyncCluster is called so tests can have non-racey
 	// timeouts
 	SyncSignal chan<- struct{}
+	// Used to signal when DeleteCluster is called
+	DeleteSignal chan<- struct{}
 }
 
 func (r *RecordingSyncer) GetInitialClusters() ([]fields.ID, error) {
@@ -937,6 +939,9 @@ func (r *RecordingSyncer) SyncCluster(pc *fields.PodCluster, labeledPods []label
 
 func (r *RecordingSyncer) DeleteCluster(id fields.ID) error {
 	r.DeleteClusterCalls = append(r.DeleteClusterCalls, id)
+	if r.DeleteSignal != nil {
+		r.DeleteSignal <- struct{}{}
+	}
 	return nil
 }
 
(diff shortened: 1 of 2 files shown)


pkg/store/consul/pcstore/consul_store_test.go
```diff
@@ -1015,6 +1020,126 @@ func TestWatchAndSync(t *testing.T) {
...
+
+	go func() {
+		err := store.WatchAndSync(syncer, quit)
+		if err != nil {
+			t.Fatalf("Couldn't start WatchAndSync(): %s", err)
+		}
+	}()
+
+	select {
+	case <-time.After(1 * time.Second):
```
mpuncel: writing a test that must take a second to pass makes me sad but short of passing an error on a channel within the failsafe `if` to the syncing routine i'm not sure how to avoid it

petertseng: Note that I also decided I could do no better in https://github.com/square/p2/pull/750/files#diff-465f690c6c15563e83acbe3a7e9a4e2cR986 or https://github.com/square/p2/pull/751/files#diff-465f690c6c15563e83acbe3a7e9a4e2cR990

rudle: Extracting the contents of the `for { select {` in `pcstore/consul_store.go` into a helper function would allow for unit testing its behaviour. The method looks a bit like channel soup to me, so idk if it's worth the trade.

pkg/store/consul/pcstore/consul_store_test.go
```diff
@@ -1015,6 +1020,126 @@ func TestWatchAndSync(t *testing.T) {
...
+	syncer := &RecordingSyncer{
+		InitialClusters: []fields.ID{pc1.ID, "some_other_ID"},
+		SyncSignal:      syncSignal,
+		DeleteSignal:    deleteSignal,
+	}
+
+	go func() {
+		err := store.WatchAndSync(syncer, quit)
+		if err != nil {
+			t.Fatalf("Couldn't start WatchAndSync(): %s", err)
```
petertseng: funny story - `Fatalf` calls `FailNow`, and https://golang.org/pkg/testing/#B.FailNow "FailNow must be called from the goroutine running the test or benchmark function, not from other goroutines created during the test."

Now, I'm not sure what happens if some other goroutine calls it. Perhaps we could try?

pkg/store/consul/pcstore/consul_store_test.go
mpuncel: I think we could do better, yeah. The behavior I don't want to happen (`DeleteCluster()` called) is in a different function though that is communicated with by a channel. I think it'd be fairly invasive to reorganize. Something we should look at in a later PR i think

pkg/store/consul/pcstore/consul_store_test.go
mpuncel: Interesting, i've observed that behavior in the past but never seen an explanation. Yeah i think it won't work in its current state.

mpuncel: In this case it happens that `err` can't be nil because only `GetInitialClusters()` returning an error can cause it, and our mock implementation unconditionally returns a `nil` error

PR #758, opened 2017-02-13
Add retry behavior to p2-finish-env-extractor.

mpuncel:
p2-finish-env-extractor is a binary that p2-preparer wires into runit to
capture the exit code of the processes that runit manages. In the case
of uuid pods p2-finish-env-extractor writes a row to a sqlite database
on disk to capture the exit code.

This scheme can result in lost information because writing to a sqlite
database always locks the entire database, so if multiple processes exit
at the same time they will compete for the ability to write to the
database.

This commit adds the ability to make p2-finish-env-extractor retry
errors writing to the database to improve the likelihood of a row making
it into the database.

### DIFF
--- a/bin/p2-finish-env-extractor/main.go
+++ b/bin/p2-finish-env-extractor/main.go
@@ -9,6 +9,7 @@ package main
 
 import (
 	"log"
+	"time"
 
 	"github.com/square/p2/pkg/logging"
 	"github.com/square/p2/pkg/preparer/podprocess"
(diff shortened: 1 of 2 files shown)


bin/p2-finish-env-extractor/main.go
```diff
@@ -30,7 +36,19 @@ func main() {
...
-	err := envExtractor.WriteFinish(*exitCode, *exitStatus)
+	var err error
+	// WriteFinish() is writing to a sqlite database that may be shared
+	// across many pods across a physical host. The nature of sqlite is
+	// that writes lock the entire database, which can cause transient
+	// failures. To reduce the likelihood of writes fighting each other, we
+	// retry writes a few times after they fail
+	for i := 0; i < *retries; i++ {
+		err = envExtractor.WriteFinish(*exitCode, *exitStatus)
+		if podprocess.IsRetryable(err) {
```
petertseng: 1. Trying to ensure that this acts corrctly when `err` nil:

https://play.golang.org/p/SzvU4OuQ0R

Check.

2. But if err is nil, shouldn't we stop retrying? As far as I can tell, this *always* retries the full number of times.

mpuncel: oh yeah whoops i removed my `break`

PR #759, opened 2017-02-14
introduce p2-shutdown

rudle:
There is a desire to gracefully shutdown pods while preserving their /intent
records. An example of when this is desirable is when performing host
maintenance on a pet host.

### DIFF
--- a/bin/p2-shutdown/main.go
+++ b/bin/p2-shutdown/main.go
@@ -1,0 +1,91 @@
+package main
+
+import (
+	"log"
+	"os"
+
+	kingpin "gopkg.in/alecthomas/kingpin.v2"
+
+	"sync"
+
+	"github.com/square/p2/pkg/manifest"
+	"github.com/square/p2/pkg/pods"
+	"github.com/square/p2/pkg/store/consul"
+	"github.com/square/p2/pkg/store/consul/flags"
+	"github.com/square/p2/pkg/types"
+)
+
+const helpMessage = `
+p2-shutdown is a command that is useful to gracefully shutdown pods on a host
+before doing maintenance. Ideally, these pods would be relocated to a different
+host but we live in a world where hosts are pets.
+`
+
+var (
+	verbose      = kingpin.Flag("verbose", "Print debugging information").Short('v').Bool()
+	dryRun       = kingpin.Flag("dry", "Dry run: do not stop any pods").Short('d').Bool()
+	shutdownPods = kingpin.Flag("pods", "The list of pods to shutdown. Leave empty for all").Short('p').Strings()
+	podRoot      = kingpin.Flag("pod-root", "The base directory for pods").Default(pods.DefaultPath).String()
+)
+
+func main() {
+	_, consulOpts, _ := flags.ParseWithConsulOptions()
+	client := consul.NewConsulClient(consulOpts)
+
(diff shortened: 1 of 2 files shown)


bin/p2-shutdown/main.go
```diff
@@ -0,0 +1,81 @@
...
+
+const helpMessage = `
+p2-shutdown is a command that is useful to gracefully shutdown pods on a host
+before doing maintenance. Ideally, these pods would be relocated to a different
+host but we live in a world where hosts are pets.
+`
+
+var (
+	verbose      = kingpin.Flag("verbose", "Print debugging information").Short('v').Bool()
+	dryRun       = kingpin.Flag("dry", "dry run").Short('d').Default("false").Bool()
```
mpuncel: i'm confused by the `Default()` part, does that mean you need `--dry true`? probably the Default isn't doing anything because `false` is the zero value of `bool`

rudle: `--dry` seems to work for me, I'll remove the unnecessary Default

bin/p2-shutdown/main.go
```diff
@@ -0,0 +1,81 @@
...
+	shutdownPods = kingpin.Flag("pods", "The list of pods to shutdown. Leave empty for all").Short('p').Strings()
+	podRoot      = kingpin.Flag("podRoot", "The base directory for pods").Required().String()
```
mpuncel: for convenience you could maybe default this to `pods.DefaultPath`. Also nit: `p2-run-hooks` calls it `--pod-root` instead of `--podRoot`

bin/p2-shutdown/main.go
```diff
@@ -0,0 +1,81 @@
...
+	reality, _, err := consulStore.ListPods(consul.REALITY_TREE, node)
+	if err != nil {
+		log.Fatalf("caught fatal error while querying datastore: %v", err)
+	}
+
+	podsToShutdown := make([]types.PodID, 0, len(*shutdownPods))
+	for _, pod := range *shutdownPods {
+		podsToShutdown = append(podsToShutdown, types.PodID(pod))
+	}
+
```
mpuncel: you should probably shut down `p2-preparer` here first, otherwise a concurrent update to /intent could start one of the pods you stop

rudle: I thought of that too - I think it's simpler to invoke this script twice with `--pods p2-preparer` followed by a invocation with no pod filtering. WDYT?

rudle:
PTAL I added some _parallelism_

bin/p2-shutdown/main.go
```diff
@@ -0,0 +1,85 @@
...
+		go func(man manifest.Manifest, podID types.PodID) {
+			success, err := pod.Halt(man)
+			if !success {
+				log.Printf("[ERROR]: at least one launchable of %s did not halt successfully.", podID)
+			}
+			if err != nil {
+				log.Printf("[ERROR]: Got error while halting pod %s. Consider retrying the command. \n %s", podID, err)
+			}
+		}(realityEntry.Manifest, pod.Id)
+	}
```
mpuncel: you'll need to wait for those goroutines here otherwise they may not execute

rudle: d'oh. This is fixed.

PR #760, opened 2017-02-16
Make unmarshaling an RC with missing replicas_desired an error.

mpuncel:
This adds to the safety of the RC farm, because it allows us to
distinguish a JSON RC missing the replicas_desired field completely from
one that has it but a 0 count for the value.

It is an error to unmarshal an RC that does not have a replicas_desired
field.

This protects us against a scenario where the json tag might be changed
such that all stored RCs will have replica counts of zero. There is
already a failsafe in place that prevents the farm from taking action of
the sum of all of the rc replica counts is zero, but this adds an
additional layer of safety.

### DIFF
--- a/pkg/rc/farm.go
+++ b/pkg/rc/farm.go
@@ -128,6 +128,7 @@ func (rcf *Farm) Start(quit <-chan struct{}) {
 func (rcf *Farm) mainLoop(quit <-chan struct{}) {
 	subQuit := make(chan struct{})
 	defer close(subQuit)
+
 	rcWatch, rcErr := rcf.rcStore.WatchNewWithRCLockInfo(subQuit, rcf.rcWatchPauseTime)
 
 START_LOOP:
@@ -277,22 +278,21 @@ func (rcf *Farm) failsafe(rcFields []rcstore.RCLockResult) {
 		}
 		panic("No RCs are scheduled at all. Create one RC to enable the farm. Panicking to escape a potentially bad situation.")
 	}
-	globalReplicaCount := 0
+
 	for _, rc := range rcFields {
 		if rc.ReplicasDesired > 0 {
+			// If at least one RC has a > 0 count, we're good
 			return
 		}
-		globalReplicaCount += rc.ReplicasDesired
 	}
-	if globalReplicaCount == 0 {
-		if err := rcf.alerter.Alert(alerting.AlertInfo{
-			Description: "All RCs have zero replicas requested",
-			IncidentKey: "zero_replicas_found",
-		}); err != nil {
-			rcf.logger.WithError(err).Errorln("Unable to deliver alert!")
-		}
-		panic("The sum of all replicas is 0. Panicking to escape a potentially bad situation")
+
+	if err := rcf.alerter.Alert(alerting.AlertInfo{
+		Description: "All RCs have zero replicas requested",
+		IncidentKey: "zero_replicas_found",
+	}); err != nil {
+		rcf.logger.WithError(err).Errorln("Unable to deliver alert!")
 	}
+	panic("The sum of all replicas is 0. Panicking to escape a potentially bad situation")
 }
 
 func (rcf *Farm) releaseDeletedChildren(foundChildren map[fields.ID]struct{}) {
--- a/pkg/rc/fields/fields.go
+++ b/pkg/rc/fields/fields.go
@@ -7,6 +7,7 @@ import (
 	"k8s.io/kubernetes/pkg/labels"
 
 	"github.com/square/p2/pkg/manifest"
+	"github.com/square/p2/pkg/util"
 )
 
 // ID is a named type alias for Resource Controller IDs. This is preferred to the raw
@@ -32,7 +33,8 @@ type RC struct {
 	// A set of labels that will be added to every pod scheduled by this controller
 	PodLabels labels.Set
 
-	// The desired number of instances of the manifest that should be scheduled
+	// The desired number of instances of the manifest that should be
+	// scheduled.
 	ReplicasDesired int
 
 	// When disabled, this controller will not make any scheduling changes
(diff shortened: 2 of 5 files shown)


mpuncel:
The motivation for this PR is that we may want to make the RC farm watch `?keys` in consul instead of getting the full set of values back. This is currently not possible because we have a failsafe that runs on every iteration of the RC watch which sums the `replicas_desired` field of every RC and panics if that sum is zero. We won't be able to do that sum if the goroutine only sees the keys not the values.

A compromise would be we can do this total at the startup of the farm which would protect us from a case where the JSON field was renamed, causing golang to unmarshal all `ReplicasDesired` fields with the zero-value.

This PR adds an additional layer of safety, making it an error to unmarshal an RC that does not have the `replicas_desired` field.

pkg/rc/farm.go
```diff
@@ -277,12 +278,17 @@ func (rcf *Farm) failsafe(rcFields []rcstore.RCLockResult) {
 		}
 		panic("No RCs are scheduled at all. Create one RC to enable the farm. Panicking to escape a potentially bad situation.")
 	}
+
 	globalReplicaCount := 0
 	for _, rc := range rcFields {
-		if rc.ReplicasDesired > 0 {
+		if rc.ReplicasDesired == nil {
+			panic(fmt.Sprintf("RC %s is missing replicas_desired count", rc.ID))
```
petertseng: It seems a bit extreme. Then again, it does match up with all other `panic`s in this function, and it's not like this function has any other mechanism (since it doesn't return anything).

I wondered about this because the other checks would only trigger as a result of the state of all RCs as a whole, whereas this check will trigger on just a single RC. Thus I could imagine a scenario where a single bad RC would stop all the other RCs from working. Would this be what we want?

pkg/rc/fields/fields.go
```diff
@@ -32,8 +33,10 @@ type RC struct {
 	// A set of labels that will be added to every pod scheduled by this controller
 	PodLabels labels.Set
 
-	// The desired number of instances of the manifest that should be scheduled
-	ReplicasDesired int
+	// The desired number of instances of the manifest that should be
+	// scheduled. A pointer is used here to safely distinguish between the
+	// count being 0 and not being present in the RC json
+	ReplicasDesired *int
```
petertseng: well, if we never expect a `nil` in a Raw RC, and we panic if we try to unmarshal such a one, doesn't it mean that in the cooked/parsed RC we should expect that this will never be `nil`, and thus can keep it a plain `int`?

pkg/rc/farm.go
```diff
@@ -277,12 +278,17 @@ func (rcf *Farm) failsafe(rcFields []rcstore.RCLockResult) {
...
+		}
+
+		if *rc.ReplicasDesired > 0 {
 			return
 		}
-		globalReplicaCount += rc.ReplicasDesired
+		globalReplicaCount += *rc.ReplicasDesired
 	}
 	if globalReplicaCount == 0 {
```
petertseng: I know you didn't write this, but it strikes me that at this point `globalReplicaCount` doesn't matter - the function would have returned if any was nonzero, so if we get to this point, `globalReplicaCount` is guaranteed to be zero and the counter and `if` might as well be removed.

pkg/rc/fields/fields.go
mpuncel: yeah i'll put it back

pkg/rc/farm.go
mpuncel: having that there probably makes the intention more understandable, you're right though

pkg/rc/farm.go
mpuncel: I was thinking that if one RC is missing `replicas_desired` then probably all of them are corrupted (e.g. a code change changed the json key). In that case we'd want the server to bail as soon as possible.

I think for now i'd like to keep this heavy handed approach and then if it poses a problem we can trust the unmarshaling errors

petertseng: 👍

pkg/rc/fields/fields.go
```diff
@@ -32,7 +33,9 @@ type RC struct {
 	// A set of labels that will be added to every pod scheduled by this controller
 	PodLabels labels.Set
 
-	// The desired number of instances of the manifest that should be scheduled
+	// The desired number of instances of the manifest that should be
+	// scheduled. A pointer is used here to safely distinguish between the
```
petertseng: a pointer is no longer used here though!

pkg/rc/fields/fields.go
```diff
@@ -89,7 +97,7 @@ func (rc RC) ToRaw() (RawRC, error) {
 		Manifest:        string(manifest),
 		NodeSelector:    nodeSel,
 		PodLabels:       rc.PodLabels,
-		ReplicasDesired: rc.ReplicasDesired,
+		ReplicasDesired: &rc.ReplicasDesired,
```
petertseng: scared me for a moment that if you change an `RC`'s `ReplicasDesired` that suddenly a `RawRC`'s `ReplicasDesired` can change, but luckily `rc` is passed by value, so it is all good.

pkg/rc/farm.go
```diff
@@ -277,8 +278,13 @@ func (rcf *Farm) failsafe(rcFields []rcstore.RCLockResult) {
 		}
 		panic("No RCs are scheduled at all. Create one RC to enable the farm. Panicking to escape a potentially bad situation.")
 	}
+
 	globalReplicaCount := 0
 	for _, rc := range rcFields {
+		if rc.ReplicasDesired == 0 {
```
petertseng: @eh? this looks suspicious. Replica counts can legitimately be zero, right?

Am I regretting asking for `ReplicasDesired` to no longer be a pointer?

I think we only want to panic if we encounter a RawRC whose pointer is nil, not if we see a zero in a cooked RC. I know that we don't have raw RCs when we get to this point, so perhaps that means a change is needed?

pkg/rc/fields/fields_test.go
```diff
@@ -1,37 +1,44 @@
...
 	Assert(t).AreEqual(rc1.Manifest.ID(), rc2.Manifest.ID(), "Manifest ID changed when serialized")
 }
 
 func TestZeroUnmarshal(t *testing.T) {
-	var rc fields.RC
+	var rc RC
 	err := json.Unmarshal([]byte(`{}`), &rc)
-	Assert(t).IsNil(err, "error unmarshaling")
+	if err == nil {
+		t.Errorf("expected an error unmarshaling an RC missing the replicas_desired field")
```
petertseng: I think I understand why you did this (and the other changes) - because `Assert(t).IsNil` uses `Fatal`, but you would rather just see `Error`, right?

mpuncel: That's one reason.

I also don't like those assertions in general because they take `interface{}` arguments, so when the types change they might not be comparable anymore and the compiler doesn't complain you just get random test failures.

PR #761, opened 2017-02-16
Mpuncel/rc key watch

mpuncel:


mpuncel:
This PR contains commits from #760 so please review that first

pkg/store/consul/consulutil/watch_test.go
```diff
@@ -207,6 +208,65 @@ func TestWatchPrefix(t *testing.T) {
...
+	// The watcher should ignore kv3a, which is outside its prefix
+	f.Client.KV().Put(kv3a, nil)
+	f.Client.KV().Delete(kv2a.Key, nil)
+	keys = <-keysChan
+	if len(keys.Keys) != 0 {
+		t.Error("watch did not ignore keys outside of prefix")
+	}
+	close(done)
+	done = nil
+	for _ = range keysChan {
```
petertseng: just `for range keysChan` will suffice... some tool will correct us, though maybe we don't run it by default. Maybe it's `go fmt -s` or something.

pkg/store/consul/consulutil/watch_test.go
```diff
@@ -207,6 +208,65 @@ func TestWatchPrefix(t *testing.T) {
...
+	kv1b := &api.KVPair{Key: "prefix/hello", Value: []byte("do")}
+	kv2a := &api.KVPair{Key: "prefix/test", Value: []byte("not")}
+	kv3a := &api.KVPair{Key: "something", Value: []byte("matter")}
+
+	// Process existing data
+	f.Client.KV().Put(kv1a, nil)
+	keysChan := WatchKeys("prefix/", f.Client.KV(), done, 0)
+	keys := <-keysChan
+	if !reflect.DeepEqual(keys.Keys, []string{kv1a.Key}) {
+		t.Error("existing data not recognized")
```
petertseng: optional: perhaps the keys that do exist can be shown, to aid debugging if something does fail in tests.

pkg/store/consul/consulutil/ops.go
```diff
@@ -96,6 +96,42 @@ func List(
...
+	keys      []string
+	queryMeta *api.QueryMeta
+	err       error
+}
+
+// SafeKeys performs a KV Keys operation that can be canceled. When the "done"
+// channel is closed, CanceledError will be immediately returned. (The HTTP RPC
+// can't be canceled, but it will be ignored.) Errors from Consul will be
+// wrapped in a KVError value.
+func SafeKeys(
```
petertseng: I don't know if I'd be asking for too much if I ask for more specificity in the name, like `CancelableKeys`?

I suppose all I'll ask for is consistency. If we have other uses of `Safe` that all mean "cancelable", then I suppose this will be OK. But if `Safe` means other things in other instances in our codebase, then I will be suspicious of this name.

pkg/rc/farm.go
```diff
@@ -210,8 +211,21 @@ START_LOOP:
...
 				rcLogger.NoFields().Infoln("Acquired lock on new replication controller, spawning")
 
+				// TODO: maybe we don't need to fetch the RC
+				// here, but that involves changing replication
+				// controller code which expands the scope of
+				// the change to watching RC keys rather than
+				// values.  Also it probably doesn't matter
+				// much since we won't actually be doing a
+				// fetch that often (only when lock not held)
+				rc, err := rcf.rcStore.Get(rcKey.ID)
```
petertseng: hmm. one might wonder what the effect will be on a bunch of small `Get`s - for example, what happens when we turn on the RC farms after an extended outage so nothing is locked, so eventually all the RCs will be gotten? (grammar?) Or what happens when one node releases all its locks?

I don't mind deferring this question until later.

pkg/rc/farm.go
```diff
@@ -267,8 +281,21 @@ START_LOOP:
...
+// that an RC cannot be unmarshaled from JSON if there is no replicas_desired
+// field which would be the most likely way that all RC counts would be zeroed
+// out.
+func (rcf *Farm) initialFailsafe() {
+	rcFields, err := rcf.rcStore.List()
+	if err != nil {
+		panic(fmt.Sprintf("unable to do initial list of RCs to evaluate failsafe check: %s", err))
+	}
+
 	if len(rcFields) == 0 {
```
petertseng: I wish this could call `failsafe(rcFields)` to avoid duplicate code, but the types are different! Doh.

pkg/store/consul/rcstore/consul_store.go
```diff
@@ -195,44 +197,78 @@ func (s *consulStore) WatchNew(quit <-chan struct{}) (<-chan []fields.RC, <-chan
...
-// locks that were not held at some recent time
+// WatchRCKeysWithLockInfo executes periodic consul watch queries for
+// replication controller keys and passes the results on a channel along with
+// information about which keys have locks on them. Only keys are returned as a
+// bandwidth performance optimization. To observe changes to the fields of a
+// replication controller, individual watches should be used. Lock info is
+// returned so that it can be used to not try to lock RCs that are already
+// locked. In other words, it's a QPS optimization over the naive solution of
+// always attempting to lock every RC and relying on the consul server to
+// decide whether the lock is own.
```
petertseng: is *owned* rather than is *own*?

pkg/store/consul/rcstore/consul_store.go
```diff
@@ -195,44 +197,78 @@ func (s *consulStore) WatchNew(quit <-chan struct{}) (<-chan []fields.RC, <-chan
...
+					continue
+				}
+			}
+
+			rcIDs, err := keysToRCIDs(watchedKeys.Keys)
+			if err != nil {
+				select {
+				case <-quit:
+					return
+				case combinedErrCh <- watchedKeys.Err:
```
petertseng: (I predict this comment will end up displaying on the wrong line - I mean to comment on line 249, after you check `err != nil`)

Checking `err != nil`, but sending `watchedKeys.Err` instead of `err`? Should it be `err` instead? If it's intentional, tell me why

pkg/store/consul/rcstore/consul_store_test.go
```diff
@@ -354,38 +355,76 @@ func TestPublishLatestRCsWithLockInfoNoLocks(t *testing.T) {
...
+		},
+		{
+			Keys: []string{"replication_controllers/malformed_uuid"},
+			Err:  true,
+		},
+	}
+
+	for i, tc := range testCases {
+		ids, err := keysToRCIDs(tc.Keys)
+		if err != nil && tc.Err == false {
```
petertseng: do you specifically like having `b == false` rather than `!b`?

pkg/store/consul/rcstore/consul_store_test.go
```diff
@@ -354,38 +355,76 @@ func TestPublishLatestRCsWithLockInfoNoLocks(t *testing.T) {
...
+			t.Errorf("Case %d: unexpected error: %s", i, err)
+			continue
+		} else if err == nil && tc.Err == true {
```
petertseng: do you specifically like having `b == true` rather than `b`?

pkg/ds/farm_test.go
```diff
@@ -1048,7 +1048,7 @@ func waitForMutateSelectorFarms(firstFarm *Farm, secondFarm *Farm, ds ds_fields.
 // Polls for a condition to happen, will return an error if it does not happen
 // before the timeout
 func waitForCondition(condition func() error) error {
-	timeout := time.After(10 * time.Second)
+	timeout := time.After(20 * time.Second)
```
petertseng: oh! now that we fixed test flakes, do we still need this?

pkg/store/consul/consulutil/ops.go
```diff
@@ -96,6 +96,42 @@ func List(
 	}
 }
 
+type ConsulKeyser interface {
```
mpuncel: lol what jokes?

petertseng: Keyser Soze

pkg/store/consul/consulutil/ops.go
mpuncel: the other ones are called `Safe` also, i agree with you that that word doesn't make sense to me

pkg/rc/farm.go
mpuncel: hmm yeah I hadn't thought about what happens when the server turns on. I do think it'll be okay though, I believe consul is very efficient at single key retrievals. The main issue if there was one would probably be the http server spending a lot of effort on tls setup but connection pooling would probably make that better? In any case I think it should be okay but probably this fetch could be removed if we fix up RC code

pkg/store/consul/rcstore/consul_store.go
mpuncel: good catch

pkg/store/consul/rcstore/consul_store_test.go
mpuncel: lol i do stupid things like that sometimes with booleans

pkg/ds/farm_test.go
mpuncel: i think you're right

mpuncel:
I've addressed all the feedback, squashed into the commits

pkg/store/consul/consulutil/watch_test.go
```diff
@@ -207,6 +208,65 @@ func TestWatchPrefix(t *testing.T) {
...
+	keys := <-keysChan
+	if !reflect.DeepEqual(keys.Keys, []string{kv1a.Key}) {
+		t.Error("existing data not recognized, wanted %s but got %s", []string{kv1a.Key}, keys.Keys)
+	}
+
+	// Get an updates when the data changes (create, modify, delete)
+	f.Client.KV().Put(kv1b, nil)
+	keys = <-keysChan
+	if !reflect.DeepEqual(keys.Keys, []string{kv1b.Key}) {
+		t.Error("keys changed inappropriately")
```
petertseng: hmm. I'm sorry to say (because it will be a bit of work) that I also meant all other places in this file where `keys.Keys` was not the expected value

mpuncel: fixed that

mpuncel:
thanks @petertseng for the thorough reviews!

PR #763, opened 2017-02-16
Fix test flake in daemon_set_test.go

mpuncel:


pkg/ds/daemon_set_test.go
```diff
@@ -309,7 +307,7 @@ func TestSchedule(t *testing.T) {
 	labeled = labeledPods(t, ds)
 	Assert(t).AreEqual(len(labeled), expectedNodes, "Expected no nodes to be unlabeled")
 
-	err = waitForPodsInIntent(consulStore, len(beforeDeletePods))
+	err = waitForPodsInIntent(consulStore, 12)
```
petertseng: as this code is written, 12 will be the right number (I'll explain why later), but I believe the test might still fail if the daemon set gets deleted (and stops replicating) before the pod makes it to all 12. So, maybe a `waitForPodsInIntent(consulStore, 12)` before the delete would be appropriate.

12 is currently the right number because the 11 nodeQuality=good nodes on which the pod was previously scheduled hit the `"Reality for this node matches this DS. No action required."` and therefore they don't get new pods in intent.

Changing the manifest or having something clear out reality will ensure that all 23 show up.

petertseng: Then again, it's not very useful to wait for a number that's already there, since it will just exit immediately.

mpuncel: > 12 is currently the right number because the 11 nodeQuality=good nodes on which the pod was previously scheduled hit the "Reality for this node matches this DS. No action required." and therefore they don't get new pods in intent.

shouldn't there be intent entries left over from one of the previous replications?

mpuncel: ohhh i think the bug is in the fake preparer!

pkg/ds/daemon_set_test.go
```diff
@@ -28,40 +28,46 @@ const (
...
+		return util.Errorf("error getting pod locations: %s", err)
+	}
+
 	timedOut := false
 
 	// This for loop runs until you either time out or len(podLocations) == desired
 	// then return the length of whatever ds.CurrentNodes() is
 	for len(podLocations) != desired && !timedOut {
 		select {
 		case <-time.Tick(100 * time.Millisecond):
```
petertseng: you changed a `<-time.Tick` to a `<-time.After` further below - here too?

pkg/ds/daemon_set_test.go
```diff
@@ -289,28 +317,33 @@ func TestSchedule(t *testing.T) {
...
-	Assert(t).IsNil(err, "Unable to list intent/")
 	ds.logger.NoFields().Info("Deleting daemon set...")
 	err = dsStore.Delete(ds.ID())
 	if err != nil {
 		t.Fatalf("Unable to delete daemon set: %v", err)
 	}
 	// behavior change: Deleting a daemon set will no longer unschedule its pods (for now)
-	numNodes = waitForNodes(t, ds, 12, desiresErrCh, dsChangesErrCh)
-	Assert(t).AreEqual(numNodes, expectedNodes, "Unexpected number of nodes labeled")
+	err = waitForNodes(ds, 12, desiresErrCh, dsChangesErrCh)
```
petertseng: so now that you've made this error... how is it that it doesn't error? there really should be 23, right???

pkg/store/consul/consultest/fake_preparer.go
```diff
@@ -51,11 +51,7 @@ func (f *FakePreparer) Enable() {
 					f.logger.Errorf("Error getting all pods: %v", err)
 					continue
 				}
-				// Make a copy to know what to delete
-				podsToDelete := make(map[consul.ManifestResult]struct{})
-				for _, manifestResult := range allPods {
```
petertseng: this could have still been valid, but it would have needed to range over Reality, not Intent.

pkg/store/consul/consultest/fake_preparer.go
```diff
@@ -66,17 +62,33 @@ func (f *FakePreparer) Enable() {
...
-						consul.REALITY_TREE,
-						manifestResult.PodLocation.Node,
-						manifestResult.PodLocation.PodID,
-					)
-					if err != nil {
-						f.logger.Errorf("Error deleting pod: %v", err)
+				for _, manifestResult := range allReality {
+					found := false
+					for _, intentPod := range allPods {
+						if intentPod.PodLocation.Node == manifestResult.PodLocation.Node {
```
petertseng: why does it only check the `Node` and not the `PodID`?

This means that if I have intent: `node1/b`, and reality: `node1/a, node1/b`, then `node1/a` doesn't get deleted when it should, right?

mpuncel: I will fix that pronto

pkg/ds/daemon_set_test.go
mpuncel: probably because 12 is on the way to 23, so for some period of time it's true

mpuncel: actually that makes no sense because it was 23 before

pkg/ds/daemon_set_test.go
```diff
@@ -289,28 +321,33 @@ func TestSchedule(t *testing.T) {
...
-	Assert(t).IsNil(err, "Unable to list intent/")
 	ds.logger.NoFields().Info("Deleting daemon set...")
 	err = dsStore.Delete(ds.ID())
 	if err != nil {
 		t.Fatalf("Unable to delete daemon set: %v", err)
 	}
 	// behavior change: Deleting a daemon set will no longer unschedule its pods (for now)
-	numNodes = waitForNodes(t, ds, 12, desiresErrCh, dsChangesErrCh)
-	Assert(t).AreEqual(numNodes, expectedNodes, "Unexpected number of nodes labeled")
+	err = waitForNodes(ds, 23, desiresErrCh, dsChangesErrCh)
```
petertseng: could be expectedNodes, but OK.

PR #764, opened 2017-02-17
Integration: Use bento box, install necessary dependencies

petertseng:


integration/common-provision.sh
```diff
@@ -33,14 +33,13 @@ openssl req -x509 -newkey rsa:2048 -keyout $CERTPATH/key.pem -out $CERTPATH/cert
 
 go version
 
+sudo yum install -y ruby rubygem-rake gcc git
```
petertseng: I want to find evidence of necessity of each of these, just a second.

integration/common-provision.sh
```diff
@@ -33,14 +33,17 @@ openssl req -x509 -newkey rsa:2048 -keyout $CERTPATH/key.pem -out $CERTPATH/cert
...
+# git is required to `go get`
+sudo yum install -y rubygem-rake gcc git
+
 # Build p2.
 cd $GOPATH/src/github.com/square/p2
 go install ./...
 cp $GOPATH/bin/p2-exec /usr/local/bin
 
-# Install ruby + rake
-sudo yum install -y ruby rubygem-rake
```
petertseng: I assumed that `rubygem-rake` would cause `ruby` to get installed as well.

integration/common-provision.sh
```diff
@@ -33,14 +33,17 @@ openssl req -x509 -newkey rsa:2048 -keyout $CERTPATH/key.pem -out $CERTPATH/cert
 
 go version
 
+# check.go assumes rake to be present (it invokes `rake install`)
+# gcc is required to `go install` certain packages, probably those that `import "C"`
+# (p2/bin/p2-exec, opencontainers/runc/libcontainer/system, mattn/go-sqlite3)
+# git is required to `go get`
+sudo yum install -y rubygem-rake gcc git
```
petertseng: moved up, because as I say, we need `gcc` to `go install`, and there is a `go install` below.

PR #768, opened 2017-02-21
This commit updates protobuf to

mpuncel:
69b215d01a5606c843240eab4937eab3acee6530 and update gRPC to v1.0.5.

Background: I'd like to upgrade the version of gRPC that Square uses,
and that requires generating proto messages with protobuf that is
consistent with what gRPC uses. (Otherwise we will get a compile error
since SupportPackageIsVersionX does not exist.)

This commit will upgrade P2's protobuf to use the latest upstream
master that is same as Square's internal fork. We also need
to upgrade gRPC to avoid the same compile error inside the p2 repo.

Run the following commands to update the dep:

```
> godep restore
> cd ~/Development/go/src/google.golang.org/grpc/
> git co v1.0.5
> cd -
> cd ~/Development/go/src/github.com/golang/protobuf/
> git co 69b215d01a5606c843240eab4937eab3acee6530
> cd -
> godep update google.golang.org/grpc/... github.com/golang/protobuf/...
> protoc --go_out=plugins=grpc:. pkg/grpc/podstore/protos/podstore.proto
```

### DIFF
--- a/Godeps/Godeps.json
+++ b/Godeps/Godeps.json
@@ -1,7 +1,7 @@
 {
 	"ImportPath": "github.com/square/p2",
 	"GoVersion": "go1.7",
-	"GodepVersion": "v75",
+	"GodepVersion": "v79",
 	"Packages": [
 		"./..."
 	],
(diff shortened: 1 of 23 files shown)


kkaneda:
Thanks! LGTM

PR #769, opened 2017-02-27
Mpuncel/unschedule pod

mpuncel:


rudle:
Have you thought about retries for this? It would be nice if I could call this repeatedly without encountering an error.

mpuncel:
I was thinking clients could check for the NotFound status. in some cases that be a "whoops i gave you the wrong uuid" and in others it might be "i know the uuid is right, don't care if it was there just want to make sure it's gone"

mpuncel:
should be easy to check with `if grpc.Code(err) == codes.NotFound`

rudle:
neat! :ship:

PR #770, opened 2017-03-01
Don't run disable/enable for oneoff pods.

mpuncel:
oneoff pods don't need to have their disable/enable scripts run because
they are oneoff jobs with immutable state. Furthermore running
disable/enable scripts may impact other pods with the same pod ID on the
same physical hardware depending on the implementation if
enable/disable.

This commit creates a (artificial at this point) distinction between
uuid pods and oneoff pods. A uuid pod is a pod whose data is stored in
the pods/ consul tree with an associated uuid. A oneoff pod is a pod
that is meant to run once and does not represent a repeatedly running
service like a web server.

These two concepts are tied together currently but now there is a
IsOneoff() function on the hoist launchable type which may change at
some point.

This commi

### DIFF
--- a/pkg/hoist/hoist_launchable.go
+++ b/pkg/hoist/hoist_launchable.go
@@ -319,6 +319,14 @@ func (hl *Launchable) Executables(
 
 		for _, relativePath := range relativeExecutablePaths {
 			var entryPointName string
+
+			// This is a hack to preserve the runit service
+			// directory layout for legacy pods. From a theoretical
+			// standpoint we would like to include all parts of the
+			// path in the service name so that there could be
+			// multiple files started with the same basename but
+			// different paths. UUID pods are new so we can adopt
+			// the scheme we want
 			if hl.IsUUIDPod {
 				entryPointName = strings.Replace(relativePath, "/", "__", -1)
 			} else {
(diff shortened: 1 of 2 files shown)


pkg/hoist/hoist_launchable.go
```diff
@@ -74,7 +74,17 @@ func (hl *Launchable) If() launch.Launchable {
 	return LaunchAdapter{Launchable: hl}
 }
 
+func (hl *Launchable) IsOneoff() bool {
```
rudle: cool, that was easy. Can you replace the other reference to the `IsUUIDPod` in this file in a separate commit? It's on line 322.

rudle:
LGTM. We should send a small PSA out to our customers when this goes live.

pkg/hoist/hoist_launchable.go
mpuncel: hmm that was just the hack so we don't change the runit directory layout for legacy customers. I sort of think of that as more of the uuid vs not-uuid split rather than oneoff vs long running

pkg/hoist/hoist_launchable_test.go
```diff
@@ -374,6 +386,18 @@ func TestStop(t *testing.T) {
 	Assert(t).IsNil(err, "Got an unexpected error when attempting to stop runit services")
 }
 
+func TestNoDisableForUUIDPods(t *testing.T) {
+	hl, sb := FakeHoistLaunchableForDirUUIDPod("failing_scripts_test_hoist_launchable")
+	defer CleanupFakeLaunchable(hl, sb)
+
+	// If enable actually gets run, we'll get an error because we chose the launchable
```
petertseng: disable?

PR #771, opened 2017-03-02
Mpuncel/update pod selector

mpuncel:


pkg/pc/control/control.go
```diff
@@ -95,7 +95,8 @@ func (pccontrol *PodCluster) Get() (fields.PodCluster, error) {
 	return pccontrol.pcStore.Get(pc.ID)
 }
 
-func (pccontrol *PodCluster) Update(annotations fields.Annotations) (fields.PodCluster, error) {
+// UpdateAnnotations replaces the annotations on the pod cluster configured for the pod clustr control structure
```
petertseng: clustr -> cluster

bin/p2-pcctl/main.go
```diff
@@ -2,10 +2,12 @@ package main
 
 import (
 	"encoding/json"
+	"errors"
 	"fmt"
 	"log"
 	"os"
 	"os/user"
+	"square/up/p2/pkg/cli"
```
petertseng: nah, github

PR #772, opened 2017-03-02
Revert "Merge pull request #739 from petertseng/gzip-as-user"

petertseng:
This reverts commit c2d478a9633c68bef0d64845d56c492b7df94890, reversing
changes made to 94d5805d6a377b16ff2e2b440efad37345aee5f9.

We're not quite ready for this to go out yet, since:

1. The **tmp directory** (to which files are downloaded) needs to be
   world-readable, and it is not on some machines.
2. We're not sure whether it's safe to make it world-readable, or why it
   is not.

### DIFF
--- a/pkg/artifact/downloader.go
+++ b/pkg/artifact/downloader.go
@@ -65,12 +65,13 @@ func (l *downloader) Download(location *url.URL, verificationData auth.Verificat
 		return err
 	}
 
-	err = artifactFile.Chmod(0644)
+	// rewind a second time to allow the archive to be unpacked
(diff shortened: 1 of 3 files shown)


petertseng:
Actually I don't need to revert this wholesale, just need to build off the sha right before it.

I hate to do this though. Should we discuss whether it's safe to make the tmpdir world-readable?

petertseng:
We do not need this today. We'll deploy from the commit immediately preceding the merge of #739.

If we need to deploy any other change before having fixed permissions, this is the ticket.

PR #773, opened 2017-03-02
Use text formatting for log messages except for in socket hook.

mpuncel:
p2-preparer's logger allows adding a hook that will write log messages
to a socket in addition to stdout/stderr. This is useful to have another
program consume the log messages from the socket.

To allow this to work, we were always using logrus's JSON formatter
since parsing json is much easier for the program on the other end of
the socket. However this is a worse experience for humans trying to read
log messages.

This commit makes the logging package use the text formatter but makes
the socket hook a special case that will always use the JSON formatter
regardless of the formatter actually set on the logger.


rudle:
This is a very nice change.

PR #776, opened 2017-03-04
Mpuncel/store interfaces

mpuncel:


pkg/store/consul/dsstore/consul_store.go
```diff
@@ -40,18 +46,18 @@ func (e CASError) Error() string {
 	return fmt.Sprintf("Could not check-and-set key %q", string(e))
 }
 
-// Make sure functions declared in the Store interface have the proper contract
-var _ Store = &consulStore{}
-
-func NewConsul(client consulutil.ConsulClient, retries int, logger *logging.Logger) Store {
-	return &consulStore{
+func NewConsul(client consulutil.ConsulClient, retries int, logger *logging.Logger) *store {
```
mpuncel: I agree with you it shouldn't be private, i'll fix all these. I do think I want to not return interfaces though. Justification: https://medium.com/@cep21/preemptive-interface-anti-pattern-in-go-54c18ac0668a.

```
Go is at its most powerful when interface definitions are small. In the standard library, most interface definitions are a single method. This allows the greatest reuse because implementing an interfaces is easy to do. When programmers code to preemptive interfaces like Auth above, the interfaces tend to explode in method count, which makes the whole point of an interface (exchangeable implementations) even more difficult to accomplish.
```

mpuncel: Basically i want this type to return a type that has the full range of capability of talking to consul as it relates to daemon sets, but if it is captured into small interface variables where it is used then it will be easier to swap implementations

rudle: ah gotcha. This work is laying groundwork for the future.
:+1:

mpuncel:
addressed conflicts and added a commit that exports all the types

bin/p2-rctl/main.go
```diff
@@ -113,15 +115,24 @@ func main() {
...
-		rcs:        rcstore.NewConsul(client, labeler, 3),
-		rls:        rollstore.NewConsul(client, labeler, nil),
-		consuls:    consul.NewConsulStore(client),
-		labeler:    labeler,
-		hcheck:     checker.NewConsulHealthChecker(client),
-		logger:     logger,
+		// rcStore copied so many times right now but might not all be
+		// the same implementation of these various interfaces in
+		// the future. Also rcWatcher isn't even used it is simply
+		// used by the roll farm to initialie an RC for a computation
```
petertseng: Oh. And you know, #792 removed that. So, at this point the only thing using rcWatcher is update_test.go I beieve.

PR #777, opened 2017-03-07
Make it easier to delete a uuid pod for which pod ID is not known.

mpuncel:
A pod's uuid contains all the information necessary to delete that pod.
For safety, p2-rm historically required you pass the uuid as well as the
pod ID to confirm that the user is deleting the pod they expect they
are.

However this makes it harder to use the tool. This commit adds a
confirmation prompt that tells the user the pod ID of the pod being
deleted to preserve safety while adding some ease.

### DIFF
--- a/bin/p2-rm/rm.go
+++ b/bin/p2-rm/rm.go
@@ -1,10 +1,12 @@
 package main
 
 import (
+	"errors"
 	"fmt"
 	"path"
 	"time"
 
+	"github.com/square/p2/pkg/cli"
 	"github.com/square/p2/pkg/labels"
 	"github.com/square/p2/pkg/rc"
 	"github.com/square/p2/pkg/rc/fields"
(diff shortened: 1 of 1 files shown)


mpuncel:
fixed `bin/p2-rm/rm.go:157: possible formatting directive in Println call`

bin/p2-rm/rm.go
```diff
@@ -147,8 +149,16 @@ func (rm *P2RM) deleteUUIDPod() error {
...
 	switch {
 	case err == nil:
-		if pod.Manifest.ID() != rm.PodID {
-			return fmt.Errorf("pod %s has a podID of %s, but %s was passed as the --pod option", rm.PodUniqueKey, pod.Manifest.ID(), rm.PodID)
+		if rm.PodID != "" {
+			if pod.Manifest.ID() != rm.PodID {
+				return fmt.Errorf("pod %s has a podID of %s, but %s was passed as the --pod option", rm.PodUniqueKey, pod.Manifest.ID(), rm.PodID)
+			}
+		} else {
+			fmt.Println(fmt.Sprintf("%s has pod ID of %s, do you wish to proceed?", rm.PodUniqueKey, pod.Manifest.ID()))
```
petertseng: I don't suppose one could use Printf here?

mpuncel: yeah, i don't like needing to remember newlines but you're probably right that that's easier. wish there was a Printlnf or something. since the build is already kicked off i'm gonna leave it

PR #781, opened 2017-03-08
Mpuncel/list pod status

mpuncel:


mpuncel:
this contains commits from https://github.com/square/p2/pull/780 so please review that first

pkg/store/consul/statusstore/podstatus/consul_store.go
```diff
@@ -20,7 +22,11 @@ type consulStore struct {
 	namespace statusstore.Namespace
 }
 
-func NewConsul(statusStore statusstore.Store, namespace statusstore.Namespace) Store {
+// TODO: this pod store is coupled with the PodStatus struct, which represents
+// the book keeping that the preparer does about a pod. In other words it only
+// makes sense if the namespace is consul.PreparerPodStatusNamespace. We should
+// probably take namespace out of all these APIs and use that constant instead.
+func NewConsul(statusStore statusstore.Store, namespace statusstore.Namespace) *consulStore {
```
rudle: Did we already talk about returning private types from this type package? It's generally a no-no and I'm not yet sufficiently motivated to support it here.

mpuncel: i'll make it public. I imagine if we were generating the godocs the return type would not be documented since it's private

PR #782, opened 2017-03-08
Mpuncel/delete pod status

mpuncel:


mpuncel:
This contains commits from #780 and #781 so review those first

PR #783, opened 2017-03-09
Mpuncel/pod status client code

mpuncel:


mpuncel:
This contains commits from #780, #781, and #782 so review those first

PR #786, opened 2017-03-10
Mpuncel/intent key watch

mpuncel:


mpuncel:
The motivation for this PR is that `WatchIntentKeys()` function could later be used by the preparer instead of `WatchPods()` in order to cut down on bandwidth usage. The preparer would track keys and modify indexes and only fetch the manifest for a given key if the modify index is newer than the last one that was seen.

This way watches firing are lighter weight during preparer steady state. Typically only one manifest will change at a time between watches and today that means all of the manifests are downloaded over the network rather than just the one that changed.

mpuncel:
This PR doesn't change preparer behavior yet because that will deserve the attention of a PR on its own. Also this one is plenty big enough already

mpuncel:
This PR is still using WatchPrefix, which is exactly the same as before. Additionally, ?keys queries in consul do not give the ModifyIndex. Unfortunately this will require a change to consul before this is feasible.

PR #787, opened 2017-03-13
Make the default consul wait time configurable in p2-preparer.

mpuncel:
In large deployments of P2, a lot of bandwidth from the consul server is
used by potentially thousands of preparer instances pulling manifests
that might be multiple Kb constantly. When no intent data for a node is
changing, this means a request completes every 5 minutes, returning the
full data.

This commit makes it easy to change the wait time on consul watches
performed by p2-preparer so that bandwidth usage during periods of
steady data can be tuned.

### DIFF
--- a/pkg/preparer/setup.go
+++ b/pkg/preparer/setup.go
@@ -103,6 +103,16 @@ type store interface {
 	)
 }
 
+// ConsulConfig encapsulates config options related to how p2-preparer
+// interacts with consul.
+// TODO: move ConsulAddress, ConsulHttps, ConsulTokenPath here.
+type ConsulConfig struct {
+	// WaitTime specifies the timeout length for HTTP watches on consul. Longer
+	// values mean longer lived requests and therefore lower QPS and bandwidth
+	// usage when there are infrequent changes to the watched data
+	WatchWaitTime time.Duration `yaml:"watch_wait_time"`
+}
+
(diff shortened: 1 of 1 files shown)


pkg/preparer/setup.go
```diff
@@ -255,10 +266,11 @@ func (c *PreparerConfig) getOpts() (consul.Options, error) {
...
 	return consul.Options{
-		Address: c.ConsulAddress,
-		HTTPS:   c.ConsulHttps,
-		Token:   token,
-		Client:  client,
+		Address:  c.ConsulAddress,
+		HTTPS:    c.ConsulHttps,
+		Token:    token,
+		Client:   client,
+		WaitTime: c.ConsulConfig.WatchWaitTime,
```
rudle: The default case (0s) would have some pretty profound effects on bandwidth, right? Unless consul uses 0 as a sentinel or something…

mpuncel: since we weren't setting this on `consul.Options` before, it was already zero. It's defaulted to 5 minutes somewhere within consul

mpuncel: https://github.com/hashicorp/consul/blob/master/consul/rpc.go#L378. Although point taken, i'll make this explicitly 5 minutes if the setting is < 5 minutes

PR #789, opened 2017-03-13
Mpuncel/ds grpc api

mpuncel:


mpuncel:
the choice of which functions to implement was informed by the work in #776 but the code doesn't require that PR to be merged in order to build

mpuncel:
I rolled back my changes to the fake dsstore because it was breaking other tests. Wrote my own fake store implementation to test the WatchDaemonSets grpc endpoint.

PR #790, opened 2017-03-13
Unconditionally install hooks on preparer startup.

mpuncel:
Prior to this commit the hook manifest (fetched from preparer manifest
config) was compared to the current manifest written to disk for hooks.
If the manifests were the same, hook installation was skipped.

This commit removes this check and just unconditionally installs the
hooks.

This is for two reasons:
1) Installing is low-cost anyway. We don't redownload artifacts if
they're present on the disk, and the rest is just linking and writing of
the current manifest file.
2) It's possible that the hooks might have been removed but the
current_manifest.yaml file for them was untouched, which will cause the
preparer to incorrectly not install them.

### DIFF
--- a/pkg/preparer/setup.go
+++ b/pkg/preparer/setup.go
@@ -568,31 +568,8 @@ func (p *Preparer) InstallHooks() error {
 		"pod": p.hooksManifest.ID(),
 	})
 
-	// Figure out if we even need to install anything.
-	// Hooks aren't running services and so there isn't a need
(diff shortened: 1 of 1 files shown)


rudle:
What if I'm developing hooks on host? Probably I have to update the preparer manifest anyway.

mpuncel:
as long as the preparer doesn't restart you could theoretically swap in different things on the host if you have the necessary privileges, but if you need to restart the preparer it will be wiped

PR #791, opened 2017-03-14
Take a context.Context argument in podstore client functions.

mpuncel:
Even though the context is currently ignored on the server side for
these functions, it's idiomatic to allow contexts to be passed all the
way through.


rudle:
ctx for the ctx throne

PR #792, opened 2017-03-16
roll: Ensure current == desire before enabling new RC; rc: extract CurrentPods

petertseng:
As noted in the comments, if we enable before it has met its desire, we
may momentarily unschedule a node (most likely to happen in times of
high consul latency).

---

~~WIP: probably not yet ready for merge for reasons I will explain in a comment.
If you want it urgently, you can merge it, because the code is believed functional.
But there is a small matter of code cleanliness to deal with first.~~

### DIFF
--- a/pkg/rc/replication_controller.go
+++ b/pkg/rc/replication_controller.go
@@ -320,12 +320,13 @@ func (rc *replicationController) eligibleNodes() ([]types.NodeName, error) {
 	return rc.scheduler.EligibleNodes(rc.Manifest, rc.NodeSelector)
 }
 
-func (rc *replicationController) CurrentPods() (types.PodLocations, error) {
-	selector := klabels.Everything().Add(RCIDLabel, klabels.EqualsOperator, []string{rc.ID().String()})
+// CurrentPods returns all pods managed by an RC with the given ID.
+func CurrentPods(rcid fields.ID, labeler Labeler) (types.PodLocations, error) {
+	selector := klabels.Everything().Add(RCIDLabel, klabels.EqualsOperator, []string{rcid.String()})
 
 	// replication controllers can only pass cachedMatch = false because their operations for matching
 	// replica counts are not necessarily idempotent.
-	podMatches, err := rc.podApplicator.GetMatches(selector, labels.POD, false)
+	podMatches, err := labeler.GetMatches(selector, labels.POD, false)
 	if err != nil {
 		return nil, err
 	}
@@ -341,6 +342,10 @@ func (rc *replicationController) CurrentPods() (types.PodLocations, error) {
 		result[i].PodID = podID
 	}
 	return result, nil
+}
+
+func (rc *replicationController) CurrentPods() (types.PodLocations, error) {
+	return CurrentPods(rc.ID(), rc.podApplicator)
 }
 
 // forEachLabel Attempts to apply the supplied function to all user-supplied labels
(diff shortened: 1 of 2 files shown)


pkg/roll/run_update.go
```diff
@@ -371,6 +371,38 @@ func (u *update) enable() error {
...
+		//
+		// Now let's say we interrupt the deploy here,
+		// and because of Consul latency the new RC doesn't get to act on its desire.
+		// Then, we rollback.
+		//
+		//  enabled, 1 desired, 2 labeled | disabled, 1 desired, 0 labeled.
+		//
+		// At this point, the formerly-old RC momentarily unschedules one node, because it has too many.
+		// This is undesirable.
+		currentPods, err := rc.New(newRC, u.consuls, u.rcs, u.sched, u.labeler, u.logger, u.alerter).CurrentPods()
```
petertseng: this is gratuitous. We create a new `rc`, passing it all these parameters, and then immediately discard it.

We do the same in `countHealthy` below.

pkg/roll/run_update.go
```diff
@@ -371,6 +371,38 @@ func (u *update) enable() error {
...
+		// disabled, 2 desired, 2 labeled | enabled, 0 desired, 0 labeled.
+		// disabled, 1 desired, 2 labeled | enabled, 1 desired, 0 labeled.
+		//
+		// Now let's say we interrupt the deploy here,
+		// and because of Consul latency the new RC doesn't get to act on its desire.
+		// Then, we rollback.
+		//
+		//  enabled, 1 desired, 2 labeled | disabled, 1 desired, 0 labeled.
+		//
+		// At this point, the formerly-old RC momentarily unschedules one node, because it has too many.
```
petertseng: Honestly though, you could argue that we could optimize here by simply setting the RCs to 2/0 at this point, if they haven't had a chance to act on their desire. This might get racy if they were in the process of doing so, though.

petertseng: You might think that race is harmless, but in fact it could erroneously leave the cluster in a state where one node is on the old version and one node on the new, which is not what is desired.

petertseng: Okay, I don't have a sequence of events to back up that "in fact", so maybe I can be proven wrong on that.

petertseng: I've added a comment explaining the alternative.

pkg/roll/run_update.go
```diff
@@ -400,7 +436,7 @@ func (u *update) countHealthy(id rcf.ID, checks map[types.NodeName]health.Result
 
 	ret.Desired = rcFields.ReplicasDesired
 
-	currentPods, err := rc.New(rcFields, u.consuls, u.rcs, u.sched, u.labeler, u.logger, u.alerter).CurrentPods()
```
petertseng: wonder if I can remove any of these fields now....

pkg/rc/replication_controller.go
```diff
@@ -320,12 +320,13 @@ func (rc *replicationController) eligibleNodes() ([]types.NodeName, error) {
 	return rc.scheduler.EligibleNodes(rc.Manifest, rc.NodeSelector)
 }
 
-func (rc *replicationController) CurrentPods() (types.PodLocations, error) {
-	selector := klabels.Everything().Add(RCIDLabel, klabels.EqualsOperator, []string{rc.ID().String()})
+// CurrentPods returns all pods managed by an RC with the given ID.
+func CurrentPods(rcid fields.ID, labeler Labeler) (types.PodLocations, error) {
```
mpuncel: nice!

mpuncel:
looks good to me, up to you whether you want to merge. we're strictly adding constraints so we can't be decreasing safety

pkg/roll/run_update.go
petertseng: I think the cleanup PR will come later.

PR #795, opened 2017-03-24
roll: Disable old RC after convergence check

petertseng:
If the new RC is already disabled, the only thing that can cause the new
RC's number of labeled nodes to meet its desired number (assuming the
desired number is too high) is for the other RC to steal some nodes from
it. If the other RC is disabled before it can possibly do that,
deadlock.

Note that my ideal solution in this situation is actually to change the
desired counts of the two RCs to match what they already have labeled,
but I was worried that a transactionless update might fail in an
inppoortune place. I suppose the risk of that is no greater than what it
already is.


petertseng:
I don't like my branch name

PR #797, opened 2017-03-24
roll: If old RC desired < labeled, increase it to match

petertseng:
Reviewers: I think I will prefer #796 over this.
Nevertheless, it is here for consideration (and for me to keep it around in case it becomes necessary).

Writing this code really makes me think "Add from one RC and subtract from another" should be in a transaction. I have unreasonable fear that we will leave RC counts in an incorrect state here. But note that this fear should already exist, since we have no transactions in current roll code.

### DIFF
--- a/pkg/roll/run_update.go
+++ b/pkg/roll/run_update.go
@@ -381,17 +381,28 @@ func (u *update) enable() error {
 		// At this point, the formerly-old RC momentarily unschedules one node, because it has too many.
 		// This is undesirable.
 		//
-		// Solution: Wait until formerly-old RC has only 1 node labeled.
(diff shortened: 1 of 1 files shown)


petertseng:
If I ever use this, it should probably get a test.

petertseng:
Honestly, I can't even be sure this is safe. What if the formerly-new RC has an in-flight request to relabel a node? I guess we can say that's not disastrous, as then the formerly-old RC ends up at, say, 3 desired 2 labeled, fine.

mpuncel:
I believe that the other change you made will address the issue by allowing the RCs to catch up. I share your concern about this approach (not because i've thought about it extensively but it's a more invasive change)

PR #798, opened 2017-03-24
Ask for confirmation before changing RC replica count in p2-rctl.

mpuncel:
The p2-rctl set-replicas command is very useful and powerfull, however
it's easy to make a typo and cause a problem. If the replica count being
set is lower than the currently set one, it will cause pod
uninstallations.

This commit adds a confirmation prompt which will give a user time to
confirm that the count they typed in was in fact what they wanted.

### DIFF
--- a/bin/p2-rctl/main.go
+++ b/bin/p2-rctl/main.go
@@ -15,6 +15,7 @@ import (
 	"gopkg.in/alecthomas/kingpin.v2"
 	klabels "k8s.io/kubernetes/pkg/labels"
 
+	"github.com/square/p2/pkg/cli"
 	"github.com/square/p2/pkg/health/checker"
 	"github.com/square/p2/pkg/labels"
 	"github.com/square/p2/pkg/logging"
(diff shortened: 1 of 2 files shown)


mpuncel:
added a bypass, the integration test needed it too

bin/p2-rctl/main.go
```diff
@@ -60,6 +61,7 @@ var (
 	cmdReplicas = kingpin.Command(cmdReplicasText, "Set desired replica count of a replication controller")
 	replicasID  = cmdReplicas.Arg("id", "replication controller uuid to modify").Required().String()
 	replicasNum = cmdReplicas.Arg("replicas", "number of replicas desired").Required().Int()
+	yes         = cmdReplicas.Flag("yes", "auto confirm the replica change (i.e. no confirmatino prompt)").Short('y').Bool()
```
petertseng: confirmatino!

mpuncel: darn

PR #801, opened 2017-03-27
[WIP] p2-watch can watch the health tree

rudle:
I wrote this code to debug an issue and I think it's generally useful. If yes, I'll polish it up for a proper review.

### DIFF
--- a/bin/p2-watch/main.go
+++ b/bin/p2-watch/main.go
@@ -6,7 +6,9 @@ import (
 	"os"
 	"os/signal"
 	"syscall"
+	"time"
 
+	"github.com/square/p2/pkg/health"
 	"github.com/square/p2/pkg/labels"
 	"github.com/square/p2/pkg/logging"
 	"github.com/square/p2/pkg/pc/fields"
@@ -16,16 +18,21 @@ import (
 	"github.com/square/p2/pkg/store/consul/pcstore"
 	"github.com/square/p2/pkg/types"
 
+	"sort"
+
 	"github.com/Sirupsen/logrus"
+	"github.com/square/p2/pkg/health/checker"
 	"github.com/square/p2/pkg/version"
 	"gopkg.in/alecthomas/kingpin.v2"
 )
 
 var (
-	nodeName     = kingpin.Flag("node", "The node to do the scheduling on. Uses the hostname by default.").String()
-	watchReality = kingpin.Flag("reality", "Watch the reality store instead of the intent store. False by default").Default("false").Bool()
-	hooks        = kingpin.Flag("hook", "Watch hooks.").Bool()
-	podClusters  = kingpin.Flag("pod-clusters", "Watch pod clusters and their labeled pods").Bool()
+	nodeName      = kingpin.Flag("node", "The node to do the scheduling on. Uses the hostname by default.").String()
+	watchReality  = kingpin.Flag("reality", "Watch the reality store instead of the intent store. False by default").Default("false").Bool()
+	hooks         = kingpin.Flag("hook", "Watch hooks.").Bool()
+	podClusters   = kingpin.Flag("pod-clusters", "Watch pod clusters and their labeled pods").Bool()
+	watchHealthF  = kingpin.Flag("health", "Watch health using ConsulHealthChecker").Bool()
+	healthService = kingpin.Arg("health-pod", "Pod to watch. Required if --health is passed").String()
 )
 
 func main() {
@@ -43,6 +50,13 @@ func main() {
 	}
 	if *podClusters {
 		watchPodClusters(client, applicator)
+	} else if *watchHealthF {
+		if *healthService == "" {
+			log.Fatal("Refusing to watch entire health tree, please set a pod ID with --health-pod")
+		}
+
+		watchHealth(*healthService, client)
+		return
 	} else {
 		podPrefix := consul.INTENT_TREE
 		if *watchReality {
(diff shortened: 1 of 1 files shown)


petertseng:
If this would print out the health of all the nodes for the named service on startup and every time any individual node updates, I could be convinced of its usefulness.

bin/p2-watch/main.go
```diff
@@ -17,15 +19,18 @@ import (
...
-	nodeName     = kingpin.Flag("node", "The node to do the scheduling on. Uses the hostname by default.").String()
-	watchReality = kingpin.Flag("reality", "Watch the reality store instead of the intent store. False by default").Default("false").Bool()
-	hooks        = kingpin.Flag("hook", "Watch hooks.").Bool()
-	podClusters  = kingpin.Flag("pod-clusters", "Watch pod clusters and their labeled pods").Bool()
+	nodeName      = kingpin.Flag("node", "The node to do the scheduling on. Uses the hostname by default.").String()
+	watchReality  = kingpin.Flag("reality", "Watch the reality store instead of the intent store. False by default").Default("false").Bool()
+	hooks         = kingpin.Flag("hook", "Watch hooks.").Bool()
+	podClusters   = kingpin.Flag("pod-clusters", "Watch pod clusters and their labeled pods").Bool()
+	watchHealth   = kingpin.Flag("health", "Watch health using ConsulHealthChecker").Bool()
+	healthService = kingpin.Arg("health-pod", "Pod to watch. Required if -health is passed").String()
```
mpuncel: s/-health/--health

bin/p2-watch/main.go
```diff
@@ -43,6 +48,33 @@ func main() {
...
+		for {
+			select {
+			case hr = <-healthResults:
+				log.Printf("health watch sez: %+v\n", hr)
+			case err = <-errCh:
+				log.Printf("[ERROR] %v\n", err)
+			case <-quitCh:
+				log.Println("huh? quitch shouldn't do that")
+			default:
+				// log.Println("backoff 1s")
```
mpuncel: remove

bin/p2-watch/main.go
```diff
@@ -43,6 +48,33 @@ func main() {
...
+				<-time.After(200 * time.Millisecond)
```
mpuncel: nit: may as well `time.Sleep`

rudle:
PTAL, this has changed a bit since the last review:
1) Methods were extracted
2) Formatting has changed (now it looks like `$host $result`)
3) earlier comments addressed

bin/p2-watch/main.go
```diff
@@ -122,3 +134,39 @@ func watchPodClusters(client consulutil.ConsulClient, applicator labels.Applicat
...
+
+	var hr map[types.NodeName]health.Result
+	var sortedHealthResults []health.Result
+	var err error
+	for {
+		select {
+		case hr = <-healthResults:
+			for _, r := range hr {
+				fmt.Printf("%s %s\n", r.Node.String(), r.Status)
+				sortedHealthResults = append(sortedHealthResults, r)
```
petertseng: confused. You don't seem to do anything with these sortedHealthResults...

Nor do you appear to actually sort anything.

rudle: Right. I guess I should sort these.

_typical lamentation about generics_

bin/p2-watch/main.go
```diff
@@ -122,3 +136,56 @@ func watchPodClusters(client consulutil.ConsulClient, applicator labels.Applicat
...
+}
+
+type nodeHealthResults []health.Result
+
+func (hrs nodeHealthResults) Len() int {
+	return len(hrs)
+}
+
+func (hrs nodeHealthResults) Less(i, j int) bool {
+	return hrs[i].Node.String() < hrs[j].Node.String()
```
rudle: This is not the best possible ordering for this data. However, it does meet the baseline of keeping the nodes in a stable order in the output.

bin/p2-watch/main.go
```diff
@@ -122,3 +136,56 @@ func watchPodClusters(client consulutil.ConsulClient, applicator labels.Applicat
...
+
+	go func() {
+		signalCh := make(chan os.Signal, 2)
+		signal.Notify(signalCh, syscall.SIGTERM, os.Interrupt)
+		received := <-signalCh
+		log.Printf("[WARN] Received %v, shutting down", received)
+		close(quitCh)
+	}()
+
+	var hr map[types.NodeName]health.Result
```
petertseng: Might not need to have this outside. I don't think it saves allocations either since healthResults is probably emitting new ones every time.

bin/p2-watch/main.go
```diff
@@ -122,3 +136,56 @@ func watchPodClusters(client consulutil.ConsulClient, applicator labels.Applicat
...
+	var err error
+	for {
+		select {
+		case hr = <-healthResults:
```
petertseng: could be `:=`, right?

bin/p2-watch/main.go
```diff
@@ -122,3 +136,56 @@ func watchPodClusters(client consulutil.ConsulClient, applicator labels.Applicat
...
+			var sortedHealthResults nodeHealthResults
+			for _, r := range hr {
+				sortedHealthResults = append(sortedHealthResults, r)
+			}
+			sort.Sort(sortedHealthResults)
+			for _, r := range sortedHealthResults {
+				fmt.Printf("%s %s\n", r.Node.String(), r.Status)
+			}
+			fmt.Printf("\n")
+		case err = <-errCh:
```
petertseng: could be `:=`, right?

bin/p2-watch/main.go
```diff
@@ -43,6 +50,13 @@ func main() {
 	}
 	if *podClusters {
 		watchPodClusters(client, applicator)
+	} else if *watchHealthF {
+		if *healthService == "" {
+			log.Fatal("Refusing to watch entire health tree, please set a pod ID with --health-pod")
+		}
+
+		watchHealth(*healthService, client)
+		return
```
petertseng: not *really* necessary, and a bit inconsistent because no other cases have it, but I guess it doesn't hurt

bin/p2-watch/main.go
petertseng: I don't know what other ordering you had in mind. I do like having the nodes in a stable order, so I think node name seems good. Were you thinking of something else?

rudle: For Square's hostname scheme, using the first letter isn't very logical (as it implies the class of the host). A better sort would probably sort by DC name, then by zone then by number.

bin/p2-watch/main.go
rudle: I put this here because I don't want behaviour to change if somebody adds code below the close of the else clause.

PR #802, opened 2017-03-28
Add Watch histograms for new-ish WatchKeys function

rudle:
As with the other watch functions, this allows us to see where we are spending
our time in Watch-land. Of particular interest to us is time spend blocking on the
output channel, it may motivate us to change the dataflow patterns.

### DIFF
--- a/pkg/store/consul/consulutil/watch.go
+++ b/pkg/store/consul/consulutil/watch.go
@@ -29,11 +29,19 @@ func WatchKeys(
 		var currentIndex uint64
 		timer := time.NewTimer(0)
 
+		listLatencyHistogram, consulLatencyHistogram, outputPairsBlocking, _ := watchHistograms("keys")
+
 		// put a lower bound on pause time to prevent tight looping queries against
 		// consul
 		if pause < 250*time.Millisecond {
 			pause = 250 * time.Millisecond
 		}
+
+		// Don't allocate these in a loop
+		var (
+			listStart        time.Time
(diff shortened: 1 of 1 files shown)


pkg/store/consul/consulutil/watch.go
```diff
@@ -43,10 +51,14 @@ func WatchKeys(
 			}
 
 			timer.Reset(pause)
+			listStart = time.Now()
 			keys, queryMeta, err := SafeKeys(clientKV, done, prefix, &api.QueryOptions{
 				WaitIndex:  currentIndex,
 				AllowStale: true,
 			})
+			consulLatencyHistogram.Update(int64(queryMeta.RequestTime))
+			listLatencyHistogram.Update(int64(time.Since(listStart) / time.Millisecond))
```
mpuncel: do you expect these two things to differ much?

pkg/store/consul/consulutil/watch.go
```diff
@@ -81,12 +93,15 @@ func WatchKeys(
...
 			currentIndex = queryMeta.LastIndex
+			outputPairsStart = time.Now()
 			select {
 			case <-done:
 				return
 			case out <- WatchedKeys{
 				Keys: keys,
 			}:
+
+				outputPairsBlocking.Update(int64(time.Since(outputPairsStart) / time.Millisecond))
```
mpuncel: this'll be interesting

PR #803, opened 2017-03-28
Add TransferReplicaCounts function to consul RC store.

mpuncel:
TransferReplicaCounts() "transfers" a number of replicas from one RC to
another utilizing consul's transaction functionality. This allows
invariants to be maintained about the total of the replica counts of two
RCs even across consul server failures.

An integration test was added to test this functionality since it is
hard to implement a fake consul server that has the full transaction
capability.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,8 +12,8 @@ install:
   - go get github.com/tools/godep
   - go get github.com/kisielk/errcheck
   - rake install
-  - wget https://releases.hashicorp.com/consul/0.5.2/consul_0.5.2_linux_amd64.zip
-  - unzip consul_0.5.2_linux_amd64.zip
+  - wget https://releases.hashicorp.com/consul/0.7.1/consul_0.7.1_linux_amd64.zip
+  - unzip consul_0.7.1_linux_amd64.zip
   - sudo mv consul /usr/bin/
 
 script:
--- a/integration/common-provision.sh
+++ b/integration/common-provision.sh
@@ -46,5 +46,6 @@ cp $GOPATH/bin/p2-exec /usr/local/bin
 
 # Install P2 test dependencies
 sudo yum -y --nogpgcheck localinstall $GOPATH/src/github.com/square/p2/integration/test-deps/*rpm
+sudo yum -y install unzip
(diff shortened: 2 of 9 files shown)


rudle:
Neat! I'm wondering how we can test the TXN rollback functionality…

bwester:
When the old RC count is decremented, it should remove the pod from one of its nodes. When the new RC count is incremented, it should grab one more node and schedule the pod there.

Do they pick the same node?

mpuncel:
@bwester today that is handled by disabling the old RC, so it won't do anything. the new RC just steals a node of its choosing

@rudle I'm not sure how to do that outside of passing in indexes as arguments instead of fetching them within the function. The downside of doing that is it becomes more consul-specific potentially. Also I think as long as the test succeeds, i've proven that the transaction was configured correctly. Testing failures is sort of testing consul itself

rudle:
Right. I didn't mean to suggest a regression test, but I would like to have more confidence that transactions work a little bit like we expect them to! Perhaps we can write some dummy methods that do weird things (mess with indexes, eg) and test that in a lab setting.

bwester:
@mpuncel: So the scope of this change is only to make the paired +/- operations atomic? That seems fine. As you point out, even this new atomic operation will still part of a larger procedure where RCs must be disabled for correctness.

I suppose I was just momentarily excited at the prospect of having an atomic, self-contained "move a node" operation that didn't require any disabling. But that's a larger scope.

mpuncel:
yeah that's outside the scope of this PR, but that could potentially be accomplished by removing the RC label for the node you want to move, and rewriting RC code to prefer scheduling "unowned" nodes over owned ones.

mpuncel:
I started thinking about working this function into the rolling update loop, and realized it should be more flexible to let you add/remove different counts. Updated to have that extra flexibility

rudle:
New changes seem fine. 

I can imagine an implementation where the replica count was one number to prevent an errand client from adding 2 and removing 3. This library function is probably not the place for such a thing.

mpuncel:
I'm going to merge once this passes. the only changes since the last review were dynamically fetching consul in the integration test so we don't need to commit the binary

PR #804, opened 2017-03-29
Make rolling update loop use rcstore's TransferReplicaCounts()

mpuncel:
This makes the updating of RC counts during a rolling update
transactional which means if the p2-master server or consul cluster
faults the invariant of RC count totals will not be violated.

This change notably removes the old check-and-set functionality on
updating RCs, but given that the RC mutation locks are held during a
rolling update we have reasonable confidence that the RC counts were not
mutated underneath the rolling update loop.

### DIFF
--- a/pkg/roll/run_update.go
+++ b/pkg/roll/run_update.go
@@ -244,19 +244,10 @@ func (u *update) rollLoop(podID types.PodID, hChecks <-chan map[types.NodeName]h
 					"nextRemove": nextRemove,
 					"nextAdd":    nextAdd,
 				}).Infof("Adding %d new nodes and removing %d old nodes", nextAdd, nextRemove)
-				if nextRemove > 0 {
-					err = u.rcs.CASDesiredReplicas(u.OldRC, oldNodes.Desired, oldNodes.Desired-nextRemove)
-					if err != nil {
-						u.logger.WithError(err).Errorln("Could not decrement old replica count")
-						break
-					}
-				}
-				if nextAdd > 0 {
-					err = u.rcs.CASDesiredReplicas(u.NewRC, newNodes.Desired, newNodes.Desired+nextAdd)
-					if err != nil {
-						u.logger.WithError(err).Errorln("Could not increment new replica count")
-						break
-					}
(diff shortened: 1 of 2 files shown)


mpuncel:
there's no integration test that covers this but i have run this in a test environment

petertseng:
looks like you are in fact going to have to implement TransferReplicaCounts in the fake store, eh?

rudle:
Oh, this is exciting. That said, I'm fearful enough that I think this should bake in staging for a while. Is it worth creating a frankenbranch? 👹

pkg/store/consul/rcstore/fake_store.go
```diff
@@ -233,5 +233,15 @@ func (s *fakeStore) LockForUpdateCreation(rcID fields.ID, session consul.Session
 }
 
 func (s *fakeStore) TransferReplicaCounts(toRCID fields.ID, replicasToAdd int, fromRCID fields.ID, replicasToRemove int) error {
-	return util.Errorf("TransferReplicaCounts not implemented in the fake store")
+	err := s.AddDesiredReplicas(toRCID, replicasToAdd)
+	if err != nil {
+		return err
+	}
+
+	err = s.AddDesiredReplicas(fromRCID, -replicasToRemove)
```
petertseng: I had half a mind to ask for `return s.AddDesired` but I know this is about symmetry. Shrug.

pkg/roll/run_update.go
```diff
@@ -244,19 +244,10 @@ func (u *update) rollLoop(podID types.PodID, hChecks <-chan map[types.NodeName]h
...
-						break
-					}
-				}
-				if nextAdd > 0 {
-					err = u.rcs.CASDesiredReplicas(u.NewRC, newNodes.Desired, newNodes.Desired+nextAdd)
-					if err != nil {
-						u.logger.WithError(err).Errorln("Could not increment new replica count")
-						break
-					}
+				err = u.rcs.TransferReplicaCounts(u.NewRC, nextAdd, u.OldRC, nextRemove)
```
petertseng: interesting change - now zeroes are explicitly CASed to the same value instead of taking no action! I think it is an improvement.

PR #805, opened 2017-03-30
Make grpc WatchMatches() implementation retry after connection loss.

mpuncel:
The raw consul applicator implementation of WatchMatches() continually
makes HTTP requests to consul to fetch the latest labels and then
applies the selector in question. This scheme is tolerant of broken or
failed connections to the server due to retries.

The gRPC implementation did not preserve this self-healing behavior if
the connection to the server dies: it would make one RPC call initially
and then loop reading values fromt he stream. This commit retries the
RPC call if the stream is broken for any reason.

### DIFF
--- a/pkg/grpc/labelstore/client/client.go
+++ b/pkg/grpc/labelstore/client/client.go
@@ -5,6 +5,8 @@ onto a grpc server call
 package client
 
 import (
+	"time"
+
 	label_protos "github.com/square/p2/pkg/grpc/labelstore/protos"
 	"github.com/square/p2/pkg/labels"
 	"github.com/square/p2/pkg/logging"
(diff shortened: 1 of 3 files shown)


mpuncel:
I tested this by:

1) starting the server in `bin/grpc/clients/label-store-server`
2) starting the client in `bin/grpc/clients/label-store-client`
3) terminating the server
4) watched the client retry every 2 seconds
5) started the server
6) watched the client reconnect
7) wrote some label-selector-matching data to consul and saw the client display it

pkg/grpc/labelstore/client/client.go
```diff
@@ -51,15 +59,31 @@ func (c Client) WatchMatches(selector klabels.Selector, labelType labels.Type, q
...
+					innerCtx, innerCancel = context.WithCancel(outerCtx)
+
+					time.Sleep(2 * time.Second)
+					watchClient, err = c.labelStoreClient.WatchMatches(innerCtx, &label_protos.WatchMatchesRequest{
+						LabelType: labelTypeToProtoLabelType(labelType),
+						Selector:  selector.String(),
+					})
+					if err != nil {
+						c.logger.WithError(err).Errorln("could not restart WatchMatches RPC, will retry")
+						innerCancel()
```
rudle: Safe to call this more than once?

pkg/grpc/labelstore/client/client.go
```diff
@@ -51,15 +59,31 @@ func (c Client) WatchMatches(selector klabels.Selector, labelType labels.Type, q
...
-				return
+				innerCancel()
+				c.logger.WithError(err).Errorln("unexpected error reading from WatchMatches stream, starting another RPC")
+
+				watchClient = nil
+
+				for watchClient == nil {
+					innerCtx, innerCancel = context.WithCancel(outerCtx)
+
+					time.Sleep(2 * time.Second)
```
rudle: Is there a best practice guide from G here? I'd expect to see an exponential vs linear backoff.

I also wonder if we need a global timeout.

mpuncel: We certainly do want an exponential backoff, however calling `WatchMatches` does not actually initiate a TCP connection to a server. `grpc.Dial()` (the output of which is the `grpc.ClientConn` we get in the `Client` constructor) handles reestablishing TCP connections and it uses exponential backoff. All i'm trying to prevent here is a CPU-sucking busy-loop.

mpuncel: I'm conflicted about a global timeout. If such a timeout were to expire, the only action we can take to signal to the caller that things are broken is to close the output channel, at which point I would expect it to just reinitialize another watch. Do we think that would handle more different error cases?

rudle: Call it an aversion to loops with non-deterministic exit conditions. I don't think such a strategy would handle any cases that are not yet handled.

mpuncel: I share your aversion, although i'm not sure what's worse. Options are:

1) make this retry forever
2) add a case where the output channel might be closed before the caller closes `quitCh`. All callers across implementations (e.g. direct consul access) would have to be modified to detect this and start another `WatchMatches`

pkg/grpc/labelstore/client/client.go
mpuncel: yeah i'll put back the comment that said that

mpuncel: it does get re-assigned on line 74 though so it shouldn't be called more than once.

PR #806, opened 2017-04-03
Add a new -w/--write flag to p2-norm

mcpherrinm:
This causes the output to be written to the file it was read from.  An error is
returned if p2-norm is reading from stdin and is passed -w / --write

### DIFF
--- a/bin/p2-norm/main.go
+++ b/bin/p2-norm/main.go
@@ -32,10 +32,10 @@ func main() {
 	var err error
 	output := os.Stdout
 	if *filename == "" || *filename == "-" {
-		data, err = ioutil.ReadAll(os.Stdin)
 		if *write {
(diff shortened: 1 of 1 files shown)


mpuncel:
one nit: lgtm if you've tested it

bin/p2-norm/main.go
```diff
@@ -29,10 +30,17 @@ func main() {
 
 	var data []byte
 	var err error
+	output := os.Stdout
 	if *filename == "" || *filename == "-" {
 		data, err = ioutil.ReadAll(os.Stdin)
+		if *write {
```
mcpherrinm: done

mcpherrinm:
Yeah I just used this on a bunch of manifests. (both with and without -w and --write)

PR #807, opened 2017-04-04
[WIP] WatchKeys is sometimes latent

rudle:
the <-rcKeyWatch block here is potentially slow and upstream code blocks on
writing to this error channel

I am open to an all encompassing discussion about the sanity of error channels.

### DIFF
--- a/pkg/rc/farm.go
+++ b/pkg/rc/farm.go
@@ -133,6 +133,17 @@ func (rcf *Farm) mainLoop(quit <-chan struct{}) {
 
 	rcKeyWatch, rcErr := rcf.rcStore.WatchRCKeysWithLockInfo(subQuit, rcf.rcWatchPauseTime)
 
+	go func(errCh <-chan error) {
+		for {
+			select {
+			case <-quit:
+				return
+			case err := <-rcErr:
+				rcf.logger.WithError(err).Errorln("Could not read consul replication controllers")
+			}
+		}
+	}(rcErr)
+
 START_LOOP:
 	for {
 		// Check the quit channel independently of the others before entering a multi-channel select.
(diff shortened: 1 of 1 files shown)


pkg/rc/farm.go
```diff
@@ -133,6 +133,15 @@ func (rcf *Farm) mainLoop(quit <-chan struct{}) {
 
 	rcKeyWatch, rcErr := rcf.rcStore.WatchRCKeysWithLockInfo(subQuit, rcf.rcWatchPauseTime)
 
+	go func(errCh chan error) {
+		select {
```
mpuncel: i'm guessing you want a `for { ... }` around this

rudle: ofc

pkg/rc/farm.go
```diff
@@ -133,6 +133,15 @@ func (rcf *Farm) mainLoop(quit <-chan struct{}) {
 
 	rcKeyWatch, rcErr := rcf.rcStore.WatchRCKeysWithLockInfo(subQuit, rcf.rcWatchPauseTime)
 
+	go func(errCh chan error) {
```
mpuncel: may as well put a direction on the channel here

pkg/rc/farm.go
```diff
@@ -133,6 +133,17 @@ func (rcf *Farm) mainLoop(quit <-chan struct{}) {
 
 	rcKeyWatch, rcErr := rcf.rcStore.WatchRCKeysWithLockInfo(subQuit, rcf.rcWatchPauseTime)
 
+	go func(errCh <-chan error) {
+		for {
+			select {
+			case <-quit:
+				return
+			case err := <-rcErr:
```
mpuncel: don't think it matters but just noticed this is using the variable from outer scope rather than the arg

PR #809, opened 2017-04-04
Allow tuning of request rate by WatchService

mpuncel:
WatchService performs a repeated list on v1/kv/health/{serviceID} using
a wait index. This means that in periods of little change for the
service's health, we will be getting a request about every minute
(because watches expire after 5 minutes even if the data doesn't
change).

However, if the data is changing frequently, the upper bound on QPS is
1/second. In the case of daemon sets for which there might be thousands
of health keys, it's quite easy to end up in a scenario where health is
changing very frequently and QPS is close to that upper bound.

This commit adds a watchDelay parameter to WatchService which guarantees
that no two List() operations will be initiated within that interval.


pkg/ds/farm.go
```diff
@@ -581,8 +581,12 @@ func (dsf *Farm) spawnDaemonSet(
 
 	if dsf.monitorHealth {
 		go func() {
-			aggregateHealth := replication.AggregateHealth(ds.PodID(), *dsf.healthChecker)
-			ticks := time.NewTicker(1 * time.Minute)
+			// daemon sets are deployed to many servers and deploys are very slow. This means that
+			// 1) the data size for each health response is huge which can suck up bandwidth
+			// 2) the metrics aren't going to change very frequently, so a low sample rate is acceptable
+			watchDelay := 1 * time.Minute
```
mpuncel: i suppose i should make this configurable?

rudle: Yeah, we may need to increase this delay as the number of daemon sets increases.

mpuncel:
@rudle i'd appreciate another look since i changed some things. when you last looked I hadn't propagated the settings through the replicator/replication structs far enough and additionally the farms now have configurable settings

mpuncel:
I've tested this in action therefore i'm confident enough to merge

PR #810, opened 2017-04-05
servicebuilder: chpst -P (run in own pgrp)

petertseng:
An analogue to https://github.com/square/prodeng/pull/54

### DIFF
--- a/pkg/runit/servicebuilder.go
+++ b/pkg/runit/servicebuilder.go
@@ -93,7 +93,7 @@ func (s ServiceTemplate) runScript() ([]byte, error) {
 $stderr.reopen(STDOUT)
 require 'yaml'
 sleep %d
-exec *YAML.load(DATA.read)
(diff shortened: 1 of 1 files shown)


pkg/runit/servicebuilder.go
```diff
@@ -124,7 +124,7 @@ func (s ServiceTemplate) logScript() ([]byte, error) {
 	ret := fmt.Sprintf(`#!/usr/bin/ruby
 require 'yaml'
 sleep %d
-exec *YAML.load(DATA.read)
```
mpuncel: i wonder if that first part should be `filepath.Basename(DefaultChpst)` or something like that

petertseng:
We must use caution when deploying this. So I'll not merge it yet.

petertseng:
let's hold off on this one, and do #814 instead. Don't want to do servicebuilder business yet. Feel free to revive if the servicebuilder business gets sorted.

PR #811, opened 2017-04-07
ds: spawn child when acquiring lock on updated DS

petertseng:
If we have just acquired a lock on a daemon set, necessarily it is not
one of our children. Therefore, any attempt to send on its updated
channel will trigger a nil dereference. This code has never worked.

Attached test demonstrates, though "Attempt to acquire unchanged daemon
sets" of #813 alleviates the problem slightly. With #813, the attached
test only sometimes dereferences nil because sometimes the lock is still
held when the second farm attempts to access and the daemon set moves to
the "unchanged" or "same" category.

A similar mistake was fixed in #753.

### DIFF
--- a/pkg/ds/farm_test.go
+++ b/pkg/ds/farm_test.go
@@ -1007,6 +1007,96 @@ func TestRelock(t *testing.T) {
 	assertLabel("node3", "first farm quit")
 }
 
+func TestDieAndUpdate(t *testing.T) {
+	retryInterval = testFarmRetryInterval
+
+	dsStore := dsstoretest.NewFake()
+	consulStore := consultest.NewFakePodStore(make(map[consultest.FakePodStoreKey]manifest.Manifest), make(map[string]consul.WatchResult))
+	applicator := labels.NewFakeApplicator()
+
+	preparer := consultest.NewFakePreparer(consulStore, logging.DefaultLogger)
+	preparer.Enable()
+	defer preparer.Disable()
+
+	session := consultest.NewSession()
+
+	allNodes := []types.NodeName{"node1", "node2", "node3"}
+	happyHealthChecker := fake_checker.HappyHealthChecker(allNodes)
+
+	mkFarm := func(logName string) (chan<- struct{}, <-chan struct{}) {
+		farm := &Farm{
+			dsStore:   dsStore,
+			dsLocker:  dsStore,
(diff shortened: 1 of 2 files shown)


petertseng:
```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x20 pc=0x489523]
goroutine 443 [running]:
panic(0x8e3a80, 0xc420016120)
	/home/travis/.gimme/versions/go1.7.4.linux.amd64/src/runtime/panic.go:500 +0x1a1
github.com/square/p2/pkg/ds.(*Farm).handleDSChanges(0xc420164180, 0xc420158048, 0x1, 0x1, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, ...)
	/home/travis/gopath/src/github.com/square/p2/pkg/ds/farm.go:291 +0x1ba3
github.com/square/p2/pkg/ds.(*Farm).mainLoop(0xc420164180, 0xc42001ca20)
	/home/travis/gopath/src/github.com/square/p2/pkg/ds/farm.go:243 +0x565
github.com/square/p2/pkg/ds.TestDieAndUpdate.func1.1(0xc420164180, 0xc42001ca20, 0xc42001d200)
	/home/travis/gopath/src/github.com/square/p2/pkg/ds/farm_test.go:956 +0x61
created by github.com/square/p2/pkg/ds.TestDieAndUpdate.func1
	/home/travis/gopath/src/github.com/square/p2/pkg/ds/farm_test.go:958 +0x404
```

In case it isn't obvious: `Updated` is assuming all updated DSes have already previously been locked by this farm. This is an incorrect assumption:

* farm F1 locks DS D1.
* F1 crash.
* D1 updated.
* F2 perceives that D1 is updated.

F2 hasn't added D1 to its children, but it should.

petertseng:
Test is the same, now with a fix.

pkg/ds/farm_test.go
```diff
@@ -1007,6 +1007,96 @@ func TestRelock(t *testing.T) {
 	assertLabel("node3", "first farm quit")
 }
 
+func TestDieAndUpdate(t *testing.T) {
```
petertseng: unfortunately, a lot of duplicate code with that of #813. May be worth deduplicating sometime.

PR #812, opened 2017-04-07
Increase safety of rolling update loop via TransferReplicaCounts().

mpuncel:
Commit feb411da55cd07df531e9614a92d12cfccaaa30d changed the rolling
update loop to batch its RC count changes into a single consul
transaction. However, there was a regression in safety introduced in
that commit which is addressed by this one.

The old code would make separate calls of CASDesiredReplicas for each
RC. That function fetches the RC, ensures that the replica count matches
one passed to the function call, and only then does a CAS operation on
the RC. This prevents a situation where the rolling update loop computes
a RC count change based on a previous count that might have changed by
the time the change is applied.

This commit makes the TransferReplicaCounts() function take similar
parameters: StartingToReplicas and StartingFromReplicas. The function
will not apply the requested update unless the counts on the ToRC and
FromRC respectively match the passed values.

### DIFF
--- a/integration/single-node-slug-deploy/check.go
+++ b/integration/single-node-slug-deploy/check.go
@@ -434,7 +434,19 @@ func verifyTransferReplicas(errCh chan<- error, tempdir string, logger logging.L
 		return
 	}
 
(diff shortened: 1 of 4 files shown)


mpuncel:
agreed, it's certainly awkard

PR #814, opened 2017-04-10
p2-exec: Setpgid to create new process group

petertseng:
An alternative to https://github.com/square/prodeng/pull/54 and #810.
The two above would require a coordinated rollout, while this change to
p2-exec can be rolled out without coordination.

### DIFF
--- a/bin/p2-exec/main.go
+++ b/bin/p2-exec/main.go
@@ -114,6 +114,24 @@ func main() {
 	if err != nil {
 		log.Fatal(err)
 	}
+
+	/*
+		from `man setpgid`:
+		* "If pid is zero, then the call applies to the current process" / "if pid is 0, the process ID of the calling process shall be used"
+		* "Also, if pgid is 0, the process ID of the indicated process shall be used."
+		So we set the current process's process group ID to be equal to that of its process ID.
+		Without this, its process group ID is equal to its parent's.
+		This means that operations that act on their entire process tree will affect not only their own pod,
+		but all pods on the same machine (in a typical setup, where they all originate from the same parent).
+
+		For example, a renice may change the entire parent process tree's nice level.
+		Worse, this new nice level may propagate into a container, causing the container to be unable to set its nice level back to 0.
+		This has been observed to cause containers to fail to start.
+	*/
+	if err := unix.Setpgid(0, 0); err != nil {
+		log.Fatal(err)
+	}
+
 	err = syscall.Exec(binPath, *cmd, os.Environ())
 	// should never be reached
 	if err != nil {


bin/p2-exec/main.go
```diff
@@ -114,6 +114,12 @@ func main() {
 	if err != nil {
 		log.Fatal(err)
 	}
+
+	// setpgid: zeroes for current process ID
+	if err := unix.Setpgid(0, 0); err != nil {
```
petertseng: Relying on:

* "If pid is zero, then the call applies to the current process"/"if pid is 0, the process ID of the calling process shall be used" (appears on both Linux's and BSD's man page for setpgid)
* "Also, if pgid is 0, the process ID of the indicated process shall be used." (appears only on Linux's man page)

That second bullet point is an interesting question - maybe we need to be explicit, then?

eam: Behavior ought to be the same on FreeBSD. From the [BSD manpage](https://www.freebsd.org/cgi/man.cgi?sektion=2&query=setpgrp) I see:

```
If pgrp is zero,	then the process id of the process
specified by pid is used instead.
```

bin/p2-exec/main.go
```diff
@@ -114,6 +114,12 @@ func main() {
 	if err != nil {
 		log.Fatal(err)
 	}
+
+	// setpgid: zeroes for current process ID
```
stouset: Nit: commentary below should be added to the code, so someone looking 3 years from now knows why this decision was made! Also we should describe why this is being done in the first place.

bin/p2-exec/main.go
```diff
@@ -114,6 +114,24 @@ func main() {
...
+		So we set the current process's process group ID to be equal to that of its process ID.
+		Without this, its process group ID is equal to its parent's.
+		This means that operations that act on their entire process tree will affect not only their own pod,
+		but all pods on the same machine (in a typical setup, where they all originate from the same parent).
+
+		For example, a renice may change the entire parent process tree's nice level.
+		Worse, this new nice level may propagate into a container, causing the container to be unable to set its nice level back to 0.
+		This has been observed to cause containers to fail to start.
+	*/
+	if err := unix.Setpgid(0, 0); err != nil {
```
rudle: I understand that this will start every process exec'd by p2-exec in its own pgroup, thus giving us certain desirable properties. 

I do not know of any verification that's been performed in the wild to verify that this sort of thing is safe. Suppose, for instance, an app is using p2-exec for its environment directory abilities and expecting (probably implicitly) that the result of this command will share a pgroup with their main app. Said another way, p2-exec is used in several places not only in runit /run files.

petertseng:
In the interests of promoting preventative care and in consideration of the necessity of this for certain to-be-deployed pods, we're merging soon.

It's true that we don't have complete knowledge of how p2-exec is being used and whether anyone implicitly depends on being in the same process group as the parent.

I surmise that few things are, because either the shell is changing their pgroup already (if it is being run from an interactive shell), or otherwise they were opening themselves up to odd behaviour anyway since the pgroup included the entire tree.

Nevertheless, rolling this out deserves much caution. The industry has seen stranger things with people depending on behaviours that are ultimately incorrect.

PR #815, opened 2017-04-10
Remove replication controller watchers from roll farm.

mpuncel:
As a result of splitting up the giant preemptive *.Store interfaces, it
became clear that the rolling update store was passing a store that
watches RCs just to instantiate a ReplicationController struct which
wouldn't even use that parameter.

Now that it is completely unused, we can delete it from the roll farm
completely.

### DIFF
--- a/bin/p2-rctl/main.go
+++ b/bin/p2-rctl/main.go
@@ -121,13 +121,10 @@ func main() {
 		baseClient: client,
 		// rcStore copied so many times right now but might not all be
 		// the same implementation of these various interfaces in
-		// the future. Also rcWatcher isn't even used it is simply
-		// used by the roll farm to initialie an RC for a computation
-		// that probably doesn't need to be defined on RC.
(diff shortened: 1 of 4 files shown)


bin/p2-rctl/main.go
```diff
@@ -117,17 +117,11 @@ func main() {
 
 	rcStore := rcstore.NewConsul(client, labeler, 3)
 	rctl := rctlParams{
-		httpClient: httpClient,
-		baseClient: client,
-		// rcStore copied so many times right now but might not all be
```
petertseng: you could probably keep these three lines if you want, an donly delete the `rcWatcher` parts

mpuncel: oh man how'd that happen

petertseng: Now you left too much in I think, since the comment still mentions rcWatcher!

PR #818, opened 2017-04-12
Mpuncel/ds grpc client

mpuncel:


pkg/grpc/daemonsetstore/daemon_set_store.go
```diff
@@ -104,10 +118,33 @@ func rawDSToProtoDS(rawDS fields.DaemonSet) (*daemonsetstore_protos.DaemonSet, e
 	}, nil
 }
 
+func ProtoDSToRawDS(protoDS *daemonsetstore_protos.DaemonSet) (fields.DaemonSet, error) {
```
mpuncel: @rudle would be sweet to codegen stuff like this

PR #820, opened 2017-04-17
Hook Linting

rudle:
Do not review yet.

### DIFF
--- a/pkg/hooks/hooks_test.go
+++ b/pkg/hooks/hooks_test.go
@@ -30,7 +30,7 @@ func TestExecutableHooksAreRun(t *testing.T) {
 	// So PodFromPodHome doesn't bail out, write a minimal current_manifest.yaml
 	ioutil.WriteFile(path.Join(podDir, "current_manifest.yaml"), []byte("id: my_hook"), 0755)
 
-	hooks := Hooks(os.TempDir(), pods.DefaultPath, &logging.DefaultLogger)
+	hooks := NewContext(tempDir, pods.DefaultPath, &logging.DefaultLogger)
 	pod, err := pods.PodFromPodHome("testNode", podDir)
 	Assert(t).IsNil(err, "the error should have been nil")
 	hooks.runHooks(tempDir, AfterInstall, pod, testManifest(), logging.DefaultLogger)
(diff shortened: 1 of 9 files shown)


rudle:
OK. This is ready for review. No functionality should be changed by this.

The commits are not super clean, I'm sorry about that.

pkg/hooks/hooks_test.go
```diff
@@ -30,10 +30,10 @@ func TestExecutableHooksAreRun(t *testing.T) {
 	// So PodFromPodHome doesn't bail out, write a minimal current_manifest.yaml
 	ioutil.WriteFile(path.Join(podDir, "current_manifest.yaml"), []byte("id: my_hook"), 0755)
 
-	hooks := Hooks(os.TempDir(), pods.DefaultPath, &logging.DefaultLogger)
+	hooks := NewContext(tempDir, pods.DefaultPath, &logging.DefaultLogger)
```
rudle: This is a curious change!

pkg/hooks/types.go
```diff
@@ -0,0 +1,112 @@
...
+)
+
+// HookType represents the stage in the Hook lifecycle
+type HookType string
+
+func (hookType HookType) String() string {
+	return string(hookType)
+}
+
+var (
```
petertseng: so, while yo'ure moving these - any chance of making these `const`s? Doesn't seem we'd want to change them.

pkg/hooks/types.go
```diff
@@ -0,0 +1,112 @@
...
+		return AfterLaunch, nil
+	case AfterAuthFail.String():
+		return AfterAuthFail, nil
+	default:
+		return HookType(""), fmt.Errorf("%s is not a valid hook type", value)
+	}
+}
+
+// DefaultPath is a directory on disk where hooks are installed by default.
+var DefaultPath = "/usr/local/p2hooks.d"
```
petertseng: should this be const? Or do we want to keep as var to allow override at build-time?

pkg/hooks/types.go
```diff
@@ -0,0 +1,112 @@
...
+	podRoot string
+	logger  *logging.Logger
+}
+
+type hookExecContext struct {
+	Path    string // path to hook's executable
+	Name    string // human-readable name of Hook
+	Timeout time.Duration
+	env     []string // unix environment in which the hook should be executed
+	logger  *logging.Logger
```
petertseng: change to pointer noted

PR #821, opened 2017-04-18
Reapply "Merge pull request #739 from petertseng/gzip-as-user" 

petertseng:
New commit means we are ready for this:

preparer: chmod TempDir on setup

As explained in code comments: If we extract as target user, we need to
allow them to see the tmpdir.

Original:

As we've seen time and time again, it's very difficult to get our custom
tar code right (see #116, #373, #374, #376, #623). We have decided
simply to call tar, dropping privileges to the RunAs user of the pod.

Using tar insulates us from any more edge cases we missed as long as
tar was the tool used to create the tarball, and dropping privileges
lets Unix permissions prevent any unintended system file modifications.

### DIFF
--- a/pkg/preparer/setup.go
+++ b/pkg/preparer/setup.go
@@ -404,6 +404,25 @@ func New(preparerConfig *PreparerConfig, logger logging.Logger) (*Preparer, erro
 		return nil, util.Errorf("Could not create preparer pod directory: %s", err)
 	}
 
+	// Artifact files are downloaded to os.TempDir().
+	// Since we extract artifact files as target user, we must allow them to access the tmpdir.
+	// We expect that there is no sensitive information in TempDir, so 755 is safe, though 711 could be considered.
+	tmpDirStat, err := os.Stat(os.TempDir())
+	if err != nil {
+		return nil, util.Errorf("Could not stat tmpdir: %s", err)
+	}
+	mode := tmpDirStat.Mode()
+	// We don't chmod if the directory is already 755.
+	// Normally there is no harm in doing so,
+	// but on Travis we don't have permission to do so (we don't run as root).
+	if mode&0755 != 0755 {
+		// keep whatever upper bit is there.
+		err = os.Chmod(os.TempDir(), (mode&07000)|0755)
+		if err != nil {
+			return nil, util.Errorf("Could not chmod tmpdir: %s", err)
+		}
+	}
+
 	var logExec []string
 	if len(preparerConfig.LogExec) > 0 {
 		logExec = preparerConfig.LogExec
--- a/pkg/artifact/downloader.go
+++ b/pkg/artifact/downloader.go
@@ -65,13 +65,12 @@ func (l *downloader) Download(location *url.URL, verificationData auth.Verificat
 		return err
 	}
 
-	// rewind a second time to allow the archive to be unpacked
-	_, err = artifactFile.Seek(0, os.SEEK_SET)
+	err = artifactFile.Chmod(0644)
 	if err != nil {
-		return util.Errorf("Could not reset artifact file position after verification: %v", err)
+		return err
 	}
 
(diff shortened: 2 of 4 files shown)


pkg/preparer/setup.go
```diff
@@ -404,6 +404,14 @@ func New(preparerConfig *PreparerConfig, logger logging.Logger) (*Preparer, erro
 		return nil, util.Errorf("Could not create preparer pod directory: %s", err)
 	}
 
+	// Artifact files are downloaded to os.TempDir().
+	// Since we extract artifact files as target user, we must allow them to access the tmpdir.
+	// We expect that there is no sensitive information in TempDir, so 0755 is safe, though 0711 could be considered.
+	err = os.Chmod(os.TempDir(), 0755)
```
petertseng: alternative: Perform this in the downloader.

Advantage: Action is closer to where it is necessary.
Disadvantage: Action performed multiple times.

petertseng: Eh. I'm going to do it in the downloader. Interesting test interactions ("operation can't be nil"), if I have it here, I wonder what that means

petertseng: OK, never mind, happens no matter where it goes.

pkg/preparer/setup.go
```diff
@@ -404,6 +404,22 @@ func New(preparerConfig *PreparerConfig, logger logging.Logger) (*Preparer, erro
...
 
+	// Artifact files are downloaded to os.TempDir().
+	// Since we extract artifact files as target user, we must allow them to access the tmpdir.
+	// We expect that there is no sensitive information in TempDir, so 755 is safe, though 711 could be considered.
+	tmpDirStat, err := os.Stat(os.TempDir())
+	if err != nil {
+		return nil, util.Errorf("Could not stat tmpdir: %s", err)
+	}
+	mode := tmpDirStat.Mode()
+	if mode&0755 != 0755 {
```
petertseng: this is my best guess as to why operation would be nil? If the mode was already 0755? Let's see...

petertseng:
I honestly think https://travis-ci.org/square/p2/builds/223785567 (conditional, but always do 0755, don't preserve high bits) only succeeded because /tmp is already x755, so the test is inconclusive. I'll see what unconditional does though.

petertseng:
OK, unconditional but keep high bitsfails, so conditional is needed.

pkg/preparer/setup.go
```diff
@@ -404,6 +404,22 @@ func New(preparerConfig *PreparerConfig, logger logging.Logger) (*Preparer, erro
...
+		// keep whatever upper bit is there.
+		err = os.Chmod(os.TempDir(), (mode&07000)|0755)
```
petertseng: wondering if anyone thinks this (keeping the highest... triad) is necessary. This was added after the most recent approval, would like a quick look at it.

pkg/preparer/setup.go
petertseng: Experimentally, this is the case, but I cannot reproduce this on my machine?!

pkg/preparer/setup.go
mpuncel: i doubt it will matter either way. can `(mode&07000)|0755` not just be written as `mode | 0755`?

mpuncel: oh that would keep it executable

pkg/preparer/setup.go
petertseng: I'm going to leave a comment saying I can't reproduce it on any machine I have but Travis.

Perhaps Travis has special characteristics on its tmpdir.

petertseng:
I've just realized what happened. "operation not permitted to be nil" is actually "expected **operation not permitted** to be nil". On Travis, we simply can't chmod the tmpdir.

PR #823, opened 2017-04-24
Hook Execution Auditing

rudle:


bin/p2-run-hooks/main.go
```diff
@@ -34,7 +35,20 @@ func main() {
...
+	auditLogger = hooks.NewFileAuditLogger(&logging.DefaultLogger)
+	if *sqlitePath != "" {
+		al, err := hooks.NewSQLiteAuditLogger(*sqlitePath, &logging.DefaultLogger)
+		if err != nil {
+			logging.DefaultLogger.Errorf("Unable to connect sqlite database at %s, printing audit logs to STDOUT %v", *sqlitePath, err)
+		}
+		auditLogger = al
+	}
+	if auditLogger == nil {
+		log.Fatalf("DEBUG: audit logger not available")
```
mpuncel: don't need `f`. also is this if statement necessary? would only happen if one of the constructors returned a `nil` reference and also `nil` error

pkg/hooks/audit_log_file.go
```diff
@@ -0,0 +1,45 @@
+package hooks
+
+import "github.com/square/p2/pkg/logging"
+
+// Returns a FileAuditLogger using the given logger
+func NewFileAuditLogger(logger *logging.Logger) *FileAuditLogger {
```
mpuncel: the word `File` is a bit confusing to me here since it could just be logging to stdout, maybe i'm not unix-y enough (everything is a file)

rudle:
Not implemented here: pruning of the SQLite hook audit logs. I imagine a 24 hour retention is going to be fine, but now is the time to speak up if that seems bad to you.

pkg/hooks/audit_logger_sqlite.go
```diff
@@ -0,0 +1,137 @@
...
+
+type SQLiteAuditLogger struct {
+	sqlite *sql.DB
+	logger *logging.Logger
+}
+
+func NewSQLiteAuditLogger(sqliteDBPath string, logger *logging.Logger) (*SQLiteAuditLogger, error) {
+	db, err := sql.Open("sqlite3", sqliteDBPath)
+	if err != nil {
+		return nil, err
```
mpuncel: probably wrapping this error would be good

pkg/hooks/audit_logger_sqlite.go
```diff
@@ -0,0 +1,137 @@
...
+var (
+	sqliteMigrations = []string{
+		`create table hook_results (
+	    id integer not null primary key autoincrement,
+	    date datetime default current_timestamp,
+	    pod_id text,
+	    pod_unique_key text,
+	    hook_name string,
+	    hook_stage string,
+	      success tinyint);`,
```
mpuncel: would it be useful to make this exit code instead of just a bool?

pkg/hooks/audit_logger_sqlite.go
```diff
@@ -0,0 +1,137 @@
...
+
+var (
+	sqliteMigrations = []string{
+		`create table hook_results (
+	    id integer not null primary key autoincrement,
+	    date datetime default current_timestamp,
+	    pod_id text,
+	    pod_unique_key text,
+	    hook_name string,
+	    hook_stage string,
```
mpuncel: might be better to call this hook_type since that's what it's called in code, although I don't love that word. Maybe we could rename the type?

pkg/hooks/audit_logger_sqlite.go
```diff
@@ -0,0 +1,137 @@
...
+		statement := sqliteMigrations[i]
+		_, err = tx.Exec(statement)
+		if err != nil {
+			return util.Errorf("Could not apply migration %d: %s", i+1, err)
+		}
+	}
+
+	_, err = tx.Exec(updateSchemaVersionStatement, int64(len(sqliteMigrations)))
+	if err != nil {
+		al.logger.WithError(err).Errorln("Could not update schema_version table")
```
mpuncel: maybe `return err` in here to defend against people adding code below here and continuing past this one

pkg/hooks/audit_logger_sqlite.go
```diff
@@ -0,0 +1,137 @@
...
+import (
+	"database/sql"
+
+	_ "github.com/mattn/go-sqlite3"
+	"github.com/square/p2/pkg/logging"
+	"github.com/square/p2/pkg/util"
+)
+
+type SQLiteAuditLogger struct {
+	sqlite *sql.DB
```
mpuncel: it probably doesn't matter that much but there should probably be a `Close()` function to call `sqlite.Close()`

pkg/hooks/audit_logger_sqlite_test.go
```diff
@@ -0,0 +1,48 @@
...
+	"database/sql"
+	"io/ioutil"
+	"path/filepath"
+	"testing"
+
+	"github.com/square/p2/pkg/logging"
+)
+
+func initSQLiteAuditLogger(t *testing.T) (AuditLogger, *sql.DB) {
+	tempDir, err := ioutil.TempDir("", "hooks_audit_log")
```
mpuncel: do you remove this directory anywhere?

pkg/hooks/hooks.go
```diff
@@ -134,21 +138,19 @@ func (h *hookContext) runHooks(dirpath string, hType HookType, pod Pod, podManif
...
+		HookEnvVar:                path.Base(dirpath),
+		HookEventEnvVar:           hType.String(),
+		HookedNodeEnvVar:          pod.Node().String(),
+		HookedPodIDEnvVar:         podManifest.ID().String(),
+		HookedPodHomeEnvVar:       pod.Home(),
+		HookedPodManifestEnvVar:   tmpManifestFile.Name(),
+		HookedConfigPathEnvVar:    path.Join(pod.ConfigDir(), configFileName),
+		HookedEnvPathEnvVar:       pod.ConfigDir(),
+		HookedConfigDirPathEnvVar: h.podRoot,
+		HookedSystemPodRootEnvVar: pod.UniqueKey().String(),
```
mpuncel: looks like you lost `HookedPodUniqueKeyEnvVar`

pkg/hooks/hooks_test.go
```diff
@@ -118,6 +120,37 @@ func TestHookRunWithTimeout(t *testing.T) {
...
+func TestHookAuditLogging(t *testing.T) {
+	tempDir, err := ioutil.TempDir("", "hook")
+	Assert(t).IsNil(err, "the error should have been nil")
+	defer os.RemoveAll(tempDir)
+
+	podDir, err := ioutil.TempDir("", "pod")
+	defer os.RemoveAll(podDir)
+	Assert(t).IsNil(err, "the error should have been nil")
+
+	ioutil.WriteFile(path.Join(tempDir, "test1"), []byte("#!/bin/sh\necho $HOOKED_POD_ID > $(dirname $0)/output"), 0755)
```
mpuncel: check for errors here and below

pkg/hooks/hooks_test.go
```diff
@@ -118,6 +120,37 @@ func TestHookRunWithTimeout(t *testing.T) {
...
+	auditLoggerLogger := logging.TestLogger()
+	buf := &bytes.Buffer{}
+	auditLoggerLogger.Logger.Out = buf
+
+	hooks := NewContext(tempDir, pods.DefaultPath, &logging.DefaultLogger, NewFileAuditLogger(&auditLoggerLogger))
+	pod, err := pods.PodFromPodHome("testNode", podDir)
+	Assert(t).IsNil(err, "the error should have been nil")
+	hooks.runHooks(tempDir, AfterInstall, pod, testManifest(), logging.DefaultLogger)
+
+	fmt.Printf("\n\n%v\n\n", buf)
```
mpuncel: remove?

pkg/preparer/setup.go
```diff
@@ -439,11 +441,25 @@ func New(preparerConfig *PreparerConfig, logger logging.Logger) (*Preparer, erro
 		}
 		hooksPodFactory := pods.NewHookFactory(filepath.Join(preparerConfig.PodRoot, "hooks"), preparerConfig.NodeName)
 		hooksPod = hooksPodFactory.NewHookPod(hooksManifest.ID())
+		hooksSqlite, ok := hooksManifest.GetConfig()["sqlite_path"]
+		if ok {
+			sqlitePath := hooksSqlite.(string)
+			err := os.MkdirAll(sqlitePath, os.ModeDir)
+			if err == nil {
+				logger.Errorf("Unable to construct a SQLite based audit-logger. Using file backed instead: %v", err)
```
mpuncel: you could consider using `WithError(err)` instead of putting it in the message

mpuncel:
super neat!

bin/p2-run-hooks/main.go
rudle: Right. This was some debug code.

pkg/hooks/audit_log_file.go
rudle: Do you prefer FileAuditLogger or LoggerAuditLogger? Maybe there's a non-derpy name for this thing that somehow implies that it logs auditfully to a logger.

mpuncel: probably FileAuditLogger is preferable LOL. i don't have a better suggestion so 👍

pkg/hooks/audit_logger_sqlite.go
rudle: Good idea!

cuzelac:
> I imagine a 24 hour retention is going to be fine, but now is the time to speak up if that seems bad to you.

If there's room, can you bump it to at least 3 days?  That would cover us in the case of something happening on the weekend and not getting around to checking it until Monday.

pkg/hooks/audit_logger_sqlite.go
rudle: Hmmmm, this is tricky. Not all references to the audit logger have an exit code, though it does seem nice to keep it around.

It's also the case that there's no platform independent way to get the exit code of an `os/exec.Cmd` which makes this request a bit cumbersome. Here's a citation for that http://stackoverflow.com/questions/10385551/get-exit-code-go

pkg/hooks/audit_logger_sqlite.go
```diff
@@ -0,0 +1,137 @@
...
+	podUniqueKey := env.HookedPodUniqueKeyEnvVar
+	hookName := ctx.Name
+	hookStage := env.HookEventEnvVar
+	// Boolean values are represented by TINYINTs in SQL. Adapt.
+	dbSuccess := 0
+	if !success {
+		dbSuccess = 1
+	}
+
+	_, err := al.sqlite.Exec(stmt, podID, podUniqueKey, hookName, hookStage, dbSuccess)
```
mpuncel: one thing to be aware of, i believe the behavior of sqlite is that writes lock the entire database, so if there are concurrent hook finishes or pruning operations happening these writes will fail. you could consider a retry here with a little backoff or something like that, or maybe we're just okay with lossiness for now

mpuncel: or you could use an in process mutex since the preparer is the only thing that should be doing writes

rudle: Surprising! I'd expect writes to block, not fail. I'll do some research to confirm this one way or the other.

rudle: https://www.sqlite.org/faq.html#q3
> (5) When any process wants to write, it must lock the entire database file for the duration of its update. But that normally only takes a few milliseconds. Other processes just wait on the writer to finish then continue about their business. 

So it claims to work as I expect. Given that you've worked with sqlite in production before, maybe you have some more information that I don't.

pkg/hooks/audit_logger_sqlite.go
rudle: Added. I guess I should call it somewhere!

pkg/hooks/audit_logger_sqlite.go
mpuncel: I've definitely observed failed writes via the CLI, i just tested by running `while true; do sqlite3 foo.db "insert into foo values(1)"; done` in one terminal window and `for i in $(seq -w 1000); do echo $i; sqlite3 foo.db "insert into heys values($i)"; done` in another, and saw some failed writes and missing sequential numbers.

anecdotally i've seen the go client do the same thing, where a runit finish process fails to save an exit code to the database

pkg/hooks/audit_logger_sqlite.go
rudle: foiled by interfaces, I am. I'll have to see about wiring this up to this implementations life-cycle.

pkg/hooks/audit_logger_sqlite.go
rudle: done

rudle:
This does pruning but still no retries. PTAL at the pruning implementation.

rudle:
Tests are finally green. PTAL!

pkg/hooks/audit_logger_sqlite.go
```diff
@@ -0,0 +1,143 @@
...
+success
+) VALUES(?,?,?,?,?)`
+	podID := env.HookedPodIDEnvVar
+	podUniqueKey := env.HookedPodUniqueKeyEnvVar
+	hookName := ctx.Name
+	hookStage := env.HookEventEnvVar
+	// Boolean values are represented by TINYINTs in SQL. Adapt.
+	dbSuccess := 0
+	if !success {
+		dbSuccess = 1
```
petertseng: so I guess this is going to be fine if we're consistent about it (?) but this will be in contravention of https://sqlite.org/datatype3.html#boolean_datatype - we have 1 false and 0 true. Makes sense with exit codes though.

pkg/hooks/audit_logger_sqlite.go
```diff
@@ -0,0 +1,143 @@
...
+	}
+
+	_, err := al.sqlite.Exec(stmt, podID, podUniqueKey, hookName, hookStage, dbSuccess)
+	if err != nil {
+		al.logger.Errorln("nope: %v", err)
```
petertseng: is the nope necessary? remove, or replace with a better?

pkg/hooks/audit_logger_sqlite.go
```diff
@@ -0,0 +1,143 @@
...
+
+	tx, err := al.sqlite.Begin()
+	if err != nil {
+		return util.Errorf("Could not start transaction for migrations: %s", err)
+	}
+
+	defer func() {
+		if err == nil {
+			// return the commit error by assigning to return variable
+			err = tx.Commit()
```
petertseng: this works?! I don't see `(err error)` in the function sig

mpuncel:
looking forward to this! good work

pkg/hooks/hooks_test.go
```diff
@@ -118,6 +119,39 @@ func TestHookRunWithTimeout(t *testing.T) {
...
+	podDir, err := ioutil.TempDir("", "pod")
+	defer os.RemoveAll(podDir)
+	Assert(t).IsNil(err, "the error should have been nil")
+
+	err = ioutil.WriteFile(path.Join(tempDir, "test1"), []byte("#!/bin/sh\necho $HOOKED_POD_ID > $(dirname $0)/output"), 0755)
+	Assert(t).IsNil(err, "Caught error while writing test hook")
+
+	// So PodFromPodHome doesn't bail out, write a minimal current_manifest.yaml
+	ioutil.WriteFile(path.Join(podDir, "current_manifest.yaml"), []byte("id: my_hook"), 0755)
+	Assert(t).IsNil(err, "Caught error while writing test hook")
```
petertseng: manifest not hook

pkg/hooks/audit_logger_sqlite.go
```diff
@@ -2,12 +2,16 @@ package hooks
...
 	"database/sql"
+	"fmt"
 
 	_ "github.com/mattn/go-sqlite3"
 	"github.com/square/p2/pkg/logging"
 	"github.com/square/p2/pkg/util"
+	"github.com/square/p2/pkg/util/param"
 )
 
+var auditLogTTL = param.Int("audit_log_sqlite_ttl", 259200)
```
petertseng: can I get a comment on how long this is in larger units? for the human!

pkg/hooks/audit_logger_sqlite.go
```diff
@@ -106,6 +112,27 @@ func (al *SQLiteAuditLogger) Close() error {
 	return al.sqlite.Close()
 }
 
+func (al *SQLiteAuditLogger) withRetries(f func() error, n int) error {
+	var err error
+	err = f()
```
petertseng: is there something that makes it so that this duplicate of 124-128 can't be removed? Code is the same in the `for` loop! Would just need to start attempts at 0.

rudle:
Thanks. It seems as though there are some permission issues on bootstrapping the sqlite DB so expect one more commit.

pkg/hooks/audit_logger_sqlite.go
petertseng: removed in latre commit

pkg/hooks/audit_logger_sqlite.go
mpuncel: oh yeah might not

pkg/hooks/audit_logger_sqlite.go
```diff
@@ -0,0 +1,184 @@
...
+	"database/sql"
+	"fmt"
+
+	_ "github.com/mattn/go-sqlite3"
+	"github.com/square/p2/pkg/logging"
+	"github.com/square/p2/pkg/util"
+	"github.com/square/p2/pkg/util/param"
+)
+
+var auditLogTTL = param.Int("audit_log_sqlite_ttl", 259200)
```
petertseng: I see Github thinks https://github.com/square/p2/pull/823#discussion_r113801096 is out of date but I say it still applies!

pkg/hooks/audit_logger_sqlite.go
```diff
@@ -0,0 +1,184 @@
...
+)
+
+// Close will terminate this AuditLogger. Re-establishing the connection is not supported, use the constructor.
+func (al *SQLiteAuditLogger) Close() error {
+	return al.sqlite.Close()
+}
+
+func (al *SQLiteAuditLogger) withRetries(f func() error, n int) error {
+	var err error
+	err = f()
```
petertseng: I see Github thinks https://github.com/square/p2/pull/823#discussion_r113801665 is out of date but I say it still applies!

pkg/hooks/audit_logger_sqlite.go
rudle: I don't understand:
`func (tx *Tx) Commit() error`

pkg/hooks/audit_logger_sqlite.go
rudle: comments go stale, math is forever.

pkg/hooks/audit_logger_sqlite.go
rudle: In the future this might represent exit code, the primary reason it doesn't today is because it's complex.

pkg/hooks/audit_logger_sqlite.go
rudle: ohhhhh, I understand this comment now.

PR #828, opened 2017-04-27
Parallelize sending updates to each watcher in consul aggregator.

mpuncel:
The ConsulAggregator is designed to allow performing "watches" on a
label selector for a certain label type, while only actually performing
one watch routine per process. The aggregator maintains a linked list of
"watchers" which represent a particular selector, and every time the
overall label watch returns new values, the list is iterated and each
watcher's selector is applied to the full dataset, with the results
being sent to the watcher's channel.

In practice, if the dataset is large (e.g. ~1.6Mb) and the number of
watchers is high (e.g. ~1700), it can take a noticeable amount of CPU
time in order to process a watch value. This is bad because a mutex
guarding access to the aggregator's linked list is held while these
computations are being performed.

This mutex must also be acquired when registering or deregistering
watchers. It's been observed in production that registering and
deregistering 1700 watchers all at the same time under the workload
described above can take several minutes due to this mutex contention.

This commit aims to mitigate the problem by parallelizing the selector
computations for each watcher. It does not attempt to make any
architectural changes to the aggregator, although there certainly are
some that could be made, for instance using a map instead of a linked
list for faster deregistration which requires a full scan of the linked
list in the worst case.

One test had to be slightly modified because it depended on the lack of
parallelism, but the intent of the test to make sure a bad client cannot
block others was preserved.

### DIFF
--- a/pkg/labels/consul_aggregator_test.go
+++ b/pkg/labels/consul_aggregator_test.go
@@ -28,7 +28,10 @@ func fakeLabeledPods() map[string][]byte {
 func TestTwoClients(t *testing.T) {
 	alterAggregationTime(100 * time.Millisecond)
 
-	fakeKV := &fakeLabelStore{fakeLabeledPods(), nil}
+	fakeKV := &fakeLabelStore{
+		data:         fakeLabeledPods(),
+		watchTrigger: nil,
+	}
 	aggreg := NewConsulAggregator(POD, fakeKV, logging.DefaultLogger, metrics.NewRegistry())
 	go aggreg.Aggregate()
 	defer aggreg.Quit()
@@ -67,7 +70,10 @@ func TestTwoClients(t *testing.T) {
 func TestQuitAggregateAfterResults(t *testing.T) {
 	alterAggregationTime(100 * time.Millisecond)
 
-	fakeKV := &fakeLabelStore{fakeLabeledPods(), nil}
+	fakeKV := &fakeLabelStore{
+		data:         fakeLabeledPods(),
+		watchTrigger: nil,
+	}
 	aggreg := NewConsulAggregator(POD, fakeKV, logging.DefaultLogger, metrics.NewRegistry())
 	go aggreg.Aggregate()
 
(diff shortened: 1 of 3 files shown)


pkg/labels/consul_aggregator_test.go
```diff
@@ -181,7 +181,7 @@ func TestIgnoreIndividualWatch(t *testing.T) {
 	Assert(t).AreEqual(int64(2), aggreg.metWatchCount.Value(), "should currently have two watchers")
 
 	// iterate twice to show that we are not waiting on other now-closed channels
```
petertseng: change comment?

pkg/labels/consul_aggregator_test.go
```diff
@@ -192,7 +192,15 @@ func TestIgnoreIndividualWatch(t *testing.T) {
 		}
 	}
 
-	Assert(t).AreEqual(aggreg.metWatchSendMiss.Value(), int64(1), "should have missed exactly one watch send")
+	// this is a range because sending matches is parallelized, so by the
+	// time we've read a value 3 values from labeledChannel2 we don't know
```
petertseng: remove "a value" ?

anthonybishopric:
FWIW I'm not sure there's a good reason to use a linked list anymore - the use case of adding and removing thousands of watches at a time was not in the original thought process, although it probably should have been.

mpuncel:
I agree that a linked list probably isn't the best choice anymore, but I actually suspect that traversing the list doesn't wind up holding the mutex for very long compared to time spent in `sendMatches()`, so it's probably not terribly high priority for us to change it. linked list traversal is only really happening when de-registering a listener, which in practice will only happen when a client of the grpc label store hangs up (i.e. doesn't happen that much in steady state).

@petertseng to answer your question, this was causing our gRPC store to cease to function for many minutes after a client hung up due to lock contention. about 1700 listeners would need to be deregistered, and the slow loop in `sendMatches()` was preventing the mutex from being acquired, which meant matches were being sent to listeners long after they had quit

pkg/labels/consul_aggregator_test.go
mpuncel: what needs to change about it?

mpuncel: oh i was looking at a different place with an identical comment :)

mpuncel: hmm it seems like your comment appears both within the diff and here but mine only shows up in one or the other! anyway you're right i was looking at a different place with that identical comment

PR #830, opened 2017-05-02
gzip: add test against escaping symlink vulnerability

petertseng:
I checked that if we revert #821 and/or #739, the newly-added test fails.

This shows that our old tar extraction code was vulnerable and that the
new extraction code is not. Any extraction code we write or use should
defend against this vulnerability.


pkg/gzip/gzip_test.go
```diff
@@ -65,3 +65,13 @@ func TestPathWithParent(t *testing.T) {
 		Assert(t).IsTrue(os.IsNotExist(err), "expected extracted file not to exist")
 	})
 }
+
+func TestEscapingSymlink(t *testing.T) {
+	testExtraction(t, "escaping_symlink.tar.gz", func(tarErr error, dest string) {
+		// UNCOMMENT:
```
petertseng: sorry

petertseng:
I'll consider adding the tests of #376, but this is simpler and shows the same problem.

PR #831, opened 2017-05-02
Mpuncel/fix races

mpuncel:


pkg/ds/daemon_set_test.go
```diff
@@ -418,12 +444,37 @@ func TestPublishToReplication(t *testing.T) {
...
 	desiresErrCh := ds.WatchDesires(quitCh, updatedCh, deletedCh)
 	dsChangesErrCh := watchDSChanges(ds, dsStore, quitCh, updatedCh, deletedCh)
 
+	// We need to be careful when shutting down in order to avoid data
+	// races, since updatedCh and deletedCh are shared between two
+	// functions (WatchDesires() and watchDSChanges()). We need to make
+	// sure that we don't close updatedCh or deletedCh until their
+	// respective output channels have closed which indicates that they
+	// won't try to send any more values
+	defer func() {
```
mpuncel: bonus: i think this is the cause of the `panic: send on closed channel` flake we have

mpuncel:
it seems travis is catching even more races! so don't review until i have a green build :)

mpuncel:
tests pass! this is ready for review

petertseng:
Haven't yet looked at whether there is something doing this, but perhaps continuing to run the racy tests without the race detector?

pkg/replication/replication.go
```diff
@@ -255,7 +259,7 @@ func (r *replication) Enact() {
 
 				go func() {
 					defer close(exitCh)
-					err = r.updateOne(node, timeoutCh, aggregateHealth)
```
rudle: I wonder if there's a linter we can turn on for "this goroutine captures free vars!" It's a dangerous property.

pkg/ds/daemon_set_test.go
```diff
@@ -104,16 +104,17 @@ func watchDSChanges(
 				errCh <- util.Errorf("Error occured when watching daemon set changes: %v", watched.Err)
 			}
 
+			dsID := ds.ID()
```
rudle: I'm curious: does this change fix a race? I can't figure out why it would.

mpuncel: it doesn't fix a race, but theoretically the ID could change between `ID()` calls (although in practice it wouldn't).

I think this just came from me playing with the solution to this race for a while and one incarnation of it didn't involve locking inside of the `ID()` function

PR #833, opened 2017-05-03
Mpuncel/audit logs

mpuncel:


pkg/store/consul/auditlogstore/consul_store.go
```diff
@@ -0,0 +1,128 @@
...
+}
+
+func NewConsulStore(consulKV ConsulKV) ConsulStore {
+	return ConsulStore{
+		consulKV: consulKV,
+	}
+}
+
+func (ConsulStore) Create(
+	txn *api.KVTxnOps,
```
mpuncel: i'm eager for feedback on this way of doing things. The workflow will look sort of like a mysql transaction: you start a transaction, apply operations on it, and then commit. Of course in this case it's all local rather than on the server

rudle: I'm interested in how the entity specific functions will look, for example will we need to change the signatures of the CRUD functions in RC stores?

mpuncel: I think we'll have 3 options:

1) modify the existing functions (e.g. AddDesiredReplicas) to create their own transactions and handle audit logging, maybe the store has a configuration option to turn this off and on.
2) create new functions alongside the old that do this
3) pass in transaction to AddDesiredReplicas() which makes the caller responsible for its audit logs

mpuncel: there are probably some combinations of those options but point is, i'm not sure exactly how it will look yet!

mpuncel: now that i think about it, passing in transactions is probably nice. I suppose the stores can still handle creating appropriate audit records, but if things accept transaction arguments it's easy to string together arbitrary actions into a single transaction

mpuncel:
I added a commit that makes the rollstore use transactions to create rolling updates, greatly reducing the amount of cleanup code required to handle partial failures and also reducing the number of consul calls. I haven't added audit logging to the rollstore, that's coming briefly

mpuncel:
I think i'm happy with this approach, but this PR is a weird hodgepodge of functionality, so i'm going to close it. I want to work on a level of indirection for consul transactions as well

PR #834, opened 2017-05-04
gzip: Always extract with --no-same-owner

petertseng:
This preserves the behaviour present before #739 / #821, for apps that
run_as root, since the custom gzip code would explicitly chown those
files to root.

So this can be considered as a fix to a regression caused by #821.

As code comment states: this helps apps that run_as root ensure that
their files are in fact owned by root rather than some unknown user
dictated by the build system.

### DIFF
--- a/pkg/gzip/gzip.go
+++ b/pkg/gzip/gzip.go
@@ -32,7 +32,11 @@ func ExtractTarGz(owner string, filename string, dest string) (err error) {
 		return util.Errorf("error setting ownership of root directory %s: %s", dest, err)
 	}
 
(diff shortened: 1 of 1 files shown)


petertseng:
updating commit message:

This preserves the behaviour present before #739 / #821, for apps that
run_as root, since the custom gzip code would explicitly chown those
files to root.

So this can be considered as a fix to a regression caused by #821.

petertseng:
I checked on CentOS6, CentOS7, and OS X (fun fact: it's not in the man page on OS X, but still works!!!)

PR #835, opened 2017-05-05
Mpuncel/roll txn

mpuncel:


pkg/store/consul/rcstore/fake_store.go
```diff
@@ -62,6 +63,10 @@ func (s *fakeStore) Create(manifest manifest.Manifest, nodeSelector labels.Selec
 	return entry.RC, nil
 }
 
+func (s *fakeStore) CreateTxn(txn *transaction.Tx, manifest manifest.Manifest, nodeSelector labels.Selector, podLabels labels.Set) (fields.RC, error) {
+	panic("transactions not implemented in fake rc store")
```
rudle: hmmmm, does this mean we don't have test coverage over the Helper functions in rollstore/consul_store.go?

pkg/store/consul/transaction/transaction.go
```diff
@@ -0,0 +1,125 @@
...
+	if !ok {
+		return util.Errorf("an unknown error occurred when applying the transaction")
+	}
+
+	return nil
+}
+
+func (c *Tx) runCommitHooks() {
+	for _, f := range c.commitHooks {
+		f()
```
mpuncel: I'd expect the logging to be done within the function since the most context will be available there

pkg/store/consul/rcstore/fake_store.go
mpuncel: i should add a comment explaining this. The decision I've made is that if a test uses transactions, it will use a real consul store instead of one of these fake storage layers. It would be complicated to implement the same transaction semantics ourselves.

I'm not sure exactly what helper functions you're asking about but all of the tests in integration_test.go use transactions against a real consul instance (i.e. they don't use this fake store)

rudle: got it. I don't know what to call the functions like `CreateRollingUpdateFromOneMaybeExistingWithLabelSelector`, that's what I was wondering about.

PR #836, opened 2017-05-05
bli/ready-file

brianrli:
ready-file flag

### DIFF
--- a/bin/p2-exec/main.go
+++ b/bin/p2-exec/main.go
@@ -32,6 +32,7 @@ var (
 		"Specifies an extra environment KEY=VALUE pair to set for the process. May be used multiple times. Takes precedence over --env if there are conflicts",
 	).StringMap()
 	launchableName = kingpin.Flag("launchable", "The key in $PLATFORM_CONFIG_PATH containing the cgroup parameters.").Short('l').String()
+	requireFile    = kingpin.Flag("require-file", "Check for the presence of a required file before execing its argument").String()
 	cgroupName     = kingpin.Flag("cgroup", "The name of the cgroup that should be created.").Short('c').String()
 	nolim          = kingpin.Flag("nolimit", "Remove rlimits.").Short('n').Bool()
 	clearEnv       = kingpin.Flag("clearenv", "Clear all environment variables before loading envDir(s).").Bool()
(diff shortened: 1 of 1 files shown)


rudle:
I can't remember: did we decide it was OK to change the command line invocation of `p2-exec` in our runit files?

mpuncel:
I do believe it's okay, and we've done it plenty of times in the past e.g. by adding --extra-env

bin/p2-exec/main.go
```diff
@@ -32,6 +32,7 @@ var (
 		"Specifies an extra environment KEY=VALUE pair to set for the process. May be used multiple times. Takes precedence over --env if there are conflicts",
 	).StringMap()
 	launchableName = kingpin.Flag("launchable", "The key in $PLATFORM_CONFIG_PATH containing the cgroup parameters.").Short('l').String()
+	readyFile      = kingpin.Flag("ready-file", "Checking for the presence of a ready-to-run file before execing its argument").String()
```
mpuncel: s/Checking/Check

bin/p2-exec/main.go
```diff
@@ -110,6 +111,12 @@ func main() {
 		}
 	}
 
+	if *readyFile != "" {
+		if _, err := os.Stat(*readyFile); os.IsNotExist(err) {
```
bwester: Should the service start if `Stat()` returns, say, an I/O error?

bin/p2-exec/main.go
```diff
@@ -32,6 +32,7 @@ var (
 		"Specifies an extra environment KEY=VALUE pair to set for the process. May be used multiple times. Takes precedence over --env if there are conflicts",
 	).StringMap()
 	launchableName = kingpin.Flag("launchable", "The key in $PLATFORM_CONFIG_PATH containing the cgroup parameters.").Short('l').String()
+	readyFile      = kingpin.Flag("ready-file", "Check for the presence of a ready-to-run file before execing its argument").String()
```
bwester: nit: Calling this a "ready file" seems strange. To me, without the context you have right now, the name doesn't convey the meaning of what it does. I would suggest something more straightforward that doesn't introduce a new term:
```
  --require-file FILENAME
      If given, require the presence of the named file before running the command.
```

bwester:
Oh, and (nit) please structure your commit message according to GitHub's and git's [recommended format](https://github.com/erlang/otp/wiki/Writing-good-commit-messages). Keep the title separate from the message.

PR #837, opened 2017-05-05
read ready file from preparer config

brianrli:


rudle:
This seems correct. You'll probably want to refer to this config value in servicebuilder.go's runscript() function so - we refer to `p2-exec` there. Of course, see my comment on #836 about the viability of changing what this file looks like as it is shared by the servicebuilder implementation in square/prodeng.

bin/p2-preparer/main.go
```diff
@@ -34,6 +34,12 @@ func main() {
 		logger.WithError(err).Fatalln("invalid parameter")
 	}
 
+	if preparerConfig.ReadyFile != "" {
+		if _, err := os.Stat(preparerConfig.ReadyFile); os.IsNotExist(err) {
+			logger.WithError(err).Fatalln("ready file not present")
```
mpuncel: Is the intent of this change to make the preparer fail to operate if the ready file is not present? As an aside, we try to give thorough commit messages explaining the intent and motivation for each change so questions like this are easy to answer when looking at old code

brianrli: yes, if I am understanding the design doc correctly, it fails if the ready file is not present

petertseng: 🤔 

I don't think we wanted the **preparer** to fail to start though. I think we wanted to make the preparer read from this config to tell p2-exec what file to check

brianrli:
preparer will now fail to start if the --require-file is not present

bin/p2-exec/main.go
```diff
@@ -110,28 +111,16 @@ func main() {
...
-	if err != nil {
-		log.Fatal(err)
+	if *requireFile != "" {
+		if _, err := os.Stat(*requireFile); os.IsNotExist(err) {
+			log.Fatalf("ready file not present: %v\n", err)
+		}
 	}
 
-	/*
-		from `man setpgid`:
```
petertseng: I am confused. why was this removed?

brianrli: not supposed to--i did something dumb but have since fixed it

bin/p2-preparer/main.go
```diff
@@ -44,6 +44,12 @@ func main() {
 		defer statusServer.Close()
 	}
 
+	if preparerConfig.RequireFile != "" {
```
petertseng: I don't think we want these six lines at all, though I could be wrong.

Because the p2-preparer is (in our deployment) started by p2-exec (which already checks the required file), it doesn't seem useful for p2-preparer to do a separate check. Or are we explicitly trying to support other environments in which p2-preparer isn't started by p2-exec?

brianrli: oh, I didn't know that p2-exec starts p2-preparer

then, in the design doc, what is the point of having the p2-preparer reading from config at all? If it's just going to pass through p2-exec?

petertseng: In my understanding, it was so that p2-exec would not need to have the required file hard-coded. That goal has been achieved in this PR, I believe: in this PR, p2-exec doesn't have the required file hard-coded, and it only knows about it by being passed the --required-file flag. The p2-preparer writes the appropriate servicebuilder/runit files so that when p2-exec runs, it has the right --required-file flag.

This was merely a suspicion that we wouldn't want to be too opinionated in open-source code. If we didn't care, it might have been acceptable to have the required file be a constant in p2-exec, and have p2-preparer be completely unaware of it. It seems that the way this PR has it (have it in p2-preparer's config) is the way we decided on, though.

petertseng:
wondering about the closing of #836 - it seems like it would have been acceptable for the addition of --require-file to p2-exec to be one commit, and then the p2-preparer business in a second commit.

bin/p2-exec/main.go
```diff
@@ -110,6 +111,12 @@ func main() {
 		}
 	}
 
+	if *requireFile != "" {
+		if _, err := os.Stat(*requireFile); os.IsNotExist(err) {
```
rudle: Please log the other error cases. I don't those sorts of things should crash `p2-exec`, but I could be convinced.

pkg/hooks/hook_env.go
```diff
@@ -74,12 +74,12 @@ func (h *HookEnv) PodFromDisk() (*pods.Pod, error) {
 		return nil, err
 	}
 
-	return pods.PodFromPodHome(types.NodeName(node), podHome)
+	return pods.PodFromPodHome(types.NodeName(node), podHome, "")
```
rudle: This pattern is coming up a lot. Consider creating a new constructor that doesn't require the caller to know the suitable default value. When we want a pod factor with a require file, we can use this 3 argument constructor.

integration/single-node-slug-deploy/check.go
```diff
@@ -1098,3 +1104,22 @@ func startLabelStoreServer(dir string) error {
...
+	if err != nil {
+		return "", err
+	}
+
+	err = os.Chmod(dir, 0777) // make file executable by all.
+	if err != nil {
+		return "", err
+	}
+
+	err = os.Chmod(f.Name(), 0777) // make file executable by all.
```
petertseng: so, any chance I could get you to use 666 here instead of 777?

brianrli: # **fixed!**

bin/p2-preparer/main.go
mpuncel: we clarified in person but for the record we decided we DO want the preparer to crash here. This is because when a machine is initially boot strapped it will be launched _without_ the require file flag in p2-exec

integration/single-node-slug-deploy/check.go
```diff
@@ -74,12 +74,17 @@ func main() {
 		log.Fatalln("Could not create temp directory, bailing")
 	}
 
+	requireFile, err := createRequireFile(tempdir)
+	if err != nil {
+		log.Fatalln("Could not create temp require file, bailing")
```
mpuncel: include the error in this message

petertseng:
never mind it wasn't added in this PR, its size was just increased

whoaaaaaa

PR #841, opened 2017-05-10
Mpuncel/audit logs

mpuncel:


pkg/grpc/auditlogstore/audit_log_store.go
```diff
@@ -0,0 +1,88 @@
...
+
+func New(auditLogStore AuditLogStore, logger logging.Logger, txner transaction.Txner) audit_log_protos.P2AuditLogStoreServer {
+	return store{
+		auditLogStore: auditLogStore,
+		logger:        logger,
+		txner:         txner,
+	}
+}
+
+func (s store) List(_ context.Context, _ *audit_log_protos.ListRequest) (*audit_log_protos.ListResponse, error) {
```
rudle: Forgive the naive question: why don't we care about the ListRequest argument?

mpuncel: right now we just always list all of the records, so the list request doesn't inform the server how to do anything. we might care about it if we want to set a max number of records or something like that

pkg/store/consul/auditlogstore/consul_store.go
```diff
@@ -0,0 +1,126 @@
...
+}
+
+func NewConsulStore(consulKV ConsulKV) ConsulStore {
+	return ConsulStore{
+		consulKV: consulKV,
+	}
+}
+
+func (ConsulStore) Create(
+	txn *transaction.Tx,
```
rudle: Call it premature generality, but I wonder about replacing the consul transaction type with a context.Context and storing the TXN using the .WithValue function. It's more abstract and might make testing or otherwise using this code easier. The drawback is pulling the TXN out of the object is slightly clumsy and we lose compiler legibility wrt the TXN being provided to all functions of this type.

PR #842, opened 2017-05-11
DEFENSIVE MIGRATION SYNTAX

rudle:
This will support us in case the hook_schema_versions table is faulty.


rudle:
Thanks. Merging on green.

PR #843, opened 2017-05-15
Make all functions take a context.Context instead of transaction.Tx

mpuncel:
This commit changes function signatures for consul store operations to
take a context.Context instead of a *transaction.Tx. This means we're
providing a more standard interface and additionally callers can specify
deadlines or cancellation signals for applying a transaction.

### DIFF
--- a/pkg/store/consul/rcstore/consul_store.go
+++ b/pkg/store/consul/rcstore/consul_store.go
@@ -1,6 +1,7 @@
 package rcstore
 
 import (
+	"context"
 	"encoding/json"
 	"errors"
 	"fmt"
@@ -130,8 +131,11 @@ func (s *ConsulStore) Create(manifest manifest.Manifest, nodeSelector klabels.Se
 }
 
 // TODO: replace Create() with this
-func (s *ConsulStore) CreateTxn(txn *transaction.Tx, manifest manifest.Manifest, nodeSelector klabels.Selector, podLabels klabels.Set) (fields.RC, error) {
-	rc, err := s.innerCreateTxn(txn, manifest, nodeSelector, podLabels)
+func (s *ConsulStore) CreateTxn(ctx context.Context, manifest manifest.Manifest, nodeSelector klabels.Selector, podLabels klabels.Set) (fields.RC, error) {
+	rc, err := s.innerCreateTxn(ctx, manifest, nodeSelector, podLabels)
+	if err != nil {
+		return fields.RC{}, err
+	}
 
 	// TODO: measure whether retries are is important in practice
 	for i := 0; i < s.retries; i++ {
@@ -196,7 +200,7 @@ func (s *ConsulStore) innerCreate(manifest manifest.Manifest, nodeSelector klabe
 }
 
 // TODO: replace innerCreate() with this function
-func (s *ConsulStore) innerCreateTxn(txn *transaction.Tx, manifest manifest.Manifest, nodeSelector klabels.Selector, podLabels klabels.Set) (fields.RC, error) {
+func (s *ConsulStore) innerCreateTxn(ctx context.Context, manifest manifest.Manifest, nodeSelector klabels.Selector, podLabels klabels.Set) (fields.RC, error) {
 	id := fields.ID(uuid.New())
 	rcp, err := s.rcPath(id)
 	if err != nil {
@@ -217,7 +221,7 @@ func (s *ConsulStore) innerCreateTxn(txn *transaction.Tx, manifest manifest.Mani
 		return fields.RC{}, util.Errorf("Could not marshal RC as json: %s", err)
 	}
 
-	err = txn.Add(api.KVTxnOp{
+	err = transaction.Add(ctx, api.KVTxnOp{
 		Verb:  api.KVCAS,
 		Key:   rcp,
 		Value: jsonRC,
--- a/pkg/store/consul/rollstore/integration_test.go
+++ b/pkg/store/consul/rollstore/integration_test.go
@@ -3,6 +3,7 @@
 package rollstore
 
 import (
+	"context"
 	"encoding/json"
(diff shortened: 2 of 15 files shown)


pkg/store/consul/transaction/transaction.go
```diff
@@ -11,90 +12,86 @@ import (
...
-func (c *Tx) Commit(txner Txner) error {
-	if c.alreadyCommitted {
-		return ErrAlreadyCommitted
+// Commit attempts to run all of the kv operations in the context's
+// transaction. The cancel function with which the context was created must
+// also be passed to guarantee that the transaction won't be applied twice
+func Commit(ctx context.Context, cancel context.CancelFunc, txner Txner) error {
+	txn, err := getTxnFromContext(ctx)
+	if err != nil {
+		return nil
```
rudle: Is this right?

pkg/store/consul/transaction/transaction.go
```diff
@@ -11,90 +12,86 @@ import (
...
 	}
 
-	c.runCommitHooks()
-	// Set this after we know err == nil because otherwise it could have
+	txn.runCommitHooks()
+
+	// Run this after we know err == nil because otherwise it could have
 	// been a retry-able TCP error for example
-	c.alreadyCommitted = true
+	cancel()
```
rudle: I can't figure out why this doesn't have a conditional around it. It seems like cancellation should be done only in case of error.

pkg/store/consul/transaction/transaction.go
mpuncel: nope!

mpuncel: will write a test case for this

pkg/store/consul/transaction/transaction.go
mpuncel: It's the opposite, i'm using `cancel()` here to mark the transaction as completed so it can't be used again. This is enforced by checking `<-ctx.Done()` in `getTxnFromContext()`.

I don't want to cancel if there was an error, because the caller might want to retry the error (e.g. if consul temporarily unavailable, don't want to go through the work of rebuilding the whole transaction)

mpuncel: I'll put this in a comment

rudle: alright, cool.

mpuncel: maybe it's an anti-pattern to use cancelation in this way, I could go back to tracking an `alreadyCommitted bool` on the transaction

pkg/store/consul/transaction/transaction.go
```diff
@@ -123,3 +124,23 @@ func txnErrorsToString(errors api.TxnErrors) string {
...
+func getTxnFromContext(ctx context.Context) (*Tx, error) {
+	select {
+	case <-ctx.Done():
+		return nil, ErrAlreadyCommitted
+	default:
+	}
+
+	txnValue := ctx.Value(contextKey)
+	if txnValue == nil {
+		return nil, util.Errorf("no transaction was opened on the passed Context")
```
petertseng: you define an ErrNoTransaction above, but then do not use it

mpuncel: removed because we're never going to check for that type of error, it indicates that the code is just written wrong!

pkg/store/consul/transaction/transaction.go
petertseng: I think this seems consistent with the use of Cancel. https://golang.org/pkg/context/#WithCancel has told me

> Canceling this context releases resources associated with it, so code should call cancel as soon as the operations running in this Context complete.

So in doing this we are in keeping with that advice - we cancel when the operation is complete. Right?

pkg/store/consul/rcstore/consul_store.go
```diff
@@ -130,8 +131,11 @@ func (s *ConsulStore) Create(manifest manifest.Manifest, nodeSelector klabels.Se
 }
 
 // TODO: replace Create() with this
-func (s *ConsulStore) CreateTxn(txn *transaction.Tx, manifest manifest.Manifest, nodeSelector klabels.Selector, podLabels klabels.Set) (fields.RC, error) {
-	rc, err := s.innerCreateTxn(txn, manifest, nodeSelector, podLabels)
+func (s *ConsulStore) CreateTxn(ctx context.Context, manifest manifest.Manifest, nodeSelector klabels.Selector, podLabels klabels.Set) (fields.RC, error) {
+	rc, err := s.innerCreateTxn(ctx, manifest, nodeSelector, podLabels)
+	if err != nil {
```
petertseng: I suppose these three lines would hvae been good to have in the pre-context version of this code too

pkg/store/consul/rollstore/integration_test.go
```diff
@@ -428,7 +429,7 @@ func TestCreateRollingUpdateFromOneMaybeExistingWithLabelSelectorFailsWhenConfli
 	time.Sleep(1 * time.Millisecond)
 
 	// Second one should fail
-	txn2 := transaction.New()
+	txn2, cancelFunc := transaction.New(context.Background())
```
petertseng: surprised that there is no complaint about "cancelFunc is assigned but never used" or whatever it is. Is this being run?

mpuncel: hmm interesting

mpuncel: it must be because there's a variable named `cancelFunc` above that is referenced

mpuncel:
Yeah context is basically a wrapper for transaction. It's conceivable we'd attach things like metrics registries to it, and it also allows cancelling and freeing up of resources acquired during building of the transaction e.g. the RC locks in the roll store

PR #846, opened 2017-05-17
Add client library wrapper for audit log store grpc operations.

mpuncel:
This will allow clients to operate using the "raw" audit log types
rather than the primitive types that are ncessarily used by generated
protobuf code.

### DIFF
--- a/pkg/grpc/auditlogstore/client/client.go
+++ b/pkg/grpc/auditlogstore/client/client.go
@@ -1,0 +1,52 @@
+package client
+
+import (
+	"context"
+
+	"github.com/square/p2/pkg/audit"
+	"github.com/square/p2/pkg/grpc/auditlogstore"
+	auditlogstore_protos "github.com/square/p2/pkg/grpc/auditlogstore/protos"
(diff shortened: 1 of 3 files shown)


pkg/grpc/auditlogstore/client/client.go
```diff
@@ -0,0 +1,52 @@
...
+}
+
+func (c Client) Delete(ctx context.Context, idsToDelete []audit.ID) error {
+	req := &auditlogstore_protos.DeleteRequest{}
+	for _, id := range idsToDelete {
+		req.AuditLogIds = append(req.AuditLogIds, id.String())
+	}
+
+	_, err := c.client.Delete(ctx, req)
+	if err != nil {
```
petertseng: this could be `return err`

mpuncel: I'm just gonna merge cuz i don't want to wait for CI. I also sort of prefer writing it this way because then if anything was ever added after `Delete()` these lines wouldn't need to change

PR #847, opened 2017-05-18
store/consul: Add MutatePod

petertseng:
This allows for transactionally transforming a number of pod manifests
in some way. A first use case is to replace a signature.

### DIFF
--- a/pkg/store/consul/kv.go
+++ b/pkg/store/consul/kv.go
@@ -4,6 +4,7 @@ package consul
 
 import (
 	"bytes"
+	"context"
 	"encoding/json"
 	"fmt"
 	"strings"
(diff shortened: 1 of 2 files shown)


pkg/store/consul/kv.go
```diff
@@ -226,6 +228,59 @@ func (c consulStore) DeletePod(podPrefix PodPrefix, nodename types.NodeName, pod
 	return writeMeta.RequestTime, nil
 }
 
+// TransformPod mutates the input context in such a way
+// that it will transactionally transform the pod manifest on the given nodes.
+// If the pod does not exist on some of the given nodes,
+// the operation skips those nodes without considering that an error.
+func (c consulStore) TransformPod(
```
mpuncel: nit: we've used the word `Mutate` elsewhere e.g. https://github.com/square/p2/blob/master/pkg/store/consul/pcstore/consul_store.go#L230

rudle:
Sweet! How did you test this?

petertseng:
Now called Mutate and with included test

petertseng:
ds test flaked on me:

```
--- FAIL: TestPublishToReplication (1.03s)
	daemon_set_test.go:501: Unexpectedly no error when no nodes are eligible
```

But it's OK, that was my first commit not my second. So I can still merge.

PR #849, opened 2017-05-19
Mpuncel/ru deletion event

mpuncel:


pkg/roll/farm.go
```diff
@@ -374,7 +394,32 @@ func (rlf *Farm) validateRoll(update roll_fields.Update, logger logging.Logger)
 
 // Tries to delete the given RU every second until it succeeds
 func (rlf *Farm) mustDeleteRU(id roll_fields.ID, logger logging.Logger) {
-	for err := rlf.rls.Delete(id); err != nil; err = rlf.rls.Delete(id) {
+	ctx, cancelFunc := transaction.New(context.Background())
+	defer cancelFunc() // technically not necessary once Commit() succeeds
+	err := rlf.rls.Delete(ctx, id)
+	if err != nil {
+		// this error is really bad because we can't recover from it
```
rudle: Can we retry this operation? It seems like there's nothing else on the context at this point, so we could create a new one and try the rollstore.Delete again.

mpuncel: the only scenarios in which `Delete()` would error are 1) if computing the key based on the ID fails, or 2) if there are more than 64 operations on the transaction. If either of those occurred here retrying isn't going to help

mpuncel: maybe i should retry as a matter of hygiene though

PR #850, opened 2017-05-22
Mpuncel/ds contention

mpuncel:


pkg/ds/farm.go
```diff
@@ -502,83 +504,73 @@ func (dsf *Farm) releaseLock(unlocker consulutil.Unlocker) {
...
+
+	allDaemonSets, err := dsStore.List()
+	if err != nil {
+		return ds_fields.DaemonSet{}, false, util.Errorf("listing daemon sets failed so could not check for contention: %s", err)
 	}
 
 	// If this daemon set has a node selector set to Everything, check the labels
 	// of other daemon sets
-	for _, child := range dsf.children {
+	for _, otherDS := range allDaemonSets {
```
rudle: d'oh

bin/p2-dsctl/main.go
```diff
@@ -129,7 +131,14 @@ func main() {
 			log.Fatalf("Error occurred: %v", err)
 		}
 
-		ds, err := dsstore.Create(manifest, minHealth, name, selector, podID, *createTimeout)
+		ctx, cancelFunc := transaction.New(context.Background())
+		defer cancelFunc()
+		ds, err := dsstore.Create(ctx, manifest, minHealth, name, selector, podID, *createTimeout)
```
rudle: Is this for the future? What's the advantage of a transaction with only one operation?

mpuncel: right now the advantage is the deferred creation of the daemon set. `Create()` creates a `DeamonSet` which is required to pass to `DSContends`, but the problem is i don't want the creation to succeed until i know there isn't contention.

I agree that there are ways to accomplish this without using a transaction but i figured we probably want this to be transactional soon anyway, i think audit log records for daemon set creations will be nice

mpuncel:
added a newline to the end of the "checking for contention" message

PR #853, opened 2017-05-23
pc/control: Only accept/use session in Create; p2-pcctl: Only create session when creating a PC

petertseng:
pc/control: Only accept/use session in Create

The constructors were all taking a session but only using them in
Create. Might as well only take in Create.

History: The session argument used to be used in Mutate as well, but
that was removed in #524. After this point, Create is the only user of a
session, so it can be the only function that takes one.

---

p2-pcctl: Only create session when creating a PC

Thus, if someone does not have permission to create a session but can
still read from the PC store, they will still be able to use all
non-Create operations, which seems desirable.

### DIFF
--- a/bin/p2-pcctl/main.go
+++ b/bin/p2-pcctl/main.go
@@ -111,7 +111,7 @@ func main() {
 		cn := fields.ClusterName(*createName)
 		podID := types.PodID(*createPodID)
 		selector := defaultSelector(az, cn, podID)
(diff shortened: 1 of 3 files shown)


petertseng:
--- FAIL: TestPublishToReplication (1.06s)
	daemon_set_test.go:501: Unexpectedly no error when no nodes are eligible

PR #855, opened 2017-05-24
Add pod cluster identifying info to RUCompletionEventDetails

mpuncel:
Previously audit log records that were created to mark the cancellation
or successful completion of a rolling update did not contain information
to identify the pod cluster the RU belonged to. This commit adds that
information because audit log systems will likely want to be able to
look up this information based on pod cluster.

### DIFF
--- a/pkg/audit/ru.go
+++ b/pkg/audit/ru.go
@@ -3,6 +3,7 @@ package audit
 import (
 	"encoding/json"
 
+	"github.com/square/p2/pkg/labels"
 	"github.com/square/p2/pkg/manifest"
 	pc_fields "github.com/square/p2/pkg/pc/fields"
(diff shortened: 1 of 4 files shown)


pkg/audit/ru.go
```diff
@@ -60,17 +64,31 @@ func NewRUCreationEventDetails(
...
 
+type Labeler interface {
+	GetLabels(labelType labels.Type, id string) (labels.Labeled, error)
+}
+
 func NewRUCompletionEventDetails(
 	rollingUpdateID roll_fields.ID,
 	succeeded bool,
 	canceled bool,
+	labeler Labeler,
```
mpuncel: i could also consider putting this function on the audit log store and making the audit log store require a labeler but felt it's slightly too early to that

PR #856, opened 2017-05-26
Return a valid grpc response when audit log deletion successful

mpuncel:


mpuncel:
added a test and fixed a race condition

mpuncel:
by fixing the mutex copy i introduced some test deadlocks! now they should be fixed

PR #858, opened 2017-05-26
Fix a panic case in WatchPodStatus() grpc handler.

mpuncel:
There was a race where a goroutine might panic sending on a closed
channel. This commit gives responsibility of closing the channel
to the goroutine writing to it.


mpuncel:
good call @petertseng ! fixed it so `innerQuit` will always be closed

PR #861, opened 2017-05-30
Do not consider a missing implicit entry point an error.

mpuncel:
This change is intended to cut down on error noise in p2-preparer. The
hoist artifact spec historically required all "entry points" for an
artifact to be located at bin/launch whether it's a directory of entry
points or a file. At a later point in time the pod manifest spec was
extended to allow specifying alternate entry points, at which point it
was considered an error if an entry point is missing.

However it is not uncommon for an artifact to have no entry points, for
instance when the intention is just to deploy some scripts to a server
and not run any processes. These cases now generate errors which pollute
preparer logs and exception systems.

This commit tries to take the best of both worlds by considering a
missing entry point to be an error only if it was explicitly enumerated
in the pod manifest. If the implicit default one is missing, it is no
longer an error.

### DIFF
--- a/pkg/hoist/hoist_launchable.go
+++ b/pkg/hoist/hoist_launchable.go
@@ -21,6 +21,17 @@ import (
 	"github.com/square/p2/pkg/util"
 )
 
+type EntryPoints struct {
+	// Paths contains the relative paths to each entry point from the root
+	// of the launchable
+	Paths []string
+	// Implicit means that the entry points were not explicitly specified
+	// in the pod manifest and instead a default was used. This is useful
+	// information since we want to treat a missing entry point as an error
+	// if and only if it was listed explicitly
+	Implicit bool
+}
+
 // A HoistLaunchable represents a particular install of a hoist artifact.
 type Launchable struct {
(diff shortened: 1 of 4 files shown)


pkg/hoist/hoist_launchable.go
```diff
@@ -294,13 +305,19 @@ func (hl *Launchable) Executables(
...
 		absEntryPointPath := filepath.Join(hl.InstallDir(), relativeEntryPoint)
 
 		entryPointInfo, err := os.Stat(absEntryPointPath)
 		if err != nil {
-			return nil, MissingEntryPoints{
-				message: util.Errorf("missing entry point %s: %s", absEntryPointPath, err).Error(),
+			if hl.EntryPoints.Implicit {
+				// We assume the deployer knows what they're doing
+				// and there aren't supposed to be any entry points
+				return nil, nil
```
petertseng: any way `Stat` might return a different sort of error, such as no permission? If so, is it still OK and intentional that an implicit entry point will cause that error to be ignored?

mpuncel: yeah i suppose i should check if it's a not exist error

PR #864, opened 2017-06-05
roll farm: only delete old RC if replica count is zero. otherwise page.

mpuncel:
Previously the RU farm would always zero out the old RC's replica count
and then delete it if the "LeaveOld" flag was not set. This can lead to
bad situations if a math error is encountered when creating an RU
because we might delete an RC that was managing pods and then those pods
become orphaned.

This commit causes the RU farm to only delete the old RC if its replica
count is zero. Otherwise it will send an alert so a human user can
correct the situation.

### DIFF
--- a/pkg/roll/integration_test.go
+++ b/pkg/roll/integration_test.go
@@ -1,15 +1,25 @@
+// +build !race
+
 package roll
 
 import (
+	"errors"
+	"sync"
 	"testing"
 	"time"
 
+	"github.com/square/p2/pkg/alerting"
 	"github.com/square/p2/pkg/audit"
 	"github.com/square/p2/pkg/labels"
 	"github.com/square/p2/pkg/logging"
+	"github.com/square/p2/pkg/manifest"
+	"github.com/square/p2/pkg/roll/fields"
 	"github.com/square/p2/pkg/store/consul/auditlogstore"
 	"github.com/square/p2/pkg/store/consul/consulutil"
+	"github.com/square/p2/pkg/store/consul/rcstore"
 	"github.com/square/p2/pkg/store/consul/rollstore"
+
+	klabels "k8s.io/kubernetes/pkg/labels"
 )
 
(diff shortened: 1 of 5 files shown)


pkg/roll/run_update.go
```diff
@@ -181,15 +188,56 @@ func (u *update) Run(quit <-chan struct{}) (ret bool) {
...
+			// scheduled. If LeaveOld is false, we expect that
+			// all nodes will have been rolled from the old RC to
+			// the new one at this point, so we can delete the
+			// old RC after it isn't managing any nodes anymore.
+			// This error likely needs manual fixing by shifting
+			// the remaining nodes off of this RC and then
+			// deleting it
+			u.logger.Errorln("could not delete old RC because its replica count is nonzero")
+
+			// return nil to avoid looping, the alert
```
petertseng: took me a while. I now see that this means `alerter.Alert` will return nil

pkg/roll/integration_test.go
```diff
@@ -56,3 +66,108 @@ func TestAuditLogCreation(t *testing.T) {
...
+
+	var wg sync.WaitGroup
+	wg.Add(1)
+	go func() {
+		defer wg.Done()
+		update.cleanupOldRC(quit)
+	}()
+
+	<-alertOut
+	<-alertOut
```
petertseng: why two, I wonder?

mpuncel: i wanted to test that it spins until there's a successful alert sent

pkg/roll/run_update.go
mpuncel: i actually violated my own personal style here. usually i prefer checking the error explicitly for clarity. will fix

pkg/roll/integration_test.go
mpuncel: added a clarifying comment

PR #865, opened 2017-06-06
Transactions in orchestrate.go

bunsenmcdubbs:
Please don't approve... just starting a PR to track progress... and solicit help.

### DIFF
--- a/pkg/grpc/auditlogstore/audit_log_store.go
+++ b/pkg/grpc/auditlogstore/audit_log_store.go
@@ -11,7 +11,6 @@ import (
 	"github.com/square/p2/pkg/store/consul/transaction"
 	"github.com/square/p2/pkg/util"
 
-	grpccontext "golang.org/x/net/context"
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/codes"
 )
@@ -38,7 +37,7 @@ func New(auditLogStore AuditLogStore, logger logging.Logger, txner transaction.T
 	}
 }
 
-func (s store) List(_ grpccontext.Context, _ *audit_log_protos.ListRequest) (*audit_log_protos.ListResponse, error) {
+func (s store) List(_ context.Context, _ *audit_log_protos.ListRequest) (*audit_log_protos.ListResponse, error) {
 	records, err := s.auditLogStore.List()
 	if err != nil {
 		return nil, grpc.Errorf(codes.Unavailable, "error listing audit log records: %s", err)
@@ -54,7 +53,7 @@ func (s store) List(_ grpccontext.Context, _ *audit_log_protos.ListRequest) (*au
 	}, nil
 }
 
-func (s store) Delete(ctx grpccontext.Context, req *audit_log_protos.DeleteRequest) (*audit_log_protos.DeleteResponse, error) {
+func (s store) Delete(ctx context.Context, req *audit_log_protos.DeleteRequest) (*audit_log_protos.DeleteResponse, error) {
 	if len(req.GetAuditLogIds()) == 0 {
 		return nil, grpc.Errorf(codes.InvalidArgument, "no audit log IDs were specified for deletion")
 	}
--- a/pkg/grpc/auditlogstore/audit_log_store.go
+++ b/pkg/grpc/auditlogstore/audit_log_store.go
@@ -11,6 +11,7 @@ import (
 	"github.com/square/p2/pkg/store/consul/transaction"
 	"github.com/square/p2/pkg/util"
 
+	grpccontext "golang.org/x/net/context"
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/codes"
 )
@@ -37,7 +38,7 @@ func New(auditLogStore AuditLogStore, logger logging.Logger, txner transaction.T
 	}
 }
 
-func (s store) List(_ context.Context, _ *audit_log_protos.ListRequest) (*audit_log_protos.ListResponse, error) {
+func (s store) List(_ grpccontext.Context, _ *audit_log_protos.ListRequest) (*audit_log_protos.ListResponse, error) {
 	records, err := s.auditLogStore.List()
 	if err != nil {
 		return nil, grpc.Errorf(codes.Unavailable, "error listing audit log records: %s", err)
@@ -53,7 +54,7 @@ func (s store) List(_ context.Context, _ *audit_log_protos.ListRequest) (*audit_
 	}, nil
 }
 
-func (s store) Delete(ctx context.Context, req *audit_log_protos.DeleteRequest) (*audit_log_protos.DeleteResponse, error) {
+func (s store) Delete(ctx grpccontext.Context, req *audit_log_protos.DeleteRequest) (*audit_log_protos.DeleteResponse, error) {
 	if len(req.GetAuditLogIds()) == 0 {
 		return nil, grpc.Errorf(codes.InvalidArgument, "no audit log IDs were specified for deletion")
 	}
--- a/pkg/grpc/auditlogstore/audit_log_store_test.go
+++ b/pkg/grpc/auditlogstore/audit_log_store_test.go
@@ -2,6 +2,7 @@ package auditlogstore
 
 import (
 	"bytes"
+	"context"
 	"encoding/json"
 	"testing"
 
@@ -11,8 +12,6 @@ import (
 	"github.com/square/p2/pkg/store/consul/auditlogstore"
 	"github.com/square/p2/pkg/store/consul/consulutil"
 	"github.com/square/p2/pkg/store/consul/transaction"
-
-	"golang.org/x/net/context"
 )
 
 func TestList(t *testing.T) {
--- a/pkg/grpc/auditlogstore/protos/audit_log_store.pb.go
+++ b/pkg/grpc/auditlogstore/protos/audit_log_store.pb.go
@@ -22,7 +22,8 @@ import fmt "fmt"
 import math "math"
 
 import (
-	context "golang.org/x/net/context"
+	"context"
+
 	grpc "google.golang.org/grpc"
 )
 
--- a/pkg/grpc/auditlogstore/protos/audit_log_store.pb.go
+++ b/pkg/grpc/auditlogstore/protos/audit_log_store.pb.go
@@ -22,8 +22,7 @@ import fmt "fmt"
 import math "math"
 
 import (
-	"context"
-
+	context "golang.org/x/net/context"
 	grpc "google.golang.org/grpc"
 )
 
--- a/pkg/grpc/podstore/podstore.go
+++ b/pkg/grpc/podstore/podstore.go
@@ -1,7 +1,6 @@
 package podstore
 
 import (
-	"context"
 	"time"
 
 	podstore_protos "github.com/square/p2/pkg/grpc/podstore/protos"
@@ -14,6 +13,7 @@ import (
 	"github.com/square/p2/pkg/types"
 
 	"github.com/hashicorp/consul/api"
+	context "golang.org/x/net/context"
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/codes"
 )
--- a/pkg/grpc/podstore/podstore.go
+++ b/pkg/grpc/podstore/podstore.go
@@ -251,9 +251,10 @@ func (s store) MarkPodFailed(ctx context.Context, req *podstore_protos.MarkPodFa
 	// should be only one system trying to write the status record so it
 	// doesn't hurt.
 	trxctx, cancelFunc := transaction.New(ctx)
+	defer cancelFunc()
 	err = s.podStatusStore.MutateStatus(trxctx, podUniqueKey, mutator)
 	if err != nil {
-		return nil, grpc.Errorf(codes.Internal, "failed to construct a consul transaction")
+		return nil, grpc.Errorf(codes.Internal, "failed to construct a consul transaction to update pod %s to failed: %s", podUniqueKey, err)
 	}
 	err = transaction.Commit(trxctx, cancelFunc, s.consulClient.KV())
 	if err != nil {
--- a/pkg/grpc/podstore/podstore.go
+++ b/pkg/grpc/podstore/podstore.go
@@ -7,9 +7,11 @@ import (
 	"github.com/square/p2/pkg/launch"
 	"github.com/square/p2/pkg/manifest"
 	"github.com/square/p2/pkg/store/consul"
+	"github.com/square/p2/pkg/store/consul/consulutil"
 	"github.com/square/p2/pkg/store/consul/podstore"
 	"github.com/square/p2/pkg/store/consul/statusstore"
 	"github.com/square/p2/pkg/store/consul/statusstore/podstatus"
+	"github.com/square/p2/pkg/store/consul/transaction"
 	"github.com/square/p2/pkg/types"
 
 	"github.com/hashicorp/consul/api"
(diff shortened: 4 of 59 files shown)


pkg/store/consul/podstore/consul_store.go
```diff
@@ -234,16 +235,21 @@ func (c *consulStore) WriteRealityIndex(ctx context.Context, podKey types.PodUni
 		return util.Errorf("Could not marshal index as json: %s", err)
 	}
 
-	indexPair := &api.KVPair{
+	//indexPair := &api.KVPair{
```
bunsenmcdubbs: TODO: remove this commented-out code

pkg/grpc/auditlogstore/audit_log_store.go
```diff
@@ -54,7 +53,7 @@ func (s store) List(_ grpccontext.Context, _ *audit_log_protos.ListRequest) (*au
 	}, nil
 }
 
-func (s store) Delete(ctx grpccontext.Context, req *audit_log_protos.DeleteRequest) (*audit_log_protos.DeleteResponse, error) {
```
mpuncel: unfortunately I dont think this will compile, this `store` type needs to implement `audit_log_protos.P2AuditLogStoreServer` and in order to implement an interface you have to take EXACTLY the same types as arguments even if the types are the same shape. In other words, `"google.golang.org/x/net/context" != "context"`

pkg/grpc/auditlogstore/audit_log_store_test.go
```diff
@@ -2,6 +2,7 @@ package auditlogstore
 
 import (
 	"bytes"
+	"context"
```
mpuncel: i think this one will be fine though! because `context.Background()` satisfies the `Context` interface from either package. confusing stuff

pkg/grpc/auditlogstore/protos/audit_log_store.pb.go
```diff
@@ -22,7 +22,8 @@ import fmt "fmt"
 import math "math"
 
 import (
-	context "golang.org/x/net/context"
```
mpuncel: oh this is why it compiles. unfortunately it's not a good idea to edit this file, it's generated automatically by running `protoc --go_out=plugins=grpc:. pkg/grpc/auditlogstore/protos/audit_log_store.proto` (after installing `protoc` on your laptop). You can follow the instructions at http://www.grpc.io/docs/quickstart/go.html if you're curious

pkg/grpc/auditlogstore/audit_log_store.go
bunsenmcdubbs: I haven't run into any issues with `rake build` yet...

pkg/store/consul/statusstore/consul_store.go
```diff
@@ -84,27 +86,34 @@ func IsStaleIndex(err error) bool {
...
+			return consulutil.NewKVError("cas", key, err)
+		}
 
-	if !success {
-		return NewStaleIndex(key, modifyIndex)
-	}
+		if !success {
+			return NewStaleIndex(key, modifyIndex)
+		}
+	*/
```
mpuncel: looks good, you can remove this comment block

pkg/grpc/podstore/podstore.go
```diff
@@ -246,7 +246,7 @@ func (s store) MarkPodFailed(_ context.Context, req *podstore_protos.MarkPodFail
 	// we don't really need the CAS properties of MutateStatus but there
 	// should be only one system trying to write the status record so it
 	// doesn't hurt.
-	err = s.podStatusStore.MutateStatus(podUniqueKey, mutator)
+	err = s.podStatusStore.MutateStatus(ctx, podUniqueKey, mutator)
```
mpuncel: I think this will always fail because `ctx` doesn't have the necessary metadata to perform a transaction. To see what i'm talking about look at `transaction.New()` which takes a context as input and derives a new context from the input one that has a value. The rest of the functions like `Add` will fail if the passed context doesn't have this value on it.

Probably I should have written it a bit differently so that the functions in the `transaction` package don't take the `context.Context` type but some different type like `transaction.Context` which is a superset of the `context.Context` interface but is guaranteed to have the data it needs. that way the compiler would complain at you here instead of just failing at run time.

All of that is a long winded way of saying that before this line you should do ```newCtx, cancel := transaction.New(ctx)
defer cancel()
```

and then pass `newCtx` to `MutateStatus`. Let me know if any of that doesn't make sense I just threw a lot at you

File: pkg/preparer/orchestrate.go [hunk 7]
```diff
@@ -435,10 +439,11 @@ func (p *Preparer) installAndLaunchPod(pair ManifestPair, pod Pod, logger loggin
 				ps.Manifest = string(manifestBytes)
 				return ps, nil
 			}
-			err := p.podStatusStore.MutateStatus(pair.PodUniqueKey, mutator)
+			err := p.podStatusStore.MutateStatus(ctx, pair.PodUniqueKey, mutator)
 			if err != nil {
 				logger.WithError(err).Errorln("Could not update manifest in pod status")
 			}
+			transaction.Commit(ctx, cancelFunc, p.client.KV())
```
mpuncel: there's an `err` to check here

pkg/preparer/orchestrate.go
```diff
@@ -418,7 +420,9 @@ func (p *Preparer) installAndLaunchPod(pair ManifestPair, pod Pod, logger loggin
 			}
 		} else {
 			// TODO: do this in a transaction
-			err = p.podStore.WriteRealityIndex(pair.PodUniqueKey, p.node)
+			// TODO dai - what do I do with the errors?
+			ctx, cancelFunc := transaction.New(context.Background())
```
mpuncel: you should add `defer cancelFunc()` here. sorry it's kind of a janky calling pattern that maybe we should rethink.

The reasoning is that if you return (e.g. on line 440) the transaction will never be closed, and that can cause a memory leak because contexts keep some state around that isn't released until the transaction is canceled. See https://golang.org/pkg/context/#WithCancel

mpuncel: also it's harmless if `cancelFunc()` is called twice, which it will be in the happy path.

pkg/preparer/orchestrate.go
```diff
@@ -476,7 +481,9 @@ func (p *Preparer) stopAndUninstallPod(pair ManifestPair, pod Pod, logger loggin
 		// processes can be viewed for some time after installation.
 		// It is the responsibility of external systems to delete pod
 		// status entries when they are no longer needed.
-		err := p.podStatusStore.MutateStatus(pair.PodUniqueKey, func(podStatus podstatus.PodStatus) (podstatus.PodStatus, error) {
+		// TODO dai - make these context aware/use tranasactions, check errors
+		ctx, cancelFunc := transaction.New(context.Background())
```
mpuncel: `defer cancelFunc()` here as well

pkg/preparer/orchestrate.go
```diff
@@ -490,6 +497,7 @@ func (p *Preparer) stopAndUninstallPod(pair ManifestPair, pod Pod, logger loggin
 			logger.WithError(err).
 				Errorln("Could not remove reality index for uninstalled pod")
 		}
+		transaction.Commit(ctx, cancelFunc, p.client.KV())
```
mpuncel: check `err`

pkg/preparer/podprocess/reporter.go
```diff
@@ -373,11 +378,14 @@ func (r *Reporter) reportLatestExits() {
 			continue
 		}
 
-		err = r.podStatusStore.SetLastExit(finish.PodUniqueKey, finish.LaunchableID, finish.EntryPoint, podstatus.ExitStatus{
+		ctx, cancelFunc := transaction.New(context.Background())
```
mpuncel: `defer cancelFunc()`

pkg/store/consul/kv_test.go
```diff
@@ -227,8 +227,13 @@ func TestAllPods(t *testing.T) {
 		t.Fatal(err)
 	}
 
+	ctx, cancelFunc := transaction.New(context.Background())
```
mpuncel: `defer cancelFunc()`

pkg/store/consul/statusstore/podstatus/consul_store_test.go
```diff
@@ -84,14 +86,20 @@ func TestDelete(t *testing.T) {
 func TestMutateStatusNewKey(t *testing.T) {
 	store := newFixture()
 
+	// TODO dai fix this test
 	key := types.NewPodUUID()
-	err := store.MutateStatus(key, func(p PodStatus) (PodStatus, error) {
+	ctx, cancelFunc := transaction.New(context.Background())
```
mpuncel: `defer cancelFunc()`

pkg/store/consul/statusstore/podstatus/consul_store_test.go
```diff
@@ -122,13 +130,19 @@ func TestMutateStatusExistingKey(t *testing.T) {
 		t.Fatalf("Unable to set up test with an existing key: %s", err)
 	}
 
-	err = store.MutateStatus(key, func(p PodStatus) (PodStatus, error) {
+	// TODO dai fix this test
+	ctx, cancelFunc := transaction.New(context.Background())
```
mpuncel: `defer cancelFunc()`

pkg/grpc/podstore/podstore.go
bunsenmcdubbs: that makes sense

Follow up question: how does `cancel` work? I see `defer cancel()` in a lot of places... Won't it error if we try to cancel an already completed transaction? Should all `..., ... = transaction.New(...)` be followed by `defer cancel()"

mpuncel: it's safe to cancel a context multiple times. and yeah i think all `transaction.New()` should be followed with that `defer`.

I can't remember what doc i was looking at that said you can cancel multiple times, but it's clear from the source code:

```
func (c *cancelCtx) cancel(removeFromParent bool, err error) {
  	if err == nil {
  		panic("context: internal error: missing cancel error")
  	}
  	c.mu.Lock()
  	if c.err != nil {
  		c.mu.Unlock()
  		return // already canceled
  	}
...
```
(https://golang.org/src/context/context.go?s=8329:8393#L335)

pkg/preparer/orchestrate.go
```diff
@@ -418,7 +420,9 @@ func (p *Preparer) installAndLaunchPod(pair ManifestPair, pod Pod, logger loggin
 			}
 		} else {
 			// TODO: do this in a transaction
-			err = p.podStore.WriteRealityIndex(pair.PodUniqueKey, p.node)
+			// TODO dai - what do I do with the errors?
```
bunsenmcdubbs: How should I handle the errors... what happens when one of these functions fails and it doesn't get added to the transaction?

pkg/grpc/podstore/podstore.go
bunsenmcdubbs: How will this transaction be committed? In `MarkPodFailed` or its callers (in which case we need to play "find the `Trxer` game again)?

pkg/preparer/orchestrate.go
mpuncel: hmm this is a tricky one. At this stage in the code, we've already started the pod and we're trying to write to reality so that we don't do it a second time, which could be bad. It's pretty important that the write succeeds.

Probably this whole bit of code from `transaction.New()` to `transaction.Commit()` should be retried until success with an exponential backoff `time.Sleep` in between failures. That way we guarantee this goroutine won't take any additional action until the write is successful

mpuncel: it's worth noting that most of these functions will only return an error if `transaction.Add` fails which should only happen if we've passed the 64 operation limit that consul imposes on transactions. We're not even close to hitting that limit here.

The `Commit()` part could fail if there is an intermittent issue talking to consul. we could consider just putting retries around that part

pkg/grpc/podstore/podstore.go
mpuncel: oh right, you're going to need a `Txner` here as well so you can call `transaction.Commit()` after `MutateStatus()`

pkg/preparer/orchestrate.go
bunsenmcdubbs: Is it worth checking for errors on `transaction.Add()` if we know it won't error? What should the error handler do in that case?

mpuncel: as a general statement i think errors should always be checked. It's one of the interesting things about go, it forces you to think through all of these cases.

I think we should still just keep looping trying to see if we get a different result from `Add()`. If we ended up in the position where that was failing, we'd hopefully know pretty quickly by looking at error rate graphs in the preparer (which count the number of Error() log lines) or by customer reports

pkg/grpc/podstore/podstore.go
bunsenmcdubbs: TODO: This is still not fixed. I'm confused why the test for this passes.

See: https://github.com/square/p2/blob/master/pkg/grpc/podstore/podstore_test.go#L426-L460

bunsenmcdubbs:
My theory is that Travis builds are failing because `go vet` is more diligent about `context` than `golang.org/x/net/context`. That said, I think the fix is as simple as adding `defer innerCancel()` to each of these faulty lines to guarantee that all paths cancel the context. 

These changes are in 6747a22

```
$ go vet ./bin/... ./pkg/...
pkg/grpc/daemonsetstore/client/client.go:77: the innerCancel function is not used on all paths (possible context leak)
pkg/grpc/daemonsetstore/client/client.go:90: this return statement may be reached without using the innerCancel var defined on line 77
exit status 1
pkg/grpc/labelstore/client/client.go:47: the innerCancel function is not used on all paths (possible context leak)
pkg/grpc/labelstore/client/client.go:53: this return statement may be reached without using the innerCancel var defined on line 47
pkg/grpc/labelstore/client/client.go:74: the innerCancel function is not used on all paths (possible context leak)
pkg/grpc/labelstore/client/client.go:64: this return statement may be reached without using the innerCancel var defined on line 74
exit status 1
pkg/grpc/podstore/client/client.go:99: the innerCancel function is not used on all paths (possible context leak)
pkg/grpc/podstore/client/client.go:84: this return statement may be reached without using the innerCancel var defined on line 99
exit status 1
```

pkg/grpc/podstore/podstore.go
```diff
@@ -246,7 +250,12 @@ func (s store) MarkPodFailed(_ context.Context, req *podstore_protos.MarkPodFail
 	// we don't really need the CAS properties of MutateStatus but there
 	// should be only one system trying to write the status record so it
 	// doesn't hurt.
-	err = s.podStatusStore.MutateStatus(podUniqueKey, mutator)
+	trxctx, cancelFunc := transaction.New(ctx)
+	err = s.podStatusStore.MutateStatus(trxctx, podUniqueKey, mutator)
+	if err != nil {
```
mpuncel: should `cancelFunc()` here (although maybe it's not technically necessary, probably the grpc library will cancel the input context?)

pkg/grpc/podstore/podstore.go
```diff
@@ -246,7 +250,12 @@ func (s store) MarkPodFailed(_ context.Context, req *podstore_protos.MarkPodFail
...
+		return nil, grpc.Errorf(codes.Internal, "failed to construct a consul transaction")
```
mpuncel: include a `%s` for `err` in here too

pkg/grpc/podstore/podstore.go
bunsenmcdubbs: i can add a `defer cancelFunc()` just in case?

mpuncel: yeah that covers more cases

bunsenmcdubbs: adding `defer cancelFunc()` is causing failures in both unit and integration tests...

PR #868, opened 2017-06-07
Retry hooks more

rudle:
3 commits for 2 lines (net). What do I win?


pkg/hooks/audit_logger_sqlite.go
```diff
@@ -121,10 +122,11 @@ func (al *SQLiteAuditLogger) withRetries(f func() error, n int) error {
 			return nil
 		}
 		attempts++
-		al.logger.WithError(err).Errorln("Caught retriable error: %d of %d", attempts, n)
```
mpuncel: i _think_ you don't need the `\n`

rudle: ah yeah, you're right.

PR #869, opened 2017-06-08
rctl has learned manifest mutate

rudle:
This new command will be useful in emergency or otherwise tricky situations that would heretofore require manual interaction with the consul DB.

### DIFF
--- a/bin/p2-rctl/main.go
+++ b/bin/p2-rctl/main.go
@@ -37,16 +37,17 @@ import (
 )
 
 const (
-	cmdCreateText     = "create"
-	cmdDeleteText     = "delete"
-	cmdReplicasText   = "set-replicas"
-	cmdListText       = "list"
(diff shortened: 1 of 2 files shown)


bin/p2-rctl/main.go
```diff
@@ -428,5 +437,26 @@ func (r rctlParams) ScheduleUpdate(oldID, newID string, want, need int, txner tr
...
+	if err != nil {
+		r.logger.WithError(err).Fatalln("Unable to open manifest for reading")
+	}
+
+	rawBytes, err := ioutil.ReadAll(f)
+	if err != nil {
+		r.logger.WithError(err).Fatalln("Unable to read manifest")
+	}
+
+	man, err := manifest.FromBytes(rawBytes)
```
mpuncel: nit: there's a `manifest.FromPath()` that will save you some work

PR #870, opened 2017-06-12
Mpunce/rc audit logs 2

mpuncel:


mpuncel:
the branch name is a little misleading, there are no audit logs yet. this PR just makes nodes be scheduled/labeled in a transaction by the RCs, and in a future PR we'll add creating audit log records

PR #873, opened 2017-06-13
Mpuncel/transactional old rc delete

mpuncel:


mpuncel:
This PR contains commits from #871 so review that one first

mpuncel:
i might want to go further with this one by also guaranteeing that the locks are still held as a part of the transaction (and maybe all of the operations the RU farm does)

mpuncel:
gonna decline this for now and reopen after making some changes

PR #872, opened 2017-06-12
Mpuncel/fix flakes

mpuncel:

PR #875, opened 2017-06-15
Mpuncel/transaction api change

mpuncel:


mpuncel:
this PR contains commits from #872 so review that first

pkg/store/consul/rollstore/consul_store.go
```diff
@@ -320,7 +320,9 @@ func (s ConsulStore) CreateRollingUpdateFromOneExistingRCWithID(
 
 	go func() {
 		<-ctx.Done()
+		fmt.Println("DONE")
 		_ = session.Destroy()
+		fmt.Println("DESTROYED")
```
mpuncel: oops

rudle:
I liked the connection between context & transactions, but I understand that the API was cumbersome.

I'm not sure it should be an error to commit a transaction twice. What ought to be an error is appending an operation to an committed transaction - can we do that?

How can we hope to ensure that cancelFunc is called everywhere it needs to be?

mpuncel:
I assume you're referring to using a `bool` instead of `ctx.Done()` to decide if a transaction was run already? I think that's a good change because `Done()` signals cancellation or deadline being hitnot necessarily completion. I also think passing in the cancel func was kind of violating the typical control flow of a context, it's weird that some code you're calling can cancel your context!

I like the suggestion about not erroring if you commit twice, but only if you add an operation. i'll make that change.

I think the only practical way to make sure that cancelFunc is called everywhere is by putting it in a `defer`. That's another reason i like not needing to pass cancel to Commit(), because it's redundant

mpuncel:
Now i'm having second thoughts about calling `Commit()` twice. What do we expect to happen? should the transaction be sent to consul again? I think you'd only want to do that if there was an `err` (as opposed to `!ok`).

rudle:
> Now i'm having second thoughts about calling Commit() twice. What do we expect to happen? should the transaction be sent to consul again? I think you'd only want to do that if there was an err (as opposed to !ok).

One could argue that it's a bug to call Commit() on a transaction that's already committed. However, I predict that threading transactions through code is difficult enough already and ensuring that Commit() is only exactly once is more difficult still. So that's the case I want to support - code reuse and possibly calling Commit() twice as a result of difficult to predict control flow. This is based on limited experience with transactions - I'd happily defer to your experience here.

All of that said, I remain convinced that the case in which code adds an operation to a transaction and expects Commit() to have side-effects is an error case.

mpuncel:
I don't anticipate a transaction being shared between goroutines since adding operations to it is generally low cost (except when you have to fetch a record to do CAS). I don't think it would make sense to commit that transaction from multiple goroutines either, so i think i'm on the "`Commit()` twice is a bug" train. I do think if we change our minds on that it will be easier to remove the error rather than impose one later on.

I'll fix `Add()` so it errors if the transaction is already committed

mpuncel:
there were some significant conflicts in the two files in `pkg/rc` so maybe those could use anotehr once over

pkg/store/consul/transaction/transaction.go
```diff
@@ -57,48 +66,67 @@ type Txner interface {
...
+	}
+
+	return nil
+}
+
 // Commit attempts to run all of the kv operations in the context's
 // transaction. The cancel function with which the context was created must
 // also be passed to guarantee that the transaction won't be applied twice
-func Commit(ctx context.Context, cancel context.CancelFunc, txner Txner) error {
+func Commit(ctx context.Context, txner Txner) (bool, *api.KVTxnResponse, error) {
```
rudle: I have to call this out as speculative because it has no callers as of this PR. I think it'll be useful though so we should keep it.

rudle:
seems good

pkg/store/consul/transaction/transaction.go
mpuncel: do you mean that nothing is using the response? right now it's used to differentiate error messages. initially i thought it would be useful in my change to make rolling updates use transactions because it might want to take different action depending on which operation had a conflict but it turned out not to be the case

rudle: Right, nothing calls this function directly right now. Is it still useful to expose all the different types of errors?

mpuncel: It ends up being more useful in the other PR because there's also a `CommitWithRetries()` function which retries with exponential backoff until a context is canceled, OR there's a transaction conflict. `err != nil` for example could be a TCP error worth retrying, whereas a transaction conflict is unlikely to resolve itself

rudle: OK. As long as it has a use case, I'm happy.

PR #876, opened 2017-06-21
Fix function signature of WatchMatches() on grpc label client.

mpuncel:
The grpc label client is intended to implement some functions from the
labels.Applicator interface which benefit from centralizing queries on a
server. Recently WatchMatches() had its function signature change to
make the aggregation rate of queries configurable, however the grpc
client did not have a similar change made to its function signature.

This commit brings them into lockstep.

### DIFF
--- a/pkg/grpc/labelstore/client/client.go
+++ b/pkg/grpc/labelstore/client/client.go
@@ -29,6 +29,16 @@ func NewClient(conn *grpc.ClientConn, logger logging.Logger) Client {
 	}
 }
 
+// this interface is just to make the compiler assert that our functions match
+// those in the direct consul applicator
+type client interface {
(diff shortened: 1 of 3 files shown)


pkg/grpc/labelstore/client/client.go
```diff
@@ -29,11 +29,22 @@ func NewClient(conn *grpc.ClientConn, logger logging.Logger) Client {
 	}
 }
 
+// this interface is just to make the compiler assert that our functions match
+// those in the direct consul applicator
+type client interface {
+	WatchMatches(selector klabels.Selector, labelType labels.Type, _ time.Duration, quitCh <-chan struct{}) (chan []labels.Labeled, error)
```
petertseng: so I take it these two WatchMatches *in this file* must now stay in sync?

Easier than checking checking whether it's in sync with a *different file* though, so I guess

mpuncel: i added a second assertion on line 40 that will make sure that those two stay in sync

PR #877, opened 2017-06-22
Mpuncel/rc audit logs

mpuncel:


pkg/rc/replication_controller.go
```diff
@@ -396,7 +401,8 @@ func (rc *replicationController) ensureConsistency(current types.PodLocations) e
 		}
 
 		rc.logger.WithField("node", pod.Node).WithField("intentManifestSHA", intentSHA).Info("Found inconsistency in scheduled manifest")
-		if err := rc.schedule(ctx, pod.Node); err != nil {
+
+		if err := rc.scheduleNoAudit(ctx, pod.Node); err != nil {
```
mpuncel: potentially we could just do audit logging in `ensureConsistency()` as well. it's going to produce an extra audit log record unnecessarily which is a downside but the upside would mean we don't need this `scheduleNoAudit()` function and we also don't need the `Context()` getter on the auditing transaction, and it would prevent someone from doing `transaction.Commit(txn.Context()` in the wrong place (and thus not creating an audit log record)

mpuncel: actually we need that `Context()` getter unless we're going to move node scheduling logic into the auditing transaction as well. i think things are fine the way they are

pkg/rc/auditing_transaction.go
```diff
@@ -0,0 +1,95 @@
...
+	"github.com/square/p2/pkg/audit"
+	pc_fields "github.com/square/p2/pkg/pc/fields"
+	"github.com/square/p2/pkg/store/consul/transaction"
+	"github.com/square/p2/pkg/types"
+	"github.com/square/p2/pkg/util"
+)
+
+type auditingTransaction struct {
+	ctx           context.Context
+	nodes         map[types.NodeName]struct{}
```
bunsenmcdubbs: why is this a map of `types.NodeName` to `struct{}`? I've been using `map[types.NodeName]bool` to create sets for fast lookup. is this analogous?

mpuncel: I could have done `bool`, but `struct{}` occupies 0 bytes of memory so it's slightly more "efficient" even though it doesn't really matter here.

source: https://dave.cheney.net/2014/03/25/the-empty-struct

pkg/rc/replication_controller.go
```diff
@@ -226,19 +229,19 @@ func (rc *replicationController) addPods(current types.PodLocations) error {
 
 	rc.logger.NoFields().Infof("Need to schedule %d nodes out of %s", toSchedule, possible)
 
-	ctx, cancelFunc := transaction.New(context.Background())
+	txn, cancelFunc := rc.newAuditingTransaction(context.Background(), currentNodes)
 	for i := 0; i < toSchedule; i++ {
 		// create a new context for every 5 nodes. This is done to make sure
```
rudle: Unrelated to the PR here, but would you mind adding a link to this issue in this comment? https://github.com/hashicorp/consul/issues/2921
We'll be able to simplify this code soon.

rudle:
What kind of metrics should we watch when this rolls out? I think it could cause a write spike, especially for pods like p2-heartbeat

mpuncel:
I don't think it will show up in our metrics, we're not increasing the number of operations performed by the RC farm, we're just adding an extra write in each operation. Bandwidth will increase but it will likely be invisible compared to other sources of bandwidth fluctuation

PR #878, opened 2017-06-26
Fix a nil pointer exception in consulutil.WatchKeys()

mpuncel:
Some metrics histograms are updated to track things like consul latency.
However in the WatchKeys() function, the queryMeta return value from the
consul List() operation was being used to update histograms before
checking that the response came back without an error. If there is an
error then queryMeta will be nil which will cause a server panic.

### DIFF
--- a/pkg/store/consul/consulutil/watch.go
+++ b/pkg/store/consul/consulutil/watch.go
@@ -56,8 +56,6 @@ func WatchKeys(
 				WaitIndex:  currentIndex,
 				AllowStale: true,
 			})
-			consulLatencyHistogram.Update(int64(queryMeta.RequestTime))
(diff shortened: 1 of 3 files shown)


pkg/store/consul/consulutil/watch.go
```diff
@@ -56,8 +56,6 @@ func WatchKeys(
 				WaitIndex:  currentIndex,
 				AllowStale: true,
 			})
-			consulLatencyHistogram.Update(int64(queryMeta.RequestTime))
-			listLatencyHistogram.Update(int64(time.Since(listStart) / time.Millisecond))
```
petertseng: this line doesn't depend on `queryMeta`, so it could have stayed, but probably better to do as you have done and keep them together.

PR #881, opened 2017-06-29
Mpuncel/transactional old rc delete no txn errors

mpuncel:


mpuncel:
this contains commits from #875  so review that first. also #877 is higher priority

mpuncel:
This now contains the commits from #914 so review that first

pkg/roll/run_update.go
```diff
@@ -139,37 +151,115 @@ func RetryOrQuit(f func() error, quit <-chan struct{}, logger logging.Logger, er
 	return true
 }
 
-func (u *update) Run(quit <-chan struct{}) (ret bool) {
+// Run causes the update to be processed either until it is complete or it is
+// cancelled via the passed quit channel. The passed context is expected to
+// have a consul transaction value stored in it and cleanup operations such as
+// deleting the old RC will be added to it when applicable.
+func (u *update) Run(ctx context.Context) (ret bool) {
 	u.logger.NoFields().Debugln("Locking")
```
rudle: I am laughing at this log line and I'm not sure why

pkg/store/consul/session.go
```diff
@@ -211,6 +176,65 @@ type unlocker struct {
...
+		Verb:    api.KVCheckSession,
+		Key:     t.key,
+		Session: t.session,
+	})
+	if err != nil {
+		return util.Errorf("could not build key unlocking transaction: %s", err)
+	}
+
+	// delete the lock (which will fail if we're not still holding it
+	// thanks to the check-session operation
```
rudle: this comment makes assumptions about what other operations are in the transaction. Is that wise?

mpuncel:
gonna rebase first and then re-review it since it's been a while since I've looked at this! will probably do some burn-in tests before merging as well

mpuncel:
I had to remove some commits during rebase (because i eventually put them in other PRs) and also had to change the way locking of the RCs is done because the LockTxn() interface had changed. still gonna review myself and burn this in

mpuncel:
I think there's an issue in here where deleting the RU (i.e. stopping the rollout) cancels the context which is used to release the RC mutation locks. This means that the RU that last handled it will keep the locks forever until the session is destroyed (which is farm-wide).

I think we probably need a different transaction context for removing the locks and we run that unconditionally when the handler exits, but still use the main context as the root of the transaction that deletes the RU (on success) and the old RC (when applicable)

rudle:
Removing locks unconditionally in error cases sounds like what sessions are for. Is it time to revisit the approach of one session per farm? I'll admit it's not straightforward to figure out which of RU, RC_old or RC_new should maintain the session but it might simplify things a bit.

mpuncel:
that sounds like a good idea

mpuncel:
added a commit to this to make a session per RU, that part could probably use a look from someone

mpuncel:
I just rebased this PR, will give it a once over to make sure i didn't do anything egregious, test it in our staging environment, and then if it works I'll merge

mpuncel:
some tests needed updating due to changes in stuff like the rcstore Watch() function

PR #882, opened 2017-07-11
Mpuncel/ds audit records

mpuncel:


mpuncel:
hmm yeah i should add an enable

pkg/store/consul/dsstore/consul_store.go
```diff
@@ -221,28 +242,96 @@ func (s *ConsulStore) MutateDS(
...
+	dsPath, err := s.dsPath(id)
+	if err != nil {
+		return fields.DaemonSet{}, util.Errorf("Error getting daemon set path: %v", err)
+	}
+
+	err = transaction.Add(ctx, api.KVTxnOp{
+		Verb:  api.KVCAS,
+		Key:   dsPath,
+		Index: metadata.LastIndex,
+		Value: rawDS,
```
rudle: Dumb question: this allows us to track manifest or node selector changes over time, right?

mpuncel: that's the intention. This change is just making all the ops use transactions so we can throw an audit log record into the transaction later

PR #884, opened 2017-07-11
Mpuncel/ds audit records 2

mpuncel:


pkg/store/consul/dsstore/auditing_store.go
```diff
@@ -0,0 +1,250 @@
...
+	"github.com/square/p2/pkg/audit"
+	"github.com/square/p2/pkg/ds/fields"
+	"github.com/square/p2/pkg/manifest"
+	"github.com/square/p2/pkg/types"
+	"github.com/square/p2/pkg/util"
+
+	klabels "k8s.io/kubernetes/pkg/labels"
+)
+
+type AuditLogStore interface {
```
rudle: Does this need to be public?

mpuncel: hmm it probably doesn't have to be, although i guess if we pretended we had godocs i don't think private types get documented?

mpuncel:
the stuff after the approval is just a rebase so i'm still considering this approved

mpuncel:
had to fix up some Commit() calls in tests

PR #887, opened 2017-07-17
Add label tree constant for config

mpuncel:
This will allow labeling config values with information that can be used
to identify the pod cluster that the config value corresponds to.


rudle:
🚀

PR #888, opened 2017-07-17
Config store

rudle:
This PR is WIP but it breaks the API somewhat.

### DIFF
--- a/pkg/store/consul/configstore/store.go
+++ b/pkg/store/consul/configstore/store.go
@@ -2,7 +2,6 @@ package configstore
 
 import (
 	"context"
-	"fmt"
 
 	"github.com/square/p2/pkg/util" // TODO this is wrong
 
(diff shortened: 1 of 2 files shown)


pkg/store/consul/configstore/store.go
```diff
@@ -97,9 +99,25 @@ func (cs *ConsulStore) PutConfig(ctx context.Context, config Fields, v *Version)
...
+		Key:         id.String(),
+		ModifyIndex: v.uint64(),
+	}
+
+	ok, _, err := cs.consulKV.DeleteCAS(kvPair, nil)
+	if !ok {
+		return util.Errorf("CAS Delete Failed! Consider retry")
+	}
+	if err != nil {
+		return util.Errorf("CAS Delete Failed: %", err)
```
petertseng: percent

rudle:
PTAL! This is a pretty solid implementation of V1 of the store. Next step for this work is to support transactions in LabelConfig. cc @bunsenmcdubbs

PR #889, opened 2017-07-17
pods: setupConfig: merge config from config store

petertseng:
No need to review yet, just letting Travis build this because I'm too lazy to do it locally nor make my own fork do Travis.

Advice requested on the question "When do the pod labels get set? Will they be available when the preparer is installing the pod?"

### DIFF
--- a/bin/p2-install-hook/main.go
+++ b/bin/p2-install-hook/main.go
@@ -53,6 +53,7 @@ func main() {
 	pod := hookFactory.NewHookPod(manifest.ID())
 
 	// for now use noop verifier in this CLI
+	// OK to not merge config for hooks, which are not part of a pod cluster.
 	err = pod.Install(manifest, auth.NopVerifier(), artifact.NewRegistry(*registryURI, uri.DefaultFetcher, osversion.DefaultDetector))
 	if err != nil {
 		log.Fatalf("Could not install manifest %s: %s", manifest.ID(), err)
--- a/bin/p2-launch/main.go
+++ b/bin/p2-launch/main.go
@@ -88,6 +88,7 @@ func main() {
 	podFactory := pods.NewFactory(*podRoot, types.NodeName(*nodeName), fetcher, "")
 	pod := podFactory.NewLegacyPod(manifest.ID())
 
+	// OK to not merge config for launched pods, which are not part of a pod cluster.
 	err = pod.Install(manifest, auth.NopVerifier(), artifact.NewRegistry(*artifactRegistryURL, fetcher, osversion.DefaultDetector))
 	if err != nil {
 		log.Fatalf("Could not install manifest %s: %s", manifest.ID(), err)
--- a/pkg/store/consul/configstore/store.go
+++ b/pkg/store/consul/configstore/store.go
@@ -1,0 +1,17 @@
+package configstore
+
+import "context"
+
+type ID string
+type Version string
+
+type Fields struct {
+	Config map[interface{}]interface{}
+	ID     ID
+}
+
+type Storer interface {
+	FetchConfig(ID) (Fields, Version, error)
+	PutConfig(context.Context, ID, Fields, Version) error
+	DeleteConfig(context.Context, ID, Version) error
+}
--- a/pkg/store/consul/configstore/store.go
+++ b/pkg/store/consul/configstore/store.go
@@ -2,7 +2,6 @@ package configstore
 
 import (
 	"context"
-	"fmt"
 
 	"github.com/square/p2/pkg/util" // TODO this is wrong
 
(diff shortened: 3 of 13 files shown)


petertseng:
> "When do the pod labels get set? Will they be available when the preparer is installing the pod?"

By the RC. It's fine, they'll be available.

Next question. Given that a pod has some labels L, with which subset of those labels should we query for a matching pod cluster?

1. Hard code. Doesn't seem sensible. The RC doesn't hard code the labels it applies (Right?)
2. Config. I guess that works
3. Take the current labels, subtract pod_id and replication_controller_id, use that. Trying to be clever but require zero config. Brittle if an unexpected label gets added.

petertseng:
Oooookay, I think we now merge the config in on install time. If the code works on the first try.

bin/p2-install-hook/main.go
```diff
@@ -53,7 +53,9 @@ func main() {
 	pod := hookFactory.NewHookPod(manifest.ID())
 
 	// for now use noop verifier in this CLI
-	err = pod.Install(manifest, auth.NopVerifier(), artifact.NewRegistry(*registryURI, uri.DefaultFetcher, osversion.DefaultDetector))
+	// TODO: should this have config store/labeler to merge hook configs?
```
petertseng: a TODO I was concerned about, but less so

bin/p2-launch/main.go
```diff
@@ -88,7 +88,8 @@ func main() {
 	podFactory := pods.NewFactory(*podRoot, types.NodeName(*nodeName), fetcher, "")
 	pod := podFactory.NewLegacyPod(manifest.ID())
 
-	err = pod.Install(manifest, auth.NopVerifier(), artifact.NewRegistry(*artifactRegistryURL, fetcher, osversion.DefaultDetector))
+	// TODO: config store/labeler? Otherwise, p2-launch ignores the config store, which doesn't seem ideal.
```
petertseng: a TODO I was concerned about, more so

petertseng:
I need #896 for this. I should probably just rebase off that branch...

petertseng:
temporarily cherry picking it in to make sure the build would pass with it, I didn't rebase directly because #896 is itself out of date, but maybe it's not missing any commit I need, so maybe not necessary

bin/p2-launch/main.go
mpuncel: hmmmmmmmmmmmmmmmmmmmmmm yeah we're really hampering `p2-launch` here because now the manifest isn't all the info you need to run a pod

rudle: Passing nil arguments to a function we control is a smell. Can you write a function with a more descriptive name and a more correct arity?

petertseng:
Hmm. `Install` and `InstallWithConfig`? I wouldn't want to imply that `Install` installs **without** config though.

petertseng:
--- FAIL: TestDeleteConfigTxn (0.44s)
	consultest.go:98: starting Consul server with port: 33739
	store_test.go:279: Expected to receive an error when fetching deleted configuration. Got: {map[] }
	consultest.go:111: stopping Consul server with port: 33739
FAIL
FAIL	github.com/square/p2/pkg/store/consul/configstore	0.551s

petertseng:
--- FAIL: TestDeleteConfig (0.00s)
	store_test.go:171: Expected to receive an error when fetching deleted configuration. Got: {map[] }

petertseng:
2017/07/26 01:54:55 Bootstrapping complete
2017/07/26 01:54:58 Status of preparer: p2-preparer OK
2017/07/26 01:54:59 grpc: addrConn.resetTransport failed to create client transport: connection error: desc = "transport: dial tcp 127.0.0.1:3000: getsockopt: connection refused"; Reconnecting to {localhost:3000 <nil>}
2017/07/26 01:55:00 grpc: addrConn.resetTransport failed to create client transport: connection error: desc = "transport: dial tcp 127.0.0.1:3000: getsockopt: connection refused"; Reconnecting to {localhost:3000 <nil>}
2017/07/26 01:55:02 grpc: addrConn.resetTransport failed to create client transport: connection error: desc = "transport: dial tcp 127.0.0.1:3000: getsockopt: connection refused"; Reconnecting to {localhost:3000 <nil>}
2017/07/26 01:55:04 grpc: addrConn.resetTransport failed to create client transport: connection error: desc = "transport: dial tcp 127.0.0.1:3000: getsockopt: connection refused"; Reconnecting to {localhost:3000 <nil>}
2017/07/26 01:55:08 grpc: addrConn.resetTransport failed to create client transport: connection error: desc = "transport: dial tcp 127.0.0.1:3000: getsockopt: connection refused"; Reconnecting to {localhost:3000 <nil>}
label-store-server tail: 
2017-07-26_01:55:08.48305 time="2017-07-26T01:55:08Z" level=info msg="Listening tcp on port 3000" Counter=0 PID=11962

pkg/store/consul/configstore/store.go
```diff
@@ -0,0 +1,224 @@
+package configstore
+
+import (
+	"context"
+	"encoding/json"
+	"path"
+
+	"github.com/square/p2/pkg/util" // TODO this is wrong
```
petertseng: I no longer remember why this is wrong

PR #891, opened 2017-07-18
Transactions in config store!

bunsenmcdubbs:
Add transaction versions of `PutConfig` and `DeleteConfig` ... `PutConfigTxn` and `DeleteConfigTxn`! With tests! That pass!

### DIFF
--- a/pkg/store/consul/configstore/store.go
+++ b/pkg/store/consul/configstore/store.go
@@ -9,6 +9,7 @@ import (
 
 	"github.com/hashicorp/consul/api"
 	"github.com/square/p2/pkg/labels"
+	"github.com/square/p2/pkg/store/consul/transaction"
 	"gopkg.in/yaml.v2"
 	klabels "k8s.io/kubernetes/pkg/labels"
 )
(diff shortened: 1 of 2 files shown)


pkg/store/consul/configstore/store.go
```diff
@@ -108,24 +106,53 @@ func (cs *ConsulStore) PutConfig(ctx context.Context, config Fields, v *Version)
...
-		ModifyIndex: v.uint64(),
+func (cs *ConsulStore) PutConfigTxn(ctx context.Context, config Fields, v *Version) error {
+	yamlConfig, err := yaml.Marshal(config.Config)
+	if err != nil {
+		return err
 	}
+	env := envelope{Config: string(yamlConfig)}
 
-	ok, _, err := cs.consulKV.DeleteCAS(kvPair, nil)
+	bs, err := json.Marshal(env)
```
mpuncel: missing an `err` check

pkg/store/consul/configstore/store.go
```diff
@@ -108,24 +106,53 @@ func (cs *ConsulStore) PutConfig(ctx context.Context, config Fields, v *Version)
...
+
+func (cs *ConsulStore) DeleteConfig(id ID, v *Version) error {
+	ok, _, err := cs.consulKV.DeleteCAS(&api.KVPair{
+		Key:         id.String(),
+		ModifyIndex: v.uint64(),
+	}, nil)
 	if !ok {
 		return util.Errorf("CAS Delete Failed! Consider retry.")
 	}
 	if err != nil {
```
mpuncel: probably should check `err` first, i think `ok` will always be false when an error occurs

pkg/store/consul/configstore/store_test.go
```diff
@@ -1,15 +1,18 @@
 package configstore
```
mpuncel: you should add ```// +build !race

package configstore
```

PR #893, opened 2017-07-18
manifest: Allow writing any config, not just manifest's ; Add function to merge two configs

petertseng:


petertseng:
There are no callers of the new functions in this PR, but the code was requested so that it may be used elsewhere. I think they will be suitable for use? At least the `MergedConfig` is tested.

pkg/manifest/manifest.go
```diff
@@ -318,13 +320,58 @@ func (manifest *manifest) Marshal() ([]byte, error) {
...
+// MergedConfig merges this manifest's config with the given map, returning a new config map.
+// The config of the manifest is not modified.
+func (manifest *manifest) MergedConfig(with map[interface{}]interface{}) (map[interface{}]interface{}, error) {
+	merged := manifest.GetConfig()
+
+	// Tried just marshalling the second config, then unmarshalling it into the same map.
+	// However, that doesn't preserve inner maps correctly.
+	mergeInto(merged, with)
+
+	return merged, nil
```
petertseng: says it can error, but in no code path will there be a non-nil error.

I used to say it is an error to merge a non-map with a map, but then I just ended up overwriting everything and not showing errors. Wonder if anyone can think of any time to error. If not, maybe I just won't return an error.

PR #894, opened 2017-07-18
Effectively revert a change made in #890

rudle:
The type of map used by pkg/manifest is map[interface{}]interface{} so it
behooves us to match that. The map[string]interface{} change was to support JSON
marshalling of config stanzas, which seems less important now than it once did.


mpuncel:
👍

PR #899, opened 2017-07-20
This isn't an error case yet

rudle:
We intend to backfill config in which case it will become an error


mpuncel:
are we sure we want this? If we already know the ID to fetch on, we expect it to exist right?

petertseng:
I'm no longer convinced I need this, but for obvious reasons am not trying it without this commit yet. Maybe after I can try without this commit.

rudle:
Closing as stale

PR #900, opened 2017-07-20
Mpuncel/podstore sugar

mpuncel:


mpuncel:
this contains the commit from @rudle's PR

PR #901, opened 2017-07-21
Pick a private IP for consul to bind to.

mpuncel:
This addresses travis errors where consul won't start with the following
error:

2017-07-20_18:14:35.22743 ==> Error starting agent: Failed to get
advertise address: Multiple private IPs found. Please configure one.

### DIFF
--- a/integration/single-node-slug-deploy/check.go
+++ b/integration/single-node-slug-deploy/check.go
@@ -8,6 +8,7 @@ import (
 	"io/ioutil"
 	"log"
 	"math/rand"
+	"net"
 	"net/http"
 	"net/url"
 	"os"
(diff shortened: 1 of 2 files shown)


mpuncel:
i don't know if this will work yet, trying to fix travis builds

integration/single-node-slug-deploy/check.go
```diff
@@ -722,10 +723,21 @@ func getConsulManifest(dir string) (string, error) {
...
+	if err != nil {
+		return "", util.Errorf("could not list addresses to find one to bind consul to: %s", err)
+	}
+	ip := strings.Split(addrs[0].String(), "/")[0]
+
 	rubyBin := filepath.Join(dir, "launch")
 	cmds := `#!/usr/bin/env ruby
 
-exec "/usr/bin/consul agent -server -bootstrap-expect 1 -data-dir /tmp/consul"`
+exec "/usr/bin/consul agent -server -bootstrap-expect 1 -data-dir /tmp/consul -bind %%%%"`
```
rudle: did you try to bind to 0.0.0.0 ?

mpuncel: according to consul docs, that's the default. So I think that's what we were doing before.

https://www.consul.io/docs/agent/options.html#_bind

mpuncel: from wikipedia: ```In the context of servers, 0.0.0.0 means "all IPv4 addresses on the local machine". If a host has two IP addresses, 192.168.1.1 and 10.1.2.1, and a server running on the host listens on 0.0.0.0, it will be reachable at both of those IPs.```

And consul has decided that they don't want to bind to multiple IP addresses, so it's a crash in that case. so we choose one arbitrarily here

PR #903, opened 2017-07-24
Close quit channels on shutdown rather than send a value

mpuncel:
The preparer captures TERM signals to buy time to get the work it's
doing into a logical stopping place before exiting. It does this by
creating a bunch of quit channels, one for each function the preparer
performs and then signalling exit via the quit channels.

Prior to this commit it would send a value on the quit channels to
signal that an exit should happen, however if a preparer function shared
the quit channel across goroutines this would cause only one of them to
quit.

This commit instead closes the channel which will cause any watcher on
the quit channel to fire and exit.


mpuncel:
yeah it's being shared inside of the finish reporter. I forget how I came across this, was a branch i had sitting around locally that i forgot about. `close` does seem like a better option though

PR #904, opened 2017-08-01
daemon sets: add logging to analyze labeling failures

mpuncel:


pkg/ds/daemon_set.go
```diff
@@ -434,11 +434,12 @@ func (ds *daemonSet) addPods() error {
 	// Get the difference in nodes that we need to schedule on and then sort them
 	// for deterministic ordering
 	toScheduleSorted := types.NewNodeSet(eligible...).Difference(types.NewNodeSet(currentNodes...)).ListNodes()
-	ds.logger.NoFields().Infof("Need to label %d nodes", len(toScheduleSorted))
+	ds.logger.Infof("Need to label %d nodes: %s", len(toScheduleSorted), toScheduleSorted)
```
rudle: This logline is going to be loooooooooooooooooong

mpuncel: mostly only when a daemon set is new (because this is only the diff between current nodes and eligible nodes).

Also we already have super long log lines https://github.com/square/p2/blob/master/pkg/ds/daemon_set.go#L567

PR #907, opened 2017-08-09
Treat hook failures like normal failures.

mpuncel:
Up until this commit, hook failures woul not stop a pod from being
installed and having reality updated. Hook failures were logged but
otherwise ignored.

This commit changes this by treating a hook failure as a critical
failure which will cause the entire pod installation routine to be
retried (the same way a download failure is treated, for example).

This means that P2 operators will have confidence that if reality is
updated for a pod, that all of its hooks must have succeeded. Previously
this was not the case and the philosophy was that P2 applications should
check for conditions they expect hooks to put in place within their
health checks, which will prevent a rolling update from rendering all
pod instances within a pod cluster to be nonfunctional

### DIFF
--- a/integration/single-node-slug-deploy/check.go
+++ b/integration/single-node-slug-deploy/check.go
@@ -52,6 +52,16 @@ const (
 	sqliteFinishDatabasePath = "/data/pods/p2-preparer/finish_data/finish.db"
 )
 
+var noAddUser = false
+
+func parseOptions(args []string) {
+	for _, arg := range args {
+		if arg == "--no-add-user" {
+			noAddUser = true
+		}
+	}
+}
+
 func main() {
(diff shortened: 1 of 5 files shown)


mpuncel:
I'll probably have to resort to manual testing to verify this change

mpuncel:
I gave this a test spin with working hooks and failing hooks, and it looks good to me

rudle:
Did you test with a hook that always times out?

mpuncel:
not yet, i'll do that

rudle:
If the timeout paths have been exercised, I'm supportive of this change.

mpuncel:
giving it a spin right now, also had to add a commit to fix the integration test (since we had a hook that reliably fails on travis)

mpuncel:
i confirmed that timeouts are also treated like failures and cause the install flow to restart

PR #909, opened 2017-08-10
Report number of nodes in critical, warning, and unknown states

spenceral:


rudle:
Looks good! Please wait for the tests to pass before merging.

petertseng:
**If** this causes too much lock contention (I have no idea whether it will!), we could consider locking the podHealth just once and doing all four counts at once instead of locking/unlocking four times. We probably don't need to worry about this though

PR #910, opened 2017-08-14
Mpuncel/transaction replication

mpuncel:


pkg/ds/daemon_set.go
```diff
@@ -559,11 +554,10 @@ func (ds *daemonSet) PublishToReplication() error {
 	// InitializeReplicationWithCheck, we will get an error
 	ds.cancelReplication()
 
-	podLocations, err := ds.CurrentPods()
+	nodes, err := ds.EligibleNodes()
```
mpuncel: i'm hoping this doesn't have unintended consequences. Prior to this commit we would label the new nodes ahead of time and then target the labeled nodes (aka `CurrentPods()`). Now they're not labeled yet, so we target the eligible ones (that match the node selector).

99% sure it's the same thing but check my work on that

pkg/store/consul/transaction/transaction.go
```diff
@@ -135,6 +136,40 @@ func TxnErrorsToString(errors api.TxnErrors) string {
 	return str
 }
 
+// CommitWithRetries retries Commit() until the transaction is applied without
+// an error. It will not retry the transaction if there is no error but the
+// transaction was rolled back (and it wouldn't make sense to because it won't
+// succeed after that point)
+//
+// An exponential backoff strategy is used until the context is cancelled with
+// a max backoff time if 10 seconds
```
rudle: typo: if -> of

pkg/ds/daemon_set.go
rudle: This seems right, and probably more correct than it was before. I suspect that this whole PR will cause PublishToReplication to target a broader set of nodes while narrowing the scope of the other mutative functions (addPods, removePods, clearPods) to be more correct.

mpuncel:
just fixed that comment typo, nothing else has changed

PR #911, opened 2017-08-16
Mpuncel/lock txn

mpuncel:


mpuncel:
This PR doesn't add any new functionality but adds some new functions that aren't used but are useful in certain circumstances. These commits are for instance used in #881 and will be useful for daemon sets ensuring that the daemon set lock is held while replications occur.

pkg/store/consul/transaction/transaction.go
```diff
@@ -35,10 +35,20 @@ type tx struct {
 	committedMu sync.Mutex
 }
 
+// New() returns a new context derived from the one passed as an argument and
+// its cancel function with the context containing metadata used to build a
+// consul transaction. If the passed context already has consul operations
+// defined, the new context will inherit those operations defined but in a
+// separate transaction.
 func New(ctx context.Context) (context.Context, context.CancelFunc) {
```
rudle: Without seeing this used in _context_, it's hard for me to say whether this behaviour overloading is clear or confusing. Could we accomplish the same goal with a `Branch()` constructor?

mpuncel: basically it would be like (shortcuts taken for brevity)
```
lockCtx := transaction.New(context.Background())
unlockCtx := transaction.New(context.Background())
checkLockedCtx := transaction.New(context.Background())

buildTxns(lockCtx, unlockCtx, checkLockedCtx)

// acquire the lock
transaction.Commit(lockCtx)

// release lock when work is done
defer transaction.Commit(unlockCtx)

// branch some transactions off the "check locked one"
checkLockedCtx1 := transaction.New(checkLockedTxn)
doSomeWork(checkLockedCtx1)

checkLockedCtx2 := transaction.New(checkLockedTxn)
doSomeWork(checkLockedCtx2)
```

If `Branch()` seems more clear i'm all for that. `Branch()` can fail if there is no transaction in the provided context, and `New()` can fail if there is one?

rudle: ah, I guess it's not so different than what we have today - I forgot about the context.Background thing.

I'm ambivalent about this. I don't see there being a use case for the "multi-method" approach for transaction.New() but the API seems fine as it is.

mpuncel:
fixed up commit history, everything else is the same

PR #912, opened 2017-08-17
Use contexts for cancellation in daemon set farm.

mpuncel:
This will be useful in order to pass a context containing a partial
consul transaction that ensures that the daemon set farm still holds the
lock that it's supposed to be when it performs consul operations. This
will be useful to protect against multiple daemon set farms writing the
same daemon set status entry with conflicting information.


mpuncel:
summary of changes:

1) using pointers a lot less. there's no benefit to them and using values avoids unintended consequences
2) DRY'd up a bunch of stuff
3) using contexts instead of quit channels

All of this will be useful for ensuring that daemon set operations check that the lock is held in consul transactions in follow up work

PR #915, opened 2017-09-05
pods: Pass fetcher from factory to pods

petertseng:
NewFactory was taking a fetcher and then not using it, instead using a
defaultFetcher.

This is inappropriate for situations where we must configure specific
settings on the fetcher in order to be able to download artifacts (for
example, CA file)

`newPodWithHome` must accept a fetcher and then use it, and `NewUUIDPod`
and `NewLegacyPod` must pass the fetcher given from `NewFactory`.

Recall that the certificates we issue require that we be able to
configure specific CAs.

This was partially done in #826, but it was not made to apply to Pods,
which internally contain Fetchers as well.

### DIFF
--- a/pkg/pods/factory.go
+++ b/pkg/pods/factory.go
@@ -92,24 +92,29 @@ func (f *factory) NewUUIDPod(id types.PodID, uniqueKey types.PodUniqueKey) (*Pod
 		return nil, util.Errorf("uniqueKey cannot be empty")
 	}
 	home := filepath.Join(f.podRoot, computeUniqueName(id, uniqueKey))
-	return newPodWithHome(id, uniqueKey, home, f.node, f.requireFile), nil
+	return newPodWithHome(id, uniqueKey, home, f.node, f.requireFile, f.fetcher), nil
 }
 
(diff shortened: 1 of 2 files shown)


petertseng:
failing test was:

```
condition failed, will retry: farm_test.go:1340: Daemon sets do not match, expected 'ecf4b07d-5556-49d6-aacf-f50f63ee5bd9', 'availability_zone=az3', got 'ecf4b07d-5556-49d6-aacf-f50f63ee5bd9', 'availability_zone=az99'
panic: test timed out after 2m0s
FAIL	github.com/square/p2/pkg/ds	120.020s
```

to be investigated

petertseng:
ok now it's:


```
2017/09/05 20:45:23 grpc: addrConn.resetTransport failed to create client transport: connection error: desc = "transport: dial tcp 127.0.0.1:3000: getsockopt: connection refused"; Reconnecting to {localhost:3000 <nil>}

2017/09/05 20:45:38 Could not start grpc label store: grpc: timed out when dialing
exit status 1
```

maybe that's actually something

PR #916, opened 2017-09-05
pods: Take/use fetcher in NewHookFactory

petertseng:
pods: Take/use fetcher in NewHookFactory

As in #915, it seems necessary to configure a fetcher, otherwise I predict (but have no hard evidence to support) that we will be unable to install any hooks.

I'm unsure of the necessity of this one but it's here if we need it.


petertseng:
Rebased, so no longer contains the commit from #915.

PR #918, opened 2017-09-19
Mpuncel/daemon set status

mpuncel:


mpuncel:
this contains commits from #914 so review that first

pkg/ds/farm.go
```diff
@@ -614,29 +625,37 @@ func (dsf *Farm) spawnDaemonSet(
...
 			// The session probably either expired or there was probably a network
 			// error, so the rest will probably fail
 			dsf.handleSessionExpiry(dsFields, dsLogger, err)
-			dsf.releaseLock(dsUnlocker)
+			return false
+		}
+		if !ok {
+			// Lock was either already acquired by another farm or it was Acquired
+			// by this farm
+			dsf.logger.Infof("Lock on daemon set '%v' was already acquired by another ds farm: %s", dsFields.ID, transaction.TxnErrorsToString(resp.Errors))
```
rudle: `Errorf()`

pkg/ds/daemon_set.go
```diff
@@ -224,12 +225,24 @@ func (ds *daemonSet) GetNodeSelector() klabels.Selector {
 	return ds.DaemonSet.NodeSelector
 }
 
+func (ds *daemonSet) Manifest() manifest.Manifest {
+	ds.mu.Lock()
+	defer ds.mu.Unlock()
```
rudle: Why read locks? Is this same mutex used to guard writes too?

pkg/ds/daemon_set.go
```diff
@@ -34,7 +37,8 @@ const (
 )
 
 var (
-	DefaultRetryInterval = 5 * time.Minute
+	DefaultRetryInterval  = 5 * time.Minute
+	statusWritingInterval = 15 * time.Second
```
rudle: Make this a config flag?

pkg/ds/farm.go
mpuncel: I don't think it's really an error, it just means that the daemon set is already claimed by another farm. we expect that to happen often

pkg/ds/daemon_set.go
mpuncel: 👍

pkg/ds/daemon_set.go
mpuncel: yeah, https://github.com/square/p2/blob/master/pkg/ds/daemon_set.go#L313

mpuncel: although maybe there's only one goroutine acting here?

mpuncel:
this has undergone a bit of turmoil since the last review, so not sure if i should still consider it approved.

the main change is that we don't CAS the status values anymore, that proved challenging because the `queryMeta` gives you the last GLOBAL index of consul not the modify index of the key you fetched. I didn't want to tear up our APIs to get the modify index out of the `api.KVPair`. Instead i'm just using a "set" not a "cas" now which I believe to be no less thorough because we do check we have the daemon set lock still and that combined with single goroutine writes should give us the same guarantees

rudle:
I agree with your reasoning. 

I recall that CAS was in the API before transactions were released and it remains around as a less expensive approach to synchronization.

rudle:
:shipit:

mpuncel:
Added a commit to this because I noticed in testing that `replication_in_progress` never became false in the status. Added an `InProgress()` function to replications to handle that

mpuncel:
reviewed the final commit with @rudle in person

PR #920, opened 2017-09-20
Add excludePods option to not shutdown pods

rudle:
This is useful for shutting down apps but not core services

### DIFF
--- a/bin/p2-shutdown/main.go
+++ b/bin/p2-shutdown/main.go
@@ -26,6 +26,7 @@ var (
 	verbose      = kingpin.Flag("verbose", "Print debugging information").Short('v').Bool()
 	dryRun       = kingpin.Flag("dry", "Dry run: do not stop any pods").Short('d').Bool()
 	shutdownPods = kingpin.Flag("pods", "The list of pods to shutdown. Leave empty for all").Short('p').Strings()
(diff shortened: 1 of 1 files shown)


bin/p2-shutdown/main.go
```diff
@@ -26,6 +26,7 @@ var (
 	verbose      = kingpin.Flag("verbose", "Print debugging information").Short('v').Bool()
 	dryRun       = kingpin.Flag("dry", "Dry run: do not stop any pods").Short('d').Bool()
 	shutdownPods = kingpin.Flag("pods", "The list of pods to shutdown. Leave empty for all").Short('p').Strings()
+	excludePods  = kingpin.Flag("exclude-pods", "The list of pods to exclude from shutdown.").Short('e').Strings()
```
mpuncel: do we want some kind of typo protection? verify that all the excluded pods exist on the system? maybe that's too restrictive

rudle: It's only for shutdown use cases, so I think the cost of being wrong is not big enough to warrant the cost.

This change was motivated by a feature request to exclude the RCE agent that executes this very script when shutting down a host for maintenance. We're moving towards a world where people don't type this in by hand.

mpuncel: 👍

PR #922, opened 2017-09-20
Mpuncel/fix replication

mpuncel:


pkg/store/consul/transaction/transaction.go
```diff
@@ -162,20 +162,24 @@ func CommitWithRetries(ctx context.Context, txner Txner) (bool, *api.KVTxnRespon
...
 
-	backoff := 100 * time.Millisecond
-	for err := f(); err != nil; err = f() {
-		select {
-		case <-ctx.Done():
-			break
-		default:
+	// Need this anonymous function because you can't use "break" in a
+	// select, you have to return :(
+	func() {
```
spenceral: Could you use a label here? 
```
Loop:
for err := f(); err != nil; err = f() {
select {
    case <-ctx.Done():
        break Loop
    ...
}
```

mpuncel: yeah that would have worked as well. I think I like functions better than labels (although i've never worked with them). Here we'd probably get away with it but you can imagine if there were lots of labels here it might be hard to know where a block of new code being added should go. I think it's easier to grok the control flow with functions

PR #925, opened 2017-09-26
Fix a bug in p2-inspect where launchable versions would be incorrect

mpuncel:


pkg/inspect/inspect.go
```diff
@@ -70,9 +70,9 @@ func AddKVPToMap(result consul.ManifestResult, source int, filterNode types.Node
 		}
 		old.IntentManifestSHA = manifestSHA
 		for launchableID, launchable := range result.Manifest.GetLaunchableStanzas() {
-			var version *launch.LaunchableVersion
+			var version launch.LaunchableVersion
```
rudle: this code looks similar to the helper function LaunchableVersion() on the LaunchableStanza type

PR #926, opened 2017-09-26
Straighten out the logic around excluded pods

rudle:
Add a test that ought to have appeared in the original PR

### DIFF
--- a/bin/p2-shutdown/main.go
+++ b/bin/p2-shutdown/main.go
@@ -46,11 +46,13 @@ func main() {
 		log.Fatalf("caught fatal error while querying datastore: %v", err)
 	}
 
+	podsToExclude := make([]types.PodID, 0, len(*excludePods))
+	for _, s := range *excludePods {
+		podsToExclude = append(podsToExclude, types.PodID(s))
+	}
+
 	podsToShutdown := make([]types.PodID, 0, len(*shutdownPods))
 	for _, pod := range *shutdownPods {
-		if listInclude(pod, *excludePods) {
-			continue
-		}
 		podsToShutdown = append(podsToShutdown, types.PodID(pod))
 	}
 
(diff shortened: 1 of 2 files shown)


bin/p2-shutdown/main.go
```diff
@@ -59,12 +56,12 @@ func main() {
 	var haltWG sync.WaitGroup
 	for _, realityEntry := range reality {
 		pod := podFactory.NewLegacyPod(realityEntry.Manifest.ID())
-		if *dryRun {
-			log.Printf("dry run, skipping this pod: %s", pod.Id)
+		if !shutdownPod(pod.Id, podsToShutdown, *excludePods) {
```
mpuncel: nit: `shouldShutdownPod()` would better indicate that this function doesn't perform the down shutting

bin/p2-shutdown/main_test.go
```diff
@@ -0,0 +1,27 @@
...
+func TestShutdownPod(t *testing.T) {
+	testCases := []struct {
+		pod            types.PodID
+		podsToShutdown []types.PodID
+		podsToExclude  []string
+		expectation    bool
+	}{
+		{types.PodID("a"), []types.PodID{"a"}, []string{}, true},
+		{types.PodID("a"), []types.PodID{}, []string{"a"}, false},
+		{types.PodID("a"), []types.PodID{}, []string{}, true},
```
mpuncel: maybe add a case where "b" is in teh shutdown list which should make the expectation for "a" false right?

rudle:
Feedback addressed. 

I guess I have different git authors on my different machines. That's annoying.

bin/p2-shutdown/main.go
```diff
@@ -84,23 +86,22 @@ func main() {
...
 	}
-	return false
-}
 
-func listInclude(pod string, pods []string) bool {
-	for _, p := range pods {
-		if p == pod {
-			return true
-		}
+	if len(podsToShutdown) > 0 {
```
petertseng: I guess if you wanted to code golf you could `return len(podsToShutdown) == 0`

shrug

PR #928, opened 2017-10-04
Mpuncel/allocate scheduler

mpuncel:
This gives us a starting point to build our allocation server endpoint


pkg/grpc/scheduler/protos/scheduler.proto
```diff
@@ -0,0 +1,17 @@
+syntax = "proto3";
+
+package scheduler_protos;
+
+service P2Scheduler {
+  rpc Allocate (AllocateRequest) returns (AllocateResponse) {}
```
spenceral: nit: I like `AllocateNodes` which I think you mentioned in person earlier

mpuncel:
addressed feedback, although I realized that maybe the implementation of the scheduler interface shouldn't be `ApplicatorScheduler` because it's going to be doing grpc stuff not just labels queries? can probably punt that to a different PR once we start the implementation tho

PR #929, opened 2017-10-04
Add groundwork for allocating nodes based on a given strategy

spenceral:


pkg/rc/fields/fields.go
```diff
@@ -49,6 +52,10 @@ type RC struct {
 
 	// When disabled, this controller will not make any scheduling changes
 	Disabled bool
+
+	// Distinguishes between dynamic, static or other strategies for allocating
+	// nodes on which the rc can schedule the manifest.
+	AllocationStrategy Strategy
```
mpuncel: you're also going to need to add this to `RawRC` and probably copy the value to/from the RC in the JSON marshaling functions. Sorry this stuff is so convoluted, i think it doesn't have to be but that's how it is :)

pkg/rc/fields/fields.go
```diff
@@ -15,6 +15,9 @@ import (
 // string format so that Go will typecheck its uses.
 type ID string
 
+// Strategy is a type alias used for node allocation strategies.
+type Strategy int
```
mpuncel: i'd probably rather make this a string so that you can tell what the strategy is by looking at the JSON without needing to correlate it with an enum

pkg/rc/replication_controller.go
```diff
@@ -256,6 +262,17 @@ func (rc *replicationController) addPods(current types.PodLocations, eligible []
 			txn, cancelFunc = rc.newAuditingTransaction(context.Background(), txn.Nodes())
 		}
 		if len(possibleSorted) < i+1 {
+			if rc.AllocationStrategy == CattleStrategy {
```
mpuncel: it's probably fine to just leave this out of the PR, you can keep it in a branch for later

spenceral:
Updated to fix build and tests

integration/single-node-slug-deploy/check.go
```diff
@@ -441,13 +441,13 @@ func verifyTransferReplicas(errCh chan<- error, tempdir string, logger logging.L
...
 	man := builder.GetManifest()
 
-	fromRC, err := rcStore.Create(man, klabels.Everything(), "some_az", "some_cn", nil, nil)
+	fromRC, err := rcStore.Create(man, klabels.Everything(), "some_az", "some_cn", nil, nil, "some_strategy")
 	if err != nil {
 		errCh <- util.Errorf("could not create RC for replica transfer test: %s", err)
 		return
 	}
 
-	toRC, err := rcStore.Create(man, klabels.Everything(), "some_az", "some_cn", nil, nil)
```
mpuncel: mental note to myself that we'll probably have to make this a real strategy once the RC farm actually looks at this (the integration test runs a real RC farm!)

spenceral: I thought about putting `PetStrategy` in instead of `"some_strategy"` but didn't know if it was overkill. I no longer check if a string is a valid strategy by doing a map lookup so we may not need to make it an actual strategy if we just let it fall through to the default case/branch

PR #932, opened 2017-10-06
Add RC Status store

spenceral:
As part of better host idling, RC's will keep track of their status to know when they are performing a node transfer. To do this they will need to Get the status to see if a node transfer is in progress, Set the status when they are performing a node transfer, and Delete the status when they are finished.  Those three behaviors are implemented and tested.

### DIFF
--- a/pkg/store/consul/statusstore/rcstatus/status.go
+++ b/pkg/store/consul/statusstore/rcstatus/status.go
@@ -1,0 +1,38 @@
+package rcstatus
+
+import (
+	"encoding/json"
+
+	"github.com/square/p2/pkg/store/consul/statusstore"
+	"github.com/square/p2/pkg/types"
+	"github.com/square/p2/pkg/util"
+)
+
+type RCStatus struct {
+	NodeTransfer NodeTransfer `json:"node_transfer"`
+}
+
+type NodeTransfer struct {
+	OldNode types.NodeName `json:"old_node"`
+	NewNode types.NodeName `json:"new_node"`
+}
+
+func statusToRCStatus(rawStatus statusstore.Status) (RCStatus, error) {
+	var rcStatus RCStatus
+
+	err := json.Unmarshal(rawStatus.Bytes(), &rcStatus)
+	if err != nil {
+		return RCStatus{}, util.Errorf("Could not unmarshal raw status as rc status: %s", err)
+	}
+
+	return rcStatus, nil
+}
+
+func rcStatusToStatus(rcStatus RCStatus) (statusstore.Status, error) {
+	bytes, err := json.Marshal(rcStatus)
+	if err != nil {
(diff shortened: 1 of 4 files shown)


pkg/store/consul/statusstore/rcstatus/status.go
```diff
@@ -0,0 +1,38 @@
...
+import (
+	"encoding/json"
+
+	"github.com/square/p2/pkg/store/consul/statusstore"
+	"github.com/square/p2/pkg/types"
+	"github.com/square/p2/pkg/util"
+)
+
+type RCStatus struct {
+	NodeTransfer NodeTransfer `json:"node_transfer"`
```
mpuncel: I wonder if we want to make this a pointer or a list so it's more obvious when it's empty? otherwise we can check that `OldNode == "" && NewNode == ""`

pkg/store/consul/statusstore/rcstatus/store_test.go
```diff
@@ -0,0 +1,92 @@
...
+	}
+
+	if !statusstore.IsNoStatus(err) {
+		t.Errorf("Expected error to be NoStatus but was %s", err)
+	}
+}
+
+func newFixture() *ConsulStore {
+	return &ConsulStore{
+		statusStore: statusstoretest.NewFake(),
```
mpuncel: Heads up for next time, lately i've been using `consulutil.NewFixture(t)` everywhere to get a consul client and then just call the normal initialier (e.g. `statusstore.NewConsul(fixture.Client)` or whatever it's called. The reason for this is that we'll likely add operations that use transaction and we're never going to make this fake status store implement transactions.

mpuncel: (this is totally fine for the functions you have so not a PR blocker)

mpuncel:
looks great! just raised some thoughts for the future but they're thinks we can address in future PRs as this functionality is actually used

pkg/store/consul/statusstore/rcstatus/status.go
spenceral: Yeah thats a good question. I think it depends on how we end up using it. Will the RCStatus always exist and the we'll check to see if NodeTransfer is empty, or will we delete the RCStatus when we're done so we'll be checking for a `NoStatus` error from Consul to see if a node transfer is not in progress?

mpuncel: Once an RC status record is created I'd think it should exist for the whole lifetime of the RC, and we'll just empty out the `NodeTransfer` part when the transfer is done. The question is just whether the `NodeTransfer` structure itself should be `nil` (which is only allowed in Go if you use a pointer) or if it should have `""` for both of its fields

spenceral: Okay in that case I agree that we should use a pointer

PR #933, opened 2017-10-09
Spenceral/implement new states

spenceral:
When RCs discovers they are scheduled to ineligible nodes, those with the cattle strategy make a call to `scheduler.AllocateNodes()` and write a node transfer status. Those with the old strategy (pet strategy) will alert. 

There are currently no tests. Thorough integration tests will be written once the the entire new replication controller behavior has finished being implemented.

### DIFF
--- a/pkg/rc/replication_controller.go
+++ b/pkg/rc/replication_controller.go
@@ -19,6 +19,8 @@ import (
 	"github.com/square/p2/pkg/scheduler"
 	"github.com/square/p2/pkg/store/consul"
 	"github.com/square/p2/pkg/store/consul/rcstore"
+	"github.com/square/p2/pkg/store/consul/statusstore"
+	"github.com/square/p2/pkg/store/consul/statusstore/rcstatus"
 	"github.com/square/p2/pkg/store/consul/transaction"
 	"github.com/square/p2/pkg/types"
 	"github.com/square/p2/pkg/util"
@@ -103,6 +105,7 @@ type replicationController struct {
 	logger logging.Logger
 
 	consulStore   consulStore
+	rcStatusStore rcstatus.ConsulStore
 	auditLogStore AuditLogStore
 	txner         transaction.Txner
 	rcWatcher     ReplicationControllerWatcher
@@ -118,6 +121,7 @@ type ReplicationControllerWatcher interface {
 func New(
 	fields fields.RC,
 	consulStore consulStore,
+	rcStatusStore rcstatus.ConsulStore,
 	auditLogStore AuditLogStore,
 	txner transaction.Txner,
 	rcWatcher ReplicationControllerWatcher,
@@ -135,6 +139,7 @@ func New(
 
 		logger:        logger,
 		consulStore:   consulStore,
+		rcStatusStore: rcStatusStore,
 		auditLogStore: auditLogStore,
 		txner:         txner,
 		rcWatcher:     rcWatcher,
@@ -498,11 +503,7 @@ func (rc *replicationController) checkForIneligible(current types.PodLocations, 
 	}
 
 	if len(ineligibleCurrent) > 0 {
-		errMsg := fmt.Sprintf("RC has scheduled %d ineligible nodes: %s", len(ineligibleCurrent), ineligibleCurrent)
-		err := rc.alerter.Alert(rc.alertInfo(errMsg))
-		if err != nil {
-			rc.logger.WithError(err).Errorln("Unable to send alert")
-		}
+		rc.transferNodes(ineligibleCurrent)
 	}
 }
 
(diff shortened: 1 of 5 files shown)


pkg/store/consul/kv.go
```diff
@@ -28,8 +28,9 @@ import (
 const (
 	TTL = 60 * time.Second
 
-	// Don't change this, it affects where pod status keys are read and written from
+	// Don't change this, it affects where status keys are read and written from
 	PreparerPodStatusNamespace statusstore.Namespace = "preparer"
+	RCStatusNamespace          statusstore.Namespace = "rc_farm"
```
rudle: I gotta nit the name because this will be hard to change. Avoid contractions, go for the explicit "replication_controller"

pkg/rc/replication_controller.go
```diff
@@ -316,6 +308,72 @@ func (rc *replicationController) addPods(current types.PodLocations, eligible []
...
+	return err
+}
+
+func (rc *replicationController) doNotEnoughCattleNodes(currentNodes []types.NodeName, eligibleNodes []types.NodeName, numScheduledNodes int) error {
+	rcStatus, _, err := rc.rcStatusStore.Get(rc.ID())
+	isNoNodeTransferInProgress := (err != nil && statusstore.IsNoStatus(err)) ||
+		(err == nil && rcStatus.NodeTransfer == nil)
+	if isNoNodeTransferInProgress {
+		newNodes, err := rc.scheduler.AllocateNodes(rc.Manifest, rc.NodeSelector, len(currentNodes)+1)
+		if err != nil || len(newNodes) < 1 {
```
rudle: We need to capture this error message if it's not nil.

pkg/rc/replication_controller.go
```diff
@@ -316,6 +308,72 @@ func (rc *replicationController) addPods(current types.PodLocations, eligible []
...
+	case txnErr != nil:
+		return txnErr
+	case !ok:
+		return util.Errorf("could not schedule pods due to transaction violation: %s", transaction.TxnErrorsToString(resp.Errors))
+	}
+
+	return err
+}
+
+func (rc *replicationController) doNotEnoughCattleNodes(currentNodes []types.NodeName, eligibleNodes []types.NodeName, numScheduledNodes int) error {
```
rudle: I don't see any code to check that the RC lock is still held. Is this code run inside of the RC lock's critical section at all?

pkg/store/consul/kv.go
spenceral: Will do

pkg/rc/replication_controller.go
spenceral: Ah I see that the lock is used when accessing the fields in the `fields.RC` struct. I'll update to use the lock.

pkg/rc/replication_controller.go
spenceral: Will do

pkg/rc/replication_controller.go
```diff
@@ -316,6 +318,70 @@ func (rc *replicationController) addPods(current types.PodLocations, eligible []
...
+	if err != nil {
+		return util.Errorf("Unable to allocate nodes: %s", err)
+	}
+
+	return nil
+}
+
+func (rc *replicationController) allocateCattleNodes(currentNodes []types.NodeName, eligibleNodes []types.NodeName) error {
+	rcStatus, _, err := rc.rcStatusStore.Get(rc.ID())
+	isNoNodeTransferInProgress := (err != nil && statusstore.IsNoStatus(err)) ||
```
mpuncel: i think you're not handling the `err !=nil && !statusstore.IsNoStatus(err)` case here (in other words, when there is a real error that prevents you from reading the status.

When there are a lot of cases I like using switch statements. you could do something like this:

```
switch {
case err == nil || statusstore.IsNoStatus(err):
  isNoNodeTransferInProgress = true
default:
  return err
}
```

pkg/rc/replication_controller.go
```diff
@@ -316,6 +318,70 @@ func (rc *replicationController) addPods(current types.PodLocations, eligible []
...
+				rc.logger.WithError(err).Errorln("Unable to send alert")
+			}
+
+			return util.Errorf(errMsg)
+		}
+
+		// Right now we do not support multiple node transfers, so we just
+		// write the first
+		rcStatus.NodeTransfer.NewNode = newNodes[0]
+		err = rc.rcStatusStore.Set(rc.ID(), rcStatus)
```
mpuncel: I don't think it's strictly necessary but it'd probably be good practice if we used a transaction here. either a "check-and-set" using the modify index from the previous `Get()` call or do it in a full on transaction (would be a check-and-set either way)

pkg/rc/replication_controller.go
```diff
@@ -316,6 +318,70 @@ func (rc *replicationController) addPods(current types.PodLocations, eligible []
...
+		if err != nil {
+			return util.Errorf("Could not write new node to store: %s", err)
+		}
+	} else if err != nil {
```
mpuncel: oh you are handling the error after all! my bad for not seeing that. I do subscribe to the philosophy of the golang style guide though: https://github.com/golang/go/wiki/CodeReviewComments#indent-error-flow . basically it's good to handle the errors as soon as they show up

pkg/rc/replication_controller.go
```diff
@@ -295,7 +293,11 @@ func (rc *replicationController) addPods(current types.PodLocations, eligible []
 				return util.Errorf("could not schedule pods due to transaction violation: %s", transaction.TxnErrorsToString(resp.Errors))
 			}
 
-			return util.Errorf(errMsg)
+			if err != nil {
+				return util.Errorf("Could not meet desires by allocating nodes: %s. Scheduled on %d nodes instead.", err, i)
+			}
+
+			return nil
```
mpuncel: I think you don't want to return here because you haven't done the `schedule()` yet

pkg/rc/replication_controller.go
```diff
@@ -277,14 +282,7 @@ func (rc *replicationController) addPods(current types.PodLocations, eligible []
...
 		if len(possibleSorted) < i+1 {
-			errMsg := fmt.Sprintf(
-				"Not enough nodes to meet desire: %d replicas desired, %d currentNodes, %d eligible. Scheduled on %d nodes instead.",
-				rc.ReplicasDesired, len(currentNodes), len(eligible), i,
-			)
-			err := rc.alerter.Alert(rc.alertInfo(errMsg))
-			if err != nil {
-				rc.logger.WithError(err).Errorln("Unable to send alert")
-			}
+			err := rc.allocateNodes(currentNodes, eligible)
```
mpuncel: If the goal of this PR is to allow node transfers, I think the right place for this is actually within `checkForIneligible()` on line 237 (of course that function should be renamed). "ineligible" means that the RC is scheduled on a node that it shouldn't be scheduled on, it seems like as written the goal of this code is being able to scale an RC beyond what has been allocated. We shouldn't even need a node transfer for that case since it's simply adding more!

pkg/rc/replication_controller.go
spenceral: really useful link. There are a few things I should refactor based on those recommendations

pkg/rc/replication_controller.go
```diff
@@ -562,6 +562,10 @@ func (rc *replicationController) checkForIneligible(current types.PodLocations,
 	}
 
 	if len(ineligibleCurrent) > 0 {
+		if rc.AllocationStrategy == fields.CattleStrategy {
+			rc.allocateCattleNodes(current.Nodes(), eligible)
+			return
```
spenceral: @mpuncel

mpuncel: yeah that looks like the right spot

pkg/rc/replication_controller.go
```diff
@@ -316,6 +318,70 @@ func (rc *replicationController) addPods(current types.PodLocations, eligible []
...
+	return nil
+}
+
+func (rc *replicationController) allocateCattleNodes(currentNodes []types.NodeName, eligibleNodes []types.NodeName) error {
+	rcStatus, _, err := rc.rcStatusStore.Get(rc.ID())
+	isNoNodeTransferInProgress := (err != nil && statusstore.IsNoStatus(err)) ||
+		(err == nil && rcStatus.NodeTransfer == nil)
+	if isNoNodeTransferInProgress {
+		rc.mu.Lock()
+		newNodes, err := rc.scheduler.AllocateNodes(rc.Manifest, rc.NodeSelector, len(currentNodes)+1)
```
mpuncel: I think you should just pass `1` here rather than `len(currentNodes)+1`. We could implement the server to know what you mean but that will require more calls between p2-resource-api and hoist and also the RC loop here has access to the full list of nodes, so it won't do a double allocate

spenceral: Sounds good, I just misunderstood the intention of that parameter. I remember debating between a couple of interpretations. I should've just asked 🙃

pkg/rc/replication_controller.go
spenceral: Are you suggesting that we should not be allocating nodes here at all? I took it as there are two cases in which we allocate: 1) the ineligible case, and 2) the not enough nodes case where we simply add one without performing a node transfer.

pkg/rc/replication_controller.go
mpuncel: yeah at some point I thought we needed the total, then realized we didn't! sorry for the churn

pkg/rc/replication_controller.go
```diff
@@ -614,3 +615,82 @@ func (rc *replicationController) unschedule(txn *auditingTransaction, node types
...
+	}
+	if err != nil {
+		return util.Errorf("Unable to transfer nodes: %s", err)
+	}
+
+	return nil
+}
+
+func (rc *replicationController) transferCattleNodes(ineligibleCurrent []types.NodeName) error {
+	if len(ineligibleCurrent) < 1 {
```
spenceral: Note: we may change this if we decide to bring this node allocation functionality to `addPods()` since there would be no current ineligible nodes in that case

spenceral:
Added tests to wrap this up, but the implementation will likely change once we finalize our node transfer strategy.

pkg/rc/replication_controller.go
```diff
@@ -498,11 +503,7 @@ func (rc *replicationController) checkForIneligible(current types.PodLocations,
 	}
 
 	if len(ineligibleCurrent) > 0 {
-		errMsg := fmt.Sprintf("RC has scheduled %d ineligible nodes: %s", len(ineligibleCurrent), ineligibleCurrent)
-		err := rc.alerter.Alert(rc.alertInfo(errMsg))
-		if err != nil {
-			rc.logger.WithError(err).Errorln("Unable to send alert")
-		}
+		rc.transferNodes(ineligibleCurrent)
```
mpuncel: `errcheck` in the build noticed that you're not checking the error that `transferNodes()` returns here

mpuncel: I found this by running `errcheck -ignoretests github.com/square/p2/pkg/... | grep -v defer` which is the line that failed in the build (but you have to unwrap it from the `exit() | wc -l`)

mpuncel:
I thought I approved this on Friday! sorry about that!

PR #934, opened 2017-10-09
Env var for restart timeout

rudle:
An internal request for exposing the restart_timeout has prompted this change. The alternative is having the customer parse the manifest (bad) & copy the knowledge of the default timeout (very bad!)

### DIFF
--- a/pkg/pods/pod.go
+++ b/pkg/pods/pod.go
@@ -42,13 +42,14 @@ var (
 )
 
 const (
-	ConfigPathEnvVar         = "CONFIG_PATH"
-	LaunchableIDEnvVar       = "LAUNCHABLE_ID"
-	LaunchableRootEnvVar     = "LAUNCHABLE_ROOT"
-	PodIDEnvVar              = "POD_ID"
-	PodHomeEnvVar            = "POD_HOME"
-	PodUniqueKeyEnvVar       = "POD_UNIQUE_KEY"
-	PlatformConfigPathEnvVar = "PLATFORM_CONFIG_PATH"
+	ConfigPathEnvVar               = "CONFIG_PATH"
+	LaunchableIDEnvVar             = "LAUNCHABLE_ID"
+	LaunchableRootEnvVar           = "LAUNCHABLE_ROOT"
+	PodIDEnvVar                    = "POD_ID"
+	PodHomeEnvVar                  = "POD_HOME"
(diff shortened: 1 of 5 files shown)


pkg/pods/pod.go
```diff
@@ -42,13 +42,14 @@ var (
...
-	PodUniqueKeyEnvVar       = "POD_UNIQUE_KEY"
-	PlatformConfigPathEnvVar = "PLATFORM_CONFIG_PATH"
+	ConfigPathEnvVar               = "CONFIG_PATH"
+	LaunchableIDEnvVar             = "LAUNCHABLE_ID"
+	LaunchableRootEnvVar           = "LAUNCHABLE_ROOT"
+	PodIDEnvVar                    = "POD_ID"
+	PodHomeEnvVar                  = "POD_HOME"
+	PodUniqueKeyEnvVar             = "POD_UNIQUE_KEY"
+	PlatformConfigPathEnvVar       = "PLATFORM_CONFIG_PATH"
+	LaunchableRestartTimeoutEnvVar = "RESTART_TIMOUT"
```
mpuncel: s/TIMOUT/TIMEOUT

pkg/pods/pod.go
```diff
@@ -602,7 +603,11 @@ func (pod *Pod) setupConfig(manifest manifest.Manifest, launchables []launch.Lau
 		if err != nil {
 			return err
 		}
-		err = writeEnvFile(launchable.EnvDir(), "LAUNCHABLE_ROOT", launchable.InstallDir(), uid, gid)
+		err = writeEnvFile(launchable.EnvDir(), LaunchableRootEnvVar, launchable.InstallDir(), uid, gid)
+		if err != nil {
+			return err
+		}
+		err = writeEnvFile(launchable.EnvDir(), LaunchableRestartTimeoutEnvVar, fmt.Sprintf("%s", launchable.GetRestartTimeout()*time.Second), uid, gid)
```
mpuncel: have you checked how this turns into a string? seems like it's already a `time.Duration` so you probably don't need to `*time.Second`

rudle: fair cop. Here's what go thinks: https://play.golang.org/p/wJQzqbhN-T

I like the explicit int64() conversion - WDYT?

rudle:
PTAL

PR #935, opened 2017-10-10
Fix usage of timers in daemon sets.

mpuncel:
This will hopefully fix a bug we've observed where the daemon set
handler responds every 5 minutes as if the timer has fired. This might
be occurring because we don't properly drain the timer's channel after
we stop it.


rudle:
Can you add a log line to the timer.C leg of the select? That would prove or disprove this theory.

PR #936, opened 2017-10-10
Refactor rc status store

spenceral:
Add transactional check and set CASTxn(). Also update
names based on go code review guide.

Cleaned up rc status store text, added CASTxn to use in the replication controller, and updated rc status struct name based on https://github.com/golang/go/wiki/CodeReviewComments#package-names. This commit also includes the NodeTransfer as a pointer change.

### DIFF
--- a/pkg/store/consul/statusstore/rcstatus/store.go
+++ b/pkg/store/consul/statusstore/rcstatus/store.go
@@ -1,6 +1,8 @@
 package rcstatus
 
 import (
+	"context"
+
 	"github.com/square/p2/pkg/rc/fields"
 	"github.com/square/p2/pkg/store/consul/statusstore"
 	"github.com/square/p2/pkg/util"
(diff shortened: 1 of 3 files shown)


pkg/store/consul/statusstore/rcstatus/store.go
```diff
@@ -25,41 +27,46 @@ func NewConsul(statusStore statusstore.Store, namespace statusstore.Namespace) C
...
+	rawStatus, err := statusToRawStatus(status)
 	if err != nil {
 		return err
 	}
 
 	return c.statusStore.SetStatus(statusstore.RC, statusstore.ResourceID(rcID), c.namespace, rawStatus)
 }
 
-func (c ConsulStore) Delete(rcID fields.ID) error {
+func (c ConsulStore) CASTxn(ctx context.Context, rcID fields.ID, modifyIndex uint64, status Status) error {
```
mpuncel: might be worth a test making sure that this function mutates a status when you pass the correct index, and fails to when you pass the wrong one

spenceral:
```
func (s *FakeStatusStore) CASStatus(
	ctx context.Context,
	t statusstore.ResourceType,
	id statusstore.ResourceID,
	namespace statusstore.Namespace,
	status statusstore.Status,
	modifyIndex uint64,
) error {
	return errors.New("CASStatus uses transactions which requires a real consul instance")
}
``` 
Looks like to test `CASTxn()`  I'll need to create a new fake then

spenceral: Done.

mpuncel:
Test looks great! just that minor point about stopping the fixture

PR #937, opened 2017-10-10
Mpuncel/scheduler grpc client

mpuncel:


pkg/grpc/scheduler/client/client.go
```diff
@@ -70,5 +70,20 @@ func (c *Client) AllocateNodes(man manifest.Manifest, nodeSelector klabels.Selec
...
+	}
+
+	req := &scheduler_protos.DeallocateNodesRequest{
+		NodeSelector:  selector.String(),
+		NodesReleased: nodeStrings,
+	}
+
+	_, err := c.schedulerClient.DeallocateNodes(context.TODO(), req)
+	if err != nil {
+		return util.Errorf("DeallocateNOdes gRPC call failed: %s", err)
```
spenceral: small typo "DeallocateNOdes"

pkg/grpc/scheduler/client/client.go
```diff
@@ -1,18 +1,48 @@
...
+	if err != nil {
+		return nil, util.Errorf("could not marshal manifest for EligibleNodes gRPC request: %s", err)
+	}
+
+	req := &scheduler_protos.EligibleNodesRequest{
+		Manifest:     string(manifestStr),
+		NodeSelector: sel.String(),
+	}
+
+	resp, err := c.schedulerClient.EligibleNodes(context.TODO(), req)
```
spenceral: I'm not an expert on contexts. Is context.TODO() something you anticipate replacing?

mpuncel: it's the same thing as `context.Background()` except for there are some tools that can find `context.TOD()`. you're supposed to use it in places where the context might be threaded through someday, e.g. we add a `context.Context` argument to everything on the `Scheduler` interface. i'm not sure if we will do that because I don't think we'd ever want to cancel the gRPC, perhaps i should just just `context.Background()`

pkg/grpc/scheduler/client/client.go
```diff
@@ -45,8 +45,28 @@ func (c *Client) EligibleNodes(man manifest.Manifest, sel klabels.Selector) ([]t
...
-	return nil, util.Errorf("AllocateNodes() not yet implemented")
+func (c *Client) AllocateNodes(man manifest.Manifest, nodeSelector klabels.Selector, allocationCount int) ([]types.NodeName, error) {
+	manifestStr, err := man.Marshal()
+	if err != nil {
+		return nil, util.Errorf("could not marshal manifest for AllocateNodes gRPC request: %s", err)
+	}
+	req := &scheduler_protos.AllocateNodesRequest{
+		Manifest:       string(manifestStr),
+		NodeSelector:   nodeSelector.String(),
+		NodesRequested: int64(allocationCount),
```
spenceral: nit: thoughts on changing the `allocationCount` param name to `nodesRequested`? It'll be consistent with the proto message and `allocationCount` did confuse me in my last PR. I thought it might refer to the total allocations desired. You're choice

mpuncel: yeah i think `nodesRequested` will be clearer

PR #938, opened 2017-10-16
Mpuncel/daemon set redesign

mpuncel:


mpuncel:
I think there might be a bug in how manifest updates are handled here

pkg/replication/replication.go
```diff
@@ -257,9 +315,36 @@ func (r *replication) Enact() {
 	}
 	sort.Sort(order)
 
-	nodeQueue := make(chan types.NodeName)
+	nodeQueue := r.nodeQueue
+	if nodeQueue == nil {
```
rudle: How would it come to pass that nodeQueue is not set? Is this the code that implements the "old" behavior?

mpuncel: yeah, it will be unset if you use `InitializeReplication()` rather than `InitializeDaemonSetReplication`. I can make this more explicit here with a `r.UseNodeQueue()` or something like that

pkg/ds/daemon_set.go
```diff
@@ -300,24 +303,41 @@ func (ds *daemonSet) WatchDesires(
 		ds.publishStatus(ctx)
 	}()
 
+	nodesToAdd := make(chan []types.NodeName, 100)
```
rudle: This could be 10x as big without causing problems right?

mpuncel: yeah I imagine so

mpuncel: i'll comment it too

pkg/ds/daemon_set.go
```diff
@@ -410,24 +465,53 @@ func (ds *daemonSet) WatchDesires(
...
+				nodesToAdd <- addedNodes
 
 			case <-timer.C:
+				if ds.Disabled {
+					paused = true
+					continue
+				}
+
+				if paused {
+					unpauseReplication <- struct{}{}
```
rudle: doesn't this have the side effect of unpausing replication? It seems at adds with the comment below

pkg/ds/daemon_set.go
```diff
@@ -450,51 +534,17 @@ func (ds *daemonSet) reportEligible() error {
...
-			return util.Errorf("Unable to remove pods from intent tree: %v", err)
-		}
-	}
-
-	return nil
-}
-
 // addPods schedules pods for all unscheduled nodes selected by ds.nodeSelector
-func (ds *daemonSet) addPods() error {
+func (ds *daemonSet) addPods() ([]types.NodeName, error) {
```
rudle: Is it too much to ask for a rename of this function? It has no side-effects, it's just doing computation.

mpuncel: not at all!

pkg/ds/daemon_set.go
mpuncel: it's the comment that's wrong, it does make sense to unpause here because the daemon set is not disabled (since we would have hit the `continue` above otherwise)

mpuncel: I think technically i shouldn't have to check pausing/unpausing here because that would only happen on the update channel, but there already was a disabled check so just keeping in that spirit to change as few things as possible

mpuncel:
I realized i forgot to test pausing/unpausing in my burn-in test, once I do that i'm confident in merging it

PR #943, opened 2017-10-18
Support pod level cgroup in p2-exec

brianrli:
p2-exec takes a "pod" argument, that creates a parent "pod level" cgroup wrapping the launchable

/cgroup/CPU/p2/nodename/{pod}/{cgroupname}
/cgroup/Memory/p2/nodename/{pod}/{cgroupname}

as of right now, it pulls the parameters for this "pod level" cgroup from _RESOURCE_LIMITS_PATH_, searching for a key of the same name

### DIFF
--- a/bin/p2-exec/main.go
+++ b/bin/p2-exec/main.go
@@ -16,6 +16,7 @@ import (
 	"golang.org/x/sys/unix"
 
 	"github.com/square/p2/pkg/cgroups"
+	"github.com/square/p2/pkg/pods"
 	"github.com/square/p2/pkg/util"
 	"github.com/square/p2/pkg/version"
 	"gopkg.in/alecthomas/kingpin.v2"
@@ -39,7 +40,7 @@ var (
 	workDir        = kingpin.Flag("workdir", "Set working directory.").Short('w').String()
 	umask          = kingpin.Flag("umask", "Set the process umask. Use octal notation ex. 0022").Short('m').Default(umaskDefault).String()
 	umaskDefault   = ""
-	podID          = kingpin.Flag("pod-cgroup", "Nest this launchable cgroup beneath a pod level cgroup. Looks in $PLATFORM_CONFIG_PATH_TWO for parameters").Short('p').String()
+	podID          = kingpin.Flag("pod-cgroup", "Nest this launchable cgroup beneath a pod level cgroup. Looks in $RESOURCE_LIMITS_PATH for parameters").Short('p').String()
 
 	cmd = kingpin.Arg("command", "the command to execute").Required().Strings()
 )
@@ -83,7 +84,7 @@ func main() {
 
 	cgroupPath := ""
 	if *podID != "" {
-		cgConfig, err := cgGetConfig("RESOURCE_LIMITS", "")
+		cgConfig, err := cgGetPodConfig(pods.ResourceLimitsConfigPathEnvVar)
 		if err != nil {
 			log.Fatal(err)
 		}
@@ -106,13 +107,13 @@ func main() {
 		if err != nil && !os.IsExist(err) {
 			log.Fatal(err)
 		}
-		cgroupPath = "p2" + "/" + node
+		cgroupPath = filepath.Join(cgroupPath, node)
 
 		err = cgCreate(*podID, cgroupPath, cgConfig)
 		if err != nil {
 			log.Fatalf("Could not create pod %v cgroup %v\n", *podID, err)
 		}
-		cgroupPath = cgroupPath + "/" + *podID
+		cgroupPath = filepath.Join(cgroupPath, *podID)
 	}
 
 	if *launchableName == "" && *cgroupName != "" {
@@ -122,7 +123,7 @@ func main() {
 		log.Fatalf("Specified launchable name %q, but no cgroup name was specified", *launchableName)
 	}
 	if *launchableName != "" && *cgroupName != "" {
-		cgPlatConfig, err := cgGetConfig("PLATFORM_CONFIG_PATH", *launchableName)
+		cgPlatConfig, err := cgGetLaunchableConfig(pods.PlatformConfigPathEnvVar, *launchableName)
 		if err != nil {
 			log.Fatal(err)
 		}
@@ -277,40 +278,47 @@ func cgCreate(cgroupName string, targetDirectory string, config cgroups.Config) 
 	return cg.AddPID(config.Name, 0)
 }
 
-// given ENV variable "configPath" return parameters for the new cgroup being created
-func cgGetConfig(configPath string, configKey string) (cgroups.Config, error) {
+func cgGetPodConfig(configPath string) (cgroups.Config, error) {
 	if config := os.Getenv(configPath); config != "" {
 		confBuf, err := ioutil.ReadFile(config)
 		if err != nil {
 			return cgroups.Config{}, err
 		}
-		if configPath == "RESOURCE_LIMITS" {
-			cgMap := make(map[string]cgroups.Config)
-			err = yaml.Unmarshal(confBuf, cgMap)
-			if err != nil {
-				return cgroups.Config{}, err
-			}
-			if _, ok := cgMap["cgroup"]; !ok {
-				return cgroups.Config{}, util.Errorf("Key %q in %s does not contain cgroup parameters\n", configKey, configPath)
-			}
-			cgConfig := cgMap["cgroup"]
-			return cgConfig, nil
-		} else {
-			cgMap := make(map[string]map[string]cgroups.Config)
-			err = yaml.Unmarshal(confBuf, cgMap)
-			if err != nil {
-				return cgroups.Config{}, err
-			}
-
-			if _, ok := cgMap[configKey]; !ok {
-				return cgroups.Config{}, util.Errorf("Key %q not found in %s", configKey, configPath)
-			}
-			if _, ok := cgMap[configKey]["cgroup"]; !ok {
-				return cgroups.Config{}, util.Errorf("Key %q in %s does not contain cgroup parameters\n", configKey, configPath)
-			}
-			cgConfig := cgMap[configKey]["cgroup"]
-			return cgConfig, nil
+		cgMap := make(map[string]cgroups.Config)
+		err = yaml.Unmarshal(confBuf, cgMap)
+		if err != nil {
+			return cgroups.Config{}, err
 		}
+		if _, ok := cgMap["cgroup"]; !ok {
+			return cgroups.Config{}, util.Errorf("%s does not contain cgroup parameters\n", configPath)
+		}
+		cgConfig := cgMap["cgroup"]
+		return cgConfig, nil
+	} else {
+		return cgroups.Config{}, util.Errorf("No %s found in environment", configPath)
+	}
+}
+
+func cgGetLaunchableConfig(configPath string, configKey string) (cgroups.Config, error) {
+	if config := os.Getenv(configPath); config != "" {
+		confBuf, err := ioutil.ReadFile(config)
+		if err != nil {
+			return cgroups.Config{}, err
+		}
+		cgMap := make(map[string]map[string]cgroups.Config)
+		err = yaml.Unmarshal(confBuf, cgMap)
+		if err != nil {
+			return cgroups.Config{}, err
+		}
+
+		if _, ok := cgMap[configKey]; !ok {
+			return cgroups.Config{}, util.Errorf("Key %q not found in %s", configKey, configPath)
+		}
+		if _, ok := cgMap[configKey]["cgroup"]; !ok {
+			return cgroups.Config{}, util.Errorf("Key %q in %s does not contain cgroup parameters\n", configKey, configPath)
+		}
+		cgConfig := cgMap[configKey]["cgroup"]
+		return cgConfig, nil
 	} else {
 		return cgroups.Config{}, util.Errorf("No %s found in environment", configPath)
 	}
--- a/bin/p2-exec/main.go
+++ b/bin/p2-exec/main.go
@@ -39,7 +39,7 @@ var (
 	workDir        = kingpin.Flag("workdir", "Set working directory.").Short('w').String()
 	umask          = kingpin.Flag("umask", "Set the process umask. Use octal notation ex. 0022").Short('m').Default(umaskDefault).String()
 	umaskDefault   = ""
-	podID          = kingpin.Flag("pod-cgroup", "Nest this launchable cgroup beneath a pod level cgroup. Looks in $CONFIG_PATH for parameters").Short('p').String()
+	podID          = kingpin.Flag("pod-cgroup", "Nest this launchable cgroup beneath a pod level cgroup. Looks in $PLATFORM_CONFIG_PATH_TWO for parameters").Short('p').String()
 
 	cmd = kingpin.Arg("command", "the command to execute").Required().Strings()
 )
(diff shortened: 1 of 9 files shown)


bin/p2-exec/main.go
```diff
@@ -214,30 +259,48 @@ func loadEnvDir(dir string) error {
 // a cgroup with the name <cgroupname> will be created, using the parameters for
 // <launchablename> found in the platform configuration
 // then, the current PID will be added to that cgroup
-func cgEnter(platconf, launchableName, cgroupName string) error {
-	platconfBuf, err := ioutil.ReadFile(platconf)
+func cgEnter(config, configKey, cgroupName string, params cgroupParams) error {
+	var configPath string
```
rudle: is this local variable used below?

bin/p2-exec/main.go
```diff
@@ -80,6 +92,27 @@ func main() {
...
 
+	if *podID != "" {
+		if *launchableName != "" {
+			log.Fatalf("Specified pod name %q, but no launchable name was specified", *podID)
+		}
+		if *cgroupName != "" {
+			log.Fatalf("Specified pod name %q, but no cgroupname name was specified", *podID)
+		}
+
+		if config := os.Getenv("CONFIG_PATH"); config != "" {
```
rudle: What is the value of this environment variable here at this point? I suspect this is the preparer's CONFIG_PATH, not the pod we are deploying.

Further to this point, is this approach sound in cases where we've not deployed this pod before? How about when we change the values of the cgroup config? I think these cases are handled already, but I'd like to be certain.

mpuncel: it's going to have `CONFIG_PATH` of the pod being deployed because of the `os.Setenv` loop on line 82, but maybe that relationship should be made more explicit. Probably worth crashing or setting some kind of cgroup default if there's no `CONFIG_PATH` in the env dir?

bin/p2-exec/main.go
```diff
@@ -80,6 +92,27 @@ func main() {
...
+			params := cgroupParams{
+				Type: PodCGroup,
+			}
+			err := cgEnter(config, *podID, *podID, params)
+			if err != nil {
+				log.Fatal(err)
+			}
+		} else {
+			log.Fatal("No CONFIG_PATH, cannot determine pod-level cgroup")
```
mpuncel: It's worth noting that sometimes we tell people to use `p2-exec` to run things outside of a pod context, so this might not be set. maybe that's okay but it's redefining p2-exec as only a thing that should be scheduled as a runit service for a P2 deployed process

bin/p2-exec/main.go
```diff
@@ -214,30 +259,48 @@ func loadEnvDir(dir string) error {
 // a cgroup with the name <cgroupname> will be created, using the parameters for
 // <launchablename> found in the platform configuration
 // then, the current PID will be added to that cgroup
-func cgEnter(platconf, launchableName, cgroupName string) error {
-	platconfBuf, err := ioutil.ReadFile(platconf)
+func cgEnter(config, configKey, cgroupName string, params cgroupParams) error {
+	var configPath string
+	if params.Type == PodCGroup {
+		configPath = "CONFIG_PATH"
```
mpuncel: `CONFIG_PATH` is where the user-specified config in their pod manifest ends up. I think the pod-level cgroup settings should go somewhere else in the manifest spec shouldn't it? so probably we want to change the schema of `PLATFORM_CONFIG_PATH` to include pod sizes and launchable sizes

mpuncel: btw that will probably unfortunately require a new env variable like `PLATFORM_CONFIG_V2` because currently the schema is a map of launchable name to cgroup config. Probably the top level keys should be `pod:` and `launchables:` something like that

bin/p2-exec/main.go
brianrli: Oh, I should have squashed all these commits into one I think; the two new commits are the rewritten version. Without the --pod-cgroup flag its behavior remains the same, so I'm not sure I actually understand how this redefines p2-exec?

bin/p2-exec/main.go
```diff
@@ -80,20 +81,55 @@ func main() {
 		}
 	}
 
+	cgroupPath := ""
+	if *podID != "" {
+		cgConfig, err := cgGetConfig("RESOURCE_LIMITS", "")
+		if err != nil {
+			log.Fatal(err)
```
mpuncel: we should consider whether we want this to actually be an error. I suspect that we've trained other people to use p2-exec in contexts where this env var won't exist. Maybe the preparer should write a `--enforce-resource-limit` flag that turns on this strict code?

bin/p2-exec/main.go
```diff
@@ -80,20 +81,55 @@ func main() {
...
+		}
+
+		cg, err := cgroups.Find()
+		if err != nil {
+			log.Fatal(err)
+		}
+
+		err = os.MkdirAll(filepath.Join(cg.CPU, "p2", node), 0755)
```
mpuncel: this seems like something that belongs in the `cgroups` package so we can reuse it for other things

bin/p2-exec/main.go
mpuncel: actually maybe the fact that pod ID is provided already covers what i'm saying. Nobody would pass the pod ID in a one-off invocation

bin/p2-exec/main.go
```diff
@@ -214,31 +250,22 @@ func loadEnvDir(dir string) error {
...
-	if err != nil {
-		return err
-	}
-
-	if _, ok := cgMap[launchableName]; !ok {
-		return util.Errorf("Unknown launchable %q in PLATFORM_CONFIG_PATH", launchableName)
-	}
-	if _, ok := cgMap[launchableName]["cgroup"]; !ok {
-		return util.Errorf("Launchable %q has malformed PLATFORM_CONFIG_PATH", launchableName)
+func cgCreate(cgroupName string, targetDirectory string, config cgroups.Config) error {
```
mpuncel: I like what you did here, it's much easier to follow what this function does now

bin/p2-exec/main.go
```diff
@@ -247,7 +274,46 @@ func cgEnter(platconf, launchableName, cgroupName string) error {
...
+}
+
+// given ENV variable "configPath" return parameters for the new cgroup being created
+func cgGetConfig(configPath string, configKey string) (cgroups.Config, error) {
+	if config := os.Getenv(configPath); config != "" {
+		confBuf, err := ioutil.ReadFile(config)
+		if err != nil {
+			return cgroups.Config{}, err
+		}
+		if configPath == "RESOURCE_LIMITS" {
```
mpuncel: given that like 80% of this function is in this if/else block, maybe it would make sense to have separate functions like `getPodCGConfig` and `getLaunchableCGConfig`.

bin/p2-exec/main.go
```diff
@@ -247,7 +274,46 @@ func cgEnter(platconf, launchableName, cgroupName string) error {
...
+				return cgroups.Config{}, util.Errorf("Key %q not found in %s", configKey, configPath)
+			}
+			if _, ok := cgMap[configKey]["cgroup"]; !ok {
+				return cgroups.Config{}, util.Errorf("Key %q in %s does not contain cgroup parameters\n", configKey, configPath)
+			}
+			cgConfig := cgMap[configKey]["cgroup"]
+			return cgConfig, nil
+		}
+	} else {
+		return cgroups.Config{}, util.Errorf("No %s found in environment", configPath)
```
mpuncel: i'd put this at the top of the function inside a `if config == ""`, that will save you some indentation and save people reading this from having to scroll up to see what the other cases were

pkg/manifest/manifest.go
```diff
@@ -29,6 +30,10 @@ type StatusStanza struct {
 	LocalhostOnly bool   `yaml:"localhost_only,omitempty"`
 }
 
+type ResourceLimitsStanza struct {
+	CGroup cgroups.Config `yaml:"cgroup,omitempty"`
```
mpuncel: Unfortunately using the `cgroups.Config` type directly, because it has a `size.ByteCount` with a `MarshalYAML()` function that marshals it to an integer number of bytes.

In other words using this type will mean that pod manifests will always have an integer number of bytes instead of something human readable like `10G`. I'm not really sure what to do about this though, we want it to marshal differently when it's in a pod manifest vs when it's in a cgroup file

mpuncel: You're more of a cgroups expert than I am, but i'm used to hearing it pluralized like "cgroups". Would that be more appropriate for the YAML key?

pkg/manifest/manifest.go
```diff
@@ -383,6 +403,14 @@ func (manifest *manifest) PlatformConfigFileName() (string, error) {
 	return string(manifest.Id) + "_" + sha + ".platform.yaml", nil
 }
 
+func (manifest *manifest) ResourceLimitsConfigFileName() (string, error) {
+	sha, err := manifest.SHA()
+	if err != nil {
+		return "", err
+	}
+	return string(manifest.Id) + "_" + sha + ".resource.yaml", nil
```
mpuncel: nit: `manifest.Id.String()`. I like those functions because if we ever change the way the type works then the compiler will warn us, whereas casting to a string might not

bin/p2-exec/main.go
```diff
@@ -80,20 +81,55 @@ func main() {
 		}
 	}
 
+	cgroupPath := ""
+	if *podID != "" {
+		cgConfig, err := cgGetConfig("RESOURCE_LIMITS", "")
```
mpuncel: use `pods.ResourceLimitsConfigPathEnvVar` instead of a magic string

bin/p2-exec/main.go
```diff
@@ -80,20 +81,55 @@ func main() {
...
 	}
 	if *launchableName != "" && *cgroupName != "" {
-		if platconf := os.Getenv("PLATFORM_CONFIG_PATH"); platconf != "" {
-			err := cgEnter(platconf, *launchableName, *cgroupName)
-			if err != nil {
-				log.Fatal(err)
-			}
-		} else {
-			log.Fatal("No PLATFORM_CONFIG_PATH, cannot determine cgroup")
+		cgPlatConfig, err := cgGetConfig("PLATFORM_CONFIG_PATH", *launchableName)
```
rudle: Is this the right ENV var? It doesn't match the docs.

bin/p2-exec/main.go
```diff
@@ -80,20 +81,55 @@ func main() {
...
+
+		err = os.MkdirAll(filepath.Join(cg.CPU, "p2", node), 0755)
+		if err != nil && !os.IsExist(err) {
+			log.Fatal(err)
+		}
+		err = os.MkdirAll(filepath.Join(cg.Memory, "p2", node), 0755)
+		if err != nil && !os.IsExist(err) {
+			log.Fatal(err)
+		}
+		cgroupPath = "p2" + "/" + node
```
rudle: `filepath.Join("p2",node)` is a nicer way to say this same thing. ditto on line 115

bin/p2-exec/main.go
```diff
@@ -80,20 +82,31 @@ func main() {
...
+			log.Fatal(err)
+		}
+		err = cgroups.CreatePodCgroup(*podID, cgConfig)
+		if err != nil {
+			log.Fatalf("Could not create pod %v cgroup %v\n", *podID, err)
+		}
+	}
+
 	if *launchableName == "" && *cgroupName != "" {
 		log.Fatalf("Specified cgroup name %q, but no launchable name was specified", *cgroupName)
```
mpuncel: is it intended that you'll have already created the pod cgroup here? would it be better to validate all the input first or does it not matter

pkg/cgroups/cgroups.go
```diff
@@ -72,6 +73,30 @@ func Find() (Subsystems, error) {
...
 
+func FindWithParentGroup(podID string) (Subsystems, error) {
+	subsys, err := Find()
+	if err != nil {
+		return Subsystems{}, err
+	}
+
+	node, err := os.Hostname()
+	if err != nil {
+		return Subsystems{}, err
```
mpuncel: i always try to wrap errors that come from other libraries with a `util.Errorf()` so we can easily tell where the error occurred

mpuncel: but this doens't need to be a merge blocker

pkg/cgroups/cgroups.go
```diff
@@ -178,6 +203,103 @@ func (subsys Subsystems) AddPID(name string, pid int) error {
...
+
+	if err != nil {
+		return util.Errorf("Could not find cgroupfs mount point: %s", err)
+	}
+
+	err = subsys.Write(config)
+	if _, ok := err.(UnsupportedError); ok {
+		// if a subsystem is not supported, just log
+		// and carry on
+		fmt.Printf("Unsupported subsystem (%s), continuing\n", err)
```
mpuncel: now that this isn't in a script anymore, it doesn't really make sense here. Maybe return a special error type all the way up and check it in `p2-exec`

pkg/pods/pod.go
```diff
@@ -713,28 +733,34 @@ func (pod *Pod) getLaunchable(launchableID launch.LaunchableID, launchableStanza
 		pod.logger.WithError(err).Warnf("Could not parse version from launchable %s.", launchableID)
 	}
 
+	podCgroup := ""
+	if *NestedCgroups {
+		err = pod.installCgroup()
+		if err != nil {
+			pod.logger.WithError(err).Errorf("Could not create pod cgroup")
```
mpuncel: what's the error scenario here? will the process be able to start successfully?

pkg/pods/pod.go
```diff
@@ -805,6 +832,20 @@ func (pod *Pod) disableAndHaltLaunchables(currentManifest manifest.Manifest) err
 	return nil
 }
 
+func (pod *Pod) installCgroup() error {
+	if *NestedCgroups {
```
mpuncel: where is this defined

mpuncel:
looks like you need a rebase as well

pkg/pods/pod.go
brianrli: At the top:
```
var (
	// ExperimentalOpencontainer permits the use of the experimental "opencontainer"
	// launchable type.
	ExperimentalOpencontainer = param.Bool("experimental_opencontainer", false)

	// NestedCgroups causes the p2-preparer to use a hierarchical cgroup naming scheme when
	// creating new launchables.
	NestedCgroups = param.Bool("nested_cgroups", false)
)
```

pkg/pods/pod.go
brianrli: Oh good point. I thought WithError actually killed it, but looks like it doesn't.

I think, in that case, it should attempt to try with just the launchable, basically flip the *NestedCgroups flag and try again?

brianrli: I think I will write a cleanup function to os.Remove everything that was created, if it was created

pkg/cgroups/cgroups.go
```diff
@@ -178,6 +203,100 @@ func (subsys Subsystems) AddPID(name string, pid int) error {
...
+		return err
+	}
+	err = os.MkdirAll(filepath.Join(subsys.Memory, "p2", node), 0755)
+	if err != nil && !os.IsExist(err) {
+		return err
+	}
+	return create(cgroupName, "", config)
+}
+
+func CreateLaunchableCgroup(cgroupName, podID string, config Config) error {
```
rudle: Can you create a similar public function that doesn't take a podID? I was confused when I found a potentially empty string being passed into this function on line 107 of p2-exec/main.go. In that file, use an if statement to call the correct version.

I'm less concerned about the private function taking an empty string.

PR #945, opened 2017-10-19
Add LockIfKeyNotExistsTxn() function

spenceral:
This will be used by the replication controller to schedule a new node
during a node transfer. By using a check and set with a modify index of
0, it is guaranteed that the intent/ key does not yet exist. If the CAS
fails, another RC has already taken over the node and we can rollback
the node transfer.

### DIFF
--- a/pkg/store/consul/kv.go
+++ b/pkg/store/consul/kv.go
@@ -232,6 +232,45 @@ func (c consulStore) SetPodTxn(ctx context.Context, podPrefix PodPrefix, nodenam
 	})
 }
 
+// LockIfKeyNotExistsTxn will lock and set a key, guaranteeing the key did not
+// exist.
+func (c consulStore) LockIfKeyNotExistsTxn(
+	ctx context.Context,
+	podPrefix PodPrefix,
+	nodename types.NodeName,
+	manifest manifest.Manifest,
+	session Session,
+) error {
+	buf := bytes.Buffer{}
+	err := manifest.Write(&buf)
+	if err != nil {
+		return err
+	}
+
+	key, err := podPath(podPrefix, nodename, manifest.ID())
+	if err != nil {
+		return err
+	}
+
+	// KVDeleteCAS with index 0 is used because it ensures that the subsequent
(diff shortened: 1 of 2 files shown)


spenceral:
If there is a better place to implement this function let me know.

pkg/store/consul/kv.go
```diff
@@ -232,6 +232,34 @@ func (c consulStore) SetPodTxn(ctx context.Context, podPrefix PodPrefix, nodenam
...
+	if err != nil {
+		return err
+	}
+	keyPair := &api.KVPair{
+		Key:         key,
+		Value:       buf.Bytes(),
+		ModifyIndex: modifyIndex,
+	}
+
+	ok, _, err := c.client.KV().CAS(keyPair, nil)
```
mpuncel: I think you're going to not only need to check the index is 0 but also lock it with a session (so that the key gets rolled back if the session dies). It might be better to use a transaction here with the "lock" operation as well as a "check-index" operation

spenceral: I had the session creation and locking in the code that will call this function. Would that go better here? I'm a little unclear on the session stuff

spenceral:
Rewrote using delete-cas

pkg/store/consul/kv.go
```diff
@@ -232,6 +232,45 @@ func (c consulStore) SetPodTxn(ctx context.Context, podPrefix PodPrefix, nodenam
...
+	}
+
+	key, err := podPath(podPrefix, nodename, manifest.ID())
+	if err != nil {
+		return err
+	}
+
+	// KVDeleteCAS with index 0 is used because it ensures that the subsequent
+	// lock will fail if the key existed.
+	err = transaction.Add(ctx, api.KVTxnOp{
```
mpuncel: might be useful to add a `TODO` for changing this to `check-not-exists` once we upgrade consul

mpuncel: don't worry about this, don't want to slow you down

PR #946, opened 2017-10-20
Fix an issue with daemon set responsiveness

mpuncel:
Commit 643738a9006582eb9d4df9f9469f26be5b97a188 attempted to collapse a
lot of duplicated code into a single function in the daemon set farm,
however an unintended side effect was that the daemon set farm is slow
to pass updates to a daemon set to the goroutine that is responsible for
carrying out the update.

The reason is that daemon set contention checks were being performed
regardless of whether the daemon set had been updated, so every time the
daemon set watch would fire a contention check would be run against
every daemon set. This is an expensive calculation because it requires
performing label select operations for every daemon set and then looking
for overlaps between sets.

This commit performs the check only if the daemon set is new or has been
updated.

### DIFF
--- a/pkg/ds/farm.go
+++ b/pkg/ds/farm.go
@@ -342,19 +342,19 @@ func (dsf *Farm) handleDSChanges(ctx context.Context, changes dsstore.WatchedDae
 	defer dsf.childMu.Unlock()
 	if len(changes.Created) > 0 {
 		for _, dsFields := range changes.Created {
-			dsf.lockAndSpawn(ctx, *dsFields)
+			dsf.lockAndSpawn(ctx, *dsFields, Created)
 		}
 	}
 
 	if len(changes.Updated) > 0 {
 		for _, dsFields := range changes.Updated {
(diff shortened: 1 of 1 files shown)


pkg/ds/farm.go
```diff
@@ -698,23 +707,28 @@ func (dsf *Farm) lockAndSpawn(ctx context.Context, dsFields ds_fields.DaemonSet)
...
-		return false
-	}
-
-	if isContended {
-		dsf.raiseContentionAlert(dsContended, dsFields)
-		dsf.logger.Errorf("Created daemon set '%s' contends with %s", dsFields.ID, dsContended.ID)
-		dsFields, err = dsf.dsStore.Disable(dsFields.ID)
+	// We only need to check for contention if it's new or changed. This is
+	// an expensive operation to run, particularly when 99% of the time a
+	// daemon set passed to thsi function hasn't changed
```
rudle: typo: -> this

PR #948, opened 2017-10-20
Spenceral/node transfer

spenceral:
I'll add some unit tests but I wanted to go ahead and throw this up to get some eyes on it.

### DIFF
--- a/pkg/rc/replication_controller.go
+++ b/pkg/rc/replication_controller.go
@@ -743,19 +743,37 @@ func (rc *replicationController) updateAllocations(ineligible []types.NodeName) 
 		return "", "", util.Errorf("Could not deallocate from %s: %s", oldNode, err)
 	}
 
-	nodesRequested := 1 // We only support one node transfer at a time right now
-	newNodes, err := rc.scheduler.AllocateNodes(man, sel, nodesRequested)
-	if err != nil || len(newNodes) < 1 {
-		errMsg := fmt.Sprintf("Unable to allocate nodes over grpc: %s", err)
-		err := rc.alerter.Alert(rc.alertInfo(errMsg), alerting.LowUrgency)
-		if err != nil {
-			rc.logger.WithError(err).Errorln("Unable to send alert")
-		}
-
-		return "", "", util.Errorf(errMsg)
+	eligible, err := rc.eligibleNodes()
+	if err != nil {
+		return "", "", err
 	}
 
-	newNode := newNodes[0]
+	current, err := rc.CurrentPods()
+	if err != nil {
+		return "", "", err
+	}
+
+	// Use an existing allocation if available
+	possible := types.NewNodeSet(eligible...).Difference(types.NewNodeSet(current.Nodes()...)).ListNodes()
+
+	var newNode types.NodeName
+	if len(possible) > 0 {
+		newNode = possible[0]
+	} else {
+		nodesRequested := 1 // We only support one node transfer at a time right now
+		newNodes, err := rc.scheduler.AllocateNodes(man, sel, nodesRequested)
+		if err != nil || len(newNodes) < 1 {
+			errMsg := fmt.Sprintf("Unable to allocate nodes over grpc: %s", err)
+			err := rc.alerter.Alert(rc.alertInfo(errMsg), alerting.LowUrgency)
+			if err != nil {
+				rc.logger.WithError(err).Errorln("Unable to send alert")
+			}
+
+			return "", "", util.Errorf(errMsg)
+		}
+
+		newNode = newNodes[0]
+	}
 
 	status := rcstatus.Status{
 		NodeTransfer: &rcstatus.NodeTransfer{
--- a/pkg/rc/replication_controller.go
+++ b/pkg/rc/replication_controller.go
@@ -93,14 +93,6 @@ type consulStore interface {
 		manifestID types.PodID,
 	) error
 
-	LockIfKeyNotExistsTxn(
-		ctx context.Context,
-		podPrefix consul.PodPrefix,
-		nodeName types.NodeName,
-		manifest manifest.Manifest,
-		session consul.Session,
-	) error
-
 	NewUnmanagedSession(session, name string) consul.Session
 }
 
--- a/pkg/rc/replication_controller.go
+++ b/pkg/rc/replication_controller.go
@@ -11,6 +11,8 @@ import (
 
 	"github.com/square/p2/pkg/alerting"
 	grpc_scheduler "github.com/square/p2/pkg/grpc/scheduler/client"
+	"github.com/square/p2/pkg/health"
+	"github.com/square/p2/pkg/health/checker"
 	"github.com/square/p2/pkg/labels"
 	"github.com/square/p2/pkg/logging"
 	"github.com/square/p2/pkg/manifest"
@@ -18,6 +20,7 @@ import (
 	"github.com/square/p2/pkg/rc/fields"
 	"github.com/square/p2/pkg/scheduler"
 	"github.com/square/p2/pkg/store/consul"
+	"github.com/square/p2/pkg/store/consul/consulutil"
 	"github.com/square/p2/pkg/store/consul/rcstore"
 	"github.com/square/p2/pkg/store/consul/statusstore"
 	"github.com/square/p2/pkg/store/consul/statusstore/rcstatus"
@@ -96,6 +99,16 @@ type consulStore interface {
 	NewUnmanagedSession(session, name string) consul.Session
 }
 
+// nodeTransfer encapsulates the information required to perform and cancel a node transfer
+// off an ineligible node
+type nodeTransfer struct {
+	newNode       types.NodeName
+	oldNode       types.NodeName
+	quit          chan struct{}
+	cancelSession context.CancelFunc
+	unlocker      consul.TxnUnlocker
+}
+
 // replicationController wraps a fields.RC with information required to manage the RC.
 // Note: the fields.RC might be mutated during this struct's lifetime, so a mutex is
 // used to synchronize access to it
@@ -106,6 +119,7 @@ type replicationController struct {
 	logger logging.Logger
 
 	consulStore   consulStore
+	consulClient  consulutil.ConsulClient
 	rcStatusStore rcstatus.ConsulStore
 	auditLogStore AuditLogStore
 	txner         transaction.Txner
@@ -113,6 +127,8 @@ type replicationController struct {
 	scheduler     Scheduler
 	podApplicator Labeler
 	alerter       alerting.Alerter
+	healthChecker checker.ConsulHealthChecker
+	nodeTransfer  nodeTransfer
 }
 
 type ReplicationControllerWatcher interface {
@@ -122,6 +138,7 @@ type ReplicationControllerWatcher interface {
 func New(
 	fields fields.RC,
 	consulStore consulStore,
+	consulClient consulutil.ConsulClient,
 	rcStatusStore rcstatus.ConsulStore,
 	auditLogStore AuditLogStore,
 	txner transaction.Txner,
@@ -130,6 +147,7 @@ func New(
 	podApplicator Labeler,
 	logger logging.Logger,
 	alerter alerting.Alerter,
+	healthChecker checker.ConsulHealthChecker,
 ) ReplicationController {
 	if alerter == nil {
 		alerter = alerting.NewNop()
@@ -140,6 +158,7 @@ func New(
 
 		logger:        logger,
 		consulStore:   consulStore,
+		consulClient:  consulClient,
 		rcStatusStore: rcStatusStore,
 		auditLogStore: auditLogStore,
 		txner:         txner,
@@ -147,6 +166,8 @@ func New(
 		scheduler:     scheduler,
 		podApplicator: podApplicator,
 		alerter:       alerter,
+		healthChecker: healthChecker,
+		nodeTransfer:  nodeTransfer{},
 	}
 }
 
@@ -200,9 +221,14 @@ func (rc *replicationController) WatchDesires(quit <-chan struct{}) <-chan error
 func (rc *replicationController) meetDesires() error {
 	rc.logger.NoFields().Infof("Handling RC update: desired replicas %d, disabled %v", rc.ReplicasDesired, rc.Disabled)
 
-	// If we're disabled, we do nothing, nor is it an error
+	// If we're disabled, quit an in progress node transfer (if any) and no-op
 	// (it's a normal possibility to be disabled)
 	if rc.Disabled {
+		rc.mu.Lock()
+		if rc.nodeTransfer.quit != nil {
+			close(rc.nodeTransfer.quit)
+		}
+		rc.mu.Unlock()
 		return nil
 	}
 
@@ -220,12 +246,30 @@ func (rc *replicationController) meetDesires() error {
 	nodesChanged := false
 	switch {
 	case rc.ReplicasDesired > len(current):
+		if rc.nodeTransfer.quit != nil {
+			// Remove nodeTransfer.newNode from eligible during a node transfer
+			// because if it is scheduled again below in addPods(), the rc
+			// will have one less node than desired. Waiting for the next loop
+			// for this case again will add a node, but is a bad experience.
+			for i, v := range eligible {
+				if v == rc.nodeTransfer.newNode {
+					eligible = append(eligible[:i], eligible[i+1:]...)
+					break
+				}
+			}
+		}
 		err := rc.addPods(current, eligible)
 		if err != nil {
 			return err
 		}
 		nodesChanged = true
 	case len(current) > rc.ReplicasDesired:
+		rc.mu.Lock()
+		if rc.nodeTransfer.quit != nil {
+			// stop the node transfer; additional node is no longer needed
+			close(rc.nodeTransfer.quit)
+		}
+		rc.mu.Unlock()
 		err := rc.removePods(current, eligible)
 		if err != nil {
 			return err
@@ -242,9 +286,12 @@ func (rc *replicationController) meetDesires() error {
 		}
 	}
 
-	err = rc.checkForIneligible(current, eligible)
-	if err != nil {
-		return util.Errorf("Error handling ineligible nodes: %s", err)
+	ineligible := rc.checkForIneligible(current, eligible)
+	if len(ineligible) > 0 {
+		err := rc.transferNodes(ineligible)
+		if err != nil {
+			return err
+		}
 	}
 
 	return rc.ensureConsistency(current)
@@ -427,8 +474,14 @@ func (rc *replicationController) removePods(current types.PodLocations, eligible
 
 func (rc *replicationController) ensureConsistency(current types.PodLocations) error {
 	rc.mu.Lock()
+	// if a node transfer is in progress, the old node could be in current
+	// but could be unscheduled during this function's execution. We
+	// ignore it to prevent rescheduling
+	oldNode := rc.nodeTransfer.oldNode
+
 	manifest := rc.Manifest
 	rc.mu.Unlock()
+
 	manifestSHA, err := manifest.SHA()
 	if err != nil {
 		return err
@@ -439,6 +492,10 @@ func (rc *replicationController) ensureConsistency(current types.PodLocations) e
 		cancelFunc()
 	}()
 	for i, pod := range current {
+		if pod.Node == oldNode {
+			continue
+		}
+
 		// create a new context for every 5 nodes. This is done to make
 		// sure we're safely under the 64 operation limit imposed by
 		// consul on transactions. This shouldn't be necessary after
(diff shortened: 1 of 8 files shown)


pkg/rc/replication_controller.go
```diff
@@ -228,12 +248,24 @@ func (rc *replicationController) meetDesires() error {
 	nodesChanged := false
 	switch {
 	case rc.ReplicasDesired > len(current):
+		if rc.isWatchingHealth {
+			// Remove newNode from eligible
```
mpuncel: might be worth including some detail in the comment for future readers

pkg/rc/replication_controller.go
```diff
@@ -228,12 +248,24 @@ func (rc *replicationController) meetDesires() error {
...
+				}
+			}
+		}
 		err := rc.addPods(current, eligible)
 		if err != nil {
 			return err
 		}
 		nodesChanged = true
 	case len(current) > rc.ReplicasDesired:
+		if rc.isWatchingHealth {
```
mpuncel: ditto

pkg/rc/replication_controller.go
```diff
@@ -633,81 +661,267 @@ func (rc *replicationController) unschedule(txn *auditingTransaction, node types
...
+			return err
+		}
 	} else {
-		err = rc.transferPetNodes(ineligibleCurrent)
+		newNode = status.NodeTransfer.NewNode
 	}
-	if err != nil {
-		return util.Errorf("Unable to transfer nodes: %s", err)
+
+	if rc.isWatchingHealth {
```
mpuncel: should this check happen earlier in this function? is there a case where this can be true but we haven't run `scheduleWithSession()` etc?

pkg/rc/replication_controller.go
```diff
@@ -633,81 +661,267 @@ func (rc *replicationController) unschedule(txn *auditingTransaction, node types
...
+		NodeTransfer: &rcstatus.NodeTransfer{
+			OldNode: oldNode,
+			NewNode: newNode,
+		},
 	}
+
 	writeCtx, writeCancel := transaction.New(context.Background())
 	defer writeCancel()
-	err = rc.rcStatusStore.CASTxn(writeCtx, rc.ID(), modifyIndex, status)
+	err = rc.rcStatusStore.CASTxn(writeCtx, rc.ID(), 0, status)
```
mpuncel: I think we shouldn't assume that the status doesn't exist because we might want to track other information in the status key someday (for example maybe we want to store the full list of current nodes to reduce the cost of looking that up)

mpuncel: We could consider using a different status key in consul for that though to make coordination here easier

pkg/rc/replication_controller.go
```diff
@@ -633,81 +661,267 @@ func (rc *replicationController) unschedule(txn *auditingTransaction, node types
...
+	rc.mu.Lock()
+	manifest := rc.Manifest
+	rc.mu.Unlock()
+
+	manifestBytes, err := manifest.Marshal()
+	if err != nil {
+		return err
+	}
+
+	key := path.Join(string(consul.INTENT_TREE), newNode.String(), manifest.ID().String())
```
mpuncel: this string computation should probably be put in `pkg/store/consul` somewhere with all the safety checks around empty hostname and pod ID, etc

pkg/rc/replication_controller.go
```diff
@@ -633,81 +661,267 @@ func (rc *replicationController) unschedule(txn *auditingTransaction, node types
...
+	name := fmt.Sprintf("wait-on-%s-health", newNode)
```
mpuncel: i'd maybe put the RC ID in here too

pkg/rc/replication_controller.go
```diff
@@ -633,81 +661,267 @@ func (rc *replicationController) unschedule(txn *auditingTransaction, node types
...
+	// Call this function to clean up the session when the node
+	// transfer finishes or is stopped
+	rc.nodeTransfer.sessionCancel = ctxCancel
+
+	writeCtx, writeCancel := transaction.New(context.Background())
+	defer writeCancel()
+
+	txnUnlocker, err := session.LockIfKeyNotExistsTxn(writeCtx, key, manifestBytes)
+	if err != nil {
+		return nil
```
mpuncel: did you mean `return err` here

pkg/rc/replication_controller.go
```diff
@@ -633,81 +661,267 @@ func (rc *replicationController) unschedule(txn *auditingTransaction, node types
...
+	}
+	rc.nodeTransfer.unlocker = txnUnlocker
+
+	ok, resp, err := transaction.Commit(writeCtx, rc.txner)
+	switch {
+	case err != nil:
+		return err
+	case !ok:
+		return util.Errorf("could not schedule new node due to a transaction violation: %s", transaction.TxnErrorsToString(resp.Errors))
```
mpuncel: I think here is probably the right place to `return nil` actually and just log the error along with a message like "new RC likely scheduled this node already" or something. The reason is that this function has other cases where it will return an error so we don't want to just assume everything went okay in the step above this

pkg/rc/replication_controller.go
```diff
@@ -633,81 +661,277 @@ func (rc *replicationController) unschedule(txn *auditingTransaction, node types
...
 
 	return nil
 }
 
-func (rc *replicationController) transferPetNodes(ineligibleCurrent []types.NodeName) error {
-	errMsg := fmt.Sprintf("RC has scheduled %d ineligible nodes: %s", len(ineligibleCurrent), ineligibleCurrent)
-	err := rc.alerter.Alert(rc.alertInfo(errMsg), alerting.LowUrgency)
+func (rc *replicationController) watchHealth() {
+	defer func() {
+		rc.isWatchingHealth = false
```
mpuncel: you're going to have a data race on `isWatchingHealth` because multiple goroutines write to it. You could guard it with a mutex to fix that

pkg/rc/replication_controller.go
```diff
@@ -633,81 +661,277 @@ func (rc *replicationController) unschedule(txn *auditingTransaction, node types
...
+		select {
+		case err := <-errCh:
+			rc.logger.WithError(err).Errorln("Node transfer health checker sent error")
+		case currentHealth := <-resultCh:
+			if currentHealth.Status == health.Passing {
+				err := rc.finishTransfer()
+				if err != nil {
+					rc.logger.WithError(err).Errorln("Could not complete transfer transaction")
+				}
+				return
```
mpuncel: you should also close `healthQuitCh` and `errCh` here. I'd recommend just doing that in a `defer` right after you create the channel so you don't have to keep track of all of the places that have a `return`

pkg/rc/replication_controller.go
```diff
@@ -633,81 +661,277 @@ func (rc *replicationController) unschedule(txn *auditingTransaction, node types
...
+				if err != nil {
+					rc.logger.WithError(err).Errorln("Could not complete transfer transaction")
+				}
+			}
+			return
+		}
+	}
+}
+
+func (rc *replicationController) finishTransfer() error {
```
mpuncel: just wanted to say I like the ordering of your functions, reading this PR has been pretty easy

pkg/rc/replication_controller.go
```diff
@@ -633,81 +661,277 @@ func (rc *replicationController) unschedule(txn *auditingTransaction, node types
...
+	if rc.isWatchingHealth {
+		return nil
 	}
 
+	rc.nodeTransfer = nodeTransfer{
+		newNode: newNode,
+		quit:    make(chan struct{}),
+	}
+
+	go rc.watchHealth()
```
mpuncel: `watchHealth()` does some stuff and then calls `finishTransfer()` It might be clearer if you called something like `rc.doNodeTransfer()` here which is just something like 

```
func doNodeTransfer() {
  watchHealth()
  finishTransfer()
}
``` (obviously that's pseudocode)

That I think more clearly shows the intent here and also it means you don't have to call `finishTransfer` in multiple places from `watchHealth`

File: pkg/rc/replication_controller.go [hunk 13]
```diff
@@ -633,81 +661,277 @@ func (rc *replicationController) unschedule(txn *auditingTransaction, node types
...
+		return err
+	case !ok:
+		errMsg := fmt.Sprintf(
+			"Transaction violation trying to unlock node transfer session and label new node. New RC may have scheduled node: %s",
+			transaction.TxnErrorsToString(resp.Errors),
+		)
+		err := rc.alerter.Alert(rc.alertInfo(errMsg), alerting.LowUrgency)
+		if err != nil {
+			rc.logger.WithError(err).Errorln("Unable to send alert")
+		}
```
mpuncel: do we need to delete the node transfer here? we probably do otherwise the new node will be ignored potentially forever right?

pkg/rc/replication_controller.go
spenceral: Yeah there's no use having it so late

pkg/rc/replication_controller.go
spenceral: It gets deleted by `quitTransfer` which is called by a defer in `watchHealth`, which is the function that calls this `finishTransfer` here. Your question is just further evidence in support of your previous comment. Something should be done about the organization of these functions to make this more clear and explicit

pkg/rc/replication_controller.go
spenceral: I think ultimately my decision to do it this way (can assume non-existence of status; always delete the status when transfer is done) was because of frustration with `CASTxn`. It seemed like a lot do a `Get`, `CASTxn`, and then a `transaction.Commit` just to try and eliminate history of the node transfer. Its particularly verbose since each of those can return `error`. I preferred just being able to delete and use 0 as the index. That is my argument in favor of having a separate status. That couldv'e been laziness and I understand that might just defer complication by requiring a separate status down the road.

If we keep this as the only rc status, should I try to identify where a `Set` would work, or could a `CAS` without a transaction ever be appropriate? Or is `CASTxn` the best route?

mpuncel: as a general statement, I feel like using transactional operations even if they're the only operation is still good because it makes it easier to add more operations to the transaction later on if it makes sense.

I think the argument for keeping separate status records is pretty strong, we can accomplish this by using different namespaces when configuring the status store. We should probably make the namespace string mention node transfers

pkg/rc/replication_controller.go
spenceral: if we don't return an error here the calling function won't realize that this RC failed to schedule a new node. It will keep executing and spin up the goroutine and everything

mpuncel: oh I see. In that case keep it the way it is

spenceral:
Added unit tests and addressed previous comments. This is now ready for review

pkg/rc/replication_controller.go
```diff
@@ -96,6 +99,16 @@ type consulStore interface {
 	NewUnmanagedSession(session, name string) consul.Session
 }
 
+// nodeTransfer encapsulates the information required to perform and cancel a node transfer
+// off an ineligible node
+type nodeTransfer struct {
+	newNode       types.NodeName
+	oldNode       types.NodeName
+	quit          chan struct{}
+	sessionCancel func()
```
mpuncel: nit: use `context.CancelFunc` (which is an alias for `func()` but makes it clearer what it's for)

pkg/rc/replication_controller.go
```diff
@@ -625,81 +663,280 @@ func (rc *replicationController) unschedule(txn *auditingTransaction, node types
...
+func (rc *replicationController) watchHealth() bool {
+	rc.mu.Lock()
+	podID := rc.Manifest.ID()
+	rc.mu.Unlock()
+
+	resultCh := make(chan health.Result)
+	errCh := make(chan error)
+	healthQuitCh := make(chan struct{})
+	defer func() {
+		close(resultCh)
```
mpuncel: unfortunately i think closing this isn't safe, because the way `WatchNodeService` is written it might try to send on this channel after it is closed.

I hate to give you more work for this PR, but probably `WatchNodeService()` should create `resultCh` and `errCh` itself, return them, and then take care of closing them after it sees `healthQuitCh` close.

Then you would call it like this:

```
defer close(healthQuitCh)
resultCh, errCh := rc.healthChecker.WatchNodeService(healthQuitCh)
```
which would guarantee that we have a clean shutdown.

pkg/rc/replication_controller_test.go
```diff
@@ -929,3 +933,191 @@ func TestRCDoesNotFixMembership(t *testing.T) {
 		t.Fatalf("the RC should have alerted since it has some current nodes that aren't eligible and is unable to correct this. There were %d alerts", len(alerter.Alerts))
 	}
 }
+
+func TestTransferRolledBackByQuitCh(t *testing.T) {
```
mpuncel: These are great tests, they do a good job of documenting the behavior we expect as well

mpuncel:
Looks great! I think just that `WatchNodeService()` thing needs to be fixed. Unfortunately that one is a blocker because it will cause panics (although the RC farm catches them with a `recover()` so it wouldn't be the end of the world).

pkg/rc/replication_controller.go
spenceral: I'll open up a new PR refactoring `WatchNodeService` to use `WatchSingle` and to avoid the panic case

pkg/rc/replication_controller.go
```diff
@@ -242,9 +284,18 @@ func (rc *replicationController) meetDesires() error {
...
+	ineligible := rc.checkForIneligible(current, eligible)
+	if len(ineligible) > 0 {
+		err := rc.transferNodes(ineligible)
+		if err != nil {
+			return err
+		}
+	}
+
+	// transferNodes() may have changed set of current pods so we must refresh
+	current, err = rc.CurrentPods()
```
spenceral: Found a race condition here that I could never reproduce locally but happened in the CI. The node transfer finished before `rc.ensureConsistency` got called so the ineligible node was scheduled again since it was still in `current`. Not sure if that would happen in reality but it happened in the test.

spenceral:
Fixed the unsafe channel close with changes from https://github.com/square/p2/pull/950 so this is ready for review again

pkg/rc/replication_controller.go
mpuncel: ah that's interesting. It's definitely possible for this to happen in practice (although it probably wouldn't be on the same iteration of the main loop as the node transfer was started).

We should probably make sure that `ensureConsistency()` ignores the old node of a node transfer

pkg/rc/replication_controller.go
```diff
@@ -633,81 +669,276 @@ func (rc *replicationController) unschedule(txn *auditingTransaction, node types
...
+func (rc *replicationController) watchHealth() bool {
+	rc.mu.Lock()
+	podID := rc.Manifest.ID()
+	rc.mu.Unlock()
+
+	healthQuitCh := make(chan struct{})
+	defer close(healthQuitCh)
+
+	// these channels are closed by WatchPodOnNode
+	resultCh, errCh := rc.healthChecker.WatchPodOnNode(rc.nodeTransfer.newNode, podID, healthQuitCh)
```
mpuncel: thanks for making this change, I think it reads much better this way as well

pkg/rc/replication_controller.go
```diff
@@ -633,81 +669,276 @@ func (rc *replicationController) unschedule(txn *auditingTransaction, node types
...
+	rc.isWatchingHealth = true
+	ok := rc.watchHealth()
+	if ok {
+		err := rc.finishTransfer()
+		if err != nil {
+			rc.logger.WithError(err).Errorln("could not do final node transfer transaction")
+		}
+	}
+
+	rc.isWatchingHealth = false
```
mpuncel: I think these lines also are going to cause a race with the places where you do
```
if rc.IsWatchingHealth {
  close(rc.nodeTransfer.quit)
}
```
Probably there should be a mutex that guards the `isWatchingHealth` and `nodeTransfer` variables so that these operations are atomic. You might also be able to get away without using `isWatchingHealth` and just checking `nodeTransfer.quit != nil` or something like that.

pkg/rc/replication_controller.go
spenceral: Yeah actually this doesn't solve the problem. The old node could still be unscheduled between when ensureConsistency() is called finishes. I'll fix this

pkg/rc/replication_controller.go
spenceral: good call

pkg/rc/replication_controller.go
mpuncel: I also think it's okay if we don't make another `CurrentPods()` call here at all. We're never going to be able to call it at exactly the right time to detect changes, so I think we just need to make sure that we make safe choices with the initial `CurrentPods()` call at the beginning of the loop, and plan on the next iteration dealing with changes appropriately.

I think as long as we exclude the new node from `eligible` and the old node from `current` as it pertains to `ensureConsistency()` then we're good

pkg/rc/replication_controller.go
```diff
@@ -633,81 +669,276 @@ func (rc *replicationController) unschedule(txn *auditingTransaction, node types
...
+	err := rc.scheduler.DeallocateNodes(sel, []types.NodeName{oldNode})
+	if err != nil {
+		return "", "", util.Errorf("Could not deallocate from %s: %s", oldNode, err)
 	}
 
 	nodesRequested := 1 // We only support one node transfer at a time right now
-	rc.mu.Lock()
-	newNodes, err := rc.scheduler.AllocateNodes(rc.Manifest, rc.NodeSelector, nodesRequested)
-	rc.mu.Unlock()
+	newNodes, err := rc.scheduler.AllocateNodes(man, sel, nodesRequested)
```
mpuncel: I'm wondering if this spot should check if there already is a newly allocated node. I'm not sure how that would happen, I guess if the allocator server makes the allocation on a previous iteration but we end up with an error.

You could feed the diff of current and eligible down to here to see if there is already an unused allocation to use for the transfer rather than calling indiscriminately.

I'm not sure how important it is to check for this, I think it might mess up the rolling update logic if there's an extra allocation lying around unused

pkg/rc/replication_controller.go
```diff
@@ -633,81 +669,276 @@ func (rc *replicationController) unschedule(txn *auditingTransaction, node types
...
+		select {
+		case err := <-errCh:
+			rc.logger.WithError(err).Errorln("Node transfer health checker sent error")
+		case currentHealth := <-resultCh:
+			if currentHealth.Status == health.Passing {
+				return true
+			}
+		case <-rc.nodeTransfer.quit:
+			return false
+		case <-time.After(5 * time.Minute):
```
mpuncel: I'd log something here so we know we hit this timeout

mpuncel:
This looks good once the `ensureConsistency()` and `isWatchingHealth` races are fixed

pkg/rc/replication_controller.go
```diff
@@ -228,12 +244,28 @@ func (rc *replicationController) meetDesires() error {
...
+				}
+			}
+		}
 		err := rc.addPods(current, eligible)
 		if err != nil {
 			return err
 		}
 		nodesChanged = true
 	case len(current) > rc.ReplicasDesired:
+		if rc.nodeTransfer.quit != nil {
```
mpuncel: I still think you need to do something like

```
rc.nodeTransfer.mu.Lock()
if rc.nodeTransfer.quit != nil {
    close(rc.nodeTransfer.quit)
}
rc.nodeTransfer.mu.Unlock()
```

and then below acquire the lock before doing `rc.nodeTransfer = nodeTransfer{}` which nils out the channel. Otherwise you might panic closing a nil channel

spenceral:
I think I fixed the races (completely removed `isWatchingHealth` in favor of `nodeTransfer.quit != nil`) and changed transfer behavior to use an existing allocation if possible. There is an associated test for that behavior.

pkg/rc/replication_controller.go
spenceral: Do you think I need a new mutex specifically for nodeTransfers or can I just use the `rc.mu` that the rc struct already has

mpuncel: you could use the same one, yeah. They're protecting different data but there's not a lot of contention on that one so it's fine

pkg/rc/replication_controller.go
```diff
@@ -427,8 +474,14 @@ func (rc *replicationController) removePods(current types.PodLocations, eligible
 
 func (rc *replicationController) ensureConsistency(current types.PodLocations) error {
 	rc.mu.Lock()
+	// if a node transfer is in progress, the old node could be in current
+	// but could be unscheduled during this function's execution. We
+	// ignore it to prevent rescheduling
+	oldNode := rc.nodeTransfer.oldNode
```
spenceral: My previous change still allowed a race because a completed node transfer would make `rc.nodeTransfer.oldNode` nil. Then if the ineligible node was still in `current` it wouldn't be detected. I believe this fixes it

PR #950, opened 2017-10-24
Spenceral/watch node health

spenceral:


pkg/health/checker/checker.go
```diff
@@ -64,25 +64,40 @@ func NewConsulHealthChecker(client consulutil.ConsulClient) ConsulHealthChecker
...
+		case err := <-wsErrs:
+			errCh <- err
+		case kvPair := <-wsOut:
+			// convert to health result
+			res, err := kvpToResult(*kvPair)
 			if err != nil {
 				errCh <- err
 			} else {
-				resultCh <- consulWatchToResult(kvCheck)
+				resultCh <- *res
```
mpuncel: there's still a chance for a panic because we might make it to this line and then the caller calls `close(resultCh)` just before we send.

This can be solved by creating/closing `resultCh` within this function and then returning it. something like this:

```
resultCh := make(chan health.Result)
go func() {
  defer close(resultCh)
  for {
    select {
    }
  }
}

return resultCh
```

mpuncel: (you'll need to do the same with `errCh` as well because it has the same problem. Additionally you should do these writes inside of a select like this:

```
select {
  case resultCh <- *res:
  case <-quitCh:
    return
}
```
 This protects against the case where the caller might stop reading `resultCh` (or `errCh`) and causing this goroutine to block forever and never notice the cancellation signal

spenceral: I was going to try and follow the pattern of `WatchSingle` and have this function close `resultCh`. The caller only closes `errCh` and `quitCh`. That is why I included `close(resultCh)` in this function. 

However, since there's not a way to enforce that (i.e. not permit caller to close `resultCh`) I could refactor. Thoughts?

mpuncel: that works, although I personally think it's cleaner to return channels you're going to close rather than pass them as arguments. You're correct that `WatchSingle` and its cousins don't really do it right

spenceral: I'll rewrite it. I'm gathering that I can't receive from `WatchSingle` in the select and then try and send those values to channels without wrapping those sends in their own select.

mpuncel: It's possible for the caller to close the quit channel and then wait for the result channel to close, draining all of the values off of it, so maybe you don't need to wrap them all in selects. It does guard against the case that someone makes a mistake later on though

mpuncel: The main issue is my first comment in this chain, we need to guarantee that writes will not be attempted on a channel that might be closed (the easiest way to accomplish this is to make the goroutine that writes the channel also be the one that closes it)

spenceral:
updated

pkg/health/checker/checker.go
```diff
@@ -64,28 +62,49 @@ func NewConsulHealthChecker(client consulutil.ConsulClient) ConsulHealthChecker
...
-			kvCheck, err := c.consulStore.GetHealth(serviceID, nodename)
-			if err != nil {
-				errCh <- err
-			} else {
-				resultCh <- consulWatchToResult(kvCheck)
+	key := consul.HealthPath(podID.String(), nodename)
+
+	wsOut := make(chan *api.KVPair) // closed by WatchSingle
+	wsQuit := make(chan struct{})
+	defer close(wsQuit)
```
mpuncel: you probably want this inside the goroutine, otherwise it's going to be closed immediately!

mpuncel:
Looks good after that one comment!

pkg/health/checker/checker.go
spenceral: 🙃

PR #951, opened 2017-10-25
Add allocation strategy field to pod clusters

mpuncel:
This strategy field will be a long-term place to store the desired
configuration for the RC strategy for RCs created within this pod
cluster (since RCs are ephemeral).

### DIFF
--- a/pkg/pc/fields/fields.go
+++ b/pkg/pc/fields/fields.go
@@ -4,6 +4,7 @@ import (
 	"encoding/json"
 	"reflect"
 
+	rc_fields "github.com/square/p2/pkg/rc/fields"
 	"github.com/square/p2/pkg/types"
 
 	"k8s.io/kubernetes/pkg/labels"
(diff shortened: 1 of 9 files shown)


pkg/pc/fields/fields.go
```diff
@@ -53,6 +54,12 @@ type PodCluster struct {
 	// Selector to identify the pods that are members of this pod cluster
 	PodSelector labels.Selector
 
+	// AllocationStrategy tweaks certain characteristic about how pods
+	// within this cluster are managed. For example the "pet" strategy will
+	// never transfer a pod from one node to another without human
+	// intervention whereas the "cattle" strategy will
+	AllocationStrategy rc.Strategy `json:"allocation_strategy"`
```
spenceral: `RawPodCluster` contains the json encoding/decoding stuff for the other fields. Is it an intentional choice to include that here?

mpuncel: IDK what i was thinking, I missed a lot of stuff. should be fixed now!

mpuncel: actually i'm adding to the tests too

PR #954, opened 2017-10-26
Add --update-strategy command to p2-pcctl

mpuncel:
This new command allows changing the strategy on a pod cluster, which
is recommended to serve as a long-term storage mechanism for the default
strategy for an RC. It's recommended to copy the strategy for an RC from
its pod cluster

### DIFF
--- a/bin/p2-pcctl/main.go
+++ b/bin/p2-pcctl/main.go
@@ -32,6 +32,7 @@ const (
 	cmdDeleteText            = "delete"
 	cmdUpdateAnnotationsText = "update-annotations"
 	cmdUpdateSelectorText    = "update-selector"
+	cmdUpdateStrategyText    = "update-strategy"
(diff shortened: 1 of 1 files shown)


bin/p2-pcctl/main.go
```diff
@@ -91,6 +92,20 @@ var (
 	updateSelector = cmdUpdateSelector.Flag("selector", "The label selector to use for the pod cluster's pod selector (e.g. \"pod_id=foo,availability_zone=bar\")").Required().String()
 )
 
+// "update-strategy" command
+var (
+	cmdUpdateStrategy = kingpin.Command(cmdUpdateStrategyText, "Update a pod cluster's allocation strategy. This determines whether RCs created for this pod cluster will automatically change node membership when hsots are down")
```
rudle: typo: hsots -> hosts

PR #956, opened 2017-10-26
Add udpate strategy p2-rctl command

spenceral:


pkg/store/consul/rcstore/consul_store.go
```diff
@@ -663,6 +663,14 @@ func (s *ConsulStore) UpdateManifest(id fields.ID, man manifest.Manifest) error
 	return s.retryMutate(id, manifestUpdater)
 }
 
+func (s *ConsulStore) UpdateStrategy(id fields.ID, strategy fields.Strategy) error {
+	strategyUpdater := func(rc fields.RC) (fields.RC, error) {
```
mpuncel: maybe here you should check that you got a real strategy

mpuncel: although I didn't do that for pod clusters either so maybe i'm a hypocrite :P

spenceral: 👍

rudle: didn't you? https://github.com/square/p2/pull/954/files#diff-21e734f407a99256a47190a2322142b1R106

mpuncel: ah yeah i meant i didn't do it in a store function, but I also didn't write a store function I just used `MutatePC()`.

pkg/store/consul/rcstore/consul_store.go
```diff
@@ -663,6 +663,18 @@ func (s *ConsulStore) UpdateManifest(id fields.ID, man manifest.Manifest) error
 	return s.retryMutate(id, manifestUpdater)
 }
 
+func (s *ConsulStore) UpdateStrategy(id fields.ID, strategy fields.Strategy) error {
+	if strategy != fields.DynamicStrategy && strategy != fields.StaticStrategy {
```
spenceral: checked

PR #960, opened 2017-10-31
Add happy path logs and error checking to node transfer

spenceral:


pkg/rc/replication_controller.go
```diff
@@ -893,9 +896,14 @@ func (rc *replicationController) watchHealth() bool {
 	for {
 		select {
 		case err := <-errCh:
-			rc.logger.WithError(err).Errorln("Node transfer health checker sent error")
+			if err != nil {
+				rc.logger.WithError(err).Errorln("Node transfer health checker sent error")
+			} else {
+				rc.logger.Errorln("Node transfer health checker sent error")
```
mpuncel: probably say "sent nil error" right?

pkg/rc/replication_controller.go
```diff
@@ -893,9 +896,14 @@ func (rc *replicationController) watchHealth() bool {
 	for {
 		select {
 		case err := <-errCh:
```
mpuncel: One possible reason we saw a nil error here might be that `errCh` closed, you can check that by doing `err, ok := <-errCh` and then checking `ok`. It will be false if the channel was closed

pkg/rc/replication_controller.go
```diff
@@ -288,6 +288,7 @@ func (rc *replicationController) meetDesires() error {
 
 	ineligible := rc.checkForIneligible(current, eligible)
 	if len(ineligible) > 0 {
+		rc.logger.Infoln("Ineligible node found, starting node transfer")
```
mpuncel: maybe add a `%s` for `ineligible` too? golang slices print decently well

pkg/rc/replication_controller.go
```diff
@@ -892,10 +895,16 @@ func (rc *replicationController) watchHealth() bool {
 
 	for {
 		select {
-		case err := <-errCh:
-			rc.logger.WithError(err).Errorln("Node transfer health checker sent error")
+		case err, ok := <-errCh:
+			// TODO rewrite once we understand why nil errors were sent to this chan
```
spenceral: I know this looks weird and I'm not actually using the `ok` from  `err, ok`. But I've left it this way _just in case_ the nil error did not actually come from a closed channel. Once we learn why the error was nil I'll rewrite this.

PR #961, opened 2017-10-31
Spenceral/node transfer fixes

spenceral:


pkg/rc/replication_controller.go
```diff
@@ -736,6 +736,7 @@ func (rc *replicationController) updateAllocations(ineligible []types.NodeName)
 	if err != nil {
 		return "", "", util.Errorf("Could not deallocate from %s: %s", oldNode, err)
 	}
+	rc.logger.Infof("Deallocate ineligible node %s", oldNode)
```
mpuncel: Is this supposed to say "deallocated"

spenceral: 👍

PR #965, opened 2017-11-02
pkg/rc: stop using an RC pointer passed to a watch function

mpuncel:
Prior to this commit, the main replication controller loop would start
by calling the Watch() function on the RC store, passing it a pointer to
an RC struct and also a mutex. Whenever the watch goroutine detected a
change to the RC, it would lock the mutex and then mutate the RC
pointer.

This meant that whenever the main RC loop tried to access a field on the
RC pointer, it would need to grab a mutex while it did so. This is
untenable as new references to the field are added all the time and it's
hard to remember to hold the mutex, creating data races.

Furthermore, it's hard to reason about what happens if the RC is mutated
in the middle of a meetDesires() invocation.

To address this, this PR changes the way Watch() on the rc store works
so that it returns a channel of RC structs (not pointers). Whenever a
value is read from the channel, this value is passed to meetDesires()
where it is processed without fear of mutation occurring during the
call.

Now the only mutex needed by the replication controller handler is to
guard the node transfer struct which is still used by multiple
goroutines.

### DIFF
--- a/pkg/rc/replication_controller.go
+++ b/pkg/rc/replication_controller.go
@@ -37,8 +37,6 @@ const (
 )
 
 type ReplicationController interface {
-	ID() fields.ID
-
 	// WatchDesires causes the replication controller to watch for any changes to its desired state.
 	// It is expected that a replication controller is aware of a backing rcstore against which to perform this watch.
 	// Upon seeing any changes, the replication controller schedules or unschedules pods to meet the desired state.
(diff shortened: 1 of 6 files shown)


pkg/rc/replication_controller.go
```diff
@@ -178,22 +182,20 @@ func New(
 }
 
 func (rc *replicationController) ID() fields.ID {
```
spenceral: Is this function worth keeping around? When you call `rc.rcWatcher.Watch` below you ignore this function and just use `rc.rcID` directly.

PR #968, opened 2017-11-07
Mpuncel/node transfer audit logs

mpuncel:


pkg/audit/node_transfer.go
```diff
@@ -0,0 +1,157 @@
...
+
+const (
+	// NodeTransferStartEvent denotes the start of a node transfer
+	NodeTransferStartEvent EventType = "NODE_TRANSFER_START"
+
+	// NodeTransferCompletionEvent denotes the successful completion of a
+	// node transfer
+	NodeTransferCompletionEvent EventType = "NODE_TRANSFER_COMPLETION"
+
+	// NodeTrasnferRollbackEvent denotes a node transfer being rolled back
```
spenceral: `NodeTrasnferRollbackEvent` small typo

pkg/rc/replication_controller.go
```diff
@@ -785,8 +786,31 @@ func (rc *replicationController) updateAllocations(ineligible []types.NodeName)
 		},
 	}
 
+	rc.mu.Lock()
```
spenceral: I just remembered I never reviewed your rc watch pr. I'll get on that. Sorry about not doing it sooner!

pkg/rc/replication_controller.go
```diff
@@ -785,8 +786,31 @@ func (rc *replicationController) updateAllocations(ineligible []types.NodeName)
...
+	rc.mu.Unlock()
+	if err != nil {
+		return "", "", util.Errorf("could not generate node transfer start audit log details: %s", err)
+	}
+
 	writeCtx, writeCancel := transaction.New(context.Background())
 	defer writeCancel()
+
+	err = rc.auditLogStore.Create(
+		writeCtx,
```
spenceral: The error message in `util.Errorf("Could not commit CASTxn: %s", err)` on line 821 (I couldn't figure out how to add a comment to a line you didn't change) could be updated since the transaction is no longer just the CASTxn

pkg/rc/replication_controller.go
```diff
@@ -902,17 +910,60 @@ func (rc *replicationController) doBackgroundNodeTransfer() {
 
 	rc.logger.Infof("Node transfer from %s to %s complete.", rc.nodeTransfer.oldNode, rc.nodeTransfer.newNode)
```
spenceral: This log I wrote is actually misleading since the node transfer may not have been successful. I didn't catch that until your changes

mpuncel:
added a commit to this, i realized that audit log records will be difficult to get use out of if we can't search them based on pod cluster information

PR #969, opened 2017-11-07
Dai/health service

bunsenmcdubbs:
Defines interfaces for clients using the newly created health service

### DIFF
--- a/pkg/health/client/client.go
+++ b/pkg/health/client/client.go
@@ -1,0 +1,24 @@
+package client
+
+import (
+	"context"
+
+	"github.com/square/p2/pkg/health"
+)
+
+type HealthRequest struct {
+	Url      string
+	Protocol string
+}
+
+type HealthResponse struct {
+	HealthRequest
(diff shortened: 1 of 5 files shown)


pkg/health/client/client.go
```diff
@@ -0,0 +1,23 @@
+package client
+
+import (
+	"context"
+
+	"github.com/square/p2/pkg/health"
+)
+
+type HealthRequest struct {
+	Url      string
```
mpuncel: any reason you didn't want to use the `url.URL` type? just easier to swap out protocols, etc?

bunsenmcdubbs: I initially used `url.URL` but there seems to be more usage of the `string` type in the code I was looking at and using. The `url.URL` type also has a `protocol` field which feels redundant (and sometimes harder to work with)

bunsenmcdubbs:
**TODO** this needs to be paired with a commit in p2-master before deploying

PR #971, opened 2017-11-08
Add a parallelism limit to shutdown

rudle:
This unscientific change is a response to claims that when p2-shutdown is
invoked internally, we don't always see the lifecycle scripts executed.

### DIFF
--- a/bin/p2-shutdown/main.go
+++ b/bin/p2-shutdown/main.go
@@ -84,15 +84,7 @@ func main() {
 				haltWG.Done()
 				rateLimiter <- struct{}{}
 			}(haltWG, rateLimiter)
-			for {
-				select {
-				case <-rateLimiter:
-					break
-				default:
-					log.Printf("[INFO]: Max parallelism of %d for pod shutdown has been reached. Waiting.", *parallelFactor)
-					time.Sleep(1 * time.Second)
-				}
-			}
+			rateLimit(rateLimiter)
 			success, err := pod.Halt(man)
 			if !success {
 				log.Printf("[ERROR]: at least one launchable of %s did not halt successfully.", podID)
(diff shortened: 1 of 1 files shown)


rudle:
This was hard to test inline, so I wrote a playground to verify my rate limiting strategy was sound.
https://play.golang.org/p/6dbgpxg9EV

bin/p2-shutdown/main.go
```diff
@@ -72,16 +79,25 @@ func main() {
...
+		go func(man manifest.Manifest, podID types.PodID, haltWG *sync.WaitGroup, rateLimiter chan struct{}) {
+			defer func(haltWG *sync.WaitGroup, rateLimiter chan struct{}) {
+				haltWG.Done()
+				rateLimiter <- struct{}{}
+			}(haltWG, rateLimiter)
+			select {
+			case <-rateLimiter:
+			default:
+				log.Printf("[INFO]: Max parallelism of %d for pod shutdown has been reached. Waiting.", *parallelFactor)
+				time.Sleep(1 * time.Second)
```
mpuncel: Isn't this just sleeping a second and then ignoring the rate limiting?

rudle: darnit

bin/p2-shutdown/main.go
```diff
@@ -72,16 +79,28 @@ func main() {
...
-			defer haltWG.Done()
+		go func(man manifest.Manifest, podID types.PodID, haltWG *sync.WaitGroup, rateLimiter chan struct{}) {
+			defer func(haltWG *sync.WaitGroup, rateLimiter chan struct{}) {
+				haltWG.Done()
+				rateLimiter <- struct{}{}
+			}(haltWG, rateLimiter)
+			for {
+				select {
+				case <-rateLimiter:
+					break
```
mpuncel: review blocked on adding that anonymous function or GOTO we discussed

rudle:
We found the root cause of the issue related to lifecycle scripts,  so this change carries more risk than reward at this point.

PR #975, opened 2017-11-16
Mpuncel/old rc node transfer

mpuncel:


mpuncel:
forgot to do the RU part of this, closing

PR #976, opened 2017-11-16
Mpuncel/old rc node transfer

mpuncel:


pkg/roll/integration_test.go
```diff
@@ -113,6 +120,72 @@ func TestCleanupOldRCHappy(t *testing.T) {
...
+	switch err {
+	case rcstore.NoReplicationController:
+		t.Fatal("old RC was deleted even though it wasn't zeroed out")
+	case nil:
+	default:
+		t.Fatalf("unexpected error when checking that old RC was deleted: %s", err)
+	}
+}
+
 // errorOnceChannelAlerter returns an error the first time you call Alert() and
```
mpuncel: i tried writing a test that an RC doesn't get deleted if it has a 0 replica count but still has current pods. unfortunately i'm not sure how to write that test without restructuring the code, basically i want to prove that the `RetryOrQuit()` call runs at least once and didn't add anything to the transaction? Both of which are not inspectable things. I think it's okay

pkg/rc/replication_controller_test.go
```diff
@@ -1425,6 +1425,266 @@ func TestNodeTransferFailsOnScheduleStep(t *testing.T) {
...
+	current := make(types.PodLocations, 0)
+
+	eligible := []types.NodeName{"node1", "node2", "node3", "node4", "node5"}
+
+	err := rc.addPods(rcFields, current, eligible)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	// now confirm that 5 pods were scheduled
```
spenceral: Looks like a copy pasted comment from the previous test. Not correct here

mpuncel:
I think there are still some edge cases to sort out here

mpuncel:
OK i believe this to be complete now. will do some testing.

High level changes:

1) disabled RCs now perform node transfers
2) if the scheduleWithSession() step encounters a transaction conflict, it continues to waiting for healthy. Rationale is that the exact manifest doesn't matter, it just needs the new node to be healthy
3) rolling updates exclude ineligible nodes from health count for dynamic RCs (because an ineligible node can be unscheduled at any moment)

pkg/rc/replication_controller.go
```diff
@@ -443,8 +446,68 @@ func (rc *replicationController) removePods(rcFields fields.RC, current types.Po
...
+				// (because it will rely on a different enabled RC to
+				// do it implicitly)
+				//
+				// One final safety restriction we put in place
+				// is that a disabled RC should only unschedule
+				// ineligible nodes if its replica count is
+				// zero, which we expect to occur at the end of
+				// a rolling update. This protection is
+				// necessary to prevent the rolling update from
+				// counting this ineligible (presumably
```
rudle: I don't agree that ineligible nodes will be healthy in all or even most cases. I don't think it's material for the logic tho.

pkg/rc/replication_controller.go
```diff
@@ -443,8 +446,68 @@ func (rc *replicationController) removePods(rcFields fields.RC, current types.Po
...
+				// after the replica count change is executed,
+				// cluster might fall below the minimum health
+				// requirement of the rolling update.
+				//
+				// If the RCs replica count is zero, we can
+				// reliably assume that the rolling update is
+				// over so it is always safe to unschedule the
+				// old node. This is enforced at the beginning
+				// of the removePods() function in the first
+				// return
```
spenceral: WDYT about breaking up the comment and putting smaller comments closer to where the relevant code is? I had to reread it and think to determine what part of this comment is executed by the `if disabled: break`. It might be simpler to have it only say something like "Don't unschedule eligible nodes since that is implicitly performed by the new RC during roll when this RC is disabled" here,  add a small, relevant comment above `if !ok`, etc..

pkg/roll/run_update.go
```diff
@@ -684,7 +713,31 @@ func (u *update) countHealthy(id rcf.ID, checks map[types.NodeName]health.Result
 		ret.Unknown = ret.Desired - ret.Current
 	}
 
+	var eligibleNodes []types.NodeName
+	if rcFields.AllocationStrategy == rcf.DynamicStrategy {
+		fmt.Println("calling bro")
```
rudle: 🕶

pkg/rc/replication_controller_test.go
```diff
@@ -1425,6 +1428,266 @@ func TestNodeTransferFailsOnScheduleStep(t *testing.T) {
...
+
+	// Set the count to zero, and now we should see the ineligible nodes become unscheduled
+	rcFields.ReplicasDesired = 0
+
+	err = rc.removePods(rcFields, currentPods, eligible)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	// now confirm that 5 pods were scheduled
```
spenceral: incorrect comment. Test confirms that 3 pods were scheduled

pkg/rc/replication_controller.go
```diff
@@ -255,6 +243,7 @@ func (rc *replicationController) meetDesires(rcFields fields.RC) error {
 	nodesChanged := false
 	switch {
 	case rcFields.ReplicasDesired > len(current):
+		rc.nodeTransferMu.Lock()
 		if rc.nodeTransfer.quit != nil {
```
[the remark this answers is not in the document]
mpuncel: we unconditionally reach the unlock on 259 so it's fine. I also want to protect the `rc.nodeTransfer.newNode` accesses

pkg/rc/replication_controller.go
```diff
@@ -410,12 +404,23 @@ func (rc *replicationController) alertInfo(rcFields fields.RC, msg string) alert
...
 
 func (rc *replicationController) removePods(rcFields fields.RC, current types.PodLocations, eligible []types.NodeName) error {
+	// If the RC is disabled, we don't want to remove any pods with one
+	// exception: if the ReplicasDesired count is zero and their are "ineligible"
+	// nodes. In this case, we want to unschedule all ineligible nodes.
+	//
+	// An ineligible node is defined as a node that appears in "current" but not
+	// in "eligible". This means that the scheduler has decided this node is no
+	// longer eligible.
+	if rcFields.ReplicasDesired != 0 && rcFields.Disabled {
```
mpuncel: hmm actually i think we want to just rely on the transfer finishing instead of having this now

pkg/roll/run_update.go
```diff
@@ -684,7 +713,31 @@ func (u *update) countHealthy(id rcf.ID, checks map[types.NodeName]health.Result
...
+		eligibleNodes, err = u.scheduler.EligibleNodes(rcFields.Manifest, rcFields.NodeSelector)
+		if err != nil {
+			return ret, util.Errorf("could not enumerate eligible nodes for %s: %s", rcFields.ID, err)
+		}
+	}
+
 	for _, pod := range currentPods {
+		if rcFields.AllocationStrategy == rcf.DynamicStrategy {
```
spenceral: nit: you could instead check for AllocationStrategy != Dynamic and continue to reduce the indentation. Kind of in the style of https://github.com/golang/go/wiki/CodeReviewComments#indent-error-flow

pkg/roll/run_update.go
```diff
@@ -684,7 +713,31 @@ func (u *update) countHealthy(id rcf.ID, checks map[types.NodeName]health.Result
...
+			nodeEligible := false
+			for _, eligibleNode := range eligibleNodes {
+				if eligibleNode == pod.Node {
+					nodeEligible = true
+					break
+				}
+			}
+			if !nodeEligible {
+				continue
+			}
```
spenceral: I can't tell what this if block does. Doesn't the code go the next iteration of the loop anyway?

mpuncel: i'll add a comment to clarify. Basically I don't want to increment `ret.Healthy` or `ret.Real` if the node isn't eligible

spenceral: Nevermind. I totally misinterpreted the truncated file on github and thought the loop ended 🙃

pkg/roll/run_update.go
mpuncel: I don't think that would reduce indentation and it would make the if block longer since I would have to include the section that counts Healthy

spenceral: You're right. This is the same loop I misread the end of. Ignore me

mpuncel:
getting a test timeout in travis and not locally 🤔

PR #980, opened 2017-12-05
node transfers: only write label for new node if it was locked

mpuncel:
Background: the RC handler routine is expected to kick off a "node
transfer" if the scheduler tells it that one of its "current" nodes are
not eligible to run the pod anymore. This is complex to get right
because there might be 2 RCs that share the same resources (nodes) from
the scheduler when a rolling update is in progress.

What should happen when a node is ineligible is that the RC will request
a new node from the scheduler to replace it, from then on we want one of
the RCs to schedule the pod on that node, but we don't really care which
one during a deploy situation. As a result, an RC scheduling a node
transfer will attempt to schedule the new node using a lock and then
later clean up that lock when the node transfer is complete.

However, if the lock operation fails because the node is already
scheduled, we assume that another RC already scheduled it in a race, and
we want the node transfer to continue without errors. Therefore, when
the node transfer is complete (determined by the new pod being healthy),
we want the RC to only unlock the intent key if it initially scheduled
it in the first place.

This commit fixes a bug where even if the intent key is not locked by
this RC, we still write a label at the end as if it was. This means that
the RC that actually scheduled the pod will have its label overwritten
inappropriately

### DIFF
--- a/pkg/rc/replication_controller.go
+++ b/pkg/rc/replication_controller.go
@@ -521,6 +521,10 @@ func (rc *replicationController) removePods(rcFields fields.RC, current types.Po
 }
 
 func (rc *replicationController) ensureConsistency(rcFields fields.RC, current []types.NodeName, eligible []types.NodeName) error {
+	if rcFields.Disabled {
+		return nil
+	}
+
 	manifest := rcFields.Manifest
 
 	eligibleCurrent := types.NewNodeSet(current...).Intersection(types.NewNodeSet(eligible...)).ListNodes()
--- a/pkg/rc/replication_controller.go
+++ b/pkg/rc/replication_controller.go
@@ -30,6 +30,7 @@ import (
 	"github.com/square/p2/pkg/types"
 	"github.com/square/p2/pkg/util"
 
+	"github.com/hashicorp/consul/api"
 	"github.com/pborman/uuid"
 )
 
@@ -104,22 +105,14 @@ type consulStore interface {
 // nodeTransfer encapsulates the information required to perform and cancel a node transfer
 // off an ineligible node
 type nodeTransfer struct {
-	newNode       types.NodeName
-	oldNode       types.NodeName
-	id            rcstatus.NodeTransferID
-	quit          chan struct{}
-	session       consul.Session
-	cancelSession context.CancelFunc
-	unlockArgs    unlockArgs
+	newNode types.NodeName
+	oldNode types.NodeName
+	id      rcstatus.NodeTransferID
+	quit    chan struct{}
 
 	// rollbackReason indicates why the quit channel was closed. Closers of
 	// the quit channel should set this first for audit logging purposes
 	rollbackReason audit.RollbackReason
-}
-
-type unlockArgs struct {
-	key   string
-	value []byte
 }
 
 type replicationController struct {
@@ -181,7 +174,6 @@ func New(
 		podApplicator: podApplicator,
 		alerter:       alerter,
 		healthChecker: healthChecker,
-		nodeTransfer:  nodeTransfer{},
 	}
 }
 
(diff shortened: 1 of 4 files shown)


pkg/rc/replication_controller.go
```diff
@@ -940,14 +940,15 @@ func (rc *replicationController) finishTransfer(rcFields fields.RC) error {
...
-	txn.AddNode(rc.nodeTransfer.newNode)
+		// We only want to add an audit record for the new node and set its label if we actually wrote it
+		txn.AddNode(rc.nodeTransfer.newNode)
 
-	labelKey := labels.MakePodLabelKey(rc.nodeTransfer.newNode, rcFields.Manifest.ID())
-	err = rc.podApplicator.SetLabelsTxn(txn.Context(), labels.POD, labelKey, rc.computePodLabels(rcFields))
-	if err != nil {
-		return err
+		labelKey := labels.MakePodLabelKey(rc.nodeTransfer.newNode, rcFields.Manifest.ID())
+		err = rc.podApplicator.SetLabelsTxn(txn.Context(), labels.POD, labelKey, rc.computePodLabels(rcFields))
```
mpuncel: I'm realizing there's still a race here, where we might have locked the key but then the other RC already wrote these labels. I think that will resolve itself on its own eventually though, the other RC will notice it doesn't have enough nodes to meet desired and schedule it again (which will be a no-op on intent but fix the labels).

Alternatively i could make this be a check-and-set with the index set to 0 so it will fail if the labels are written already, but then i'd have to either set up a new transaction that doesn't write the labels or i'd have to put it in two separate transactions and set up a state machine of some kind stored in the rc status structure

pkg/rc/replication_controller.go
```diff
@@ -194,6 +186,42 @@ func (rc *replicationController) WatchDesires(quit <-chan struct{}) <-chan error
...
+			status, _, err = rc.rcStatusStore.Get(rc.rcID)
+			switch {
+			case err == nil:
+				needToFetchStatus = false
+			case statusstore.IsNoStatus(err):
+				needToFetchStatus = false
+			case err != nil:
+				errOutChannel <- err
+				rc.logger.WithError(err).Errorln("cannot begin servicing RC because status could not be fetched")
+				time.Sleep(1 * time.Second)
```
spenceral: When is it correct/better to use exponential backoff and when does one not bother?

mpuncel: we should probably be doing it everywhere in all error scenario, here included, i'll add it

pkg/rc/replication_controller.go
```diff
@@ -194,6 +186,42 @@ func (rc *replicationController) WatchDesires(quit <-chan struct{}) <-chan error
...
 	go func() {
+		defer func() {
+			rc.nodeTransferMu.Lock()
+			if rc.nodeTransfer.quit != nil {
+				rc.nodeTransfer.rollbackReason = "RC farm quitting"
+				close(rc.nodeTransfer.quit)
+			}
+			rc.nodeTransferMu.Unlock()
+		}()
+		// Fetch the RC's status to find out of a node transfer is underway
```
spenceral: Is this necessary? If a node transfer is underway, there will be an ineligible node, so in `meetDesires` this rc will call transferNodes which will call `rcStatusStore.Get` anyway

spenceral: When `scheduleNewTransferNodes` fails, for instance, it resets `rc.nodeTransfer` which is why we've been relying on that`rcStatusStore.Get` at the beginning of `transferNodes`

mpuncel: It's necessary because now if a node transfer is underway, we'll have an extra node (since both old and new nodes are labeled), and we need to be able to prevent `removePods()` from removing anything.

I could replace this fetch with one in `removePods()`, but it's probably cleaner to initialize at startup and then anywhere in this file can trust `rc.nodeTransfer` as long as the mutex is held

spenceral: Understood. But as a followup, what then is stopping `removePods()` from getting rid of something after `rc.nodeTransfer` is wiped after a failed `scheduleNewTransferNodes`

pkg/rc/replication_controller.go
```diff
@@ -699,7 +742,7 @@ func (rc *replicationController) transferNodes(rcFields fields.RC, ineligible []
 	rc.nodeTransfer.id = nodeTransferID
 	rc.nodeTransferMu.Unlock()
 
-	err = rc.scheduleWithSession(rcFields, newNode)
+	err = rc.scheduleNewNodeForTransfer(rcFields, newNode, current)
```
spenceral: nit: the error message on line 751 still says "Could not schedule _with session_"

pkg/rc/replication_controller.go
mpuncel: Good catch! We should not clear the data when a failure happens, so we'll retry with the same node in a few minutes.

PR #982, opened 2017-12-08
Add rebuild reality logic to p2-preparer setup

esuen:
- if a node is missing the reality tree on startup, apply rebuild reality logic to prevent orphaned pods
  - an orphaned pod will have its data still in /data/apps but not exist in intent
  - also fix a bug in pkg/pods/factory.go PodFromPodHomeWithReqFile where it wasn't parsing the uuid of pods correctly


esuen:
Edited commit message, force pushed and creating a new PR.

PR #983, opened 2017-12-08
Add rebuild reality logic to p2-preparer setup

esuen:
- if a node is missing the reality tree on startup, apply rebuild reality logic to prevent orphaned pods
  - an orphaned pod will have its data still in /data/apps but not exist in intent
  - also fix a bug in pkg/pods/factory.go PodFromPodHomeWithReqFile where it wasn't parsing the uuid of pods correctly

### DIFF
--- a/pkg/pods/factory.go
+++ b/pkg/pods/factory.go
@@ -3,7 +3,6 @@ package pods
 import (
 	"fmt"
 	"path/filepath"
-	"strings"
 	"time"
 
 	"github.com/square/p2/pkg/logging"
@@ -14,7 +13,6 @@ import (
 	"github.com/square/p2/pkg/util"
 
 	"github.com/Sirupsen/logrus"
-	"github.com/pborman/uuid"
 )
 
 const DefaultPath = "/data/pods"
@@ -139,13 +137,10 @@ func PodFromPodHome(node types.NodeName, home string) (*Pod, error) {
 }
 
 func PodFromPodHomeWithReqFile(node types.NodeName, home string, requireFile string) (*Pod, error) {
-	// Check if the pod home is namespaced by a UUID by splitting on a hyphen and
-	// checking the last part. If it parses as a UUID, pass it to newPodWithHome.
-	// Otherwise, pass a nil uniqueKey
-	homeParts := strings.Split(filepath.Base(home), "-")
-
+	// Check if the pod home is namespaced by a UUID and pass it to newPodWithHome
+	// uniqueKey can be nil if pod home is not namespaced by a UUID
 	var uniqueKey types.PodUniqueKey
-	podUUID := uuid.Parse(homeParts[len(homeParts)-1])
+	podUUID := types.HomeToPodUUID(home)
 	if podUUID != nil {
 		uniqueKey = types.PodUniqueKey(podUUID.String())
 	}
--- a/pkg/preparer/setup.go
+++ b/pkg/preparer/setup.go
@@ -36,6 +36,7 @@ import (
 	"github.com/square/p2/pkg/store/consul/podstore"
 	"github.com/square/p2/pkg/store/consul/statusstore"
 	"github.com/square/p2/pkg/store/consul/statusstore/podstatus"
+	"github.com/square/p2/pkg/store/consul/transaction"
 	"github.com/square/p2/pkg/types"
 	"github.com/square/p2/pkg/uri"
 	"github.com/square/p2/pkg/util"
@@ -77,6 +78,7 @@ type Preparer struct {
 	hooks                  Hooks
 	Logger                 logging.Logger
 	podFactory             pods.Factory
+	podRoot                string
 	authPolicy             auth.Policy
 	maxLaunchableDiskUsage size.ByteCount
 	finishExec             []string
@@ -521,6 +523,7 @@ func New(preparerConfig *PreparerConfig, logger logging.Logger) (*Preparer, erro
 		hooks:                  hooks.NewContext(preparerConfig.HooksDirectory, preparerConfig.PodRoot, &logger, auditLogger),
 		podStatusStore:         podStatusStore,
 		podStore:               podStore,
+		podRoot:                preparerConfig.PodRoot,
 		client:                 client,
 		Logger:                 logger,
 		podFactory:             pods.NewFactory(preparerConfig.PodRoot, preparerConfig.NodeName, fetcher, preparerConfig.RequireFile),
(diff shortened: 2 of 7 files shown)


esuen:
Not ready for review. Making change to compare reality tree with pods on disk instead of checking if reality tree is wiped.

pkg/pods/factory.go
```diff
@@ -142,12 +141,14 @@ func PodFromPodHomeWithReqFile(node types.NodeName, home string, requireFile str
...
 	// checking the last part. If it parses as a UUID, pass it to newPodWithHome.
 	// Otherwise, pass a nil uniqueKey
-	homeParts := strings.Split(filepath.Base(home), "-")
+	podID := filepath.Base(home)
 
 	var uniqueKey types.PodUniqueKey
-	podUUID := uuid.Parse(homeParts[len(homeParts)-1])
-	if podUUID != nil {
-		uniqueKey = types.PodUniqueKey(podUUID.String())
+	if len(podID) > types.PodUUIDLength {
```
mpuncel: what issue were you fixing here? it seems like if a pod ID is really long you'll mistake it for a uuid pod even if it isn't one

pkg/preparer/setup.go
```diff
@@ -650,6 +652,71 @@ func getArtifactRegistry(preparerConfig *PreparerConfig) (artifact.Registry, err
...
+func (p *Preparer) BuildRealityAtLaunch() error {
+	// check if reality tree does not exist, if true then reimage occurred and may need
+	// to rebuild reality tree for orphaned pods
+	// a pod is orphaned if they exist on disk but not in intent
+	sub := p.Logger.SubLogger(nil)
+	realityResults, _, err := p.store.ListPods(consul.REALITY_TREE, p.node)
+	if err != nil {
+		sub.WithError(err).Errorln("Could not check reality: %s", err)
+		return err
+	} else if len(realityResults) == 0 {
```
mpuncel: nit: since you returned on line 663 you don't really need this `else` and you can write the if on the next line as suggested here: https://github.com/golang/go/wiki/CodeReviewComments#indent-error-flow

pkg/pods/factory.go
esuen: that check is just so that it doesn't check for uuid in the directory name since it'll be too short.
Before the code was splitting the dir name by "-", so if you have a dir name of an app such as "pod-xxx-xxx-xx", homeParts will be ["pod","xxx","xxx","xx"] and homeParts[len(homeParts)-1] will only be the last part of the uuid. So I think the uuid check is invalid.

pkg/preparer/setup.go
```diff
@@ -650,6 +652,71 @@ func getArtifactRegistry(preparerConfig *PreparerConfig) (artifact.Registry, err
...
+		manifestFilePaths, err := filepath.Glob("/data/pods/*/current_manifest.yaml")
+		if err != nil {
+			sub.WithError(err).Errorln("Could not check filepaths of pods installed on disk: %s", err)
+			return err
+		}
+		for _, fp := range manifestFilePaths {
+			podIDStr := strings.SplitN(fp, "/", 5)[3]
+			var podUUID uuid.UUID
+			if len(podIDStr) > types.PodUUIDLength {
+				podUUID = uuid.Parse(podIDStr[len(podIDStr)-types.PodUUIDLength:])
```
mpuncel: probably this should be extracted to a function since it's doing the same thing as `PodWithPodHome`

pkg/pods/factory.go
esuen: See https://godoc.org/github.com/pborman/uuid#Parse

mpuncel: oh i see, if it doesn't parse as a uuid you handle it fine, i'm still confused what was wrong before though

pkg/preparer/setup.go
```diff
@@ -650,6 +652,71 @@ func getArtifactRegistry(preparerConfig *PreparerConfig) (artifact.Registry, err
 	return artifact.NewRegistry(url, fetcher, osversion.DefaultDetector), nil
 }
 
+func (p *Preparer) BuildRealityAtLaunch() error {
```
mpuncel: this function would be nice to have a unit test for, you can set up a consul instance from a test with `consulutil.NewFixture()` and then seed it with whatever data you want, instantiate a preparer, run this function and check that reality has what you expect

pkg/pods/factory.go
esuen: It was not looking at the whole uuid in the dir name, i think the logic is wrong.

mpuncel: ah I see, whoops

mpuncel: also lol my comment 8 minutes ago was made before I refreshed, so it looks like i was responding to you but I didn't see your comments when i wrote that

pkg/preparer/setup.go
esuen: BuildRealityAtLaunch looks reads the filesystem in the path "/data/pods/*/current_manifest.yaml" to check what pods are installed on the disk. Is there a way to mock this in a test or should I create this file hierarchy on my local machine?

esuen:
- Added unit test for BuildRealityAtLaunch that tests a regular app and a uuid namespaced app that should be written to reality
- podRoot added to preparer so that it can be access in BuildRealityAtLaunch

pkg/preparer/setup.go
esuen: This is resolved.

pkg/preparer/setup.go
```diff
@@ -650,6 +653,83 @@ func getArtifactRegistry(preparerConfig *PreparerConfig) (artifact.Registry, err
...
+				if !ok {
+					err := util.Errorf("status record transaction rolled back: %s", transaction.TxnErrorsToString(resp.Errors))
+					sub.WithError(err).Errorln("Could not write uuid index to reality store")
+					return err
+				}
+			} else if err != nil {
+				sub.WithError(err).Errorln("Unexpected error reading pod: %s", err)
+				return err
+			}
+		} else if _, ok := intentMap[home]; !ok {
```
mpuncel: Is this intent check even necessary? are you trying to prevent a case where the preparer was killed halfway through an install and therefore we shouldn't write reality yet?

esuen: This check was here so that reality is only written to pods on disk not in intent. If it existed in intent then I thought it didn't need to be written to reality.

mpuncel: yeah I think it's good to check intent, but i think it would be exceedingly rare for a pod to not be in reality but still be in intent, but probably good to be thorough

esuen:
@mpuncel can you merge this? I'm done with testing.

PR #986, opened 2017-12-12
Mpuncel/ru node transfer fix

mpuncel:


pkg/roll/update_test.go
```diff
@@ -1107,3 +1234,78 @@ func TestRollLoopStallsIfUnhealthy(t *testing.T) {
...
+	if !oldRC.Disabled {
+		t.Fatal("expected the old RC to be disabled")
+	}
+	newRC, err = rcStore.Get(oldRC.ID)
+	if err != nil {
+		t.Fatal(err)
+	}
+	if !newRC.Disabled {
+		t.Fatal("expected the new RC to be disabled")
+	}
```
mpuncel: this test isn't finished yet

spenceral:
`current` is higher than replica count because on an RC performing a node transfer because we're labeling at the same time we schedule now, yeah?

spenceral:
For the sake of discussion, what would be your thoughts on incrementing replicas_desired when a node transfer starts and decrementing it when one finishes? To me, it seems that it would be nice for the replica count manipulation to be isolated to the RC

mpuncel:
you're correct about the reason current is higher than replica count.

I hadn't thought about tweaking the replica counts during/after a node transfer, but I think that would be hard to reason about when a rolling update is happening (although node transfers are already somewhat difficult to reason about during rolling updates).

The change in this PR is kind of complex, but it's worth noting that the check I had to change (that makes sure the counts are equal when enabling an RC) is something we added after probably a year of using rolling updates. It is meant to catch an exceedingly rare race condition, so even if this change seems complex, it's just a tweak to an edge case detector which makes me feel more comfortable with it than making a large design change to node transfers at this point

spenceral:
I do get that that would be pretty fundamental change to transfers relatively late in the implementation, but I think we should keep it in our minds in the event that we discover additional problems/complexity later for which that might be an appropriate solution.

 Thanks for the additional context on the race condition fixed here.

pkg/roll/run_update.go
```diff
@@ -268,6 +278,7 @@ func (u *update) Run(ctx context.Context) (ret bool) {
 	u.logger.NoFields().Debugln("Enabling")
 	err = u.enable(checkRCLocksCtx)
 	if err != nil {
+		u.logger.WithError(err).Errorln("could not enable TCs")
```
spenceral: RCs?

PR #987, opened 2017-12-13
p2-rm: allow flag to check for missing RC

mpuncel:
Typically p2-rm assumes that a pod either does not have a
replication_contoller_id label (and thus is safe to unschedule) or it
does have one, and decrementing the replica count on the managing RC is
the proper course of action.

However in practice a pod will occasionally end up in a state where it
has the replication_controller_id label but no RC with that ID exists,
which means that we should use the manual method of removing the pod.
This commit adds a --remove-orphan flag which will make p2-rm do an
extra check that the replication controller ID returns actually exists

Additionally this commit makes p2-rm remove the labels for any pods it
deletes

### DIFF
--- a/pkg/hoist/hoist_launchable_test.go
+++ b/pkg/hoist/hoist_launchable_test.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"os"
 	"path"
+	"strings"
 	"testing"
 
 	"github.com/square/p2/pkg/launch"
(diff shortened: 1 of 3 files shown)


pkg/hoist/hoist_launchable_test.go
```diff
@@ -234,9 +235,11 @@ func TestDisable(t *testing.T) {
 	disableOutput, err := hl.disable()
 	Assert(t).IsNil(err, "Got an unexpected error when calling disable on the test hoist launchable")
 
-	expectedDisableOutput := "disable invoked\n"
+	expectedDisableOutput := "disable invoked"
 
-	Assert(t).AreEqual(disableOutput, expectedDisableOutput, "Did not get expected output from test disable script")
+	if !strings.Contains(disableOutput, "disable invoked") {
```
petertseng: you maybe wanted to use expectedDisableOutput here

PR #988, opened 2017-12-13
daemon set farm: update replication timeout

mpuncel:
The daemon set runner responds to some changes to daemon sets such as
the manifest, but it did not until now respond to changes to the
timeout. This is something that someone might want to take effect before
the daemon set server is restarted.

### DIFF
--- a/pkg/replication/replication.go
+++ b/pkg/replication/replication.go
@@ -77,6 +77,9 @@ type Replication interface {
 
 	// SetManifest() can be used to change the manifest while a replication is in progress
 	SetManifest(manifest.Manifest)
+
+	// SetTimeout() is used to change the timeout used for the replication while it is in progress
+	SetTimeout(timeout time.Duration)
 }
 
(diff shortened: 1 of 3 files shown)


pkg/replication/replication.go
```diff
@@ -425,9 +427,17 @@ func (r *replication) SetManifest(man manifest.Manifest) {
 	r.manifest = man
 }
 
+func (r *replication) SetTimeout(timeout time.Duration) {
+	// not changing the manifest technically but this will work, don't expect
+	// timeout changes to be highly concurrent with manifest changes
+	r.mu.Lock()
+	r.timeout = timeout
```
rudle: I'm reading the code in Enact() and I believe that this timeout will be re-read from the *replication each time the loop ticks. Does that match your assessment?

mpuncel: yeah, each time a node is pulled off the queue

PR #989, opened 2017-12-14
RC checks for missing artifacts in meetDesires call

esuen:
- get artifact url with artifact registry
- do get request to artifact url
- if not 200 status response then create low urgency alert

If I do Head request to artifact url then it will always return 200 even artifact url is invalid.
With Get request, 200 on artifact existing and downloadable, 404 if artifact url is invalid.

### DIFF
--- a/pkg/rc/replication_controller.go
+++ b/pkg/rc/replication_controller.go
@@ -10,6 +10,7 @@ import (
 	klabels "k8s.io/kubernetes/pkg/labels"
 
 	"github.com/square/p2/pkg/alerting"
+	"github.com/square/p2/pkg/artifact"
 	"github.com/square/p2/pkg/audit"
 	grpc_scheduler "github.com/square/p2/pkg/grpc/scheduler/client"
 	"github.com/square/p2/pkg/health"
@@ -142,6 +143,8 @@ type replicationController struct {
 	// nodeTransferMu protects access to nodeTransfer because it is used
 	// across goroutines
 	nodeTransferMu sync.Mutex
+
+	artifactRegistry artifact.Registry
 }
 
 type ReplicationControllerWatcher interface {
@@ -162,6 +165,7 @@ func New(
 	logger logging.Logger,
 	alerter alerting.Alerter,
 	healthChecker checker.ConsulHealthChecker,
+	artifactRegistry artifact.Registry,
 ) ReplicationController {
 	if alerter == nil {
 		alerter = alerting.NewNop()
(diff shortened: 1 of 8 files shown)


pkg/rc/replication_controller.go
```diff
@@ -264,6 +265,26 @@ func (rc *replicationController) WatchDesires(quit <-chan struct{}) <-chan error
...
 func (rc *replicationController) meetDesires(rcFields fields.RC) error {
 	rc.logger.NoFields().Infof("Handling RC update: desired replicas %d, disabled %v", rcFields.ReplicasDesired, rcFields.Disabled)
+	podID := rcFields.Manifest.ID()
+	launchableStanzas = rcFields.Manifest.GetLaunchableStanzas()
+	for launchableID, launchableStanza := range launchableStanzas {
+		artifactUrl, _, err := rc.artifactRegistry.LocationDataForLaunchable(podID, launchableID, launchableStanza)
+		if err != nil {
+			rc.logger.WithError(err).Errorln("Unable to retrieve location for launchable")
+		}
+		res, err := http.Get(artifactUrl)
```
rudle: Would a HEAD request suffice? Artifacts can potentially be gigabytes in size, so it would be a huge advantage to use HEAD.

esuen:
Fixing compile time errors, don't review yet.

pkg/rc/replication_controller.go
esuen: When I try HEAD, it always gives me a 200 response whether the artifact url is valid or not.

esuen: I'll try debugging the head request, I agree it should not be GET.

pkg/rc/replication_controller.go
```diff
@@ -162,6 +165,7 @@ func New(
 	logger logging.Logger,
 	alerter alerting.Alerter,
 	healthChecker checker.ConsulHealthChecker,
+	artifactRegistry artifact.Registry,
```
mpuncel: you need to pass this `artifactRegistry` into the `&replicationController{}` that this function returns

pkg/rc/replication_controller.go
```diff
@@ -264,6 +269,26 @@ func (rc *replicationController) WatchDesires(quit <-chan struct{}) <-chan error
...
+			rc.logger.WithError(err).Errorln("Unable to retrieve location for launchable")
+		}
+		exists, err := rc.artifactRegistry.CheckArtifactExists(artifactUrl)
+		if err != nil {
+			rc.logger.WithError(err).Errorln("Unexpected error when checking if artifact exists")
+		}
+		if !exists {
+			if err := rc.alerter.Alert(alerting.AlertInfo{
+				Description: "This RC is missing an artifact",
+				IncidentKey: "rc_missing_artifact",
```
mpuncel: i'd put the RC ID in the incident key, because pagerduty uses these to deduplicate incidents. we probably want an incident per RC that is missing an artifact

mpuncel: you should probably use the `alertInfo()` function as well which puts additional helpful details into the page, like the RC's ID. You might have to modify it to take the incident key as an argument

PR #990, opened 2017-12-14
Read only deploys

stouset:


rudle:
Nice, that turned out to be a pretty clean change.

stouset:
In retrospect, I need to check if we need to exempt a few directories for
things like tmp and log (or make sure they’re symlinks outside the install
directory)

PR #994, opened 2017-12-19
Add function to add pod resource limits to manifest

spenceral:
This change adds a manifest builder function that creates a stanza for pod level resource limits. 

This will be used for manifest upgrades so that all deployed manifests will have pod level resource limits.

### DIFF
--- a/pkg/manifest/manifest.go
+++ b/pkg/manifest/manifest.go
@@ -18,6 +18,7 @@ import (
 	"github.com/square/p2/pkg/types"
 	"github.com/square/p2/pkg/uri"
 	"github.com/square/p2/pkg/util"
+	"github.com/square/p2/pkg/util/size"
 	"golang.org/x/crypto/openpgp/clearsign"
 	"gopkg.in/yaml.v2"
 )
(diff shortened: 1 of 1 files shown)


pkg/manifest/manifest.go
```diff
@@ -213,6 +221,22 @@ func (manifest *manifest) SetStatusLocalhostOnly(localhostOnly bool) {
...
+func (manifest *manifest) SetResourceLimits(limits *ResourceLimitsStanza) {
+	if limits != nil {
+		manifest.ResourceLimits = *limits
+		return
+	}
+
+	cpus := 0
+	memory := size.ByteCount(0)
+	for _, stanza := range manifest.LaunchableStanzas {
+		cpus += stanza.CgroupConfig.CPUs
```
rudle: This is an OK in general, however it might be dangerous to make this the default. Consider that some apps don't have cgroup limits for all their launchables, and that java apps don't set memory limits at all. Under such conditions, this function will generate a cgroup that is too small to even boot the process.

Our backfill at Square will have to proceed by querying the current allocation data and backfilling it that way. We'll have to make this a manual step for our users.

spenceral: Okay, I had a hunch that - summing launchable limits - might have been a little too simple for all cases. I'll change this function to create a resource limits stanza based on the arguments passed. Callers of this function will have to do the work themselves to determine appropriate limits.

PR #997, opened 2018-01-03
Add missing log fields to hook messages

mpuncel:
Most preparer log messages have the associated "pod" key using the
logrus logger, however inadvertently when executing hooks a created
sublogger was not being passed through, which caused some fields to be
dropped. For example, when a hook fails there's no "pod=" key to tell
the user which pod the hook failed for.

This commit plumbs the sublogger through to every place in the hook
runner that needs it


rudle:
Good change. The _test files in the hook package don't compile anymore, so they require attention.

PR #998, opened 2018-01-04
Implement p2-start and p2-stop

esuen:
- p2-start and p2-stop modeled after p2-restart but checks if user is authorized to run the script with authPolicy

### DIFF
--- a/bin/p2-start/main.go
+++ b/bin/p2-start/main.go
@@ -1,0 +1,130 @@
+package main
+
+import (
+	"fmt"
+	"os"
+	"os/user"
+	"path/filepath"
+	"strconv"
+
+	"github.com/Sirupsen/logrus"
+	"github.com/square/p2/pkg/auth"
+	"github.com/square/p2/pkg/constants"
+	"github.com/square/p2/pkg/p2start"
+	"github.com/square/p2/pkg/pods"
+	"github.com/square/p2/pkg/types"
+	"github.com/square/p2/pkg/version"
+	"gopkg.in/alecthomas/kingpin.v2"
+)
+
+const (
+	fullDescription = `Safely start and enable an existing pod.
+
+EXAMPLES
+
+$ p2-start mypod
+
+$ p2-start --pod-dir /custom/pod/home
+
+`
+)
+
+var (
+	app = kingpin.New("p2-start", fullDescription)
+
+	nodeName = app.Flag("node-name", "The name of this node (default: hostname)").String()
+	podDir   = app.Flag("pod-dir", "The directory where the pod to be started is located. ").String()
+	podName  = app.Arg("pod-name", fmt.Sprintf("The name of the pod to be started. Looks in the default pod home '%s' for the pod", pods.DefaultPath)).String()
+)
+
+func main() {
+	app.Version(version.VERSION)
+	kingpin.MustParse(app.Parse(os.Args[1:]))
+
+	pods.Log.Logger.Formatter = &logrus.TextFormatter{
+		DisableTimestamp: false,
+		FullTimestamp:    true,
+		TimestampFormat:  "15:04:05.000",
+	}
+	logger := pods.Log
+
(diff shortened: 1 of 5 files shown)


bin/p2-start/main.go
```diff
@@ -0,0 +1,124 @@
...
+	}
+	email := user + p2start.DefaultDomain
+	if !authPolicy.Authorize(email, manifest.ID().String()) {
+		logger.WithError(err).Fatalln("Could not authorize user, may not have permission to run script")
+	}
+
+	logger = logger.SubLogger(logrus.Fields{"pod": pod.Id})
+	logger.NoFields().Infoln("Finding services to start")
+
+	ls, err := pod.Launchables(manifest)
```
rudle: What specific case does this guard against? Does `pod.Launch()` (or `pod.Halt()`) throw an error if there are no launchables?

esuen: Hmm, not sure what would happen if a pod had 0 launchables or if a pod with 0 launchables could exist. I simply copied the code over from p2-restart.

pkg/auth/auth_policy.go
```diff
@@ -36,6 +36,8 @@ type Policy interface {
 	// returned.
 	AuthorizeApp(manifest Manifest, logger logging.Logger) error
 
+	Authorize(email, appUser string) bool
```
rudle: is this an email? At Square, I bet this is an LDAP handle. How about calling this "username" or "handle" instead

bin/p2-start/main.go
rudle: Hm. That's probably a bug too then!

pkg/auth/auth_policy.go
esuen: It will be an email in the form of ldap + "@squareup.com", if that is considered an username then I am up for renaming this.

bin/p2-start/main.go
```diff
@@ -0,0 +1,124 @@
...
+	logger.NoFields().Infoln("Starting pod")
+
+	ok, err := pod.Launch(manifest)
+	if err != nil {
+		logger.WithError(err).Fatalln("Could not start pod")
+	} else if !ok {
+		logger.NoFields().Warningln("Some services did not come up - check output for details")
+	}
+
+	logger.NoFields().Infoln("Start successful")
```
mpuncel: probably this should be inside an `else`

pkg/auth/auth_policy.go
mpuncel: it is an email because that's how PGP works

mpuncel: but in practice we generate the email by adding a username and a domain

bin/p2-stop/main.go
```diff
@@ -0,0 +1,124 @@
...
+	)
+	if err != nil {
+		logger.WithError(err).Fatalln("error configuring user auth")
+	}
+
+	user := os.Getenv("SUDO_USER")
+	if user == "" {
+		logger.Fatalln("Could not get SUDO_USER")
+	}
+	email := user + p2stop.DefaultDomain
```
rudle: golang prefers fmt.Sprintf() for building strings. I also suspect you need an "@" in here somewhere

pkg/auth/auth_policy.go
rudle: wild. OK, good enough.

bin/p2-start/main.go
esuen: So this guards against an error case in pod.Launch(), it calls pod.buildRunitServives with an empty ServiceTemplate which causes in an error in servicebuilder.go write function "Could not remove old servicebuilder template when new path should have no services" on line 180.

bin/p2-stop/main.go
esuen: The default domain in the build script was set to "@squareup.com".

PR #1001, opened 2018-01-09
Rate limit artifact checker

esuen:


pkg/rc/replication_controller.go
```diff
@@ -196,6 +197,11 @@ func (rc *replicationController) WatchDesires(quit <-chan struct{}) <-chan error
 	errOutChannel := make(chan error)
 	channelsClosed := make(chan struct{})
 
+	// get random wait time to check for missing artifacts bewteen 1 - 60 minutes
+	randomTime := rand.Intn(60) + 1
+	shouldInitialCheck := time.After(fmt.Sprintf("%dm", randomTime))
```
mpuncel: Does `time.After()` take a string? I think you probably want `randomTime * time.Minute` here

pkg/rc/replication_controller.go
```diff
@@ -196,6 +197,11 @@ func (rc *replicationController) WatchDesires(quit <-chan struct{}) <-chan error
...
+	shouldPeriodicCheck := time.Tick(fmt.Sprintf("%dm", randomTime+60))
```
mpuncel: i'd probably make this a `time.NewTimer(randomTime * time.Minute + time.Hour)` to start off with and then call `timer.Reset(time.Hour)` each time it fires, that way we get exactly hour spacing after the random time

pkg/rc/replication_controller.go
```diff
@@ -236,6 +242,12 @@ func (rc *replicationController) WatchDesires(quit <-chan struct{}) <-chan error
 		}
 
 		for rcFields := range rcChanges {
+			select {
+			case <-shouldInitialCheck:
+				rc.checkMissingArtifacts(rcFields)
+			case <-shouldPeriodicCheck:
+				rc.checkMissingArtifacts(rcFields)
```
mpuncel: as I said in my last comment i'd probably do a `shouldPeriodicCheck.Reset(time.Hour)` here

pkg/rc/replication_controller.go
mpuncel: what you have is fine too, it's just weird for there to be a random initialization time AND a random interval after that, but maybe i'm being too particular

pkg/rc/replication_controller.go
```diff
@@ -236,6 +242,12 @@ func (rc *replicationController) WatchDesires(quit <-chan struct{}) <-chan error
...
+			}
```
mpuncel: you should include a `default:` case here as well. The `default` case will fire if neither of the other channels are ready, which is desirable here because we might have to wait an hour for either of the channels to fire and we don't want to block that long

pkg/rc/replication_controller.go
esuen: Oops, I meant something like that, but what you said can inspired me to do this with one timer.

PR #1002, opened 2018-01-10
P2 start p2 stop fix

esuen:
had to fix some errors in the authorize fail case and p2start pkg incorrectly used in p2start script.


mpuncel:
did something funny happen with the commits here?

PR #1003, opened 2018-01-11
Pod Cgroups

rudle:
This is maybe as nice as the commit history will get.

There's a lot here, so let me go over the strategy.

Cgroups for pods are optional, for now. If the manifest has a `resource_limits` key at the top level, the preparer shall parse the file and write the limits to $pod/config/resource_limits.yaml. This file will be referenced in the p2-exec command line so that the cgroup is created (possibly redundantly) when the app starts (or restarts). 

The cgroup directory structure is nested underneath a new top-level group (called "p2") and then under the hostname. This fixes a long-standing bug where we couldn't properly cgroup things that ran inside of containers. Launchable cgroups are underneath pod cgroups where you'd expect them to be.

### DIFF
--- a/bin/p2-exec/main.go
+++ b/bin/p2-exec/main.go
@@ -16,6 +16,9 @@ import (
 	"golang.org/x/sys/unix"
 
 	"github.com/square/p2/pkg/cgroups"
+	"github.com/square/p2/pkg/manifest"
+	"github.com/square/p2/pkg/pods"
+	"github.com/square/p2/pkg/types"
 	"github.com/square/p2/pkg/util"
 	"github.com/square/p2/pkg/version"
 	"gopkg.in/alecthomas/kingpin.v2"
@@ -31,14 +34,15 @@ var (
 		"extra-env",
 		"Specifies an extra environment KEY=VALUE pair to set for the process. May be used multiple times. Takes precedence over --env if there are conflicts",
 	).StringMap()
-	launchableName = kingpin.Flag("launchable", "The key in $PLATFORM_CONFIG_PATH containing the cgroup parameters.").Short('l').String()
-	requireFile    = kingpin.Flag("require-file", "Check for the presence of a required file before execing its argument").String()
-	cgroupName     = kingpin.Flag("cgroup", "The name of the cgroup that should be created.").Short('c').String()
-	nolim          = kingpin.Flag("nolimit", "Remove rlimits.").Short('n').Bool()
-	clearEnv       = kingpin.Flag("clearenv", "Clear all environment variables before loading envDir(s).").Bool()
-	workDir        = kingpin.Flag("workdir", "Set working directory.").Short('w').String()
-	umask          = kingpin.Flag("umask", "Set the process umask. Use octal notation ex. 0022").Short('m').Default(umaskDefault).String()
-	umaskDefault   = ""
+	launchableName       = kingpin.Flag("launchable", "The key in $PLATFORM_CONFIG_PATH containing the cgroup parameters.").Short('l').String()
+	podID                = kingpin.Flag("podID", "Setting this optional flag will nest the process under the cgroup for this pod, creating it if necessary.").Short('p').String()
+	requireFile          = kingpin.Flag("require-file", "Check for the presence of a required file before execing its argument").String()
+	launchableCgroupName = kingpin.Flag("cgroup", "The name of the cgroup that should be created for the executable. You probably want this to match your executable name.").Short('c').String()
+	nolim                = kingpin.Flag("nolimit", "Remove rlimits.").Short('n').Bool()
+	clearEnv             = kingpin.Flag("clearenv", "Clear all environment variables before loading envDir(s).").Bool()
+	workDir              = kingpin.Flag("workdir", "Set working directory.").Short('w').String()
+	umask                = kingpin.Flag("umask", "Set the process umask. Use octal notation ex. 0022").Short('m').Default(umaskDefault).String()
+	umaskDefault         = ""
 
 	cmd = kingpin.Arg("command", "the command to execute").Required().Strings()
 )
@@ -80,15 +84,35 @@ func main() {
 		}
 	}
 
-	if *launchableName == "" && *cgroupName != "" {
-		log.Fatalf("Specified cgroup name %q, but no launchable name was specified", *cgroupName)
+	if *podID != "" && *launchableName != "" && *launchableCgroupName != "" {
+		hostname, err := os.Hostname()
+		if err != nil {
+			log.Fatal(err)
+		}
+		if resourceLimitsConf := os.Getenv(pods.ResourceLimitsPathEnvVar); resourceLimitsConf != "" {
+			if _, err := os.Stat(resourceLimitsConf); err == nil {
+				err := createPodCgroup(resourceLimitsConf, types.PodID(*podID), types.NodeName(hostname))
+				if err != nil {
+					log.Fatal(err)
+				}
+			} else if os.IsNotExist(err) {
+			} else { // implies err != nil && err != os.IsNotExist(err)
+				log.Fatal(err)
+			}
+		} // TODO implement launchable cgroups based on ResourceLimitsPath also. Presently, they are implemented in an entirely separate conditional
+	} else if *podID != "" {
+		log.Fatal("Missing required arguments when specifying podID.")
 	}
-	if *launchableName != "" && *cgroupName == "" {
+
+	if *launchableName == "" && *launchableCgroupName != "" {
+		log.Fatalf("Specified cgroup name %q, but no launchable name was specified", *launchableCgroupName)
+	}
+	if *launchableName != "" && *launchableCgroupName == "" {
 		log.Fatalf("Specified launchable name %q, but no cgroup name was specified", *launchableName)
 	}
-	if *launchableName != "" && *cgroupName != "" {
+	if *launchableName != "" && *launchableCgroupName != "" {
 		if platconf := os.Getenv("PLATFORM_CONFIG_PATH"); platconf != "" {
-			err := cgEnter(platconf, *launchableName, *cgroupName)
+			err := cgEnter(platconf, *launchableName, *launchableCgroupName)
 			if err != nil {
 				log.Fatal(err)
 			}
@@ -232,9 +256,9 @@ func cgEnter(platconf, launchableName, cgroupName string) error {
 		return util.Errorf("Launchable %q has malformed PLATFORM_CONFIG_PATH", launchableName)
 	}
 	cgConfig := cgMap[launchableName]["cgroup"]
-	cgConfig.Name = cgroupName
+	cgConfig.Name = cgroups.CgroupID(cgroupName)
 
-	cg, err := cgroups.Find()
+	cg, err := cgroups.DefaultSubsystemer.Find()
 	if err != nil {
 		return util.Errorf("Could not find cgroupfs mount point: %s", err)
 	}
@@ -247,7 +271,7 @@ func cgEnter(platconf, launchableName, cgroupName string) error {
 	} else if err != nil {
 		return util.Errorf("Could not set cgroup parameters: %s", err)
 	}
-	return cg.AddPID(cgConfig.Name, 0)
+	return cg.AddPID(cgConfig.Name.String(), 0)
 }
 
 // generalized code to remove rlimits on both darwin and linux
@@ -288,5 +312,44 @@ func nolimit() error {
 	if ret != 0 && err != nil {
 		return util.Errorf("Could not set RLIMIT_RSS: %s", err)
 	}
+	return nil
+}
+
+// This function could do the translation from manifest to resource_limits_path, I don't believe the file is necessary before this point (but it is necessary after)
+func createPodCgroup(resourceLimitsPath string, podID types.PodID, hostname types.NodeName) error {
+	limits, err := ioutil.ReadFile(resourceLimitsPath)
+	// this file is optional
+	if os.IsNotExist(err) {
+		return nil
+	}
+	if err != nil {
+		return err
+	}
+
+	cfg := &manifest.ResourceLimitsConfigFileSchema{}
+	err = yaml.Unmarshal(limits, cfg)
+	if err != nil {
+		return err
+	}
+	if cfg == nil {
+		return util.Errorf("Error unserializing cgroup file, please check the contents of %s", resourceLimitsPath)
+	}
+
+	podLimits, ok := cfg.PodLimits[podID]
+	if !ok {
+		return util.Errorf("Did not find any pod limits in file at: %s, instead found %+v", resourceLimitsPath, podLimits)
+	}
+	if podLimits.CPUs == 0 {
+		return util.Errorf("Expected cgroup config to contain a non-zero value for CPU. %+v", podLimits)
+	}
+	if podLimits.Memory == 0 {
+		return util.Errorf("Expected cgroup config to contain a non-zero value for CPU. %+v", podLimits)
+	}
+
+	err = cgroups.CreatePodCgroup(podID, hostname, podLimits, cgroups.DefaultSubsystemer)
+	if err != nil {
+		return err
+	}
+
 	return nil
 }
--- a/integration/single-node-slug-deploy/check.go
+++ b/integration/single-node-slug-deploy/check.go
@@ -21,6 +21,7 @@ import (
 	"strings"
 	"time"
 
+	"github.com/square/p2/pkg/cgroups"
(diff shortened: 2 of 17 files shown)


bin/main.go
```diff
@@ -0,0 +1,33 @@
+package main
+
+import (
+	"fmt"
```
mpuncel: terrible import

rudle: what is this code, how did I

pkg/cgroups/cgroups.go
```diff
@@ -1,26 +1,27 @@
...
 
 // maps cgroup subsystems to their respective paths
 type Subsystems struct {
 	CPU    string
 	Memory string
+	Prefix string
 }
 
-var Default Subsystems = Subsystems{
+var DefaultSubsystems = Subsystems{
```
rudle: how did this ever work? parser bug?

petertseng: no parser bug... the name is `Default` and the type is `Subsystems`. Amazing.

bin/p2-exec/main.go
```diff
@@ -21,6 +21,10 @@ import (
 	"gopkg.in/alecthomas/kingpin.v2"
 	"gopkg.in/yaml.v2"
 )
+import (
```
mpuncel: nit: there are now 2 import blocks in this file

bin/p2-exec/main.go
```diff
@@ -31,14 +35,16 @@ var (
...
-	nolim          = kingpin.Flag("nolimit", "Remove rlimits.").Short('n').Bool()
-	clearEnv       = kingpin.Flag("clearenv", "Clear all environment variables before loading envDir(s).").Bool()
-	workDir        = kingpin.Flag("workdir", "Set working directory.").Short('w').String()
-	umask          = kingpin.Flag("umask", "Set the process umask. Use octal notation ex. 0022").Short('m').Default(umaskDefault).String()
-	umaskDefault   = ""
+	launchableName       = kingpin.Flag("launchable", "The key in $PLATFORM_CONFIG_PATH containing the cgroup parameters.").Short('l').String()
+	podID                = kingpin.Flag("podID", "Setting this optional flag will nest the process under the cgroup for this pod, creating it if necessary.").Short('p').String()
+	resourceLimitsPath   = kingpin.Flag("resourceLimitsPath", "The path to the file where the cgroup config for the pod is set. Mandatory if podID is set.").Short('r').String()
+	requireFile          = kingpin.Flag("require-file", "Check for the presence of a required file before execing its argument").String()
+	launchableCgroupName = kingpin.Flag("cgroup", "The name of the cgroup that should be created.").Short('c').String()
```
mpuncel: nit: the help text could probably use some love

bin/p2-exec/main.go
```diff
@@ -80,15 +86,29 @@ func main() {
...
+		hostname, err := os.Hostname()
+		if err != nil {
+			log.Fatal(err)
+		}
+		err = createPodCgroup(*resourceLimitsPath, types.PodID(*podID), types.NodeName(hostname))
+		if err != nil {
+			log.Fatal(err)
+		}
+	} else if *podID != "" {
+		log.Fatalf("Missing required arguments when specifying podID.")
```
mpuncel: you'll probably get a go vet error for the unused formatting

bin/p2-exec/main.go
```diff
@@ -290,3 +310,26 @@ func nolimit() error {
...
+
+	cfg := &manifest.ResourceLimitsStanza{}
+	err = yaml.Unmarshal(limits, cfg)
+	if err != nil {
+		return err
+	}
+	if cfg == nil {
+		return util.Errorf("Error unserializing cgroup file, please check the contents of %s", resourceLimitsPath)
+	}
+
```
mpuncel: do you need to check that `cfg.Cgroup` has nonzero values for CPU,memory, etc?

pkg/cgroups/subsystem.go
```diff
@@ -0,0 +1,58 @@
...
+
+func (ps *ProcSubsystemer) Find() (Subsystems, error) {
+	// For details about how this file is structured, refer to `man proc` or
+	// https://www.kernel.org/doc/Documentation/filesystems/proc.txt section 3.5
+	mountInfo, err := os.Open("/proc/self/mountinfo")
+	if err != nil {
+		return Subsystems{}, err
+	}
+	defer mountInfo.Close()
+	if ps.CachedSubsystems != nil {
```
mpuncel: nit: you can do this before opening the proc file no? or are you trying to confirm that it is still there

pkg/cgroups/cgroups.go
```diff
@@ -187,3 +159,74 @@ func appendIntToFile(filename string, data int) error {
...
+	ss, err := s.Find()
+	if err != nil {
+		return err
+	}
+
+	err = os.MkdirAll(ss.Prefix, 0755)
+	if err != nil && !os.IsExist(err) {
+		return err
+	}
+	err = os.MkdirAll(ss.Prefix, 0755)
```
mpuncel: looks like you have this twice?

pkg/cgroups/cgroups.go
```diff
@@ -187,3 +159,74 @@ func appendIntToFile(filename string, data int) error {
 	_, err = fd.WriteString(strconv.Itoa(data))
 	return err
 }
+
+// Subsystemer is anything that can find a set of Subsystems
+type Subsystemer interface {
+	Find() (Subsystems, error)
+}
+
+func ensurePrefix(s Subsystemer) error {
```
mpuncel: is this used?

pkg/cgroups/cgroups_test.go
```diff
@@ -0,0 +1,63 @@
...
+
+	if err = os.Chmod(fs.tmpdir, os.ModePerm); err != nil {
+		return Subsystems{}, err
+	}
+	return Subsystems{CPU: filepath.Join(fs.tmpdir, "cpu"), Memory: filepath.Join(fs.tmpdir, "memory")}, nil
+}
+
+// createTmpDir is useful only for tests, it creates the tmpdir
+func (fs *FakeSubsystemer) createTmpdir() error {
+	tmpdir, err := ioutil.TempDir("", "")
```
mpuncel: do tempdirs automatically get cleaned up?

pkg/cgroups/cgroups_test.go
```diff
@@ -0,0 +1,63 @@
...
+	expectCgroupFileToContain(t, "2048\n", filepath.Join(fs.tmpdir, "memory", "p2", hostname.String(), podID.String(), "memory.memsw.limit_in_bytes"))
+	expectCgroupFileToContain(t, "1024\n", filepath.Join(fs.tmpdir, "memory", "p2", hostname.String(), podID.String(), "memory.soft_limit_in_bytes"))
+}
+
+func expectCgroupFileToContain(t *testing.T, s string, file string) {
+	actual, err := ioutil.ReadFile(filepath.Join(file))
+	if err != nil {
+		t.Fatalf("err: %v", err)
+	}
+	if string(actual) != s { // We set the hard limit to 2x the request
```
mpuncel: i'm confused what the comment is telling me

pkg/pods/pod.go
```diff
@@ -563,6 +585,17 @@ func (pod *Pod) setupConfig(manifest manifest.Manifest, launchables []launch.Lau
 		return util.Errorf("Error writing platform config file for pod %s: %s", manifest.ID(), err)
 	}
 
+	resourceLimitFilename, err := manifest.ResourceLimitsConfigFileName()
```
mpuncel: Do we need to support manifests that don't have this set for the time being?

mpuncel:
Approved but with some questions that might be PR blockers depending on the answer

pkg/pods/pod.go
rudle: Good point. I tested it with and without the manifest stanza and it worked, but I'll double checked that it's doing exactly what we want it to.

pkg/cgroups/cgroups.go
```diff
@@ -31,50 +32,16 @@ func (err UnsupportedError) Error() string {
...
-
-	return ret, nil
+type CgroupID string
+
+func (c *CgroupID) String() string {
+	return string(*c)
 }
 
+// SetCPU will set a number of CPU limits in the cgroup specified by the argument
 // set the number of logical CPUs in a given cgroup, 0 to unrestrict
```
spenceral: I don't quite follow these two comment lines

pkg/cgroups/cgroups.go
```diff
@@ -114,9 +81,10 @@ func (subsys Subsystems) SetCPU(name string, cpus int) error {
 	return nil
 }
 
+// SetMemory will set several memory limits in the cgroup specified in the argument
 // set the memory limit on a cgroup, 0 to unrestrict
```
spenceral: Same thing here. Is each line a separate sentence?

pkg/cgroups/cgroups.go
```diff
@@ -178,6 +146,10 @@ func (subsys Subsystems) AddPID(name string, pid int) error {
 	return appendIntToFile(filepath.Join(subsys.CPU, name, "cgroup.procs"), pid)
 }
 
+func (subsys Subsystems) GetPrefix() string {
```
spenceral: What is the justification for having the getter, given that both `subsys.GetPrefix()` and `subsys.Prefix` are exported? I would've expected a getter only if you had an unexported `subsys.prefix`.

nit: https://golang.org/doc/effective_go.html#Getters suggests it is not idiomatic to include `Get` in the getter name

rudle: right. I think I wanted to keep this private but then needed to make the field public for test. I'll remove this getter.

pkg/cgroups/cgroups.go
rudle: d'oh

pkg/manifest/manifest.go
```diff
@@ -72,8 +67,11 @@ type Manifest interface {
 	WriteConfig(out io.Writer) error
 	PlatformConfigFileName() (string, error)
 	WritePlatformConfig(out io.Writer) error
+	WriteResourceLimitsConfig(out io.Writer) error
 	GetLaunchableStanzas() map[launch.LaunchableID]launch.LaunchableStanza
 	GetResourceLimitsStanza() ResourceLimitsStanza
```
spenceral: You can get rid of `GetResourceLimitsStanzas()` (and the accompanying implementation) since you've added `GetResourceLimits()`

pkg/manifest/manifest.go
```diff
@@ -384,6 +390,18 @@ func (manifest *manifest) WritePlatformConfig(out io.Writer) error {
 	return nil
 }
 
+func (manifest *manifest) WriteResourceLimitsConfig(out io.Writer) error {
```
spenceral: What is the reasoning behind writing the resource limits out to a separate config file? They'll now exist in the cgroup files themselves, the config file, and the manifest. Why all three?

pkg/manifest/manifest.go
```diff
@@ -384,6 +390,18 @@ func (manifest *manifest) WritePlatformConfig(out io.Writer) error {
...
+	bytes, err := yaml.Marshal(manifest.ResourceLimits)
+	if err != nil {
+		return util.Errorf("Could not write resource limits for %s: %s", manifest.ID(), err)
```
spenceral: Should you distinguish this error message from the one on line 400 so we can differentiate between marshaling and writing errors more easily? Or do you think including the `err` at the end will be sufficient for that?

pkg/pods/pod.go
```diff
@@ -76,6 +77,9 @@ type Pod struct {
 
 	// Pod will not start if file is not present
 	RequireFile string
+
+	// subsystemer is a tool for this pod to find its cgroup subsystem controller and metadata. Optionally nil, overridden in test
```
spenceral: teeny-nit: is this comment a little misleading? I'm under the impression that `subsystemer` cannot technically be `nil`, only zero-valued

pkg/pods/pod.go
```diff
@@ -563,6 +585,17 @@ func (pod *Pod) setupConfig(manifest manifest.Manifest, launchables []launch.Lau
...
 
+	resourceLimitFilename, err := manifest.ResourceLimitsConfigFileName()
+	if err != nil {
+		return util.Errorf("Could not determine resource file name for pod %s: %s", manifest.ID(), err)
+	}
+	resourceLimitPath := filepath.Join(pod.ConfigDir(), resourceLimitFilename)
+	err = writeFileChown(resourceLimitPath, resourceLimitData.Bytes(), uid, gid)
+	if err != nil {
+		return util.Errorf("Error writing resource limit path for pod %s: %s", manifest.ID(), err)
+
```
spenceral: extra line

pkg/hoist/hoist_launchable.go
```diff
@@ -34,25 +35,28 @@ type EntryPoints struct {
...
+	P2Exec            string                     // Struct that can be used to build a p2-exec invocation with appropriate flags
+	ExecNoLimit       bool                       // If set, execute with the -n (--no-limit) argument to p2-exec
+	PodCgroupConfig   cgroups.Config             // PodCgroupConfig
+	ResourceLimitPath string                     // ResourceLimitPath points to the resource limits config file
+	CgroupConfig      cgroups.Config             // Cgroup parameters to use with p2-exec
+	CgroupConfigName  string                     // The string in PLATFORM_CONFIG to pass to p2-exec
+	CgroupName        string                     // The name of the cgroup to run this launchable in
+	RequireFile       string                     // Do not run this launchable until this file exists
+	RestartTimeout    time.Duration              // How long to wait when restarting the services in this launchable.
+	RestartPolicy_    runit.RestartPolicy        // Dictates whether the launchable should be automatically restarted upon exit.
```
spenceral: Why does this have an underscore?

pkg/preparer/orchestrate.go
```diff
@@ -401,6 +401,12 @@ func (p *Preparer) installAndLaunchPod(pair ManifestPair, pod Pod, logger loggin
 		return false
 	}
 
+	err = pod.CreateCgroupForPod()
+	if err != nil {
+		logger.WithError(err).
+			Errorln("Pod Cgroup creation failed, proceeding")
```
spenceral: So if we fail to create cgroups we allow a pod to run on a machine with unrestricted resources? That seems like lying about the available resources on the machine and allowing it to be a bad neighbor.

What do you think about failing the install/launch and letting it retry (since the key would still be in intent)?

pkg/hoist/hoist_launchable.go
rudle: This isn't my code, but I believe this avoids the golang gotcha that a field name cannot match a function name.

pkg/manifest/manifest.go
rudle: The resource_limits file is intended to replace the inflexible schema found in platform_config. This new file should support both pod and launchable cgroups and be able to versioned easily. The code here does not meet those goals – I will change it such that it does.

pkg/preparer/orchestrate.go
rudle: This is a fair point, I didn't consider that the preparer would retry.

rudle:
This code bomb is ready for another review

bin/p2-exec/main.go
```diff
@@ -80,15 +84,35 @@ func main() {
 		}
 	}
 
-	if *launchableName == "" && *cgroupName != "" {
-		log.Fatalf("Specified cgroup name %q, but no launchable name was specified", *cgroupName)
+	if *podID != "" && *launchableName != "" && *launchableCgroupName != "" {
```
spenceral: I don't understand the check for `*launchableName` and `*launchableCgroupName`. They don't seem to be used in this branch. 

Is their usage what the TODO on line 102 is referring to?

rudle: I guess so, I can remove them until we implement that feature.

integration/single-node-slug-deploy/check.go
```diff
@@ -266,9 +269,72 @@ func verifyLegacyPod(errCh chan error, tempDir string, config *preparer.Preparer
...
 }
 
+func verifyCgroups() error {
+	hostname, err := os.Hostname()
+	if err != nil {
+		return err
+	}
+	if _, err := os.Stat(fmt.Sprintf("/cgroup/cpu/p2/%s/hello/hello", hostname)); err != nil {
+		log.Printf("ERROR: %v", err)
+		// return err
```
spenceral: What's this?

rudle: hmmmmmmmmmmmmmmmmmmmmmmmm

PR #1005, opened 2018-01-16
Add retry with backoff to finishTransfer() call

spenceral:
Before this change, if there was an error building the final node
transfer transaction, whether due to a network partition or otherwise,
the node transfer would fail but would not rollback. This prevented
future node transfers from occurring.

This change adds a retry around the finishTransfer() call so that
something like a temporary network partition will not stop a node
transfer. In the event that the node transfer still fails after the
retries, there is likely an underlying problem so we now alert.

### DIFF
--- a/pkg/rc/replication_controller.go
+++ b/pkg/rc/replication_controller.go
@@ -3,6 +3,7 @@ package rc
 import (
 	"context"
 	"fmt"
+	"math"
 	"math/rand"
 	"os"
 	"sync"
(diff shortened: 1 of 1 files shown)


pkg/rc/replication_controller.go
```diff
@@ -952,13 +953,28 @@ func (rc *replicationController) scheduleNewNodeForTransfer(rcFields fields.RC,
...
+			time.Sleep(backoff)
+		}
+		rc.logger.WithError(err).Errorln("could not do final node transfer transaction after retries, rolling back")
+		alertErr := rc.alerter.Alert(rc.alertInfo(rcFields, err.Error()), alerting.LowUrgency)
+		if alertErr != nil {
+			logger.WithError(alertErr).Errorln("Unable to send alert")
 		}
 
-		rc.logger.WithError(err).Errorln("could not do final node transfer transaction, rolling back")
+		rc.nodeTransfer = nodeTransfer{}
```
mpuncel: I think you need to protect this by grabbing `nodeTransferMu` or else there's a data race

mpuncel:
looks good once you put in the mutex thing

spenceral:
Added the mutex

pkg/rc/replication_controller.go
```diff
@@ -952,13 +953,30 @@ func (rc *replicationController) scheduleNewNodeForTransfer(rcFields fields.RC,
...
+		for i := 0; i < 10; i++ {
+			err = rc.finishTransfer(rcFields, logger)
+			if err == nil {
+				logger.Infoln("Node transfer complete.")
+				return
+			}
+
+			rc.logger.WithError(err).Errorln("could not do final node transfer transaction, retrying")
+
+			backoff := time.Duration(math.Pow(2, float64(i))) * time.Second
```
rudle: the maximum value of this backoff is something 1024 seconds or ~17 minutes. WDYT about capping it at something like 60s?

spenceral: I think its 512 seconds but to your point I don't really have an opinion. I was unsure how to decide things like the number of allowed retries or the duration of the backoff so I picked them somewhat arbitrarily. 

I am definitely open to changing either. Whats the benefit to a smaller cap like 60s?

rudle: Fair enough, I had an off by one in my calculations.

For me, it's an argument of marginal utility. What's the likelihood that waiting 8 minutes generates fewer errors than waiting 1 minute? Given that someone is waiting for the node transfer to complete, raising the alarm earlier is better.

spenceral: SGTM. Thoughts on the 10 retries then? I think a similar argument can be made. Whats the likelihood that 10 retries will succeed if ~6 do not?

rudle: I think 10 is a fine number to start with. We'll retry automatically and be alerted within 10 minutes. I'd like the system to retry on its own until an unacceptable time threshold is met.

PR #1006, opened 2018-01-17
rc farm: Add error logging to a place that was missing it

mpuncel:


mpuncel:
fixed a racey test timeout in the new RU test

PR #1007, opened 2018-01-18
Add logs and use node transfer logger

spenceral:


pkg/rc/replication_controller.go
```diff
@@ -937,7 +939,6 @@ func (rc *replicationController) scheduleNewNodeForTransfer(rcFields fields.RC,
 	ok, _, err := auditingTransaction.Commit(rc.txner)
 	switch {
 	case err != nil:
-		logger.WithError(err).Errorln("could not schedule new node")
```
spenceral: I moved this log to where `scheduleNewNodeForTransfer` is called so that other errors returned in this function would also trigger a log.

PR #1010, opened 2018-01-23
Treat failure to delete old RC as an RU failure

mpuncel:


mpuncel:
I feel that this is difficult to write a unit test for and is straightforward enough that it probably doesn't need one?

rudle:
:shipit:

pkg/roll/run_update.go
```diff
@@ -480,7 +484,11 @@ func (u *update) cleanupOldRC(ctx context.Context) {
 			"rc-deletion-txn"+u.ID().String(),
```
mpuncel: we could probably make the transaction stuff return a special error type and only send this alert if the error is because the transaction got beyond 64 operations (consul's max). That was the intent here, all other errors don't make sense to alert for now that we retry them

PR #1011, opened 2018-01-29
use new health service with rolling updates

esuen:
Changes required for rolling updates to use the health service client for health checks while still using the consul health tree in parallel. These changes will test reliability of health service client.
- Created a separate healthChecker type in checker.go called ShadowTrafficHealthChecker that will utilize both the health service client and old consul tree checking code
- Old health checker was renamed from ConsulHealthChecker to HealthChecker so that in the future when the health service client is deemed reliable we can rename ShadowTrafficHealthChecker to HealthChecker
- WatchService and Service functions defined for new HealthChecker to be used by rolling updates
- WatchService uses both health service client and consul health tree
- Service uses either consul or health service client

### DIFF
--- a/bin/p2-watch/main.go
+++ b/bin/p2-watch/main.go
@@ -1,6 +1,7 @@
 package main
 
 import (
+	"context"
 	"fmt"
 	"log"
 	"os"
@@ -31,7 +32,7 @@ var (
 	watchReality  = kingpin.Flag("reality", "Watch the reality store instead of the intent store. False by default").Default("false").Bool()
 	hooks         = kingpin.Flag("hook", "Watch hooks.").Bool()
 	podClusters   = kingpin.Flag("pod-clusters", "Watch pod clusters and their labeled pods").Bool()
-	watchHealthF  = kingpin.Flag("health", "Watch health using ConsulHealthChecker").Bool()
+	watchHealthF  = kingpin.Flag("health", "Watch health using HealthChecker").Bool()
 	healthService = kingpin.Arg("health-pod", "Pod to watch. Required if --health is passed").String()
 )
 
@@ -138,12 +139,14 @@ func watchPodClusters(client consulutil.ConsulClient, applicator labels.Applicat
 }
 
 func watchHealth(service string, client consulutil.ConsulClient) {
-	hc := checker.NewConsulHealthChecker(client)
+	hc := checker.NewHealthChecker(client)
 	healthResults := make(chan map[types.NodeName]health.Result)
 	errCh := make(chan error)
 	quitCh := make(chan struct{})
+	ctx, cancelFunc := context.WithCancel(context.Background())
 	watchDelay := 1 * time.Second
-	go hc.WatchService(*healthService, healthResults, errCh, quitCh, watchDelay)
+	go hc.WatchService(ctx, *healthService, healthResults, errCh, watchDelay)
+	defer cancelFunc()
 
 	go func() {
 		signalCh := make(chan os.Signal, 2)
--- a/pkg/health/checker/checker.go
+++ b/pkg/health/checker/checker.go
@@ -1,29 +1,38 @@
 package checker
 
 import (
+	"context"
 	"encoding/json"
+	"fmt"
+	"net"
+	"net/url"
 	"time"
 
 	"github.com/square/p2/pkg/health"
+	hclient "github.com/square/p2/pkg/health/client"
+	"github.com/square/p2/pkg/labels"
+	"github.com/square/p2/pkg/manifest"
+	rcfields "github.com/square/p2/pkg/rc/fields"
 	"github.com/square/p2/pkg/store/consul"
 	"github.com/square/p2/pkg/store/consul/consulutil"
 	"github.com/square/p2/pkg/types"
 	"github.com/square/p2/pkg/util"
 
 	"github.com/hashicorp/consul/api"
+	klabels "k8s.io/kubernetes/pkg/labels"
 )
 
-type ConsulHealthChecker interface {
+type HealthChecker interface {
 	WatchPodOnNode(
 		nodename types.NodeName,
 		podID types.PodID,
 		quitCh <-chan struct{},
 	) (chan health.Result, chan error)
 	WatchService(
+		ctx context.Context,
 		serviceID string,
 		resultCh chan<- map[types.NodeName]health.Result,
 		errCh chan<- error,
-		quitCh <-chan struct{},
 		watchDelay time.Duration,
 	)
 	WatchHealth(
@@ -35,34 +44,90 @@ type ConsulHealthChecker interface {
 	Service(serviceID string) (map[types.NodeName]health.Result, error)
 }
 
-// Subset of consul.Store
-type healthStore interface {
-	GetHealth(service string, node types.NodeName) (consul.WatchResult, error)
-	GetServiceHealth(service string) (map[string]consul.WatchResult, error)
+type ShadowTrafficHealthChecker interface {
+	WatchService(
+		ctx context.Context,
+		serviceID string,
+		resultCh chan<- map[types.NodeName]health.Result,
+		errCh chan<- error,
+		watchDelay time.Duration,
+		useHealthService bool,
+		status manifest.StatusStanza,
+	)
+	Service(
+		serviceID string,
+		useHealthService bool,
+		status manifest.StatusStanza,
+	) (map[types.NodeName]health.Result, error)
+}
+
+type HealthClient interface {
+	HealthCheckEndpoints(ctx context.Context, req *hclient.HealthEndpointsRequest) (map[string]health.HealthState, error)
+	HealthMonitor(ctx context.Context, req *hclient.HealthRequest, respCh chan *hclient.HealthResponse) error
+	HealthMonitorEndpoints(ctx context.Context, req *hclient.HealthEndpointsRequest, respCh chan *hclient.HealthResponse) error
+}
+
+type ResourceClient interface {
+	GetRCIDsForPod(pod types.PodID) ([]rcfields.ID, error)
 }
 
 type healthKV interface {
 	List(prefix string, opts *api.QueryOptions) (api.KVPairs, *api.QueryMeta, error)
 }
 
-type consulHealthChecker struct {
-	client      consulutil.ConsulClient
-	kv          healthKV
-	consulStore healthStore
-}
-type consulHealth interface {
-	Node(types.NodeName, *api.QueryOptions) ([]*api.HealthCheck, *api.QueryMeta, error)
+// Subset of consul.Store
+type healthStore interface {
+	GetHealth(service string, node types.NodeName) (consul.WatchResult, error)
+	GetServiceHealth(service string) (map[string]consul.WatchResult, error)
 }
 
-func NewConsulHealthChecker(client consulutil.ConsulClient) ConsulHealthChecker {
-	return consulHealthChecker{
-		client:      client,
-		kv:          client.KV(),
-		consulStore: consul.NewConsulStore(client),
+type ReplicationControllerStore interface {
+	Get(id rcfields.ID) (rcfields.RC, error)
+}
+
+type LabelReader interface {
+	GetMatches(klabels.Selector, labels.Type) ([]labels.Labeled, error)
+}
+
+type shadowTrafficHealthChecker struct {
+	healthClient     HealthClient
+	resourceClient   ResourceClient
+	consulClient     consulutil.ConsulClient
+	kv               healthKV
+	consulStore      healthStore
+	rcStore          ReplicationControllerStore
+	labelReader      LabelReader
+	useHealthService bool
+}
+
+func NewShadowTrafficHealthChecker(hClient HealthClient, resourceClient ResourceClient, cClient consulutil.ConsulClient, rcStore ReplicationControllerStore, labelReader LabelReader, useHealthService bool) ShadowTrafficHealthChecker {
+	return shadowTrafficHealthChecker{
+		healthClient:     hClient,
+		resourceClient:   resourceClient,
+		consulClient:     cClient,
+		kv:               cClient.KV(),
+		consulStore:      consul.NewConsulStore(cClient),
+		rcStore:          rcStore,
+		labelReader:      labelReader,
+		useHealthService: useHealthService,
 	}
 }
 
-func (c consulHealthChecker) WatchPodOnNode(
+type healthChecker struct {
+	consulClient consulutil.ConsulClient
+	kv           healthKV
+	consulStore  healthStore
+}
+
+func NewHealthChecker(cClient consulutil.ConsulClient) HealthChecker {
+	return healthChecker{
+		consulClient: cClient,
+		kv:           cClient.KV(),
+		consulStore:  consul.NewConsulStore(cClient),
+	}
+}
+
+func (h healthChecker) WatchPodOnNode(
 	nodename types.NodeName,
 	podID types.PodID,
 	quitCh <-chan struct{},
@@ -75,7 +140,7 @@ func (c consulHealthChecker) WatchPodOnNode(
 	wsOut := make(chan *api.KVPair) // closed by WatchSingle
 	wsQuit := make(chan struct{})
 
-	go consulutil.WatchSingle(key, c.client.KV(), wsOut, wsQuit, errCh)
+	go consulutil.WatchSingle(key, h.consulClient.KV(), wsOut, wsQuit, errCh)
 
 	go func() {
 		defer close(wsQuit)
@@ -156,7 +221,7 @@ func publishLatestHealth(inCh <-chan api.KVPairs, quitCh <-chan struct{}, result
 			}
 
 			// here, we prepare to write the value.
-			// First we drain the resultChan of any stale health results
+			// First we drain the resultCh of any stale health results
 			select {
 			case _, ok = <-resultCh:
 				if !ok {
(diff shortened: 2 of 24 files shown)


pkg/health/checker/checker.go
```diff
@@ -5,14 +5,171 @@ import (
...
+		rcStore: rcStore
+	}
+}
+
+func (h healthServiceChecker) WatchPodOnNode(
+	nodename types.NodeName,
+	podID types.PodID,
+	quitCh <-chan struct{},
+) (chan health.Result, chan error) {
+	return consulHealthChecker(h.(healthServiceChecker)).WatchPodOnNode(nodename, podID, quitCh)
```
rudle: What's the long term fate for this type (`consulHealthChecker`)? What will happen when we move to only health checking via health service? I predicted that this code would introduce a new type that "covered" both consulHealthChecker and healthServiceHealthChecker and we'd use that everywhere. Can you remind me why that's not possible?

esuen: That's probably the better approach in the long term. That way when consulHealthChecker code can be removed when we move to only the health service. I'll make those changes.

esuen:
Ready for review, there's some minor TODOs I need to look at and some testing to be done.

esuen:
fixing checker, don't review

esuen:
you may review again..

pkg/health/checker/checker.go
```diff
@@ -41,28 +69,37 @@ type healthStore interface {
...
+	rcStore        ReplicationControllerStore
+	labelReader    LabelReader
 }
 
-func NewConsulHealthChecker(client consulutil.ConsulClient) ConsulHealthChecker {
-	return consulHealthChecker{
-		client:      client,
-		kv:          client.KV(),
-		consulStore: consul.NewConsulStore(client),
+func NewHealthChecker(hClient HealthClient, resourceClient ResourceClient, cClient consulutil.ConsulClient, rcStore ReplicationControllerStore, labelReader LabelReader) HealthChecker {
```
rudle: WDYT about a new constructor for clients who want to specify all of these arguments? The behaviour is special, so a special constructor isn't unreasonable. How about NewShadowTrafficHealthChecker(…). It has the benefit of being a bit clearer about intent and doesn't require the caller to specify so many `nil`s. At the very least, this constructor should declare which arguments are optional and which aren't.

pkg/health/checker/checker.go
```diff
@@ -213,75 +278,174 @@ func (c consulHealthChecker) WatchHealth(
...
-			timer.Reset(watchDelay)
-			results, queryMeta, err := c.client.KV().List(consul.HealthPath(serviceID, "/"), &api.QueryOptions{
-				WaitIndex: curIndex,
-			})
-			if err != nil {
-				select {
-				case <-quitCh:
-					return
-				case errCh <- consulutil.NewKVError("list", consul.HealthPath(serviceID, "/"), err):
+	if !useHealthService {
```
rudle: Is there any way to get this to send health check requests to both consul & the health service? I don't yet have enough confidence that the health service returns timely & correct health checks, which is why I want this code to send traffic to the trusted source (_consul_) and the new service so we can see how it behaves under production load levels.

esuen: Would this always send to both or only when useHealthService is true?

pkg/health/checker/checker.go
esuen: Are you suggesting to rename NewHealthChecker to NewShadowTrafficHealthChecker with optional parameters or to create an additional constructor.

Would it be implicitly or explicitly specified that certain parameters are optional? Golang doesn't seem to have a way to specify optional parameters unless the function takes a struct and the fields in the struct are optional I think.

mpuncel: I think he's suggesting a totally separate constructor, like `NewShadowTrafficHealthChecker` and leave the old `NewHealthChecker` alone for now. Then when we're confident in the health API service we can do a rename and drop the old consul related arguments

rudle: ^ yeah, this sounds right. It's easier to find callers of the old constructor than it is to find callers of a shared constructor that use deprecated arguments.

pkg/health/checker/checker.go
rudle: We can use the boolean to toggle whether or not this code sends to both.

pkg/health/checker/checker.go
rudle: In all the call sites where there are several nils passed in to the constructor, I suggest using a constructor that doesn't require the caller to pass in those nils. The NewShadowTrafficHealthChecker can call into the main constructor and pass in the nil arguments itself. This is a style request, it doesn't have much bearing on correctness.

Here's an example: https://play.golang.org/p/SaCDXIJO13-

esuen: Cool, thanks. Why the name "ShadowTraffic"?

pkg/health/checker/checker.go
```diff
@@ -265,23 +345,110 @@ func (c consulHealthChecker) WatchService(
...
+				if err != nil {
+					select {
+					case <-quitCh:
+						return
+					case errCh <- err:
+					}
+					continue
+				}
+				for _, endpoint := range statusEndpoints {
+					healthReq := &hclient.HealthRequest{
```
rudle: Does this need to change given the new health service proto? It's kind of annoying that the protos are closed source - it creates a bit of a circular dependency.

esuen: This doesn't need to change since this is calling HealthMonitor and it still uses HealthRequest messages.

pkg/health/checker/checker.go
mpuncel: "shadowing" traffic is the technique of duplicating all of the traffic going to destination A and sending it also to destination B. Usually this is done to see if destination B can handle the load that destination A can, and it's done when you are evaluating replacing A with B.

That's basically what we're doing here, we want to see if the health service can handle being accessed every time we used to be looking in consul

pkg/health/checker/checker.go
```diff
@@ -41,28 +69,45 @@ type healthStore interface {
...
+		consulClient:   cClient,
+		kv:             cClient.KV(),
+		consulStore:    consul.NewConsulStore(cClient),
+		rcStore:        rcStore,
+		labelReader:    labelReader,
 	}
 }
 
-func (c consulHealthChecker) WatchPodOnNode(
+func NewShadowTrafficHealthChecker(cClient consulutil.ConsulClient) HealthChecker {
```
mpuncel: This looks great, but you have the constructor names backwards :) this one should be the old untouched `NewConsulHealthChecker` and the other one (which does traffic shadowing to the health service) should be called `NewShadowTrafficHealthChecker`

pkg/health/checker/checker.go
```diff
@@ -176,10 +221,42 @@ func publishLatestHealth(inCh <-chan api.KVPairs, quitCh <-chan struct{}, result
 	return errCh
 }
 
+func (h healthChecker) getStatusEndpoints(
+	serviceID string,
+	status manifest.StatusStanza,
+) ([]string, error) {
+	rcIDs, err := h.resourceClient.GetRCIDsForPod(types.PodID(serviceID))
```
mpuncel: this function depends on fields that are only set if you called the constructor with more arguments. This means that if you 1) use the old constructor and 2) call `WatchService()` or `Service()` with `useHealthService=true` you're going to get a nil pointer exception!

This is making me think that the shadow checker and the non-shadow checker should probably be different types (not just `healthChecker`) and then their behavior can differ without needing to pass extra arguments like `useHealthChecker` to the functions.

A lot of what they do will be common, you can probably extract those into package level functions pretty easily (I hope). These package level functions would have to take all the struct fields as arguments.

pkg/health/checker/checker.go
```diff
@@ -176,10 +221,42 @@ func publishLatestHealth(inCh <-chan api.KVPairs, quitCh <-chan struct{}, result
...
+		// get hostnames for rc
+		labeled, err := h.labelReader.GetMatches(rc.NodeSelector, labels.NODE)
+		if err != nil {
+			return nil, err
+		}
+		// create status endpoints
+		statusEndpoints = make([]string, 0, len(labeled))
+		statusPath := status.Path
+		if status.Path == "" {
+			statusPath = "_status"
```
mpuncel: nit: this might be a nice detail to hide in the status stanza struct, e.g. `statusStanza.Path()` and it applies the default if necessary

pkg/health/checker/checker.go
```diff
@@ -265,23 +345,110 @@ func (c consulHealthChecker) WatchService(
...
+					}
+					continue
+				}
+				for _, endpoint := range statusEndpoints {
+					healthReq := &hclient.HealthRequest{
+						Url:      endpoint,
+						Protocol: "HTTPS",
+					}
+
+					err = h.healthClient.HealthMonitor(context.Background(), healthReq, respChan)
```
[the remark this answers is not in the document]
mpuncel: also the quit channel pattern is a dated go practice, contexts are preferable now

pkg/health/checker/checker.go
```diff
@@ -213,28 +290,31 @@ func (c consulHealthChecker) WatchHealth(
...
 ) {
 	defer close(resultCh)
-	var curIndex uint64 = 0
-
 	if watchDelay < time.Second {
 		watchDelay = time.Second
 	}
 
 	timer := time.NewTimer(0)
+	respChan := make(chan *hclient.HealthResponse, 1)
```
mpuncel: this channel is never closed, which means we're going to leak resources every time `WatchService` is called. We need to make sure that it's closed once we expect no values to be sent on it anymore.

Long winded explanation warning: In my experience (and I think I read in a blog post somewhere), the best way to get channel closing right is to only have the sender on the channel do the closing. In this case, the sender is the `h.healthClient.HealthMonitor()` function, and the receiver is this function (`WatchService()`).

What this means is that I think that instead of declaring `respChan` here in this function, you should have`h.healthClient.HealthMonitor()` create it and return it. Then, `HealthMonitor()` should wait for the passed context to be canceled, and when it is, close the channel and exit the goroutine. This guarantees that no values are sent after the channel is closed (which causes a panic)

pkg/health/checker/checker.go
```diff
@@ -265,23 +345,110 @@ func (c consulHealthChecker) WatchService(
...
+						case <-quitCh:
+							return
+						case errCh <- err:
+						}
+					} else {
+						select {
+						case resp, ok := <-respChan:
+							if ok {
+								out := make(map[types.NodeName]health.Result)
+								nodeID := types.NodeName(resp.HealthRequest.Url)
```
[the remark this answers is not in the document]
mpuncel: ah apparently that's going to require some changes to the health client in another repo :(

pkg/health/checker/checker.go
```diff
@@ -265,23 +345,110 @@ func (c consulHealthChecker) WatchService(
 				case resultCh <- out:
 				}
 			}
+			if useHealthService {
```
mpuncel: you'll also be able to get rid of this `if` if you use different types for the shadow and not-shadow checker

bin/p2-watch/main.go
```diff
@@ -138,12 +140,14 @@ func watchPodClusters(client consulutil.ConsulClient, applicator labels.Applicat
...
 func watchHealth(service string, client consulutil.ConsulClient) {
-	hc := checker.NewConsulHealthChecker(client)
+	hc := checker.NewHealthChecker(client)
 	healthResults := make(chan map[types.NodeName]health.Result)
 	errCh := make(chan error)
 	quitCh := make(chan struct{})
+	ctx, cancelFunc := context.WithCancel(context.Background())
 	watchDelay := 1 * time.Second
-	go hc.WatchService(*healthService, healthResults, errCh, quitCh, watchDelay)
+	go hc.WatchService(*healthService, healthResults, errCh, ctx, watchDelay, false, manifest.StatusStanza{})
```
mpuncel: nit: `ctx` should be the first parameter:

```
The Context should be the first parameter, typically named ctx:
```
https://golang.org/pkg/context/

pkg/health/checker/checker.go
```diff
@@ -41,28 +68,45 @@ type healthStore interface {
...
+	}
 }
 
-func NewConsulHealthChecker(client consulutil.ConsulClient) ConsulHealthChecker {
-	return consulHealthChecker{
-		client:      client,
-		kv:          client.KV(),
-		consulStore: consul.NewConsulStore(client),
+func NewHealthChecker(cClient consulutil.ConsulClient) HealthChecker {
+	return healthChecker{
```
mpuncel: it still makes me nervous that these two constructors return structs of the same type. If I instantiate a `healthChecker` by calling `NewHealthChecker()`, and then I pass `true` for `useHealthService`, i'm going to end up with a nil pointer exception at runtime. If you use different types to implement the functionality, the "normal" health checker could just ignore that argument

esuen: Thanks for the reminder, I was going to get around to that.

esuen:
- added new type for shadowHealthChecker.
- put ctx param as first param in function signatures
- Will be looking into making the grpc call for monitoring multiple endpoints better. Instead of for looping multiple monitor calls, I think it should be one call with all the endpoints

pkg/health/checker/checker.go
esuen: What if there's a 
```golang
defer close(resultCh)
```
in HealthMonitor?

esuen:
Will need a final review on this.

esuen:
TODO: Remove logging code in checker.go

- Created a separate healthChecker type in checker.go called ShadowTrafficHealthChecker that will utilize both the health service client and old consul tree checking code
- Old health checker was renamed from ConsulHealthChecker to HealthChecker so that in the future when the health service client is deemed reliable we can rename ShadowTrafficHealthChecker to HealthChecker
- Modified references in scripts p2-watch, p2-replicate, p2-rctl, p2-rctl-server, p2-inspect, p2-ds-farm to reference health checker signature or name changes as well as references to WatchService and Service calls
- References for ds/farm.go, ds/daemon_set.go, rc/farm.go, rc/replication_controller.go as well
- References for checker_test.go, fake_checker.go, store_test.go, replication/common_setup_test.go, pkg/roll/update_test.go as well
- References for pkg/health/store/store.go, pkg/replication/health_aggregator.go, pkg/replication/replicator.go, pkg/roll/farm.go as well
- Changes to pkg/manifest.go, add helper functions for checker to access the StatusStanza
- update pkg/roll/run_update.go to use ShadowTrafficHealthChecker

- Majority of changes in checker.go and health app
- New Health Checker type defined
- WatchService and Service functions defined for new HealthChecker to be used by rolling updates
- WatchService uses both health service client and consul health tree
- Make request to health service and wait for responses in a respChan and then pass back to the roll farm
- Service uses either consul or health service client. For health service client its a simple request response and return response

pkg/roll/run_update.go
```diff
@@ -330,9 +336,13 @@ func (u *update) Run(ctx context.Context) (ret bool) {
 	hQuit := make(chan struct{})
 	defer close(hQuit)
 	watchDelay := 1 * time.Second
-	go u.hcheck.WatchService(string(newFields.Manifest.ID()), hChecks, hErrs, hQuit, watchDelay)
+	config := newFields.Manifest.GetConfig()
+	useHealthService, _ := config["use_health_service"].(bool)
```
rudle: what if config["use_health_service"] isn't present? That's what the second return value is for.

esuen: useHealthService defaults to false if it's not present.

mpuncel: Are you sure we want this config to be in app pod manifests? it should probably be a setting on the roll farm right?

esuen: I thought we test it with specific apps first but I'm open to making it a global config. What does @rudle think?

rudle: Probably we need both, right? app specific so we can test applications like slug & p2-heartbeat and then roll farm wide so we can turn it on and off with the rest of the apps at Square.

mpuncel: ah okay, I'm just out of the loop

mpuncel: I would still check `ok` to make sure that the field is actually a bool, so you don't crash the server if someone has a typo

esuen: The ok is not used and no panic occurs even if useHealthService is not a bool. See https://tour.golang.org/methods/15, "If not, ok will be false and t will be the zero value of type T, and no panic occurs."

It's only when you do
```golang
useHealthService := config["use_health_service"].(bool)
```
for it to panic when the value is of type not bool.

mpuncel: The question is, what do you want to do if someone _intended_ to set it to true, but messed up up with a typo, e.g. `truee`. You're going to treat it as false. Perhaps that's fine, but it's worth thinking about

mpuncel: I do feel better that it at least won't crash the server, probably for a flag like this it's okay to not introduce errors here which will be difficult to recover from

esuen: I'd think I want only `true` to be treated as true and everything else be false otherwise there needs to be code to interpret what the user implied which is I don't think can be fully correct.

esuen:
-Added a field to ShadowTrafficHealthChecker struct useHealthService bool that will take precedence over the useHealthService passed to `Service` and `WatchService` of ShadowTrafficHealthChecker.
- Updated references to ShadowTrafficHealthChecker in `bin/p2-rctl/main.go` and `bin/p2-rctl-server/main.go`

pkg/roll/run_update.go
spenceral: You can check out https://golang.org/pkg/strconv/#ParseBool to error on non-booleans

esuen:
Close/reopen to rerun travis-ci.

esuen:
This can use a quick final review since last time, changes made for the suggestions from last review. See previous comments for details on the changes.

PR #1013, opened 2018-01-31
Drastically increase the artifact checker cycle period.

rudle:
We aren't _that_ interested in missing artifacts that we need alerts every hour


spenceral:
What made you settle on a day as opposed to say 1 week?

rudle:
I figure that we might deploy p2-master on a less-than-weekly cadence and that would cause this alert to never fire.

PR #1015, opened 2018-02-09
Mpuncel/fix opencontainer

mpuncel:


esuen:
Looks like a simple refactor change. I don't see obvious errors or logical errors and I am not seasoned enough yet to give Golang style suggestions. Otherwise LGTM.

PR #1016, opened 2018-02-12
gate the pod arg to p2-exec on the param

rudle:
This doesn't have a behavioural effect (that depends on the resource file which
depends on the top level manifest key resource_limits:) but it does allow us to
roll out a version of this code that understands the argument without calling
it, providing us with n-1 version backward compat

### DIFF
--- a/pkg/hoist/hoist_launchable.go
+++ b/pkg/hoist/hoist_launchable.go
@@ -20,6 +20,7 @@ import (
 	"github.com/square/p2/pkg/types"
 	"github.com/square/p2/pkg/user"
 	"github.com/square/p2/pkg/util"
+	"github.com/square/p2/pkg/util/param"
 )
 
 type EntryPoints struct {
@@ -32,6 +33,10 @@ type EntryPoints struct {
 	// if and only if it was listed explicitly
 	Implicit bool
 }
+
+// IncludePodIDArg will include a --podID argument to p2-exec.
+// Delaying the use of this argument from the rollout of the binary that understands it allows for n-1 version compatibility
+var IncludePodIDArg = param.Bool("include_pod_id_arg", false)
 
 // A HoistLaunchable represents a particular install of a hoist artifact.
 type Launchable struct {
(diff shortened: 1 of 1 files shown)


pkg/hoist/hoist_launchable.go
```diff
@@ -33,6 +34,10 @@ type EntryPoints struct {
 	Implicit bool
 }
 
+// IncludePodIDArg will include a --podID argument to p2-exec.
+// Delaying the use of this argument from the rollout of the binary that understands it allows for n-1 version compatibility
+var IncludePodIDArg = param.Bool("include_pod_id_arg", false)
```
rudle: I'd like to re-use the same config value (nested_cgroups), but I cannot import the pods package here.

This isn't a huge deal as this param should be short lived.

pkg/hoist/hoist_launchable.go
```diff
@@ -374,10 +379,22 @@ func (hl *Launchable) Executables(
...
+					User:             hl.RunAs,
+					EnvDirs:          []string{hl.PodEnvDir, hl.EnvDir()},
+					ExtraEnv:         map[string]string{launch.EntryPointEnvVar: relativePath},
+					NoLimits:         hl.ExecNoLimit,
+					CgroupConfigName: hl.CgroupConfigName,
+					PodID:            &hl.PodID,
+					CgroupName:       hl.CgroupName,
+					RequireFile:      hl.RequireFile,
+				}
+			}
```
esuen: Could you modify p2ExecArgs PodID member with something like ```p2ExecArgs.PodID = &hl.PodID``` over declaring everything again? Otherwise change looks good.

pkg/hoist/hoist_launchable.go
```diff
@@ -374,10 +379,22 @@ func (hl *Launchable) Executables(
 				ExtraEnv:         map[string]string{launch.EntryPointEnvVar: relativePath},
 				NoLimits:         hl.ExecNoLimit,
 				CgroupConfigName: hl.CgroupConfigName,
-				PodID:            &hl.PodID,
 				CgroupName:       hl.CgroupName,
 				RequireFile:      hl.RequireFile,
 			}
+			if *IncludePodIDArg {
+				p2ExecArgs = p2exec.P2ExecArgs{
```
spenceral: I believe you can dry this out a bit with
```
if *IncludePodIDArg {
    p2ExecArgs.PodID = &hl.PodID
}```
and the rest of the fields will carry over from the preceding initialization.

rudle:
Thanks for the review. PTAL

PR #1017, opened 2018-02-13
Mpuncel/allow pods to runit signal

mpuncel:


rudle:
This seems OK to me, but I wonder if it allows those who have access to the app user the ability to run arbitrary code that they couldn't run before. It seems like this change should be combined with @stouset's change to extract the artifact as root owned.

mpuncel:
closing as stale, and not sure if we want it anymore

PR #1018, opened 2018-02-13
Add no_halt_on_update launchable option to launchable stanza spec

mpuncel:
When set, this option instructs P2 to skip the "sv stop" portion of the
normal manifest update process. This is useful for processes that are
designed to gracefully reload without service interruption. However, P2
does not arrange for a signal to be sent to the process, pods that need
this functionality should implement their own signal handling in
bin/post-activate.

There are certain circumstances in which the no_halt_on_update setting
will be ignored, for instance if an operator is shutting down all pods
on the host or if a pod is being uninstalled (manifest deleted from the
intent tree).

### DIFF
--- a/pkg/launch/launchable.go
+++ b/pkg/launch/launchable.go
@@ -49,6 +49,16 @@ type LaunchableStanza struct {
 	// "always".
 	RestartPolicy_ runit.RestartPolicy `yaml:"restart_policy,omitempty"`
 
+	// NoHaltOnUpdate instructs the preparer to skip stopping the
+	// launchable's processes when it is being updated. This is useful for
+	// processes that are designed to be updated via binary overwrite and
+	// SIGHUP for example.
+	// NOTE: under certain circumstances the process still might be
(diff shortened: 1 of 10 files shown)


pkg/launch/launchable.go
```diff
@@ -49,6 +49,16 @@ type LaunchableStanza struct {
...
 
+	// NoHaltOnUpdate instructs the preparer to skip stopping the
+	// launchable's processes when it is being updated. This is useful for
+	// processes that are designed to be updated via binary overwrite and
+	// SIGHUP for example.
+	// NOTE: under certain circumstances the process still might be
+	// stopped, for example if a host operator calls p2-shutdown
+	// NOTE: P2 does not arrange for signals to your app to signify update,
+	// that should be implemented by the pod itself via bin/post-activate
+	NoHaltOnUpdate bool `yaml:"no_halt_on_update,omitempty"`
```
mpuncel: i'm happy to discuss the naming here. Notably this setting shouldn't be ported to container launchables because containers should be immutable

PR #1019, opened 2018-02-15
This alert is not yet helpful, there are many missing artifacts that …

rudle:
…we cannot restore

This alert is causing oncall fatigue. It will be helpful someday, but until then
we will replace it with a log message


rudle:
Please review

PR #1020, opened 2018-02-16
Added metrics for ru convergence time

akun1012:


esuen:
This is open src and public, so you probably don't want to reference the internal ticket.

pkg/roll/fields/update.go
```diff
@@ -27,8 +27,8 @@ type Update struct {
 	//
 	// The acceptable value of Minimum is subject to several constraints:
 	//   - the update will not even start until the minimum is satisfied, so if
-	//     the current number of live nodes is than the minimum, it will block
-	//     forever.
+	//     the current number of live nodes is smaller than the minimum, it will
```
esuen: nit: "less" instead of "smaller"

akun1012:
@esuen Thanks.

This PR could is with build failure now. Will modify, and notify when it's ready to review.

akun1012:
Build passed. Please review.

rudle:
Looks good! Please squash these commits into one before merging. One note: imports should always be added with the code that references them, please configure your editor to organize imports on file save. You might need to install `goimports`.

PR #1021, opened 2018-02-20
add new metric for node transfer convergence time

akun1012:


akun1012:
Check passed, please review.

PR #1022, opened 2018-02-22
Mpuncel/opencontainer fixes

mpuncel:


mpuncel:
At this point I'm not sure if the travis build will succeed, so there will likely be some churn unless it magically works on the first try.

integration/travis.sh
```diff
@@ -69,8 +69,4 @@ popd
 sudo curl -L https://github.com/opencontainers/runc/releases/download/v1.0.0-rc4/runc.amd64 -o /usr/local/bin/runc
 sudo chmod +x /usr/local/bin/runc
 
-# this is a lie, but the opencontainer code in P2 depends on this file existing
-# to configure runc, so just fudge it for the integration test
-sudo echo 'CentOS Linux release 7.4.1708 (Core)' > /etc/redhat-release
```
rudle: so this was always failing?

mpuncel: hmm no added that in this PR, probably did bad squashing

PR #1023, opened 2018-02-22
Read only in pod manifest

stouset:
To be honest, I'm a little out of my depth here.

That said, I'm attempting to have p2-preparer be configurable to set default readonly policy for deployed pods. Pods will be able to override this policy by setting the `ReadOnly` flag in their manifest (as they currently do today). To allow for a graceful changeover, the preparer manifest will supported a flag for whether or not readonly deploys should be the default, a list of PodIDs that should be deployed readonly by default, and a list of PodIDs that should not be deployed readonly by default.

We'll start out by having ReadOnly default to false, and manually deploying the preparer with a whitelist of readonly-clean pods. This list will grow over time. In the event of an error on our part, pod owners can explicitly deploy with ReadOnly set to false to override this, so they don't have to wait on a preparer deploy.

As we gradually roll out more apps ReadOnly, we'll build up a list of apps that *cannot* currently be deployed this way and add them to the blacklist. For now this won't do anything, as the default policy is false anyway. However, once we have an authoritative list of non-readonly apps, we can flip the default policy from false to true. Once this is done, we can remove apps from the whitelist until it's empty, then delete most of the code in this PR (and have the known-incompatible apps simply maintain that fact in their manifest).

We export a `HOOKED_POD_READ_ONLY` flag so that hooks can act upon this information even if the on-disk manifest doesn't have the information necessary to know if the pod is readonly or not (since it now depends upon the state of the pod manifest *and* the preparer manifest).

### DIFF
--- a/pkg/hooks/hook_env.go
+++ b/pkg/hooks/hook_env.go
@@ -3,6 +3,7 @@ package hooks
 import (
 	"fmt"
 	"os"
+	"strconv"
 
 	"github.com/square/p2/pkg/config"
 	"github.com/square/p2/pkg/manifest"
@@ -79,7 +80,14 @@ func (h *HookEnv) PodFromDisk() (*pods.Pod, error) {
 
 // Initializes a pod based on the hooked pod manifest and the system pod root
 func (h *HookEnv) Pod() (*pods.Pod, error) {
-	factory := pods.NewFactory(os.Getenv(HookedSystemPodRootEnvVar), HookedNodeEnvVar, uri.DefaultFetcher, "", pods.NewReadOnlyPolicy(false, nil, nil))
+	readonly, err := strconv.ParseBool(os.Getenv(HookedPodReadOnly))
+	if err != nil {
+		// if we can't parse the boolean, it may just be unset; default
+		// to readonly being off
+		readonly = false
+	}
+
+	factory := pods.NewFactory(os.Getenv(HookedSystemPodRootEnvVar), HookedNodeEnvVar, uri.DefaultFetcher, "", pods.NewReadOnlyPolicy(readonly, nil, nil))
 
 	podID, err := h.PodID()
 	if err != nil {
--- a/pkg/hooks/hook_env.go
+++ b/pkg/hooks/hook_env.go
@@ -79,7 +79,7 @@ func (h *HookEnv) PodFromDisk() (*pods.Pod, error) {
 
 // Initializes a pod based on the hooked pod manifest and the system pod root
 func (h *HookEnv) Pod() (*pods.Pod, error) {
-	factory := pods.NewFactory(os.Getenv(HookedSystemPodRootEnvVar), HookedNodeEnvVar, uri.DefaultFetcher, "")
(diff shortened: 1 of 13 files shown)


pkg/hooks/hook_env.go
```diff
@@ -79,7 +79,7 @@ func (h *HookEnv) PodFromDisk() (*pods.Pod, error) {
 
 // Initializes a pod based on the hooked pod manifest and the system pod root
 func (h *HookEnv) Pod() (*pods.Pod, error) {
```
mpuncel: this function gets imported and called by hooks, it's just convenience around constructing a `Pod` based on environment variables available to the hook. You probably want to pass the result of `HOOKED_POD_READ_ONLY` instead of `false` as the default to the read only policy? it's somewhat hacky since we're not going to actually configure a blacklist and a whitelist but it will only run in the context of the hooked pod so it should be okay

pkg/manifest/manifest.go
```diff
@@ -256,23 +257,33 @@ func (manifest *manifest) RunAsUser() string {
 }
 
 func (manifest *manifest) UnpackAsUser() string {
-	if manifest.ReadOnly {
```
mpuncel: you'll have to check that `manifest.ReadOnly` isn't nil before dereferencing it otherwise it'll panic

mpuncel: you probably just want `manifest.GetReadOnly()` here?

pkg/manifest/manifest.go
```diff
@@ -256,23 +257,33 @@ func (manifest *manifest) RunAsUser() string {
...
+	if manifest.ReadOnly != nil {
+		return *manifest.ReadOnly
+	}
+
+	return false
+}
+
+func (manifest *manifest) SetReadOnlyIfUnset(readonly bool) {
+	if manifest.ReadOnly == nil {
+		*manifest.ReadOnly = readonly
```
mpuncel: if it's not too much difficulty, we try to treat the `manifest` as an immutable representation of what is actually in the pod manifest, I'd rather see wherever the unpacking code is decide what user to use based on pod setting and manifest setting rather than pushing that logic into the manifest

pkg/hooks/hook_env.go
stouset: Fixed

pkg/manifest/manifest.go
stouset: Fixed

rudle:
I can't get this code to compile. Can you make sure that all of the test files build? 
`go test github.com/square/p2/pkg/...`

rudle:
@stouset Ping! Please fix merge conflicts and compilation errors.

stouset:
@rudle It's been rebased. Tests are failing, but they don't seem to be relevant to my changes. Can you take a look and advise?

rudle:
@stouset  I can reproduce the failure @07f7c067 
Try `go test github.com/square/p2/pkg/preparer/...` and let me know what you see.

PR #1024, opened 2018-02-23
Mpuncel/opencontainer post install

mpuncel:


mpuncel:
Tests pass now, I needed to add the `-e` flags to the p2-exec call so the cgroup could be set correctly.

pkg/opencontainer/containers.go
```diff
@@ -158,9 +163,12 @@ func (l *Launchable) Executables(serviceBuilder *runit.ServiceBuilder) ([]launch
 		Exec: append(
 			[]string{l.P2Exec},
 			p2exec.P2ExecArgs{ // TODO: support environment variables
```
rudle: they are now supported, right?

mpuncel: hmm, i'll clarify this because the answer is "not really" because `runc` does not propagate environment variables, you have to specify your full environment in config.json which might be something the preparer should do for you someday

PR #1025, opened 2018-02-28
Relocate the launchable PostInstall() invocation after env setup

mpuncel:
This will allow a launchable's PostInstall() script to depend on
environment being written out to disk.

This change is good conceptually, but given that a single function
writes the pod-wide env dir and all of the launchable env dirs, it came
out a little ugly code-wise.

The code in this commit optimized minimal disturbance of hoist
launchable code (which doesn't make use of PostInstall()) at the expense
of cleanliness.


mpuncel:
This is ugly, but I think it's the best way to make this work for opencontainer launchables without overhauling critical preparer code

PR #1028, opened 2018-03-08
Health service into rolling update fixes

esuen:
Fix up logic in WatchService of checker.go
Add refresh logic in run_update.go for the case of node transfer or other reason causing the current pods in an RU to change.


esuen:
There is some working code with TODO label that will be removed after review.
TODO stuff will be removed after this approach is deemed acceptable.

rudle:
I think this looks OK. I can't get the tests to pass on my machine, can you take a look?

esuen:
Added support for no status port in manifest, treat app as healthy.

rudle:
ok, still looks good

PR #1030, opened 2018-03-13
Mpuncel/container processing

mpuncel:
This PR is a step toward removing the bin/pre-launch hack from the opencontainer launchable type. The main other tasks to finally remove it are:

1) put an init system in the containers (or require that containers do it themselves). This is necessary because currently it's not safe to fork processes inside the container due to the "pid 1 problem"
2) Make p2-exec rewrite the environment fields in config.json to contain the contents of the env dirs
3) add a bind mount for CONFIG_PATH

This would allow in-container processes to customize their behavior based on parameters they care about environment and config.


pkg/preparer/setup.go
```diff
@@ -525,6 +538,17 @@ func New(preparerConfig *PreparerConfig, logger logging.Logger) (*Preparer, erro
 
 	podFactory := pods.NewFactory(preparerConfig.PodRoot, preparerConfig.NodeName, fetcher, preparerConfig.RequireFile)
 	podFactory.SetOSVersionDetector(osVersionDetector)
+
+	if preparerConfig.OpenContainerConfig.ExpectAtLeastOneBindMountEnvVar {
```
mpuncel: Sadly this is necessary because I believe if the variable is renamed and the build flag isn't, the build will happily succeed but the variable won't be set! so we can set this bool in our config we expect there to be any "bind mount environment variables" and at least see a runtime error if the build flag didn't work right

PR #1031, opened 2018-03-19
Allow resource_limits of 0 in pod manifests

spenceral:
Right now, a cgroup limit of 0 for either CPUs or memory in the manifest
(or the omission of the cgroup limit) translates to an actual cgroup
limit of -1, i.e. unrestricted. To facilitate an automated adoption of
reosurce_limits based on existing launchable limits, we should allow
manifest limits of 0 (at least for now). Otherwise we will have to
impose new limits that did not previously exist and that will slow down
the migration to pod resource_limits.


spenceral:
https://github.com/square/p2/blob/master/pkg/cgroups/cgroups.go#L55 is where a cpu manifest limit of 0 is translated to an unrestricted cgroup limit.

PR #1032, opened 2018-03-22
Add naive retries with backoff to getMatches

spenceral:
Lately we've seen network errors cause this function to fail with some
frequency. This changes adds naive retries with a simple linear backoff
to reduce the return of errors due to fleeting network failures. At the
same time, it tries to not dramatically increase latency (where latency
is defined as the time it takes getMatches to return, successfully or
not).

### DIFF
--- a/pkg/labels/http_applicator.go
+++ b/pkg/labels/http_applicator.go
@@ -260,20 +260,18 @@ func (h *httpApplicator) getMatches(selector labels.Selector, labelType Type, ca
 	}
 	req.Header.Add("Accept", "application/json")
 
-	resp, err := h.client.Do(req)
-	if err != nil {
-		return nil, err
+	var resp *http.Response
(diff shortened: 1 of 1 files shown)


pkg/labels/http_applicator.go
```diff
@@ -313,6 +314,22 @@ func (h *httpApplicator) getMatches(selector labels.Selector, labelType Type, ca
 	return labeled, nil
 }
 
+// The caller is responsible for closing the resp.Body
```
mpuncel: you may as well `defer resp.Body.Close()` in here because the caller doesn't even use the body (except to close it)

spenceral: Yeah you're right. Also you made me notice that the `resp.Body.Close()` call on 270 would panic because `resp` is nil in that case.

PR #1034, opened 2018-03-27
changed buffered channel

akun1012:
This commit is a complementary for this [commit](https://github.com/square/p2/commit/ca7a05bc4b683a252ee84ecf147e61a316f30639) 

Ensuring that the preparer is always using the latest pod manifest


pkg/preparer/orchestrate.go
```diff
@@ -89,11 +89,11 @@ func (p *Preparer) WatchForPodManifestsForNode(quitAndAck chan struct{}) {
 	// buffer this with 1 manifest so that we can make sure that the
 	// consumer is always reading the latest manifest. Before writing to
 	// this channel, we will attempt to drain it first
-	podChan := make(chan []consul.ManifestResult, 1)
+	podChan := make(chan []consul.ManifestResult)
 
 	go p.store.WatchPods(consul.INTENT_TREE, p.node, quitChan, errChan, podChan)
 
-	podChanMap := make(map[podWorkerID]chan ManifestPair)
+	podChanMap := make(map[podWorkerID]chan ManifestPair, 1)
```
mpuncel: i think you actually want to buffer the `chan ManifestPair` on line 123

pkg/preparer/orchestrate.go
```diff
@@ -89,11 +89,11 @@ func (p *Preparer) WatchForPodManifestsForNode(quitAndAck chan struct{}) {
 	// buffer this with 1 manifest so that we can make sure that the
 	// consumer is always reading the latest manifest. Before writing to
 	// this channel, we will attempt to drain it first
```
mpuncel: you should move this comment to the place the actual buffering is

akun1012:
Disabled race testing for orchestrate test.

pkg/preparer/orchestrate_test.go
```diff
@@ -1,3 +1,5 @@
+// +build !race
```
mpuncel: what race did you hit? looks like this file doesn't depend on consulutil which has known races, so if the race detector showed a problem it's probably a real problem

PR #1036, opened 2018-03-28
Missing a status.Port == 0 check when refresh happens on nodeIDs

esuen:


pkg/health/checker/checker.go
```diff
@@ -443,12 +443,16 @@ func (h shadowTrafficHealthChecker) WatchService(
...
 				case nodeIDs = <-nodeIDsCh:
-					for _, nodeID := range nodeIDs {
-						healthResults[nodeID] = health.Result{
-							ID:      types.PodID(serviceID),
-							Node:    nodeID,
-							Service: serviceID,
-							Status:  health.Passing,
+					// app is considered always healthy when
+					// status port is not set
+					if status.Port == 0 {
```
rudle: I wonder if this code would be better factored if the `status.Port` check lived in the loop that iterates over nodeIDs instead of in this select loop.

rudle:
Please include a regression test for this case

pkg/health/checker/checker.go
esuen: The loop over nodeIDs for setting healthResults only needs to be done if the app is always healthy, in the the default case, nodeIDs still get set to the new values. The monitoring connection is reset every 10 minutes and will be using the latest values of nodeIDs when making a new connection.

rudle: I was expressing a desire to have this knowledge (status.Port == 0 -> always healthy) be encapsulated at a lower layer so it doesn't have to appear twice in the health checker. For example, if the healthClient captured this knowledge and the health checker was naive about it, we wouldn't have had yesterday's bug.

This is a major refactor that I can't reasonably suggest doing without a test suite, so don't worry about it for this PR.

pkg/health/checker/checker_test.go
```diff
@@ -24,6 +26,157 @@ func (f fakeConsulStore) GetServiceHealth(service string) (map[string]consul.Wat
...
+}
+
+func (f fakeHealthClient) HealthMonitor(ctx context.Context, req *hc.HealthRequest, resultCh chan *hc.HealthResponse) error {
+	timer := time.NewTimer(0)
+	for {
+		select {
+		case <-ctx.Done():
+			return nil
+		case <-timer.C:
+			timer.Reset(f.MonitorDelay)
```
rudle: changing this to a smaller time (like 200 millis) makes the test run faster and it still passes. Is that a safe change to make?

pkg/health/checker/checker_test.go
```diff
@@ -24,6 +26,157 @@ func (f fakeConsulStore) GetServiceHealth(service string) (map[string]consul.Wat
...
+		healthResult, ok := result[nodeID]
+		if !ok {
+			t.Fatalf("Expected nodeID %s to be in results but not found", nodeID)
+		}
+		if expected[nodeID] != healthResult.Status {
+			t.Fatalf("Expected hCheck status from WatchService to be %s but got %s instead", expected[nodeID], healthResult.Status)
+		}
+	}
+
+	cancel()
```
rudle: Better to use defer upon where the ctx is created

pkg/health/checker/checker_test.go
```diff
@@ -24,6 +26,157 @@ func (f fakeConsulStore) GetServiceHealth(service string) (map[string]consul.Wat
...
+
+	result := make(map[types.NodeName]health.Result)
+
+LOOP1:
+	for {
+		select {
+		case <-ctx.Done():
+			return
+		case result = <-resultCh:
+			if len(result) == 3 {
```
rudle: Is this ever not 3? Should we fail the test if not?

esuen: - 3 should be len(nodeIDs)
- result will be < 3 when it receives the first 2 messages
- there will be 3 messages
- a message is sent when WatchService receives a message from health service
- an initial message is sent for each monitor

pkg/health/checker/checker_test.go
esuen: - It shouldn't make the test run faster because initial messages are sent right away, the initial timer is set to 0
- this always passes when there isn't a regression
- when there is a regression and the oldResults are received is when this test will fail

rudle: Fair enough. The test sometimes takes 5s to pass on my machine, other times it takes 0.3s. It seems independent of my random changes :)

rudle: Any idea why this test sometimes takes 5s to run?

esuen: It's consistently 5s for me. Probably just the timing of when the oldResults are sent.

esuen:
sorry, I hijacked this small fix PR into an ongoing improvements and continued efforts to migrating to the new health service PR.

esuen:
Separated into two PRs
- https://github.com/square/p2/pull/1037
- https://github.com/square/p2/pull/1038

PR #1038, opened 2018-03-30
Bug fix: pod considered healthy after node refresh 

esuen:
- fix for health service bug that will treat all pods as healthy after nodes were refreshed
- regression test and test coverage for checker code

### DIFF
--- a/pkg/health/checker/checker.go
+++ b/pkg/health/checker/checker.go
@@ -384,6 +384,14 @@ func statusURLToNodeName(s string) (types.NodeName, error) {
 	return types.NodeName(host), nil
 }
 
+func healthResultsCopy(healthResults map[types.NodeName]health.Result) map[types.NodeName]health.Result {
+	out := make(map[types.NodeName]health.Result)
+	for nodeID, healthResult := range healthResults {
+		out[nodeID] = healthResult
+	}
+	return out
+}
+
 func (h shadowTrafficHealthChecker) WatchService(
 	ctx context.Context,
 	serviceID string,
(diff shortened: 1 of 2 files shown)


esuen:
don't review yet, looking into something.

pkg/health/checker/checker.go
```diff
@@ -396,16 +404,22 @@ func (h shadowTrafficHealthChecker) WatchService(
 	useOnlyHealthService bool,
 	status manifest.StatusStanza,
 ) {
+	alwaysHealthy := false
+	// nodes are considered always healthy when status port is not set
+	// when always healthy, WatchService ONLY sends the healthy status repeatedly on a timer set to delay
+	if status.Port == 0 {
+		alwaysHealthy = true
```
spenceral: nit: you could write `alwaysHealthy := status.Port == 0`

pkg/health/checker/checker.go
```diff
@@ -437,27 +451,33 @@ func (h shadowTrafficHealthChecker) WatchService(
...
+					// it doesn't matter for correctness, but the go test --race will complain
+					resultsCopy := healthResultsCopy(healthResults)
 					select {
 					case <-ctx.Done():
 						return
-					case resultCh <- healthResults:
+					case resultCh <- resultsCopy:
 					}
 				case nodeIDs = <-nodeIDsCh:
-					for _, nodeID := range nodeIDs {
```
spenceral: Was this the source of the bug? Sending `health.Passing` whenever a node id was received in `nodeIDsCh`?

PR #1039, opened 2018-04-03
Remaining health service changes needed for full migration

esuen:


esuen:
WIP, feel free to browse

pkg/health/checker/checker.go
```diff
@@ -256,18 +393,169 @@ func publishLatestHealth(inCh <-chan api.KVPairs, quitCh <-chan struct{}, result
...
+func (h shadowTrafficHealthChecker) WatchService(
+	ctx context.Context,
+	serviceID string,
+	resultCh chan<- map[types.NodeName]health.Result,
+	errCh chan<- error,
+	watchDelay time.Duration,
+	useHealthService bool,
+	useOnlyHealthService bool,
+	status manifest.StatusStanza,
+	checkRCNodes bool,
```
rudle: What is this for? Is the client expected to know whether they are checking an RC or a DS? Why not make them separate functions that share a helper?

esuen: I think only the client would know whether to check RC, DS or both nodes when calling WatchService or Service. This could be a parameter when initializing the checker.

pkg/health/checker/checker.go
```diff
@@ -22,11 +23,16 @@ import (
 	klabels "k8s.io/kubernetes/pkg/labels"
 )
 
+// originally from pkg/ds but import cycle not allowed
+const (
+	DSIDLabel = "daemon_set_id"
```
rudle: This is a sign that this package is poorly coupled to the daemon set package.

Could we integrate the node fetching in a way that's less tightly coupled? What if serviceNodesDS and serviceNodesRC was a function of type `function() ([]types.NodeName, error)` passed in by the client?

pkg/health/checker/checker.go
```diff
@@ -45,7 +51,35 @@ type HealthChecker interface {
...
+	WatchPodOnNode(
+		ctx context.Context,
+		nodeID types.NodeName,
+		podID types.PodID,
+		useHealthService bool,
+		useOnlyHealthService bool,
+		status manifest.StatusStanza,
+	) (chan health.Result, chan error)
 	WatchService(
+		ctx context.Context,
```
rudle: I thought the idea of the ShadowTrafficHealthChecker was that it would implement the same interface as the HealthChecker. How is the migration off of this ShadowTrafficHealthChecker suppose to proceed?

If you need a different type to support health checks for daemon sets, it's ok to make a new type. I'd like this interface to maintain similarity with the HealthChecker interface.

pkg/health/checker/checker.go
```diff
@@ -142,54 +177,156 @@ func NewHealthChecker(cClient consulutil.ConsulClient) HealthChecker {
...
+	podID types.PodID,
+	useHealthService bool,
+	useOnlyHealthService bool,
+	status manifest.StatusStanza,
+) (chan health.Result, chan error) {
+	resultCh := make(chan health.Result)
+	errCh := make(chan error)
+	if !useOnlyHealthService && !h.useOnlyHealthService {
+		resultCh, errCh = consulWatchPodOnNode(ctx, nodeID, podID, h.kv)
+	}
```
rudle: Am I understanding correctly that this function will not send any traffic to the health service unless it is used exclusively? This is not the desired outcome. Instead, the shadow health checker should send maintain health requests at the health service so we can *observe how the health service scales*.

pkg/ds/farm.go
```diff
@@ -565,13 +565,17 @@ func (dsf *Farm) spawnDaemonSet(
 	ctx, cancel := context.WithCancel(ctx)
 
 	desiresCh := ds.WatchDesires(ctx, updatedCh, deletedCh)
-
+	config := dsFields.Manifest.GetConfig()
+	// defaults to false if not set
+	useHealthService, _ := config["use_health_service"].(bool)
```
mpuncel: this will crash the daemon set farm if it's a string for example right?

esuen: Right, I could put it in a switch that checks for type and only set if the type is bool, otherwise default to false like this https://play.golang.org/p/KCvc9DgKfHI

This doesn't panic if the key in the config map doesn't exist either, the value will be nil when trying to access a key that doesn't exist.

pkg/health/checker/checker.go
esuen: That function will send traffic to the health service if either useHealthService or useOnlyHealthService is true. Health checks to consul will be enabled unless useOnlyHealthService is true.

pkg/health/checker/checker.go
esuen: So ShadowTrafficHealthChecker should check all nodes for WatchService and Service and the checker type for daemon sets could only check DS nodes for WatchService and Service.

esuen: For the migration I was thinking that everything will eventually be using the shadowChecker. And when we want to finalize the migration, the consul code and health service flags will be removed and the shadowChecker will be renamed.

pkg/health/checker/checker.go
esuen: I suppose serviceNodes can be a function passed to the constructor of the healthChecker.

esuen:
Addressed comments, ready for review again.

pkg/health/checker/checker.go
esuen: Only one? I see, so that if we are relying on consul data then we shouldn't accept data from the health service. We should only be testing the load against the health service and throwing out the data we receive.

PR #1043, opened 2018-04-18
Refactor of health checker code, cleaner interchangeable checker types

esuen:
- Create one simple interchangeable interface HealthChecker to be used by two types "consulHealthChecker" and "healthChecker"
- consulHealthChecker uses existing code to check a consul tree for health checks
- healthChecker uses a health service for health checks
- remove "WatchHealth" function from checker code and pkg/health/store as it is not used
- new healthChecker type takes `serviceNodes func(serviceID string) ([]types.NodeName, error)` for WatchService and Service
- roll/farm.go takes `healthClientFactory` interface type that implements `NewHealthClient` so that each update will have its own instance of a health service client


pkg/health/checker/checker.go
```diff
@@ -361,224 +180,347 @@ func watchConsulHealth(
...
+					Node:    nodeID,
+					Service: podID.String(),
+					Status:  resp.Health,
+				}
+				select {
+				case <-ctx.Done():
+					return
+				case resultCh <- healthResult:
+				}
+			case <-tick:
```
rudle: What's the intention of this case? Is it to provide a best effort view of health in case the health service is slow? Why is that important?

pkg/health/checker/checker.go
```diff
@@ -361,224 +180,347 @@ func watchConsulHealth(
...
+	}
+	respCh := make(chan *hclient.HealthResponse, 1)
+	healthResult := health.Result{
+		ID:      podID,
+		Node:    nodeID,
+		Service: podID.String(),
+		Status:  health.Unknown,
+	}
+	tick := time.Tick(time.Second * 2)
+	go func() {
```
rudle: This loop doesn't seem to be covered by any tests. Can you add some to ensure correctness?

pkg/health/checker/checker.go
```diff
@@ -361,224 +180,347 @@ func watchConsulHealth(
...
-				// when the nodes are always healthy, these
-				// are the only messages sent
-				oldResultsDelay = time.Second * 2
-				for _, nodeID := range nodeIDs {
-					healthResults[nodeID] = health.Result{
-						ID:      types.PodID(serviceID),
-						Node:    nodeID,
-						Service: serviceID,
-						Status:  health.Passing,
+) (chan health.Result, chan error) {
```
rudle: Does this code need a special case for always healthy?

pkg/health/checker/checker.go
```diff
@@ -361,224 +180,347 @@ func watchConsulHealth(
...
+	return result, nil
 }
 
-func (h shadowTrafficHealthChecker) WatchService(
+type HealthClient interface {
+	HealthCheckEndpoints(ctx context.Context, req *hclient.HealthEndpointsRequest) (map[string]health.HealthState, error)
+	HealthMonitor(ctx context.Context, req *hclient.HealthRequest, respCh chan *hclient.HealthResponse) error
+}
+
+type healthChecker struct {
```
rudle: These would be easier to find if they were at the top of the file with the other types

esuen:
Added healthCheckerFactory logic to farms

rudle:
Because you force pushed, I can't easily see if any tests were added in your latest push. Can you let me know when the new checker code is covered by tests so I can review again?

pkg/health/checker/checker.go
esuen: Yes, probably.

pkg/health/checker/checker.go
esuen: To mimic the behavior of the old consul checker and also I think it's possible that messages are lost over the network. Health service only sends updates when there's a change.

PR #1045, opened 2018-04-30
Mpuncel/add docker support

mpuncel:


mpuncel:
there are still some rough edges here, the code might not be in a mergeable state. In particular, I think p2-launch won't work for docker pods because it doesn't initialized a docker client.

Also, installation of docker containers is awkward because it didn't fit nicely into the existing launchable interface

mpuncel:
also it doesn't arrange for writable mounts for things like logging, so if the container does any writes it will crash

pkg/pods/pod.go
```diff
@@ -466,17 +472,30 @@ func (pod *Pod) Install(manifest manifest.Manifest, verifier auth.ArtifactVerifi
...
+				return err
+			}
+		} else if launchable.Type() == "docker" {
+			// TODO: auth?
+			resp, err := pod.DockerClient.ImagePull(context.TODO(), stanza.Image.Name, dockertypes.ImagePullOptions{})
+			if err != nil {
+				pod.logLaunchableError(launchable.ServiceID(), err, fmt.Sprintf("could not pull docker image: %s", err))
+				return util.Errorf("could not pull docker image: %s", err)
+			}
+			// TODO: anything we want in this response body?
```
bwester: The Docker client returns an HTTP response body and expects the caller to parse it? What kind of an API is that?!

That (lack of) design has implications here: `err` checks for network errors and `resp.StatusCode` checks whether the operation succeeded or failed. From the REST [API reference](https://docs.docker.com/engine/api/v1.30/#operation/ImageCreate), if the call fails, the response body contains the error message. It looks like no response body is returned on success.

pkg/launch/launchable.go
```diff
@@ -59,10 +59,13 @@ type LaunchableStanza struct {
...
-	// should be launched under runit. Only launchables of type "hoist"
-	// make use of this field, and if empty, a default of ["bin/launch"]
-	// is used
+	// In the case of hoist launchables, specifies which files or
+	// directories (relative to launchable root) should be launched under
+	// runit. if empty, a default of ["bin/launch"] is used.
+	//
+	// In the case of docker containers, it specifies which commands
+	// should be started in the docker container. If none are specified, the
+	// docker image's ENTRYPOINT is used.
```
bwester: Perhaps this is best flushed out in a design doc, but what do we do about pods following the Hoist convention of having multiple launch scripts? Listing entry points here is one option, but that requires synchronizing this list to what's actually deployed (as opposed to "run everything in this directory"). Another option (perhaps not mutually exclusive) might be to have a single Docker entry point that runs a supervisor within the container (e.g., runit).

Thoughts?

pkg/docker/docker.go
```diff
@@ -0,0 +1,222 @@
...
+	CgroupMemorySize size.ByteCount
+
+	// RestartPolicy_ determines what behavior docker should have when the pod exits.
+	// Typically it will indicate that docker should always restart the container
+	// for long-running services (e.g. servers)
+	RestartPolicy_ runit.RestartPolicy
+}
+
+func (l *Launchable) Disable() error {
+	// TODO: implement
```
bwester: Even if they aren't implemented right now, how do you envision disable commands working?

Personally, I say deprecate it. SIGTERM is already a nice way to signal that a process should cleanly exit, and if the container really needs advance notice (e.g., to migrate from Hoist) potentially the entry point could be a Hoist compatibility wrapper that calls a disable script then terminates the launchable.

There are other ways to approach this though. Thoughts?

pkg/launch/launchable.go
```diff
@@ -73,6 +76,15 @@ type LaunchableStanza struct {
 	// can be used to query a configured artifact registry which will provide the artifact
 	// URL. Version may not be used in conjunction with Location
 	Version LaunchableVersion `yaml:"version,omitempty"`
+
+	// Image: the name of the container image to run. This only applies when the launchable
+	// type is "docker"
+	Image DockerImage `yaml:"image,omitempty"`
```
bwester: Do you think "version" would be suitable for specifying an image? There are a lot of other systems that look at a launchable's "version" field for informative or display purposes. Those too will  have to be changed to query an "image" field.

pkg/docker/docker.go
```diff
@@ -0,0 +1,222 @@
...
+	switch l.RestartPolicy_ {
+	case runit.RestartPolicyAlways:
+		// keep it the way it is
+	case runit.RestartPolicyNever:
+		// keep it the way it is
+		restartPolicy = "no"
+	default:
+		return util.Errorf("invalid restart policy: %s", l.RestartPolicy_)
+	}
+	hostConfig := &container.HostConfig{
```
bwester: What happens to the container's stdout/stderr logs? Collected by dockerd I suppose. If we wanted to supply our own log handler, how might that work?

pkg/docker/docker.go
```diff
@@ -0,0 +1,222 @@
...
+		NetworkMode: "host",
+		RestartPolicy: container.RestartPolicy{
+			Name:              restartPolicy,
+			MaximumRetryCount: 0, // This is ignored unless Name is "on-failure"
+		},
+		AutoRemove: false,
+		Resources: container.Resources{
+			CPUPeriod:    cgroups.CPUPeriod,
+			CPUQuota:     int64(l.CPUQuota * cgroups.CPUPeriod),
+			CgroupParent: l.ParentCgroupID,
```
bwester: Would `p2-cgroup-info` need updating to be able to handle Docker's cgroup naming scheme?

pkg/launch/launchable.go
mpuncel: I meant to roll back this comment change actually, because I didn't end up using this.

I wonder if we can just not support overriding the entry point at all and use whatever is in the image?

pkg/launch/launchable.go
mpuncel: good point, I'll look into that. The word "version" does seem a little weird for specifying the image name as well as its version, though, so I might investigate changing all the places that use the version field to call something that each launchable type can implement

pkg/pods/pod.go
mpuncel: yeah definitely a weird API. looking through their code it does look like they return an error from `ImagePull` if the response code is `< 200` or `>= 400`, and that error includes the response body. So I don't think there's ever anything useful in the response?

pkg/docker/docker.go
mpuncel: yeah I agree with that

pkg/docker/docker.go
```diff
@@ -0,0 +1,222 @@
...
+		User:  l.RunAs, // TODO: this will break most of the time since the user doesn't exist in the container
+		Image: l.Image,
+	}
+
+	restartPolicy := "always"
+	switch l.RestartPolicy_ {
+	case runit.RestartPolicyAlways:
+		// keep it the way it is
+	case runit.RestartPolicyNever:
+		// keep it the way it is
```
spenceral: Is this comment correct?

mpuncel: nope!

pkg/docker/docker.go
mpuncel: I'm not that familiar with it, but that's probably the case

pkg/docker/docker.go
mpuncel: it looks like docker captures them by default and formats them as JSON, and you can get at them by running `docker logs CONTAINER`. The `HostConfig` struct here also arranges for you to configure different log drivers, they have some out of the box like splunk

pkg/docker/docker.go
```diff
@@ -0,0 +1,221 @@
...
+func (l *Launchable) RestartPolicy() runit.RestartPolicy {
+	return l.RestartPolicy_
+}
+
+func (l *Launchable) ServiceID() string {
+	return l.ServiceID_
+}
+
+func (l *Launchable) Stop(_ *runit.ServiceBuilder, _ runit.SV, _ bool) error {
+	// TODO: not implemented
```
esuen: Guessing you forgot to remove this comment.

PR #1047, opened 2018-05-09
Add node_requirements key to pod manifests

spenceral:
The node requirements key is based on Kubernetes's podSpec nodeSelector.
It is map of key value pairs that can be added to a replication
controller's node selector. For a pod to then be scheduled on a node,
the node needs to have each of the key-value pairs as labels.


spenceral:
As an example, this could be used to allow engineers to specify to which OS their pods should be scheduled
```
node_requirements
  - osversion: cent7
```
Assuming that nodes have been labeled with `osversion`.

PR #1050, opened 2018-05-17
Add docker support

esuen:
Preliminary PR, needs feedback, forked from mpuncel's branch.


pkg/docker/docker.go
```diff
@@ -134,8 +135,13 @@ func (l *Launchable) Launch(_ *runit.ServiceBuilder, _ runit.SV) error {
 	default:
 		return util.Errorf("invalid restart policy: %s", l.RestartPolicy_)
 	}
+
+	binds := []string{
+		fmt.Sprintf("/data/pods/%s:/data/pods/%s", l.RunAs, l.RunAs),
```
spenceral: I notice we hard code '/data/pods` in a few places. I wonder if it makes sense to declare constants or variables with that represent these common paths?

spenceral: Also should these be filepaths instead of strings?

esuen: And they should be strings if you see https://github.com/esuen/p2/blob/add-docker-support/vendor/github.com/docker/docker/api/types/container/host_config.go#L333.

integration/common-provision.sh
```diff
@@ -65,14 +65,6 @@ pushd /tmp
...
 cp /usr/local/share/go/src/github.com/square/p2/integration/hello/config.json opencontainer-hello_def456/
 
-# install docker using instructions from
-# https://docs.docker.com/install/linux/docker-ce/centos/#set-up-the-repository
-sudo yum install -y yum-utils device-mapper-persistent-data lvm2
-sudo yum-config-manager \
-    --add-repo \
-    https://download.docker.com/linux/centos/docker-ce.repo
-sudo yum install -y docker-ce
-
```
mpuncel: did you mean to remove this?

pkg/docker/docker.go
```diff
@@ -0,0 +1,277 @@
...
+	// the container
+	CgroupMemorySize size.ByteCount
+
+	// RestartPolicy_ determines what behavior docker should have when the pod exits.
+	// Typically it will indicate that docker should always restart the container
+	// for long-running services (e.g. servers)
+	RestartPolicy_ runit.RestartPolicy
+}
+
+func (l *Launchable) Disable() error {
```
mpuncel: not related to this PR, but maybe we do want this actually, e.g. kubernetes allows containers to specify a PreStop hook: https://kubernetes.io/docs/concepts/containers/container-lifecycle-hooks/

mpuncel: I have a use case where this might be nice

pkg/docker/docker.go
```diff
@@ -0,0 +1,277 @@
...
+}
+
+func (l *Launchable) Launch(_ *runit.ServiceBuilder, _ runit.SV) error {
+	if l.DockerClient == nil {
+		return util.Errorf("docker client was not initialized, can't launch docker launchable")
+	}
+
+	ctx := context.TODO()
+	envVars := []string{}
+	envFiles, err := ioutil.ReadDir(l.EnvDir())
```
mpuncel: should we do this for the pod's env dir as well?

mpuncel: that might be tricky if there are paths in there that don't apply in a container filesystem... something to think about

pkg/docker/docker.go
```diff
@@ -0,0 +1,277 @@
...
+	if err != nil {
+		return util.Errorf("could not read from %s: %v", l.EnvDir(), err)
+	}
+	for _, file := range envFiles {
```
mpuncel: is there any code in p2-exec that can be reused here? it probably knows how to turn a directory into a map[string]string or could be trivially modified to do so

pkg/docker/docker.go
```diff
@@ -0,0 +1,277 @@
...
+	for _, file := range mountFiles {
+		mounts := Mounts{}
+		if !file.IsDir() {
+			// read content and load into bindings var
+			mountFilePath := filepath.Join(dockerMountsDir, file.Name())
+			data, err := ioutil.ReadFile(mountFilePath)
+			if err != nil {
+				return util.Errorf("could not read file %s: %v", mountFilePath, err)
+			}
+			yaml.Unmarshal(data, mounts)
```
mpuncel: check error

pkg/docker/docker.go
```diff
@@ -0,0 +1,277 @@
...
+			for _, mount := range mounts {
+				if mount.Type != "bind" {
+					continue
+				}
+				bindings = append(bindings, fmt.Sprintf("%s:%s", mount.Source, mount.Target))
+			}
+		}
+	}
+	hostConfig := &container.HostConfig{
+		Binds:       bindings,
```
mpuncel: I wonder if we want to use the `Mounts` field instead of `Binds`. From https://docs.docker.com/storage/bind-mounts/, it sounds like --mounts is newer and more explicit and generally suggested for new users.

It also sounds like Binds will create the directory in the container even if the source folder doesn't exist?

integration/common-provision.sh
esuen: Yes, since docker will be installed at the imaging step unless this is the part that does it? @eam knows more details about.

pkg/docker/docker.go
esuen: could you give me an example of a path that doesn't apply in a container filesystem?

pkg/launch/launchable.go
```diff
@@ -59,10 +59,13 @@ type LaunchableStanza struct {
...
 
-	// Specifies which files or directories (relative to launchable root)
-	// should be launched under runit. Only launchables of type "hoist"
-	// make use of this field, and if empty, a default of ["bin/launch"]
-	// is used
+	// In the case of hoist launchables, specifies which files or
+	// directories (relative to launchable root) should be launched under
+	// runit. if empty, a default of ["bin/launch"] is used.
+	//
+	// In the case of docker containers, it specifies which commands
```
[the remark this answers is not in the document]
esuen: Should this be rolled back to what was there previously? i.e. just 
```
// Specifies which files or directories (relative to launchable root)
// should be launched under runit. Only launchables of type "hoist"
// make use of this field, and if empty, a default of ["bin/launch"]
// is used
```

integration/common-provision.sh
mpuncel: this is run to set up your vagrant virtual machine when you run `rake integration`, so this basically _is_ the imaging process

pkg/docker/docker.go
mpuncel: SECRETS_PATH could be different inside and outside the container, for example

pkg/launch/launchable.go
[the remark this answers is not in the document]
mpuncel: yeah 👍

integration/common-provision.sh
esuen: sorry for the confusion, rolling back my deletion.

pkg/docker/docker.go
esuen: You're right, I'm depending on the path in the container to exist and be correct. So far it is good with podHome and secrets bindings. I believe careful consideration will be needed when a new env variables wants to be introduced with a dockerized launchable.

esuen: "should we do this for the pod's env dir as well?"
thanks for catching that. Initially I meant to only do pod's env dir and realized I was only doing the launchable's env dir.

pkg/docker/docker.go
esuen: The code in p2-exec in func loadEnvDir seems to be similar code except it does some validation checks. These checks include ensuring there's no equal sign in the filename, taking only the first line of data in the file, discarding trailing white space, and replace null characters with newline. Would we want that validation here as well?

pkg/docker/docker.go
esuen: From that link:
*"By contrast, when you use a volume, a new directory is created within Docker’s storage directory on the host machine, and Docker manages that directory’s contents."*
I don't think we want or need that.

*"It also sounds like Binds will create the directory in the container even if the source folder doesn't exist?"*
From that link: *"The file or directory does not need to exist on the Docker host already. It is created on demand if it does not yet exist. Bind mounts are very performant, but they rely on the host machine’s filesystem having a specific directory structure available."*
From my understanding, for binds, the source directory needs to exist. Even without docker the directories we want to bind need to exist for the app to run.

bin/p2-launch/main.go
```diff
@@ -87,9 +93,27 @@ func main() {
...
 
-	err = pod.Install(manifest, auth.NopVerifier(), artifact.NewRegistry(*artifactRegistryURL, fetcher, osversion.DefaultDetector))
+	containerRegistryAuthStr := ""
+	if *containerRegistryJsonKeyFile != "" {
+		data, err := ioutil.ReadFile(*containerRegistryJsonKeyFile)
+		if err != nil {
+			log.Fatalf("could not read container json key file %s: %s", *containerRegistryJsonKeyFile, err)
+		}
+		authConfig := dockertypes.AuthConfig{
+			Username: "_json_key",
```
mpuncel: is this a common thing all docker users will have? also should this be a constant in pkg/docker?

bin/p2-launch/main.go
```diff
@@ -87,9 +93,27 @@ func main() {
...
+			Password: string(data),
+		}
+		encodedJSON, err := json.Marshal(authConfig)
```
mpuncel: check `err`

pkg/docker/docker.go
```diff
@@ -0,0 +1,310 @@
...
+	// docker pull is idempotent and cheap, so don't worry about this
+	return false
+}
+
+func (l *Launchable) Launch(_ *runit.ServiceBuilder, _ runit.SV) error {
+	if l.DockerClient == nil {
+		return util.Errorf("docker client was not initialized, can't launch docker launchable")
+	}
+
+	ctx := context.TODO()
```
mpuncel: might have been me, but I think this should be `context.Background()` instead of `TODO()` (they're the same thing, but it's not really TODO

pkg/docker/docker.go
```diff
@@ -0,0 +1,310 @@
...
+	for _, file := range mountFiles {
+		mounts := Mounts{}
+		if !file.IsDir() {
+			// read content and load into bindings var
+			mountFilePath := filepath.Join(dockerMountsDir, file.Name())
+			data, err := ioutil.ReadFile(mountFilePath)
+			if err != nil {
+				return util.Errorf("could not read file %s: %v", mountFilePath, err)
+			}
+			err = yaml.Unmarshal(data, mounts)
```
mpuncel: don't you need `&mounts`? I'd be surprised if it gets filled in if this isn't a pointer

pkg/docker/docker.go
```diff
@@ -0,0 +1,310 @@
...
+			if err != nil {
+				return util.Errorf("error unmarshalling data: %v", err)
+			}
+			for _, mount := range mounts {
+				if mount.Type != "bind" {
+					continue
```
mpuncel: probably log that you're skipping it

pkg/docker/docker.go
```diff
@@ -0,0 +1,310 @@
...
+func (l *Launchable) RestartPolicy() runit.RestartPolicy {
+	return l.RestartPolicy_
+}
+
+func (l *Launchable) ServiceID() string {
+	return l.ServiceID_
+}
+
+func (l *Launchable) Stop(_ *runit.ServiceBuilder, _ runit.SV, _ bool) error {
+	// TODO: not implemented
```
mpuncel: seems like this comment is outdated

pkg/manifest/manifest.go
```diff
@@ -517,13 +517,24 @@ func ValidManifest(m Manifest) error {
...
+				return fmt.Errorf("'%s': launchable must contain a 'location' or 'version'", launchableID)
+			case stanza.Location != "" && stanza.Version.ID != "":
+				return fmt.Errorf("'%s': launchable must not contain both 'location' and 'version'", launchableID)
+			}
+			continue
+		}
+
+		if stanza.LaunchableType == "docker" {
+			if stanza.Image.Name == "" {
+				return fmt.Errorf("docker launchables must contain an image")
```
mpuncel: nit: included `launchableID` in this message

pkg/preparer/setup.go
```diff
@@ -545,28 +553,51 @@ func New(preparerConfig *PreparerConfig, logger logging.Logger) (*Preparer, erro
...
+	if preparerConfig.ContainerRegistryJsonKeyFile != "" {
+		data, err := ioutil.ReadFile(preparerConfig.ContainerRegistryJsonKeyFile)
+		if err != nil {
+			return nil, util.Errorf("could not read container json key file %s: %s", preparerConfig.ContainerRegistryJsonKeyFile, err)
+		}
+		authConfig := dockertypes.AuthConfig{
+			Username: "_json_key",
+			Password: string(data),
+		}
+		encodedJSON, err := json.Marshal(authConfig)
```
mpuncel: check `err`. Also this code looks the same as what's in `p2-launch`, can it be centralized into a library somewhere?

pkg/docker/docker.go
esuen: Thanks for catching that!

pkg/docker/docker.go
esuen: It doesn't seem like the launchable has access to the logger so I'll probably not do this unless I determine it's worthwhile to pass the logger to the launchable type.

pkg/docker/docker.go
esuen: I recall it did work before during testing oddly but this is the correct way.

PR #1051, opened 2018-05-23
Spenceral/improve node transfers

spenceral:


pkg/rc/replication_controller.go
```diff
@@ -346,6 +346,28 @@ func (rc *replicationController) meetDesires(rcFields fields.RC) error {
...
+	// check if a node transfer node has been removed out of band
+	status, _, err := rc.rcStatusStore.Get(rc.rcID)
+	if err != nil && !statusstore.IsNoStatus(err) {
+		rc.logger.WithError(err).Errorln("Unable to fetch RC status to check for in progress node transfer")
+		return err
+	} else if !statusstore.IsNoStatus(err) {
+		hasOldNode := false
+		for _, node := range current.Nodes() {
+			if status.NodeTransfer.OldNode == node {
+				hasOldNode = true
```
esuen: You can break out of the for loop here.

pkg/rc/replication_controller.go
```diff
@@ -346,6 +346,28 @@ func (rc *replicationController) meetDesires(rcFields fields.RC) error {
...
+			}
+		}
+		if !hasOldNode {
```
esuen: In what cases does the old node get removed during a node transfer?

pkg/rc/replication_controller.go
```diff
@@ -1124,6 +1151,45 @@ func (rc *replicationController) finishTransfer(rcFields fields.RC, logger loggi
...
+	if err != nil {
+		return util.Errorf("could not add RC status deletion to cancel node transfer transaction: %s", err)
+	}
+
+	ok, resp, err := txn.CommitWithRetries(rc.txner)
+	switch {
+	case err != nil:
+		err := util.Errorf(
+			"RC status deletion returned err after timeout: %s",
+			err,
```
esuen: I think this `err` will try to reference the inner scope `err` but you mean to reference the outer scope `err` on line 1171.

pkg/rc/replication_controller_test.go
```diff
@@ -832,6 +832,16 @@ func TestNoOpIfNodeTransferInProgress(t *testing.T) {
...
+	err = applicator.SetLabel(labels.NODE, "old.123", "nodeQuality", "good")
+	Assert(t).IsNil(err, "expected no error labeling old.123")
+
+	rcFields.ReplicasDesired = 1
+
+	err = rc.meetDesires(rcFields)
+	if err != nil {
+		t.Fatal(err)
+	}
+
```
esuen: What is the purpose of the added code in this test file?

esuen:
LGTM but I think @mpuncel should weigh in here.

pkg/rc/replication_controller.go
spenceral: A p2 operator may want to forcefully remove a node while a transfer was in progress (perhaps it didn't finish in time or the replica on the new node was not coming up healthy). Currently, the operator would have to be sure to clean up things like labels manually but with this code change that will be taken care off.

pkg/rc/replication_controller_test.go
spenceral: This labels the old node in a staged node transfer as belonging to the RC. Without this, the RC would see that it had an in progress node transfer but then see that it was not on the "old node", so it would try to roll back the transfer.

spenceral:
Closing in favor of a node transfer rewrite.

PR #1052, opened 2018-05-24
hook env: don't error if POD_UNIQUE_KEY is empty

mpuncel:
POD_UNIQUE_KEY is empty for most pods, so it doesn't make sense to
return an error if it's empty, just return an empty string


mpuncel:
hmm `errcheck` isn't failing locally for me

spenceral:
I'm having the same `errcheck` issue on my PR (and similarly, it works locally)

esuen:
In order to fix the `errcheck` issue you need to add something to your Rakefile, I should have already applied this change to the master branch.

Your errcheck in Rakefile should look like
```
task :errcheck do
  e "exit $(errcheck -ignoretests -ignore \"fmt:.*\" github.com/square/p2/pkg/... | grep -v defer | wc -l || 1)"
end
```

PR #1053, opened 2018-05-29
Add log lines when daemon set handling goroutine exits

mpuncel:
Occasionally we see problems in the DS farm where the main control loop
gets blocked sending updates to daemon set handlers via channels. This
is likely because the handler goroutines are exiting, but logs don't
currently reveal the cause. This commit adds logs to all of the places
where these goroutines might exit which should shed light on the issue

### DIFF
--- a/pkg/ds/daemon_set.go
+++ b/pkg/ds/daemon_set.go
@@ -361,6 +361,7 @@ func (ds *daemonSet) WatchDesires(
 					// so that the timer would be stopped after
 					err = nil
 				case <-ctx.Done():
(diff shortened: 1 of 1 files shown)


pkg/ds/daemon_set.go
```diff
@@ -361,6 +361,7 @@ func (ds *daemonSet) WatchDesires(
 					// so that the timer would be stopped after
 					err = nil
 				case <-ctx.Done():
+					ds.logger.Warnln("goroutine exiting because ")
```
spenceral: Can you elaborate on this message

PR #1054, opened 2018-06-06
missing step of copying image to disk. skip missing artifact check fo…

esuen:
…r docker launchable


pkg/pods/pod.go
```diff
@@ -495,7 +494,12 @@ func (pod *Pod) Install(manifest manifest.Manifest, verifier auth.ArtifactVerifi
 				return util.Errorf("could not pull docker image: %s", err)
 			}
 			defer reader.Close()
-			io.Copy(os.Stdout, reader)
+			b, err := ioutil.ReadAll(reader)
+			if err != nil {
+				pod.logLaunchableError(launchable.ServiceID(), err, "error reading docker pull output")
+			}
+			// log the output of docker pull, there could be errors here that will cause future steps to fail
+			pod.logInfo(string(b))
```
mpuncel: do we actually want to log this? it will probably create a bunch of noise in the logs

mpuncel: i'm really unclear on what this returns, is it purely informational? I assume it's all the "pulling FS layer" output. the reader is only non-nil if the function didn't return an error, but maybe it's possible to get an error later on? do we have to parse the output of this thing to determine if there is an error? what a mess

mpuncel: we might need to do something like this: https://stackoverflow.com/questions/44452679/golang-docker-api-parse-result-of-imagepull

mpuncel: alternatively we could just log it and see how noisy it really is. we probably don't need to parse for errors, we'll see an error when the container is started and then we can go back and look at the logs. tl;dr I think this PR is good

esuen:
squashed commits, wrote proper commit message.

PR #1055, opened 2018-06-06
Convert docker usernames to uid:gid format

mpuncel:
This obviates the need to provide the username to uid/gid mapping in a
container's /etc/passwd file


mpuncel:
I didn't know it was an option until I bugged other people about it! the api is weird, it should probably have Uid and Gid int fields

PR #1056, opened 2018-06-06
Account for env variable conflicts between pod and launchable env vars

mpuncel:
Environment variables are configured via P2 hooks by placing files into
the "pod env dir" and each "launchable env dir". The former is useful
for setting environment variables on all processes in a pod, and the
latter is useful for when environment variables differ by launchable.

With hoist launchables, conflicts are addressed by preferring the
launchable-specific environment variables over the pod-wide ones. This
commit makes docker launchables the same.

### DIFF
--- a/pkg/docker/docker.go
+++ b/pkg/docker/docker.go
@@ -148,17 +148,10 @@ func (l *Launchable) Launch(_ *runit.ServiceBuilder, _ runit.SV) error {
 
 	ctx := context.Background()
 
-	// get pod env vars from disk
-	podEnvVars, err := loadEnvVars(l.PodEnvDir)
+	envVars, err := loadEnvVars(l.PodEnvDir, l.EnvDir())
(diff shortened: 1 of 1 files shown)


pkg/docker/docker.go
```diff
@@ -326,9 +319,53 @@ func loadEnvVars(envDir string) ([]string, error) {
...
+	for k, v := range envVarMap {
+		envVars = append(envVars, fmt.Sprintf("%s=%s", k, v))
+	}
+
+	return envVars, nil
+}
+
+// converts a username to the <uid>:<gid> format that Docker expects. This obviates the need
+// to set up username -> id mappings in a container's /etc/passwd file
+func dockerUser(username string) (string, error) {
```
esuen: should this function exist in this PR as well?

mpuncel:
yep that was a mistake

PR #1061, opened 2018-06-21
Modify p2-preparer to read pod whitelist from pod_whitelist_file

akun1012:
Introducing pod whitelist file, which contains a list of while list pods. Particular uses case is after an host is re imaged, there may still be some sticky apps' manifest remain in node's intent store,  previously, once p2-prepare boosts, it starts to watch intent store immediately and download and install the app binaries, even without other fundamentals of the host are ready. 
So here we introduces pod whitelist file, when p2-prepare is initialized, and before start normally working, we force p2-prepare to keep checking pod whitelist file, install all the pods in the whitelist, keep checking for update, until the pod whiltelist file is empty, which services as a signal that p2-prepare can proceed to play with it's full functionality.

### DIFF
--- a/bin/p2-preparer/main.go
+++ b/bin/p2-preparer/main.go
@@ -85,7 +85,7 @@ func main() {
 		logger.WithError(err).Fatalf("Could not do initial build reality at launch: %s", err)
 	}
 
-	podWhiteList, err := prep.CheckPodWhitelist(preparerConfig)
+	podWhiteList, err := prep.ProceedPodWhitelist(preparerConfig)
 	if err != nil {
 		logger.WithError(err).Fatalf("Error occurs when checking pod whitelist %+v", podWhiteList)
 	}
--- a/bin/p2-preparer/main.go
+++ b/bin/p2-preparer/main.go
@@ -1,8 +1,10 @@
 package main
 
 import (
+	"io/ioutil"
 	"os"
 	"os/signal"
+	"strings"
 	"sync"
 	"syscall"
 
@@ -11,6 +13,7 @@ import (
 
 	"github.com/square/p2/pkg/logging"
 	"github.com/square/p2/pkg/preparer"
+	"github.com/square/p2/pkg/types"
 	"github.com/square/p2/pkg/util/param"
 	"github.com/square/p2/pkg/version"
 	"github.com/square/p2/pkg/watch"
@@ -44,6 +47,7 @@ func main() {
 		defer statusServer.Close()
 	}
 
+	requiredPods := make(map[types.PodID]bool)
 	if preparerConfig.RequireFile != "" {
 		_, err := os.Stat(preparerConfig.RequireFile)
 		if os.IsNotExist(err) {
@@ -51,6 +55,14 @@ func main() {
 		}
 		if err != nil {
 			logger.WithError(err).Fatalln("Could not check for require file's existence")
+		}
+
+		required, err := loadTokens(preparerConfig.RequireFile)
+		if err != nil {
+			logger.WithError(err).WithField("path", preparerConfig.RequireFile).Fatalln("Could not read required file")
+		}
+		for pod := range required {
+			requiredPods[types.PodID(pod)] = true
 		}
 	}
 
@@ -83,6 +95,14 @@ func main() {
 	err = prep.BuildRealityAtLaunch()
 	if err != nil {
 		logger.WithError(err).Fatalf("Could not do initial build reality at launch: %s", err)
+	}
+
+	// check if any podID existing in require file(p2-may-run), if so, install them first
+	if len(requiredPods) == 0 {
+		err = prep.InstallRequiredPods(requiredPods, preparerConfig)
+		if err != nil {
+			logger.WithError(err).Fatalln("Could not install required pods")
+		}
 	}
 
 	go prep.WatchForPodManifestsForNode(quitMainUpdate)
@@ -123,4 +143,13 @@ func waitForTermination(logger logging.Logger, quitMainUpdate chan struct{}, qui
 	}
 	quitMainUpdate <- struct{}{}
 	<-quitMainUpdate // acknowledgement
+}
+
+func loadTokens(path string) ([]string, error) {
+	tokens, err := ioutil.ReadFile(path)
+	if err != nil {
+		return nil, err
+	}
+
+	return strings.Split(string(tokens), " "), nil
 }
--- a/bin/p2-preparer/main.go
+++ b/bin/p2-preparer/main.go
@@ -1,19 +1,20 @@
 package main
 
 import (
-	"io/ioutil"
 	"os"
 	"os/signal"
-	"strings"
 	"sync"
 	"syscall"
+	"time"
 
 	"github.com/Sirupsen/logrus"
 	"gopkg.in/alecthomas/kingpin.v2"
 
+	"github.com/square/p2/pkg/constants"
 	"github.com/square/p2/pkg/logging"
 	"github.com/square/p2/pkg/preparer"
 	"github.com/square/p2/pkg/types"
+	"github.com/square/p2/pkg/util"
 	"github.com/square/p2/pkg/util/param"
 	"github.com/square/p2/pkg/version"
 	"github.com/square/p2/pkg/watch"
@@ -47,7 +48,6 @@ func main() {
 		defer statusServer.Close()
 	}
 
-	requiredPods := make(map[types.PodID]bool)
 	if preparerConfig.RequireFile != "" {
 		_, err := os.Stat(preparerConfig.RequireFile)
 		if os.IsNotExist(err) {
@@ -55,14 +55,6 @@ func main() {
 		}
 		if err != nil {
 			logger.WithError(err).Fatalln("Could not check for require file's existence")
-		}
-
-		required, err := loadTokens(preparerConfig.RequireFile)
-		if err != nil {
-			logger.WithError(err).WithField("path", preparerConfig.RequireFile).Fatalln("Could not read required file")
-		}
-		for pod := range required {
-			requiredPods[types.PodID(pod)] = true
 		}
 	}
 
@@ -97,11 +89,38 @@ func main() {
 		logger.WithError(err).Fatalf("Could not do initial build reality at launch: %s", err)
 	}
 
-	// check if any podID existing in require file(p2-may-run), if so, install them first
-	if len(requiredPods) == 0 {
-		err = prep.InstallRequiredPods(requiredPods, preparerConfig)
-		if err != nil {
-			logger.WithError(err).Fatalln("Could not install required pods")
+	whiteListPods := make(map[types.PodID]bool)
+	if preparerConfig.PodWhitelistFile != "" {
+		// Keep loopinng the white list of pods in pod whitelist file, and install the non existing pods
+		// exit while the white list is empty
+		for {
+			_, err := os.Stat(preparerConfig.PodWhitelistFile)
+			// if the whitelist file does not exist, end the loop and proceed
+			if os.IsNotExist(err) {
+				logger.WithError(err).Warningf("Pod whilelist file does not exist")
+				break
+			}
+
+			podWhitelist, err := util.LoadTokens(preparerConfig.PodWhitelistFile)
+			if err != nil {
+				logger.WithError(err).WithField("path", preparerConfig.PodWhitelistFile).Fatalln("Could not read pod whitelist file")
+			}
+
+			// if the pod white list is empty, jump out of the loop, then p2-preparer proceeds to start with full functionality
+			if len(podWhitelist) == 0 {
+				break
+			}
+
+			for pod := range podWhitelist {
+				whiteListPods[types.PodID(pod)] = true
+			}
+
+			err = prep.InstallWhiteListPods(whiteListPods, preparerConfig)
+			if err != nil {
+				logger.WithError(err).Fatalln("Could not install whitelist pods")
+			}
+
+			time.Sleep(constants.P2WhitelistCheckInterval)
 		}
 	}
 
@@ -143,13 +162,4 @@ func waitForTermination(logger logging.Logger, quitMainUpdate chan struct{}, qui
 	}
 	quitMainUpdate <- struct{}{}
 	<-quitMainUpdate // acknowledgement
-}
-
-func loadTokens(path string) ([]string, error) {
-	tokens, err := ioutil.ReadFile(path)
-	if err != nil {
-		return nil, err
-	}
-
-	return strings.Split(string(tokens), " "), nil
 }
--- a/bin/p2-preparer/main.go
+++ b/bin/p2-preparer/main.go
@@ -5,16 +5,12 @@ import (
 	"os/signal"
 	"sync"
 	"syscall"
-	"time"
 
 	"github.com/Sirupsen/logrus"
 	"gopkg.in/alecthomas/kingpin.v2"
 
-	"github.com/square/p2/pkg/constants"
 	"github.com/square/p2/pkg/logging"
 	"github.com/square/p2/pkg/preparer"
-	"github.com/square/p2/pkg/types"
-	"github.com/square/p2/pkg/util"
 	"github.com/square/p2/pkg/util/param"
 	"github.com/square/p2/pkg/version"
 	"github.com/square/p2/pkg/watch"
@@ -89,39 +85,9 @@ func main() {
 		logger.WithError(err).Fatalf("Could not do initial build reality at launch: %s", err)
 	}
 
-	whiteListPods := make(map[types.PodID]bool)
-	if preparerConfig.PodWhitelistFile != "" {
-		// Keep loopinng the white list of pods in pod whitelist file, and install the non existing pods
-		// exit while the white list is empty
-		for {
-			_, err := os.Stat(preparerConfig.PodWhitelistFile)
-			// if the whitelist file does not exist, end the loop and proceed
-			if os.IsNotExist(err) {
-				logger.WithError(err).Warningf("Pod whilelist file does not exist")
-				break
-			}
-
-			podWhitelist, err := util.LoadTokens(preparerConfig.PodWhitelistFile)
-			if err != nil {
-				logger.WithError(err).WithField("path", preparerConfig.PodWhitelistFile).Fatalln("Could not read pod whitelist file")
-			}
-
-			// if the pod white list is empty, jump out of the loop, then p2-preparer proceeds to start with full functionality
-			if len(podWhitelist) == 0 {
-				break
-			}
-
-			for pod := range podWhitelist {
-				whiteListPods[types.PodID(pod)] = true
-			}
-
-			err = prep.InstallWhiteListPods(whiteListPods, preparerConfig)
-			if err != nil {
-				logger.WithError(err).Fatalln("Could not install whitelist pods")
-			}
-
-			time.Sleep(constants.P2WhitelistCheckInterval)
-		}
+	podWhiteList, err := prep.CheckPodWhitelist(preparerConfig)
+	if err != nil {
+		logger.WithError(err).Fatalf("Error occurs when checking pod whitelist %+v", podWhiteList)
 	}
 
 	go prep.WatchForPodManifestsForNode(quitMainUpdate)
--- a/pkg/preparer/orchestrate.go
+++ b/pkg/preparer/orchestrate.go
@@ -5,6 +5,8 @@ import (
 	"fmt"
 	"time"
 
+	"os"
+
 	"github.com/Sirupsen/logrus"
 	"github.com/square/p2/pkg/artifact"
 	"github.com/square/p2/pkg/auth"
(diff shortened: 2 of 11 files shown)


pkg/preparer/orchestrate.go
```diff
@@ -79,6 +81,90 @@ func (p podWorkerID) String() string {
...
+		p.Logger.WithError(err).Errorln("could not check intent")
+	}
+	errorChan := make(chan error)
+	quit := make(chan struct{})
+
+	for _, intentResult := range intentResults {
+		if _, ok := requiredPods[intentResult.Manifest.ID()]; ok {
+			// check pod health here, to prevent reinstall. This is useful when
+			// the required pod is already running, and the p2-prepare restarted
+			if p.checkPodHealth(intentResult, preparerConfig) {
```
mpuncel: why does the normal intent/reality diff not catch this?

akun1012: This happens before p2-prepare normally start actually. So p2-prepare not watching the intent store now. I planed to make p2-prepare to finish install all required pods, then start the thread to watch for intent/reality diffs.

bin/p2-preparer/main.go
```diff
@@ -52,6 +56,14 @@ func main() {
 		if err != nil {
 			logger.WithError(err).Fatalln("Could not check for require file's existence")
 		}
+
+		required, err := loadTokens(preparerConfig.RequireFile)
+		if err != nil {
+			logger.WithError(err).Fatalln("Could not read required file")
```
spenceral: What do you think about including the require file path in this error log so that is more descriptive?

bin/p2-preparer/main.go
```diff
@@ -85,6 +97,13 @@ func main() {
 		logger.WithError(err).Fatalf("Could not do initial build reality at launch: %s", err)
 	}
 
+	if requiredPods != nil {
```
spenceral: You've allocated and instantiated `requiredPods` with `make()`, it cannot be `nil`. What are you trying to do with this check? You'll have to do it another way.

bin/p2-preparer/main.go
```diff
@@ -85,6 +97,13 @@ func main() {
...
+		err = prep.InstallRequiredPods(requiredPods, preparerConfig)
+		if err != nil {
+			logger.WithError(err).Fatalln("Could not do initial required pods")
```
spenceral: Could you change "do" to "install"?

spenceral:
If the require file is a non zero byte file, will the preparer's `WatchForPodManifestsForNode` loop still install any and all pods in intent? If yes, what do you think about the requirement that the preparer not install any pods that are not in the require file when the require file does not have a zero length?

That would allow external to the preparer processes, imaging for example, control when the preparer is installed and then also **when** and what pods it is allowed to install.

akun1012:
@spenceral 
"If the require file is a non zero byte file, will the preparer's WatchForPodManifestsForNode loop still install any and all pods in intent?"
The answer is no, although watchPods() will signal all pods from intent for the first time, but p2-prepare will compare the intent and reality menifests SHAs, if same, no installation will happen for the pods.

spenceral:
But what if the pods are not in reality?

akun1012:
If the pods not in reality, means it's not installed. Then I will install by WatchForPodManifestsForNode, which should not happen if the podID is in "require file".

spenceral:
But what if we don't want to install it yet? The require file is created and written to by an external process. That external process may not want the preparer to install any non-required pods until it is ready. With this change, you have implemented an ordering, installing required pods before other pods, but you are not preventing the installation of non-required pods. 

Do you think that is a valid requirement?

akun1012:
Updated. Now p2-prepare will just install pods listed in pod_whilelist_file (path in p2-manifest), and continuously check this file, install newly added white listed pods, or update the white list pods if there is a new version in intent. Until the pod_whitelist_file being empty, p2-prepare then proceeds to perform full functionality to what intent/reality store, install and launch other pods.

bin/p2-preparer/main.go
```diff
@@ -1,21 +1,27 @@
...
 
 	"github.com/square/p2/pkg/logging"
 	"github.com/square/p2/pkg/preparer"
+	"github.com/square/p2/pkg/types"
 	"github.com/square/p2/pkg/util/param"
 	"github.com/square/p2/pkg/version"
 	"github.com/square/p2/pkg/watch"
 )
 
+const P2WHITELIST_CONFIGCHECK_INTERVAL = 10 * time.Second
```
spenceral: Can you move this to `pkg/constants/constants.go`

spenceral: Also, what led you to settle on 10 seconds for the interval out of curiosity?

spenceral: And, see https://github.com/golang/go/wiki/CodeReviewComments#mixed-caps for const naming convention.

akun1012: I think it does not need to check every second or two, and 1 min may be too long if user already sent signal to let p2-prepare proceed. Let me know if you have a better one.

bin/p2-preparer/main.go
```diff
@@ -85,6 +90,44 @@ func main() {
...
 	}
 
+	whiteListPods := make(map[types.PodID]bool)
+	// Keep loopinng the white list of pods in pod whitelist file, and install the non existing pods
+	// exit while the white list is empty
+	if preparerConfig.PodWhitelistFile != "" {
+		for {
+			_, err := os.Stat(preparerConfig.PodWhitelistFile)
+			if os.IsNotExist(err) {
+				logger.WithError(err).Warningf("Pod whilelist path is not valid")
```
spenceral: Is a warning the most semantically correct response to this error? A warning allows continued execution. The preparer will still attempt to read the file that we have just verified does not exist.

bin/p2-preparer/main.go
```diff
@@ -85,6 +90,44 @@ func main() {
...
+			if err != nil {
+				logger.WithError(err).Warningf("Could not check for pod whitelist file's existence")
+			}
+
+			podWhitelist, err := loadTokens(preparerConfig.PodWhitelistFile)
+			if err != nil {
+				logger.WithError(err).WithField("path", preparerConfig.PodWhitelistFile).Fatalln("Could not read pod whitelist file")
+			}
+
+			// if the pod white list is empty, jump out of the loop, then p2-prepare proceeds to start with full functionality
```
spenceral: s/p2-prepare/p2-preparer

bin/p2-preparer/main.go
```diff
@@ -85,6 +90,44 @@ func main() {
...
+			if len(podWhitelist) == 0 {
+				break
+			}
+
+			for pod := range podWhitelist {
+				whiteListPods[types.PodID(pod)] = true
+			}
+
+			if len(whiteListPods) != 0 {
```
spenceral: What is the purpose of this check? Is it possible for the number of white listed pods to become 0 between the check on line 112 and here?

bin/p2-preparer/main.go
```diff
@@ -124,3 +167,12 @@ func waitForTermination(logger logging.Logger, quitMainUpdate chan struct{}, qui
 	quitMainUpdate <- struct{}{}
 	<-quitMainUpdate // acknowledgement
 }
+
+func loadTokens(path string) ([]string, error) {
```
spenceral: nit: would it make more organizational sense to put this kind of generic function in `pkg/util`?

pkg/preparer/orchestrate.go
```diff
@@ -79,6 +79,89 @@ func (p podWorkerID) String() string {
 	return fmt.Sprintf("%s-%s", p.podID.String(), p.podUniqueKey)
 }
 
+func (p *Preparer) InstallWhiteListPods(whiteListPods map[types.PodID]bool, preparerConfig *PreparerConfig) error {
+	intentResults, _, err := p.store.ListPods(consul.INTENT_TREE, p.node)
+	realityResults, _, err := p.store.ListPods(consul.REALITY_TREE, p.node)
+	var intents []consul.ManifestResult
+	if err != nil {
+		p.Logger.WithError(err).Errorln("could not check intent")
```
spenceral: 1) The error here may be from checking reality, so the error message is misleading.
2) Should we return the error here? An inability to check intent seems fatal worthy given the intention of this function. As it stands, I believe this function would return nil which might mislead the caller into thinking this function installed the white listed pods (as one would naturally expect given the name and the nil return value)

pkg/preparer/orchestrate.go
```diff
@@ -79,6 +79,89 @@ func (p podWorkerID) String() string {
...
+		pod, err = p.podFactory.NewUUIDPod(intentManifest.ID, intentManifest.PodUniqueKey)
+		if err != nil {
+			manifestLogger.WithError(err).Errorln("Could not initialize pod")
+			errorChan <- util.Errorf("failed to initialize pod: %s, error: %v", intentManifest.Intent.ID(), err)
+		}
+	}
+
+	ok, err := p.preparePod(*intentManifest, pod, manifestLogger)
+	if !ok || err != nil {
+		errorChan <- util.Errorf("failed to install pod: %s, error: %v", intentManifest.Intent.ID(), err)
```
spenceral: `err` may be `nil` here.

pkg/preparer/orchestrate.go
```diff
@@ -222,83 +304,14 @@ func (p *Preparer) handlePods(podChan <-chan ManifestPair, quit <-chan struct{})
...
-						nextLaunch.Reality = nil
-					default:
-						manifest, err := manifest.FromBytes([]byte(status.Manifest))
-						if err != nil {
-							manifestLogger.WithError(err).Errorln("Error parsing reality manifest from pod status")
-							break
-						}
-						nextLaunch.Reality = manifest
-					}
+				ok, err := p.preparePod(nextLaunch, pod, manifestLogger)
```
spenceral: What is the advantage of two return values here? They seem to be used for the same purpose. `!ok` is never treated differently than `err != nil` so I don't see the purpose of two different failure cases.

bin/p2-preparer/main.go
akun1012: good catch!

pkg/preparer/orchestrate.go
akun1012: Because in some cases when there is no error, we also need to break the process.

pkg/preparer/orchestrate.go
akun1012: fixed

spenceral: `err` is set to a new value on line 84 and its value on line 83 is never read or consumed. If there is an error reading intent, but not reality, `err` will be nil even though there _was_ an error.

pkg/preparer/orchestrate.go
```diff
@@ -79,6 +79,90 @@ func (p podWorkerID) String() string {
...
+	for _, intentManifest := range intents {
+		intentManifest := &ManifestPair{
+			Intent:       intentManifest.Manifest,
+			ID:           intentManifest.Manifest.ID(),
+			PodUniqueKey: intentManifest.PodUniqueKey,
+		}
+
+		go p.handleWhiteListPod(intentManifest, errorChan, quit)
+	}
+	for i := 0; i < len(intents); i++ {
```
spenceral: small style nit: since you don't use `i`, you could always write
```
for range intents {
}
```

bin/p2-preparer/main.go
```diff
@@ -85,6 +89,41 @@ func main() {
...
+			_, err := os.Stat(preparerConfig.PodWhitelistFile)
+			// if the whitelist file does not exist, end the loop and proceed
+			if os.IsNotExist(err) {
+				logger.WithError(err).Warningf("Pod whilelist file does not exist")
+				break
+			}
+
+			podWhitelist, err := util.LoadWhitelistPods(preparerConfig.PodWhitelistFile)
+			if err != nil {
+				logger.WithError(err).WithField("path", preparerConfig.PodWhitelistFile).Fatalln("Could not read pod whitelist file")
```
spenceral: Style nit: error strings should not be capitalized. https://github.com/golang/go/wiki/CodeReviewComments#error-strings

pkg/preparer/orchestrate.go
```diff
@@ -79,6 +79,90 @@ func (p podWorkerID) String() string {
...
+	if err != nil {
+		p.Logger.WithError(err).Errorln("could not check intent")
+		return err
+	}
+	errorChan := make(chan error)
+	quit := make(chan struct{})
+
+	for _, intentResult := range intentResults {
+		intentSHA, _ := intentResult.Manifest.SHA()
+		if _, ok := whiteListPods[intentResult.Manifest.ID()]; ok {
```
spenceral: From https://github.com/golang/go/wiki/CodeReviewComments#indent-error-flow, you could reduce indentation with:
```
ok := whiteListPods[intentResult.Manifest.ID()]
if !ok {
        break
} 
// rest of loop body, now indented one less level.
```
This idea is shown in the fourth example of the link above.

pkg/util/readfile.go
```diff
@@ -0,0 +1,15 @@
+package util
+
+import (
+	"io/ioutil"
+	"strings"
+)
+
+func LoadWhitelistPods(path string) ([]string, error) {
```
spenceral: Since this function is defined in a generic `util` package, and contains _no_ white list pod logic, I think your original name `LoadTokens` (or similar, generic line reading name of your choice) is actually more appropriate

pkg/preparer/orchestrate.go
akun1012: good catch!

pkg/preparer/orchestrate.go
```diff
@@ -79,6 +79,90 @@ func (p podWorkerID) String() string {
...
+				}
+			}
+			if !installed {
+				intents = append(intents, intentResult)
+			}
+		}
+	}
+
+	for _, intentManifest := range intents {
+		intentManifest := &ManifestPair{
```
spenceral: I defer to your judgment, but I wanted point out that I found it tough to read these two lines. You declare a variable `intentManifest` on line 111, type `ManifestResult`, and then reuse the same variable name on the very next line, but now its type is `&ManifestPair`. I found it harder to reason about than I think it has to be.

pkg/preparer/orchestrate.go
```diff
@@ -79,6 +79,90 @@ func (p podWorkerID) String() string {
...
+		p.Logger.WithError(err).Errorln("could not check intent")
+		return err
+	}
+	errorChan := make(chan error)
+	quit := make(chan struct{})
+
+	for _, intentResult := range intentResults {
+		intentSHA, _ := intentResult.Manifest.SHA()
+		if _, ok := whiteListPods[intentResult.Manifest.ID()]; ok {
+			// compare intent/reality for the pod here, to prevent reinstall
```
spenceral: `resolvePair` already does this comparison. Is there any reason to do it twice?

bin/p2-preparer/main.go
```diff
@@ -85,6 +89,41 @@ func main() {
 		logger.WithError(err).Fatalf("Could not do initial build reality at launch: %s", err)
 	}
 
+	whiteListPods := make(map[types.PodID]bool)
+	// Keep loopinng the white list of pods in pod whitelist file, and install the non existing pods
+	// exit while the white list is empty
+	if preparerConfig.PodWhitelistFile != "" {
```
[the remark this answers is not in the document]
akun1012: I would move it above for loop, to indicate the intention of the loop

pkg/preparer/orchestrate.go
akun1012: Different cases, I'm checking a map, not related to the error flow in your link. 
Same usage in check.go line 345

pkg/preparer/orchestrate.go
akun1012: I just wanted to maintain a clear logic here, that we just install whitelist pods which haven't installed. Yeah, if we let `resolvePair` to decide, we still need the for loop here to compare the manifest.ID(), and put the reality manifest into the `ManifestPair`.

bin/p2-preparer/main.go
akun1012: I've got your point, but seems they are different cases. It should be lower cases when we are forming a err, which could be inserted in a whole error message in the log. But here with `fatalln`, we are initializing the error message. So should be capitalized.

pkg/preparer/orchestrate.go
spenceral: It is Go convention to reduce the indentation level of the normal path. The [counterexample](https://github.com/square/p2/blob/master/integration/single-node-slug-deploy/check.go#L345) you provide is not the same. It checks for `!ok`, whereas you check for `ok`. Notice how the `!ok` check there is an exceptional, error returning case; it is not the normal path.

Here, you are unnecessarily nesting the normal path within an if block.

pkg/preparer/orchestrate.go
spenceral: Installing only those pods which have not yet been installed is the typical, expected behavior for the preparer. I believe that redundantly performing this check makes the logic less clear because it begs the question of why the typical preparer behavior is insufficient.

spenceral:
Can you write tests?

pkg/preparer/orchestrate.go
akun1012: The counterexample is the same case here to check the existence of a key in map. Well, as the value here is bool already, I would remove "ok".

pkg/preparer/orchestrate.go
akun1012: They are not the exactly the same logic, I have made the upmost the use existing codes. Well, we can do this in either way, I will modify as you've mentioned.

pkg/preparer/orchestrate_test.go
```diff
@@ -218,9 +244,47 @@ func testPreparer(t *testing.T, f *FakeStore) (*Preparer, *fakeHooks, string) {
...
+		currentManifest: testManifest(t),
+		wipedReality:    true,
+	}
+
+	p, _, fakePodRoot := testPreparer(t, store)
+	defer p.Close()
+	defer os.RemoveAll(fakePodRoot)
+
+	err := p.InstallWhiteListPods(whiteListPods)
+	Assert(t).IsTrue(err != nil, "Pod cannot be installed.")
```
spenceral: Is this test asserting that an error occurs? If so, can the name of the test be changed to reflect that it is testing an unhappy path?

pkg/preparer/orchestrate_test.go
```diff
@@ -218,9 +244,47 @@ func testPreparer(t *testing.T, f *FakeStore) (*Preparer, *fakeHooks, string) {
...
+	s := strings.Split(err.Error(), " ")
+	podID := s[len(s)-1]
+	Assert(t).AreEqual(podID, "hello", "The podID to install should be the one in intent")
+}
+
+func TestInstallWhiteListPodsNotInTheIntent(t *testing.T) {
```
spenceral: Is this name implying that this new code change will install pods that are in the whitelist but not in intent?

spenceral:
Can you expand on the tests? Right now the tests don't interact with the existence of the whitelist file. They also do not verify that any pods are installed, or that pods _not in_ the whitelist file are not installed.

pkg/preparer/orchestrate_test.go
akun1012: Have refined the names of the tests.

pkg/preparer/orchestrate.go
```diff
@@ -79,6 +81,125 @@ func (p podWorkerID) String() string {
 	return fmt.Sprintf("%s-%s", p.podID.String(), p.podUniqueKey)
 }
 
+func (p *Preparer) ProceedPodWhitelist(preparerConfig *PreparerConfig) (map[types.PodID]bool, error) {
+	whiteListPods := make(map[types.PodID]bool)
+	if preparerConfig.PodWhitelistFile != "" {
+		// Keep loopinng the white list of pods in pod whitelist file, and install the non existing pods
```
spenceral: typo looping

pkg/preparer/orchestrate_test.go
```diff
@@ -218,9 +246,85 @@ func testPreparer(t *testing.T, f *FakeStore) (*Preparer, *fakeHooks, string) {
...
+
+	file.Write([]byte("slug"))
+
+	cfg := &PreparerConfig{
+		PodWhitelistFile: testPath,
+	}
+	defer os.Remove(testPath)
+	go func(cfg *PreparerConfig) {
+		whitelistPods, err := p.ProceedPodWhitelist(cfg)
+		Assert(t).IsTrue(len(whitelistPods) == 1, "Successfully read the pods from whitelist file")
```
spenceral: The second argument should be an error message.

pkg/preparer/orchestrate_test.go
```diff
@@ -218,9 +246,85 @@ func testPreparer(t *testing.T, f *FakeStore) (*Preparer, *fakeHooks, string) {
...
+		Assert(t).IsNil(err, "Successfully exited the infinite for loop of checking pod whitelist, when the file is deleted")
```
spenceral: The second message should be an error message.

akun1012:
@spenceral Do you think the logic is covered in `TestPrepareProceedWhiteListFileIfConfigured`?

spenceral:
No I don't think it does, I don't see any pods in intent in that test. Imagine the intent store has pods `a` and `b`, but the whitelist file only contains pod `a`. Could you write a test that verifies that pod `b` would _not_ be installed?

akun1012:
Tested on a canary node, working correctly.

PR #1064, opened 2018-06-21
add structural changes to launchable stanza

esuen:


esuen:
Thanks for approval @spenceral, but I had more changes in mind, I should've made that clear.

esuen:
Do not review yet, WIP

esuen:
Ready for review.

pkg/docker/docker.go
```diff
@@ -236,6 +267,26 @@ func (l *Launchable) Launch(_ *runit.ServiceBuilder, _ runit.SV) error {
 	if err != nil {
 		return util.Errorf("could not start container %s: %s", containerName, err)
 	}
+	// TODO: check if we need to set anything else in this
```
mpuncel: not a blocker for this PR, but the `Launch()` function is getting really big and does a lot of things, like generating multiple complex configuration structures, building up environment variables, building up mounts. it might be nice to break this up into a few smaller functions so that it's more clear at a high level all of the things `Launch()` is doing

mpuncel: probably some of those things deserve more tests as well before Docker gets used widely

PR #1066, opened 2018-06-27
Image pull and prune resources

esuen:
- actually pull docker image by sha
- remove image pull logging as it's not useful
- add docker pruning at uninstall step

### DIFF
--- a/pkg/docker/docker.go
+++ b/pkg/docker/docker.go
@@ -13,6 +13,7 @@ import (
 
 	"gopkg.in/yaml.v2"
 
+	"github.com/docker/docker/api/types/filters"
 	"github.com/square/p2/pkg/cgroups"
 	"github.com/square/p2/pkg/launch"
 	"github.com/square/p2/pkg/runit"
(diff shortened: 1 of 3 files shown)


pkg/pods/pod.go
```diff
@@ -421,6 +431,35 @@ func (pod *Pod) Uninstall() error {
...
 
+	err = pod.pruneDockerResources()
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func (pod *Pod) pruneDockerResources() error {
```
mpuncel: how much trouble do you think it would be to reorganize all of this so it's in `pkg/docker`? probably that means changing the `artifact.Downloader` interface or creating a new interface on top, but probably we shouldn't keep piling docker stuff in this file

pkg/docker/docker.go
```diff
@@ -359,6 +365,30 @@ func GetContainerRegistryAuthStr(jsonKeyFile string) (string, error) {
 	return containerRegistryAuthStr, nil
 }
 
+func PruneResources(client *dockerclient.Client) error {
```
mpuncel: is this just a global prune? like it will prune images for images from different pods?

PR #1069, opened 2018-06-28
Add FromBytesNoValidation() function to manifest package

mpuncel:
This is useful for use cases where you don't want to error on an invalid
manifest, for instance if you're parsing previously deployed manifests
that might be invalid according to newly introduced validations


esuen:
Nice.

PR #1072, opened 2018-07-03
Docker prune should not stop uninstall on failure and also should onl…

esuen:
…y run when there's a docker launchable.

### DIFF
--- a/pkg/pods/pod.go
+++ b/pkg/pods/pod.go
@@ -421,9 +421,21 @@ func (pod *Pod) Uninstall() error {
 		return err
 	}
 
+	// only do this step if this pod has a docker launchable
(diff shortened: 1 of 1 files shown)


pkg/pods/pod.go
```diff
@@ -421,9 +421,21 @@ func (pod *Pod) Uninstall() error {
...
+			break
+		}
+	}
+	if !dockerLaunchableFound {
+		return nil
+	}
 	err = docker.PruneResources(pod.DockerClient)
 	if err != nil {
-		return err
+		// do not return err if this fails, pruning resources is not crtical to uninstalling
```
spenceral: typo in critical

PR #1073, opened 2018-07-03
Node transfers v2

spenceral:


pkg/rc/replication_controller.go
```diff
@@ -335,18 +258,24 @@ func (rc *replicationController) meetDesires(rcFields fields.RC) error {
...
+		ok, err := rc.attemptNodeTransfer(rcFields, current, ineligible[0], MaxAllocateAttempts)
 		if err != nil {
 			return err
+		} else if !ok {
+			rc.logger.Infoln("node transfer conditions not met; skipping")
 		}
 	}
 
-	return rc.ensureConsistency(rcFields, current.Nodes(), eligible)
+	return rc.ensureConsistency(rcFields)
```
esuen: What was the reasoning behind moving the params from outside the function to inside?

esuen:
LGTM, did you test with a canary node?

pkg/rc/replication_controller.go
spenceral: `ensureConsistency` needs to be passed an accurate and up-to-date  list of current and eligible nodes for it to serve its purpose. A node transfer would alter those two values. Rather than compute them for a third time in `meetDesires()`, I thought it might be better to just compute them in `ensureConsistency`. I think that helps prevent anyone who makes changes in the future from ever accidentally passing outdated values to the function.

That said, it is not a change I feel particularly strongly about, just something I noticed since this is called immediately after the node transfer.

spenceral:
I have not yet tested this beyond unit testing. I'll get on that.

PR #1078, opened 2018-07-13
Allow some node transfers during rolling updates

spenceral:
Specifically, when the old RC has only ineligible nodes and replicas that it does not desire.

### DIFF
--- a/pkg/rc/replication_controller.go
+++ b/pkg/rc/replication_controller.go
@@ -267,7 +267,7 @@ func (rc *replicationController) meetDesires(rcFields fields.RC) error {
 	ineligible := rc.checkForIneligible(current, eligible)
 	if len(ineligible) > 0 && rcFields.AllocationStrategy == fields.DynamicStrategy {
 		rc.logger.Infof("ineligible nodes: %s found", ineligible)
-		ok, err := rc.attemptNodeTransfer(rcFields, current, ineligible[0], MaxAllocateAttempts)
+		ok, err := rc.attemptNodeTransfer(rcFields, current, ineligible, MaxAllocateAttempts)
 		if err != nil {
 			return err
 		} else if !ok {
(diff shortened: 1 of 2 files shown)


pkg/rc/replication_controller.go
```diff
@@ -654,23 +654,66 @@ func (rc *replicationController) unschedule(txn *auditingTransaction, rcFields f
...
+}
+
+func (rc *replicationController) canNodeTransfer(rcFields fields.RC, current types.PodLocations, ineligibles []types.NodeName) (bool, consul.Unlocker, error) {
+	currentSet := types.NewNodeSet(current.Nodes()...)
+	ineligibleSet := types.NewNodeSet(ineligibles...)
+	if rcFields.Disabled &&
+		currentSet.Equal(ineligibleSet) &&
+		rcFields.ReplicasDesired < len(current) {
+		// If the RC is disabled, all of its nodes are inelgible, and its
+		// desired replica count is greater than its current replica count,
```
esuen: I think you mean "desired replica count is less than".

spenceral: Yes, thank you.

pkg/rc/replication_controller.go
```diff
@@ -654,23 +654,66 @@ func (rc *replicationController) unschedule(txn *auditingTransaction, rcFields f
...
 	}
 
 	_, session, err := consul.SessionContext(context.Background(), rc.consulClient, fmt.Sprintf("rc-node-transfer-%s", rc.rcID))
 	if err != nil {
 		rc.logger.WithError(err).Errorln("could not create a session to acquire rc lock")
-		return false, err
+		return false, nil, err
 	}
 	// Node transfers do not mutate the RC's state, but we attempt to acquire
 	// the lock to ensure that we do not perform a transfer during a rolling
```
akun1012: We probably need to polish the annotations here to diminish the confusion. I suppose you mean something like, to ensure that we do not perform a rolling update during node transfer.

PR #1080, opened 2018-07-19
Limit node transfers during rolling updates

spenceral:
Replication controllers are permitted to perform node transfers during
rolling updates if they have more replicas than desired. Since
performing a node transfer produces an unhealthy replica, node transfers
should only occur in this case when the it would not produce more
unhealthy replicas than len(current) - ReplicasDesired. This is to
preserve cluster health.


spenceral:
This is actually unnecessary because of the requirement that all current nodes must be ineligible for a node transfer to occur during a rolling update.

PR #1082, opened 2018-07-24
Only write to swap limit file if it doesn't exist

spenceral:


eam:
This may help us avoid certain [memory leaks in the centos7 kernel.](https://bugzilla.redhat.com/show_bug.cgi?id=1507149)

PR #1084, opened 2018-08-20
Set IdleConnTimeout on http transport

spenceral:
We can also use this PR to discuss setting other fields like `MaxIdleConns` that people see as helpful to reduce the number of connections the preparer keeps open to Consul.

### DIFF
--- a/pkg/preparer/setup.go
+++ b/pkg/preparer/setup.go
@@ -184,6 +184,9 @@ type PreparerConfig struct {
 	// clients, e.g. consul client vs artifact downloader
 	HTTPTimeout time.Duration `yaml:"http_timeout"`
 
+	// IdleConnTimeout will be set on the preparer's HTTP client transport.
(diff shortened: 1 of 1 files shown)


pkg/preparer/setup.go
```diff
@@ -337,6 +340,12 @@ func (c *PreparerConfig) getClient(
 		cxnTimeout = 30 * time.Second
 	}
 
+	idleConnTimeout := c.IdleConnTimeout
+	if idleConnTimeout == time.Duration(0*time.Second) {
```
bwester: You can simplify `time.Duration(0*time.Second)` to `0` here.

bwester:
You should definitely look at copying some other settings from the default `Transport` object defined in the `http` package, even if they aren't overridable in this manner.

Also, switching from `net.Dial` to `net.DialContext` would be beneficial too. I think that without this, connection timeouts might not actually abort the low-level connection attempt.

PR #1085, opened 2018-08-22
Support unix socket http client in p2-preparer

spenceral:
~~This is an initial, MVP for deploying an Envoy sidecar launchable
alongside p2-preparer to handle HTTP traffic between preparers and
Consul. It works by using a custom Transport that sends requests to
the Envoy unix socket along with other request metadata Envoy needs to
route the request to Consul.~~

This is largely shamelessly stolen from @mpuncel's prior work in configuring an Envoy-compatibile Go HTTP client internally at Square.

This commit adds new http client sends requests to a unix socket
specified by an environment variable in the preparer's pod manifest.
When the environment variable is set in the manifest, outbound requests
to Consul will be sent to the socket.

### DIFF
--- a/pkg/preparer/setup.go
+++ b/pkg/preparer/setup.go
@@ -46,6 +46,7 @@ import (
 	"github.com/square/p2/pkg/uri"
 	"github.com/square/p2/pkg/util"
 	netutil "github.com/square/p2/pkg/util/net"
+	"github.com/square/p2/pkg/util/net/socket"
 	"github.com/square/p2/pkg/util/param"
 	"github.com/square/p2/pkg/util/size"
 )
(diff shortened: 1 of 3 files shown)


pkg/preparer/setup.go
```diff
@@ -161,6 +162,14 @@ type PreparerConfig struct {
 	ArtifactRegistryURL          string                 `yaml:"artifact_registry_url,omitempty"`
 	DockerHost                   string                 `yaml:"docker_host,omitempty"`
 	ContainerRegistryJsonKeyFile string                 `yaml:"container_json_key_file,omitempty"`
+
+	Envoy bool `yaml:"envoy,omitempty"`
+	// EnvoyConsulHost will be set as the http.Request.Host for all Consul
+	// requets when the preparer is configured to use Envoy for communication
```
esuen: typo: requests

spenceral:
Generalized the naming of the new http client package and associated parameters as one that just dials a Unix socket as opposed to an Envoy specific client.

PR #1086, opened 2018-09-10
Check service discovery synced before node swap

spenceral:
Systems that have a delay between changes to P2 pod cluster membership
and changes to the membership of their service discovery and load
balancing systems should not let those two diverge further through node
transfers.

The RC Farm now accepts the ServiceDiscoveryChecker interface, and RC's
will call its IsSyncedToCluster() function to ensure that the service
discovery system is synced enough to support a transfer.

### DIFF
--- a/bin/p2-rctl-server/main.go
+++ b/bin/p2-rctl-server/main.go
@@ -134,6 +134,7 @@ func main() {
 		alerter,
 		1*time.Second,
 		artifactRegistry,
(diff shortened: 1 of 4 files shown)


spenceral:
The travis-ci result isn't syncing, but here is [proof](https://travis-ci.org/square/p2/builds/426886982) that it completed successfully.

PR #1087, opened 2018-09-19
increase TestHookRunWithTimeout from 1 msec to 1 sec

derektamsen:
- Fix issue when timeout is 1 ms. This results in the file being cleaned up before it is fsync() to the filesystem resulting in the temp script missing and thus can't be executed.
- Ensure hook script executes longer than 1 sec.
- Use `/usr/bin/env sh` for the hook script. Using `env` is the standard why of calling environments and platform agnostic. Switching to `sh` because we don't need a full bash environment.

### DIFF
--- a/pkg/hooks/hooks_test.go
+++ b/pkg/hooks/hooks_test.go
@@ -97,11 +97,11 @@ func testManifest() manifest.Manifest {
 }
 
 func TestHookRunWithTimeout(t *testing.T) {
-	timeout := 1 * time.Millisecond
-	sleep := "1" // 1 second sleep to be executed by the script
+	timeout := 1 * time.Second
+	sleep := "5" // Wait longer than hook timeout to trigger failure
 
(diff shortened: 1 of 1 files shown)


pkg/hooks/hooks_test.go
```diff
@@ -97,11 +97,11 @@ func testManifest() manifest.Manifest {
 }
 
 func TestHookRunWithTimeout(t *testing.T) {
-	timeout := 1 * time.Millisecond
-	sleep := "1" // 1 second sleep to be executed by the script
+	timeout := 1 * time.Second
+	sleep := "5" // 1 second sleep to be executed by the script
```
esuen: nit: the comment still says 1 second

derektamsen: Thanks, I fixed the comment to reflect why the number is larger rather than the timeout value

PR #1088, opened 2018-09-25
Discard STDOUT and STDERR in p2-logbridge

spenceral:
Environments that use systemd-journald may be affected by a bug in which
logbridge write()s will fail with EPIPE if systemd-journald restarts. To
avoid this, we can update the STDOUT and STDERR file descriptors to
point at /dev/null.

[Context](https://bugs.freedesktop.org/show_bug.cgi?id=84923)

### DIFF
--- a/bin/p2-log-bridge/main.go
+++ b/bin/p2-log-bridge/main.go
@@ -27,6 +27,11 @@ func main() {
 		os.Exit(1)
 	}
 
+	if err := discardSTDOUTAndSTDERR(); err != nil {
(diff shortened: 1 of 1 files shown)


bin/p2-log-bridge/main.go
```diff
@@ -27,6 +27,11 @@ func main() {
 		os.Exit(1)
 	}
 
+	if err := discardSTDOUTAndSTDERR(); err != nil {
+		logging.DefaultLogger.WithError(err).Error("fatal error discarding STDOUT and STDIN")
```
eam: STDOUT and STDERR

PR #1089, opened 2018-09-28
Add extra Consul headers config

spenceral:
When the preparer is communicating with Consul through a unix socket,
there is now an additional manifest config that will allow the preparer
to add any additional headers to all requests to Consul.

### DIFF
--- a/pkg/preparer/setup.go
+++ b/pkg/preparer/setup.go
@@ -170,6 +170,10 @@ type PreparerConfig struct {
 	// requests when the preparer is configured to dial a unix socket for
 	// communication with Consul
 	ConsulHost string `yaml:"consul_host_header,omitempty"`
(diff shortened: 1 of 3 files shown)


pkg/preparer/setup.go
```diff
@@ -170,6 +170,10 @@ type PreparerConfig struct {
 	// requests when the preparer is configured to dial a unix socket for
 	// communication with Consul
 	ConsulHost string `yaml:"consul_host_header,omitempty"`
+	// AdditionalHeaders will all be set as headers for Consul requests when
+	// the preparer is configured to dial a unix socket for Consul
+	// communication
+	AdditionalHeaders map[string]string `yaml:"additional_headers"`
```
mpuncel: nit: i'd maybe call this `additional_consul_headers`

PR #1090, opened 2018-10-03
add feature to mark platform hooks as required in preparer config

derektamsen:
Adding a feature to add a required list of p2-preparer hooks. The executed hook will be checked against `required_hooks: []` if it fails to see if the `err` should be returned. If the p2-preparer hook is not listed it should retain the previous functionality to log a warning and continue.

Simple preparer config:
```
id: hooks
launchables:
  req-hook:
  opt-hook:
config:
  required_hooks:
  - req-hook
```
This will result in failures to `req-hook` returning an err and `opt-hook` logging but continuing.

### DIFF
--- a/pkg/hooks/hooks.go
+++ b/pkg/hooks/hooks.go
@@ -15,6 +15,11 @@ import (
 	"github.com/square/p2/pkg/manifest"
 )
 
+// RunHookResult holds the hook run result data
+type RunHookResult struct {
+	Err error
+}
+
 func NewContext(dirpath string, podRoot string, logger *logging.Logger, auditLogger AuditLogger) *hookContext {
 	return &hookContext{
 		dirpath:     dirpath,
@@ -25,7 +30,11 @@ func NewContext(dirpath string, podRoot string, logger *logging.Logger, auditLog
 }
 
 // runDirectory executes all executable files in a given directory path.
-func (h *hookContext) runDirectory(hookEnv *HookExecutionEnvironment, logger logging.Logger) error {
+func (h *hookContext) runDirectory(
+	hookEnv *HookExecutionEnvironment,
+	logger logging.Logger,
+	hooksRequired []string,
+) error {
 	entries, err := ioutil.ReadDir(h.dirpath)
 	if os.IsNotExist(err) {
 		logger.WithField("dir", h.dirpath).Debugln("Hooks not set up")
@@ -63,6 +72,15 @@ func (h *hookContext) runDirectory(hookEnv *HookExecutionEnvironment, logger log
 			continue
 		} else if err != nil {
 			h.auditLogger.LogFailure(hec, err)
+
+			// check against a list of required hooks to see if we should return err
+			for _, reqHook := range hooksRequired {
+				if hec.Name == reqHook {
+					logger.WithError(err).Errorf("Fatal error in hook %s: %s", hec.Name, err)
+					return err
+				}
+			}
+
 			logger.WithError(err).Warningf("Unknown error in hook %s: %s", hec.Name, err)
 			continue
 		}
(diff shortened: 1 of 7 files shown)


pkg/manifest/manifest_test.go
```diff
@@ -402,3 +406,32 @@ func TestSetNodeRequirements(t *testing.T) {
...
+
+	Assert(t).IsNil(err, "Should not have erred setting config for test manifest")
+
+	manifest := builder.GetManifest()
+	result := manifest.GetRequiredHooks()
+
+	expectedList := testRequiredHooksList()
+	expected := expectedList[0]
+	for _, hook := range result {
+		Assert(t).AreEqual(hook, expected, "Should contain the required hook")
```
esuen: Probably should be `for i, hook := range result {` then
` Assert(t).AreEqual(hook, expectedList[i], "Should contain the required hook")`

derektamsen: Agreed that something better should be done to validate the full array. I mostly did it as a hack due to the test framework limitations. Let me see if I can improve it by using the standard test framework.

pkg/preparer/orchestrate_test.go
```diff
@@ -565,3 +567,17 @@ func TestPreparerWillAcceptSignatureWhenKeyringIsNil(t *testing.T) {
 		"expected the preparer to verify the signature when no keyring given",
 	)
 }
+
+func TestPrepareHooksFailureRequired(t *testing.T) {
```
esuen: Is this an unrelated test that is being added for coverage? Was there context for this?

pkg/hooks/hooks.go
```diff
@@ -82,14 +99,18 @@ func (h *hookContext) Close() error {
...
 // NB: in the event of a timeout this will leak descriptors
 func (h *HookExecContext) RunWithTimeout(logger logging.Logger) error {
-	finished := make(chan struct{})
-	go func() {
-		h.Run(logger)
+	finished := make(chan RunHookResult)
+	go func(r chan RunHookResult) {
+		err := h.Run(logger)
+		r <- RunHookResult{err}
 		close(finished)
```
esuen: seems better to add `defer close(finished)` after the line `finished := make(chan RunHookResult)`

pkg/hooks/hooks.go
```diff
@@ -82,14 +99,18 @@ func (h *hookContext) Close() error {
 //
 // NB: in the event of a timeout this will leak descriptors
 func (h *HookExecContext) RunWithTimeout(logger logging.Logger) error {
-	finished := make(chan struct{})
-	go func() {
-		h.Run(logger)
+	finished := make(chan RunHookResult)
+	go func(r chan RunHookResult) {
```
esuen: I don't think you need to pass `finished` to this function. It could be a function that takes no params. You already referenced `finished` with the `close(finished)` line in the function.

pkg/manifest/manifest_test.go
derektamsen: `reflect.DeepEqual()` was the function I needed. Updated the test.

pkg/preparer/orchestrate_test.go
derektamsen: @esuen This is for coverage at this point. I found the `if err != nil..` path untested at https://github.com/square/p2/blob/master/pkg/preparer/orchestrate.go#L290. I guess I also just left this in because I initially thought I would need to add the conditional for required hooks in this function. I can remove it if it seems unnecessary now.

esuen: It's fine, can leave this.

pkg/hooks/hooks.go
derektamsen: I added a `defer` however I needed to add it in the channel func. When the defer was outside I would see panics when a hook would timeout stating that the channel was closed.

pkg/hooks/hooks.go
```diff
@@ -82,14 +99,19 @@ func (h *hookContext) Close() error {
...
 func (h *HookExecContext) RunWithTimeout(logger logging.Logger) error {
-	finished := make(chan struct{})
+	finished := make(chan RunHookResult)
+
 	go func() {
-		h.Run(logger)
-		close(finished)
+		err := h.Run(logger)
+		finished <- RunHookResult{err}
+		defer close(finished)
```
esuen: I think this should go at the top of the function.

derektamsen:
Since the design of this changed a bit from using the manifest to switching to the preparer config I am closing the this PR and opening a new one.

PR #1092, opened 2018-10-13
switch docker deps for api and client to moby

derektamsen:


derektamsen:
It seems docker/docker has not really been purged yet and moby/moby has imports with docker/docker.

PR #1097, opened 2018-11-27
add feature to mark platform hooks as required in preparer config

derektamsen:
Adding a feature to add a required list of p2-preparer hooks. The executed hook will be checked against `hooks_required: []` if it fails to see if the `err` should be returned. If the p2-preparer hook is not listed it should retain the previous functionality to log a warning and continue.

Simple preparer config:
```
id: hooks
launchables:
  req-hook:
  opt-hook:
config:
  hooks_required:
  - req-hook
```
This will result in failures to `req-hook` returning an err and `opt-hook` logging but continuing.

### DIFF
--- a/bin/p2-run-hooks/main.go
+++ b/bin/p2-run-hooks/main.go
@@ -71,8 +71,10 @@ func main() {
 		}
 	}
 
+	hooksRequired := []string{}
+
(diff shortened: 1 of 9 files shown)


derektamsen:
New PR for #1090

PR #1098, opened 2018-12-03
Adding readbool method to config

mweissbacher:
I use this method and thought it would be generally useful.

Thanks,
Michael

### DIFF
--- a/pkg/config/config_test.go
+++ b/pkg/config/config_test.go
@@ -40,6 +40,31 @@ func TestReadStringSlice(t *testing.T) {
 	Assert(t).AreEqual("c", s[2], "Third entry c")
 }
 
+func TestReadBool(t *testing.T) {
+	conf, err := readTestFile().ReadMap("bool")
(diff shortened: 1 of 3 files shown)


pkg/config/config_test.go
```diff
@@ -40,6 +40,31 @@ func TestReadStringSlice(t *testing.T) {
...
+	v, err = conf.ReadBool("f")
+	if err != nil {
+		panic(err.Error())
+	}
+	Assert(t).AreEqual(v, false, "Mismatch")
+
+	v, err = conf.ReadBool("default")
+	// Expected error, defaulting to false
+	if err == nil {
+		panic(err.Error())
```
petertseng: careful, since only get here if `err` is `nil`, the `panic` will not contain an error as useful as desired

mweissbacher: Good point. I opened a new PR here: https://github.com/square/p2/pull/1099

PR #1101, opened 2019-01-09
add fields minimum_health, current_health and in_deployment to type p…

esuen:
…odcluster


esuen:
Removed the status fields and only kept MinimumHealth. The other fields can be queried and do not need to be stored here.

esuen:
Change MinimumHealth field to MinHealthPercentage.

PR #1102, opened 2019-01-16
Add MinHealthPercentage parameter to pcstore Create function and set …

esuen:
…default value to 0 for existing calls


derektamsen:
@esuen Sorry, I see the parent is calculating N%

PR #1110, opened 2019-02-11
Ensure session cleanup is done consistently

ausmith:
There were 2 session leaks found:
* When locking to do a node transfer, we were not cleaning up the session if we failed to get the lock.
* After returning to `attemptNodeTransfer()` from `canNodeTransfer()`, we were only unlocking the lock once complete and doing nothing with the underlying session object.

Also had to make a couple changes to improve tests:
* Increase the timeout for go tests as we were consistently exceeding the 20s timeout with the `pkg/rc/*.go` tests.
* Fix up the hello integration in the repo so `go test` would succeed.
* Modify test setup order in TestNodeTransferNoopIfLockHeld to remove surprise errors from the logs when that test does fail.

### DIFF
--- a/pkg/rc/replication_controller.go
+++ b/pkg/rc/replication_controller.go
@@ -276,7 +276,7 @@ func (rc *replicationController) meetDesires(rcFields fields.RC) error {
 
 	ineligible := rc.checkForIneligible(current, eligible)
 	if len(ineligible) > 0 && rcFields.AllocationStrategy == fields.DynamicStrategy {
-		rc.logger.Infof("ineligible nodes: %s found", ineligible)
+		rc.logger.Infof("ineligible nodes: %s found, %v rcID", ineligible, rc.rcID)
 		ok, err := rc.attemptNodeTransfer(rcFields, current, ineligible, MaxAllocateAttempts)
 		if err != nil {
 			return err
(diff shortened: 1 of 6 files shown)


pkg/rc/replication_controller.go
```diff
@@ -744,10 +744,10 @@ func (rc *replicationController) canNodeTransfer(rcFields fields.RC, current typ
 	switch {
 	case consul.IsAlreadyLocked(err):
 		rc.logger.Infoln("skipping node transfer; rc mutation lock is already held")
-		return false, nil, nil
+		return false, unlocker, nil
 	case err != nil:
 		rc.logger.WithError(err).Errorln("could not acquire rc mutatation lock for node transfer")
-		return false, nil, err
+		return false, unlocker, err
```
ausmith: My suspicion is that this is the case we are hitting that leaves us with orphaned locks, but I am not confident that such is the case. The alternative is we have multiple workers all creating locks with the same RC ID.

pkg/rc/replication_controller.go
```diff
@@ -672,14 +672,14 @@ func (rc *replicationController) unschedule(txn *auditingTransaction, rcFields f
 // when a node transfer occurs
 func (rc *replicationController) attemptNodeTransfer(rcFields fields.RC, current types.PodLocations, ineligibles []types.NodeName, allocAttempts int) (bool, error) {
 	ok, unlocker, err := rc.canNodeTransfer(rcFields, current, ineligibles)
+	if unlocker != nil {
+		defer unlocker.Unlock()
+	}
```
ausmith: By moving this `defer` block, we increase the likelihood of unlocking the session if we hit one of our error cases.

esuen: Are you sure the session will be destroyed when the consul session is unlocked? I think it needs to be explicitly destroyed. https://github.com/square/p2/blob/master/pkg/store/consul/session.go#L384

ausmith: Oh, I think you're right and I got stuck on trying to ensure it gets unlocked. Tackling that now.

ausmith:
PR updated with actual fix for session leaks and tests actually prove the issue to be fixed now.

spenceral:
TIL about session leaks. Nice fixes 😅

PR #1111, opened 2019-02-15
added customizable termination grace period for disable scritps

akun1012:
Sometime disable scripts run for 2h blocking p2-preparer to proceed. We should add termination grace period and kill the disabling process when too long time consumed. Default set to 1 hour considering not affect current process, and added `termination_grace_period` field in pod manifest to enable customization.

### DIFF
--- a/pkg/manifest/manifest.go
+++ b/pkg/manifest/manifest.go
@@ -13,6 +13,7 @@ import (
 	"net/url"
 	"os"
 	"path"
+	"time"
 
 	"github.com/square/p2/pkg/artifact"
 	"github.com/square/p2/pkg/cgroups"
@@ -43,6 +44,7 @@ type Builder interface {
 	SetLaunchables(launchableStanzas map[launch.LaunchableID]launch.LaunchableStanza)
 	SetResourceLimits(limits ResourceLimitsStanza)
 	SetNodeRequirements(map[string]string)
+	SetTerminationGracePeriod(seconds int)
 }
 
 var _ Builder = builder{}
@@ -88,6 +90,7 @@ type Manifest interface {
 	Marshal() ([]byte, error)
 	SignatureData() (plaintext, signature []byte)
 	GetNodeRequirements() map[string]string
+	GetTerminationGracePeriod() time.Duration
 
 	GetBuilder() Builder
 }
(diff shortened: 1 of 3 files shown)


pkg/manifest/manifest.go
```diff
@@ -170,6 +174,10 @@ func (manifest *manifest) SetNodeRequirements(nodeRequirements map[string]string
 	manifest.NodeRequirements = nodeRequirements
 }
 
+func (manifest *manifest) SetTerminationGracePeriod(minutes int) {
+	manifest.TerminationGracePeriod = minutes
```
ausmith: Where does the documentation belong that we expect minutes rather than seconds to be passed to us?

akun1012: Not documented. Considering the termination grace period granularity, "minute" seems enough, but "second" provides more flexibility and also provides consistency with k8s termination grace period. Would change use "second".

pkg/pods/pod.go
```diff
@@ -1057,3 +1059,32 @@ func (p *Pod) withTimeWarnings(scriptType string, serviceID string, f func()) {
...
+
+	manifest, err := p.CurrentManifest()
+	if err != nil {
+		p.logger.WithError(err).Errorln("This pod does not have a manifest, cannot determine its config file")
+	}
+	gracePeriod := TerminationGracePeriod
+	if manifest.GetTerminationGracePeriod() != 0 {
+		gracePeriod = manifest.GetTerminationGracePeriod()
+	}
+	p.logger.WithField("grace termination period", gracePeriod).Warnf("The grace termination period is %v", gracePeriod)
```
ausmith: What does `.WithField().Warnf()` do for us? Is `logrus` or `p.logger` the correct logger we use?

akun1012: It add logrus entries to the warning message. Yes, we use pod.logger as a thin wrapper for logrus to provide extendable logs for p2.

pkg/pods/pod.go
```diff
@@ -1057,3 +1059,32 @@ func (p *Pod) withTimeWarnings(scriptType string, serviceID string, f func()) {
...
+	select {
+	case <-doneCh:
+		return
+	case <-time.After(gracePeriod):
+		p.logger.WithFields(logrus.Fields{
+			"launchable": serviceID,
+		}).Warnf("The %s script for %s has been running for %s, exceeded the termination grace period, and has been terminated",
+			scriptType, serviceID, gracePeriod)
+	}
+	return
```
esuen: Returning here won't stop the execution of the disable function.

PR #1112, opened 2019-02-23
modified the way to terminate disable script with commandContext

akun1012:
For hosit launchable, terminate the disable script when grace period is met;
for docker launchable, we do not need to terminate the disable process, just unblock the p2-preparer waiting for the disable script, and proceed killing the container.

Also not only for disable script, this commit enables p2 to have termination grace period on any step.

### DIFF
--- a/pkg/docker/docker.go
+++ b/pkg/docker/docker.go
@@ -114,9 +114,28 @@ type Launchable struct {
 	SuppliedEnvVars map[string]string
 }
 
-func (l *Launchable) Disable() error {
+func (l *Launchable) Disable(gracePeriod time.Duration) error {
+	doneCh := make(chan struct{})
+	errorCh := make(chan error)
+	go func() {
+		l.disable(errorCh)
+		close(doneCh)
+	}()
+
+	select {
+	case <-doneCh:
+	case err := <-errorCh:
+		return err
+	case <-time.After(gracePeriod):
+		return util.Errorf("The disable script for container %s has been running for %s, exceeded the termination grace period",
+			l.LaunchableID, gracePeriod)
+	}
+	return nil
+}
+
+func (l *Launchable) disable(errorChan chan<- error) {
 	if len(l.PreStopCmd) == 0 {
-		return nil
+		return
 	}
 	ctx := context.Background()
 	containerName := l.containerName()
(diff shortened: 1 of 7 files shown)


pkg/docker/docker.go
```diff
@@ -114,9 +114,28 @@ type Launchable struct {
 	SuppliedEnvVars map[string]string
 }
 
-func (l *Launchable) Disable() error {
+func (l *Launchable) Disable(gracePeriod time.Duration) error {
```
ausmith: Why can't the `Disable()` functionality be the same between hoist and docker with only `disable()` differing?

pkg/hoist/hoist_launchable_test.go
```diff
@@ -14,6 +15,8 @@ import (
 	. "github.com/anthonybishopric/gotcha"
 )
 
+var terminationGracePeriod = 1 * time.Hour
```
ausmith: I'm still not thrilled with us defaulting to 1 hour, but I think I lost that battle before I joined 😢

pkg/docker/docker.go
akun1012: Because they are different to be terminated, as shown in the commit description, what's the suggestions?

ausmith: I was wondering if the `Disable()` function in the hoist launchable object could more closely reflect the `Disable()` function in the docker launchable object? Specifically asking around the channel usage in the docker disable function, but not in the hoist disable function.

pkg/hoist/hoist_launchable_test.go
```diff
@@ -428,7 +431,7 @@ func TestDisableWithFailingDisable(t *testing.T) {
 func TestDisableWithPassingDisable(t *testing.T) {
 	// This test's behavior is not dependent on whether the pod is a legacy or uuid pod
 	hl, _ := FakeHoistLaunchableForDirLegacyPod("successful_scripts_test_hoist_launchable")
-	err := hl.Disable()
+	err := hl.Disable(terminationGracePeriod)
```
ausmith: I don't see any tests that are exercising the new termination grace period (and I suggest you use a lower value than 1 hour when you test it 😉 ).

pkg/hoist/hoist_launchable_test.go
```diff
@@ -582,3 +594,12 @@ func assertExpectedServices(t *testing.T, expectedServicePaths []string, executa
 		}
 	}
 }
+
+func ContainsString(test string) func(interface{}) bool {
```
esuen: What is this for?

pkg/hoist/successful_scripts_test_hoist_launchable/installs/testLaunchable_abc123/bin/disable
```diff
@@ -1,2 +1,4 @@
 #!/bin/bash
+echo "sleeping for 2 second to test termination grace period"
+sleep 0.005
```
ausmith: The echo and sleep statements don't line up. I'm good with the test being so fast that the `sleep` can be super short, but can we make the `echo` match the reality?

PR #1117, opened 2019-03-14
rc: don't retry AllocateNodes

spenceral:
Allocating a new node is not idempotent so the call should not be
retried on failure.


ausmith:
Seems reasonable to me, but I'd like another to chime in before adding my approval in case I am not considering some aspect of this change.

PR #1118, opened 2019-03-14
enabled p2-exec to run with env

akun1012:


pkg/p2exec/p2_exec.go
```diff
@@ -40,6 +41,10 @@ func (args P2ExecArgs) CommandLine() []string {
 		cmd = append(cmd, "-e", envDir)
 	}
 
+	for _, env := range args.Envs {
+		cmd = append(cmd, "env", env)
```
esuen: Should "env" be "--env"?

akun1012: No, we are using this way  to modify the environment before executing a p2 hook launchable, and it works.

ausmith: Are these traditional environment variables that are being added or are these truly just additional arguments that happen to be similar in name? Examples of traditional environment variable usage from my perspective:
```
~ export ABC=123
~ command_that_uses_abc

# or
~ API_KEY=xxx my_awesome_command
```

akun1012: the `env` here is for adding the traditional environment variables for the commands.

ausmith: So it is a custom argument that we turn into environment variables?

PR #1120, opened 2019-07-23
Add p2 balancer to audit log

esuen:


pkg/audit/p2_balancer.go
```diff
@@ -0,0 +1,180 @@
...
+	// P2BalancerFailureEvent denotes a pod move failure due
+	// to unrecoverable errors
+	P2BalancerFailureEvent EventType = "P2_BALANCER_FAILURE"
+)
+
+type CommonP2BalancerDetails struct {
+	// P2SpinJobID is a uuid that will be the same for all audit log records
+	// associated with the same pod move. It can be used to match up
+	// "in progress" events with "failure" or "completion" events
+	P2SpinJobID P2SpinJobID `json:"p2_spin_job_id"`
```
ausmith: Is this showing up bleeding too much of our internal pieces to the open source project? What would be an open source user's equivalent? I don't see other mentions of spincycle in the open source project. Do we get around this issue entirely by making it something like `ChangeJobID`?

esuen: This could be acceptable but keep in mind that spincycle is open sourced but this does expose information.

PR #1124, opened 2019-08-08
Update rakefile to use modules

mcpherrinm:
It looks like the switch from Godep in #1119 to go modules missed the Rakefile (used at least in Travis) which still referenced godep.

There's a fairly complicated `task :sync_consul_deps` which surely doesn't work on recent Consul since they don't use govendor anymore.  I haven't touched it.  I suspect it largely isn't needed anymore since both P2 and Consul use go modules now.


bwester:
Oops! Thanks for cleaning this up.

I agree that `sync_consul_deps` isn't going to be useful on recent versions of Consul, but P2 isn't using a recent version: it's still on 0.7.0. The client packages are probably due for an update. If the team wants to update gradually, it'll need the task still. If a jump from 0.7.0 → 1.5.3 is acceptable, then task is definitely no longer useful. That's up to the team.

Depending on that answer, one of two changes should be made:
1. Remove the dead, broken task
2. Fix the task to work with modules

bwester:
Switching on module-aware mode without `-mod=vendor` means the `vendor/` directory is not being used during CI. Could you remove the `vendor/` directory too?

mcpherrinm:
Pushed a 2nd commit that removes vendor/ and the vendoring code fro the Rakefile.

mcpherrinm:
Pushed a third commit that removes the broken sync_consul_deps rake task.

mcpherrinm:
the vendor/ folder is still being used because we're using it in GOPATH mode.  I dropped the commit removing vendoring.

PR #1125, opened 2019-08-09
cgroup: decrease cpu period duration

spenceral:
When processes hit their CPU quota, they are throttled
for the remainder of the time in the configured period.
Pausing pods or launchables for up to 1 second is too
high and could lead to throughput drops.

### DIFF
--- a/pkg/cgroups/cgroups.go
+++ b/pkg/cgroups/cgroups.go
@@ -11,7 +11,7 @@ import (
 )
 
 const (
-	CPUPeriod = 1000000 // one million microseconds
+	CPUPeriod = 100000 // one hundred thousand microseconds
 )
 
 // maps cgroup subsystems to their respective paths


pkg/cgroups/cgroups.go
```diff
@@ -11,7 +11,7 @@ import (
 )
 
 const (
-	CPUPeriod = 1000000 // one million microseconds
+	CPUPeriod = 100000 // one hundred thousand microseconds
```
esuen: Does this change the cpu quote of applications?
i.e. in `pkg/cgroups/cgroups.go` line 52: `quote := cpus * CPUPeriod`

mcpherrinm: it shouldn't -- quota is a fraction of period, so changing the one constant makes both scale appropriately

mcpherrinm:
For reference, docker's configuration options here:

https://docs.docker.com/config/containers/resource_constraints/#configure-the-default-cfs-scheduler

ausmith:
Is there an easy way to rerun the Travis build so we can confirm if it was just a race condition or something needs to be updated in the tests? AFAICT it looks like a timeout was hit.

derektamsen:
@ausmith kicked a new rebuild. You can click details next to the travis build and click the restart build button. You may need to be signed in to travis to see it.

PR #1127, opened 2019-09-10
 Enabled p2-launch to just install pods and skip post active/start

akun1012:
This is for p2-one-shot, which enables bare medal machines, which do not even have hoist allocations, to install pods and do not rely on p2-preparer to be fully functioning.

### DIFF
--- a/pkg/pods/pod.go
+++ b/pkg/pods/pod.go
@@ -189,11 +189,27 @@ func (pod *Pod) Launch(manifest manifest.Manifest) (bool, error) {
 		return false, err
 	}
 
+	err = pod.PostActivate(launchables)
+	if err != nil {
+		return false, err
+	}
+
+	err = pod.BuildRunitServices(launchables, manifest)
+	if err != nil {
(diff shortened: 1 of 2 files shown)


bin/p2-launch/main.go
```diff
@@ -41,6 +41,9 @@ var (
 	requireFile                   = kingpin.Flag("require-file", "If set, the p2-exec invocation(s) written for the pod will not execute until the file exists on the system").String()
 	containerRegistryJsonKeyFile  = kingpin.Flag("container-registry-json-key", "Need to set when trying to launch a pod with docker launchables").ExistingFile()
 	dockerImageDirectoryWhitelist = kingpin.Flag("docker-image-directory-whitelist", "Need to set when trying to launch a pod with docker launchables").Strings()
+
+	// options for p2-one-shot, which is meant to make bare medal machines do not rely on p2-preparer to install apps
```
ausmith: nit: `medal` -> `metal`

PR #1128, opened 2019-09-13
Spliting Launch function

arepraneeth:
We need a case to just install pods but not start it.

I have split pod.Launch into 2 more functions.
1. PostActivate runs post activate for all launchables.
2. StartLaunchables starts up all launchables.
3. Made BuildRunitServices public


akun1012:
lgtm

PR #1129, opened 2019-09-16
p2-one-shot: enable p2-launch to skip post activating and starting a pod

akun1012:


bin/p2-launch/main.go
```diff
@@ -110,12 +113,32 @@ func main() {
...
+		if err != nil {
+			log.Fatalf("Could not fetch launchables from the menifest %s: %s", manifest.ID(), err)
+		}
+		for _, launchable := range launchables {
+			err := launchable.MakeCurrent()
+			if err != nil {
+				log.Fatalf("Failed to create current symlink for pod %s: %s", manifest.ID(), err)
+			}
+		}
+		err = pod.BuildRunitServices(launchables, manifest)
```
ausmith: Did this end up being duplicated code from elsewhere or did we have to write this fresh in the end?

ausmith: Retracting this statement as I read the code.

ausmith: So from the looks of it, `pod.Launch(manifest)` was handling `pod.BuildRunitServices()` before. As such, the main difference in this block seems to be the `launchable.MakeCurrent()` loop. As best I can tell, that is normally handled by the install hook in https://github.com/square/p2/blob/master/pkg/hooks/install.go#L19. If I understand correctly, `pod.Install()` on line 111 actually put the install on the host instead of the install hook doing it. Should we try to unify these 3 logic paths or should we proceed with creating this 3rd possible pod install path?

akun1012: "hooks install" is used when setting up p2-preparer, which install the p2 hooks, and here we are installing pods. `pod.launch(manifest)` includes  `pod.BuildRunitServices()`, but we want the service builder files to be created for p2-one-shot as we skip starting the pods and want to put as much as possible information into the images.

ausmith: Ah, I understand now, we are preloading the host with these services so that they'll start on  next reboot.

PR #1131, opened 2019-09-20
p2-setup-runtime

arepraneeth:
p2-setup-runtime sets up runtime environment for an already installed pod.
It runs hooks, post-activate scripts, builds run scripts for a pod.
It will not start services of a pod.

It is useful for setting up a pod environment on a node where p2-preparer
is not running

### DIFF
--- a/bin/p2-setup-runtime/main.go
+++ b/bin/p2-setup-runtime/main.go
@@ -1,0 +1,184 @@
+package main
+
+import (
+	"os"
+
+	"github.com/sirupsen/logrus"
+
+	dockerclient "github.com/docker/docker/client"
+	"github.com/square/p2/pkg/hooks"
+	"github.com/square/p2/pkg/logging"
+	"github.com/square/p2/pkg/manifest"
+	"github.com/square/p2/pkg/pods"
+	"github.com/square/p2/pkg/types"
+	"github.com/square/p2/pkg/uri"
+	"github.com/square/p2/pkg/version"
+	"gopkg.in/alecthomas/kingpin.v2"
+)
+
+const (
+	fullDescription = `Setup runtime environment of an installed pod
+
+EXAMPLES
+
+$ p2-setup-runtime podManifest.yaml
+
+$ p2-setup-runtime https://artifactory.local/podManifest.yaml
+
+$ p2-setup-runtime --pod-root '/var/pods' https://artifactory.local/podManifest.yaml
+`
+)
+
+var (
+	app = kingpin.New("p2-setup-runtime", fullDescription)
+
+	manifestURI = app.Arg("manifest", "a path to a pod manifest").Required().URL()
+	requireFile = app.Flag(
+		"require-file",
+		"If set, the p2-exec invocation(s) written for the pod will not execute until the file exists on the system",
+	).Default("/dev/shm/p2-may-run").String()
+	podRoot       = app.Flag("pod-root", "Pod root directory").Default(pods.DefaultPath).String()
+	hooksDir      = app.Flag("hooks-dir", "Directory hooks run scripts are located").Default(hooks.DefaultPath).String()
+	dockerUrl     = app.Flag("docker-url", "Docker host url").Default(dockerclient.DefaultDockerHost).String()
+	dockerVersion = app.Flag("docker-version", "Docker Version").Default("1.21").String()
+	tlsCA         = app.Flag("tls-ca-file", "File containing the x509 PEM-encoded CA ").ExistingFile()
+	tlsCert       = app.Flag("tls-cert", "File containing tls cert").ExistingFile()
+	tlsKey        = app.Flag("tls-key", "File containing tls key").ExistingFile()
+)
+
+func dockerClient() (*dockerclient.Client, error) {
+	var client *dockerclient.Client
+	var err error
+
+	if *tlsCA != "" && *tlsCert != "" && *tlsKey != "" {
+		client, err = dockerclient.NewClientWithOpts(
+			dockerclient.WithTLSClientConfig(*tlsCA, *tlsCert, *tlsKey),
+			dockerclient.WithHost(*dockerUrl),
+			dockerclient.WithVersion(*dockerVersion),
+		)
+		if err != nil {
+			return nil, err
+		}
+	} else {
+		client, err = dockerclient.NewClientWithOpts(
+			dockerclient.WithHost(*dockerUrl),
+			dockerclient.WithVersion(*dockerVersion),
+		)
+		if err != nil {
+			return nil, err
+		}
+	}
+	return client, err
+}
+
+func initPod(manifest manifest.Manifest) (*pods.Pod, error) {
+	nodeName, err := os.Hostname()
+	if err != nil {
+		return nil, err
+	}
+
+	fetcher := uri.DefaultFetcher
+	podReadOnlyPolicy := pods.NewReadOnlyPolicy(false, nil, nil)
+	podFactory := pods.NewFactory(*podRoot, types.NodeName(nodeName), fetcher, *requireFile, podReadOnlyPolicy)
+
+	dockerClient, err := dockerClient()
+	if err != nil {
+		return nil, err
+	}
+
+	podFactory.SetDockerClient(*dockerClient)
+	pod := podFactory.NewLegacyPod(manifest.ID())
+	return pod, nil
+}
+
+func main() {
+	app.Version(version.VERSION)
+	kingpin.MustParse(app.Parse(os.Args[1:]))
+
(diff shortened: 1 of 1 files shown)


ausmith:
Why isn't this built into the p2-one-shot work?

Should we explicitly state that docker is not supported with these features and focus on hoist launchables?

arepraneeth:
@ausmith , this is part of p2-one-shot proposal.  This should even work for docker pods. It uses generic pod orchestration.

arepraneeth:
> Why isn't this built into the p2-one-shot work?
> 
> Should we explicitly state that docker is not supported with these features and focus on hoist launchables?

@ausmith , this is part of p2-one-shot proposal. This should even work for docker pods. It uses generic pod orchestration.

ausmith:
Why not bake it into the p2-one-shot binary? Why separate it into its own binary?

arepraneeth:
> Why not bake it into the p2-one-shot binary? Why separate it into its own binary?

p2-one-shot has 2 parts to it, one for image build (to download launchables and make them available) and other for image installation (to setup pod-env by running hooks and post-activate)

bin/p2-setup-runtime/main.go
```diff
@@ -0,0 +1,191 @@
...
+	nodeName, err := os.Hostname()
+	if err != nil {
+		return nil, err
+	}
+
+	fetcher := uri.DefaultFetcher
+	podReadOnlyPolicy := pods.NewReadOnlyPolicy(false, nil, nil)
+	podFactory := pods.NewFactory(*podRoot, types.NodeName(nodeName), fetcher, *requireFile, podReadOnlyPolicy)
+
+	dockerClient, err := dockerClient()
```
ausmith: I think this marks the THIRD way in our code base that we create the docker client?
https://github.com/square/p2/blob/master/bin/p2-launch/main.go#L97
https://github.com/square/p2/blob/master/pkg/preparer/setup.go#L716

Do we know that `dockerclient.NewClientWithOpts()` is the right choice here vs `NewClient` or `NewEnvClient`? The invocations look very similar on the surface.

bin/p2-setup-runtime/main.go
```diff
@@ -0,0 +1,191 @@
...
+	// run post activate
+	logger.Infoln("Running post-activate")
+	err = pod.PostActivate(launchables)
+	if err != nil {
+		logger.WithError(err).Fatalln("Failure running PostActivate for launchables")
+	}
+
+	// build startup services
+	logger.Infoln("Running build runit services")
+	err = pod.BuildRunitServices(launchables, manifest)
```
ausmith: If I'm not mistaken, the p2-launch changes for `no-start` also built the runit services. So is `p2-setup-runtime` expected to operate independently of `p2-launch`?

bin/p2-setup-runtime/main.go
```diff
@@ -0,0 +1,191 @@
...
+	//run hooks BeforeInstall
+	logger.Infoln("Running before_install hooks")
+	err = hookContext.RunHookType(hooks.BeforeInstall, pod, manifest, hooksRequired)
+	if err != nil {
+		logger.WithError(err).Fatalln("Failure running before_install hook(s)")
+	}
+
+	// run hooks AfterInstall
+	logger.Infoln("Running after_install hooks")
+	err = hookContext.RunHookType(hooks.AfterInstall, pod, manifest, hooksRequired)
```
ausmith: You're able to just run the before and after install hooks because we assume the install itself has already been run by `p2-launch`, right?

bin/p2-setup-runtime/main.go
arepraneeth: Earlier versions of initialization methods are now deprecated. so followed suggested way(NewClientWithOpts()) to initialize docker client.

bin/p2-setup-runtime/main.go
arepraneeth: yes, we assume pod is already installed. so we run all hooks as needed

bin/p2-setup-runtime/main.go
arepraneeth: cgroups used for p2 have 'hostname' in them https://github.com/square/p2/blob/master/pkg/cgroups/cgroups.go#L167 , So we need to create cgroups, runscripts only after 'hostname' is available .

akun1012:
please test it in hoisted-p2-preparer. Update the sha in hoisted-p2-preparer, and deploy the  p2-preparer to staging canary nodes.

bin/p2-setup-runtime/main.go
ausmith: Then why is p2-launch also setting up the runit services? Is this a situation where p2-launch shouldn't be doing it or there are cases that p2-setup-runtime needs to overwrite what p2-launch did?

arepraneeth: p2-launch too can try set up run-scripts if host-env is available, running buildRunScripts via p2-setup-runtime will ensure they have proper run-scripts.

arepraneeth:
> please test it in hoisted-p2-preparer. Update the sha in hoisted-p2-preparer, and deploy the p2-preparer to staging canary nodes. Let me know if you need help for the testing.

Done, it is now tested on one of staging nodes.

bin/p2-setup-runtime/main.go
```diff
@@ -0,0 +1,191 @@
...
+
+	dockerclient "github.com/docker/docker/client"
+	"github.com/square/p2/pkg/hooks"
+	"github.com/square/p2/pkg/logging"
+	"github.com/square/p2/pkg/manifest"
+	"github.com/square/p2/pkg/pods"
+	"github.com/square/p2/pkg/types"
+	"github.com/square/p2/pkg/uri"
+	"github.com/square/p2/pkg/version"
+	"gopkg.in/alecthomas/kingpin.v2"
```
akun1012: group the imports via goimports

bin/p2-setup-runtime/main.go
```diff
@@ -0,0 +1,191 @@
...
+
+	// build startup services
+	logger.Infoln("Running build runit services")
+	err = pod.BuildRunitServices(launchables, manifest)
+	if err != nil {
+		logger.WithError(err).Fatalln("Failure building run scripts")
+	}
+
+	// run AfterLaunch hooks
+	logger.Infoln("Running after_launch hooks")
```
akun1012: why we are running after_launch hooks before lauching(starting) a pod, this part is useless

ausmith: I didn't think about this closely on my initial review, but I now have a similar question to Zac. What is the goal/desire for running after hooks? If the application isn't running, are there any after hooks that we care about? If we don't run after hooks, what are the potential ramifications?

bin/p2-setup-runtime/main.go
arepraneeth: file was already formatted using `goimports`. Is there any change  you are expecting ?

bin/p2-setup-runtime/main.go
```diff
@@ -0,0 +1,191 @@
+package main
+
+import (
+	"os"
+
+	"github.com/sirupsen/logrus"
```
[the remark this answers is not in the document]
arepraneeth: file was already formatted using goimports. Is there any change you are expecting ?

bin/p2-setup-runtime/main.go
arepraneeth: running all hooks is a p2 promise, in here we do all of the pod.Launch  except for starting of pod services.
P2-preparer runs after_launch hooks irrespective of return status from `pod.StartLaunchables`
https://github.com/square/p2/blob/master/pkg/pods/pod.go#L202 ,
https://github.com/square/p2/blob/master/pkg/preparer/orchestrate.go#L582-L607 

so i felt it is good to run post_launch hooks even if you do not start services rather than completely skipping them. This closely mimics how p2-preparer would have set the pod.

arepraneeth: I added all hooks that are part of p2-preparer's 'launch' process, after_uninstall is not part of it so did not add it here. Yes prune images hooks does not make any impact for p2-one-shot project. All I wanted is to keep expectations of p2-setup-runtime pod to be same as p2-preparer's way of doing things. p2-preparer runs after_launch hooks even if startup of services fail, that means it is okay to run after_launch hooks even if we do not start services.

ausmith: I agree with @akun1012 that we shouldn't be doing the after_launch hooks since we are not launching. If a future after_launch hook were added that depends upon the application being up and running, then p2-one-shot would get stuck on this step, which would probably cause some madness in the imaging process (where we want to use this).

arepraneeth: okay, Updated pr.

PR #1132, opened 2019-09-24
refined p2-install-hook

akun1012:
tested and works

### DIFF
--- a/bin/p2-install-hook/main.go
+++ b/bin/p2-install-hook/main.go
@@ -1,12 +1,12 @@
 package main
 
 import (
-	"fmt"
 	"log"
 	"os"
 	"path/filepath"
 
 	"gopkg.in/alecthomas/kingpin.v2"
+	"gopkg.in/yaml.v2"
 
 	"github.com/square/p2/pkg/artifact"
 	"github.com/square/p2/pkg/auth"
@@ -15,6 +15,7 @@ import (
 	p2manifest "github.com/square/p2/pkg/manifest"
 	"github.com/square/p2/pkg/osversion"
 	"github.com/square/p2/pkg/pods"
+	"github.com/square/p2/pkg/preparer"
 	"github.com/square/p2/pkg/types"
 	"github.com/square/p2/pkg/uri"
 	"github.com/square/p2/pkg/version"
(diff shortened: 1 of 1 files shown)


bin/p2-install-hook/main.go
```diff
@@ -49,11 +50,19 @@ func main() {
 
 	if *isPreparerManifest {
 		config := manifest.GetConfig()
```
ausmith: Shouldn't this line be using the alias of `p2manifest.GetConfig()` for clarity?

bin/p2-install-hook/main.go
```diff
@@ -49,11 +50,19 @@ func main() {
...
-		hooksManifest, ok := config["hooks_manifest"]
-		if ok {
-			manifest = hooksManifest.(p2manifest.Manifest)
-		} else {
-			fmt.Printf("Error to read hooks manifest from P2-Preparer manifest: %v", manifest)
+		preparerConfigBytes, err := yaml.Marshal(config)
+		if err != nil {
+			log.Printf("Unalbe to marshal preparer config: %v", err)
```
ausmith: nit: `Unalbe` -> `Unable`

bin/p2-install-hook/main.go
akun1012: no, manifest is not an alias, it's the real manifest we are using

akun1012:
@ausmith Have tried to use p2manifest.FromBytes(), which not gonna work, as normal p2 manifest does not have hooksManifest. We need to unmarshal the config to preparerConfig, which contains. the hooksManfiest. This approach is the neat one using shared functions.

bin/p2-install-hook/main.go
```diff
@@ -49,11 +50,19 @@ func main() {
 
 	if *isPreparerManifest {
 		config := manifest.GetConfig()
-		hooksManifest, ok := config["hooks_manifest"]
-		if ok {
-			manifest = hooksManifest.(p2manifest.Manifest)
-		} else {
-			fmt.Printf("Error to read hooks manifest from P2-Preparer manifest: %v", manifest)
+		preparerConfigBytes, err := yaml.Marshal(config)
```
ausmith: Following further inspection of `manifest.GetConfig()`, it looks like it already does the `Marshal` and `Unmarshal` steps that you are attempting here. https://github.com/square/p2/blob/master/pkg/manifest/manifest.go#L185-L208 The only difference between the two is that you are using `preparer.UnmarshalConfig()` instead of `yaml.Unmarshal()`. As best I can tell, the reason you want to explicitly use the `preparer.UnmarshalConfig()` over just a `yaml.Unmarshal()` is because it adds the hostname, consul address, hooks directory, and pod root if those values are missing from the manifest. https://github.com/square/p2/blob/master/pkg/preparer/setup.go#L265-L291

As you specifically want the hooks data and you do not want that info to be missing, and you also don't want the configuration to change unexpectedly, should we be using `manifest.Marshal(config, err)` instead of just `yaml.Marshal(config)`? https://github.com/square/p2/blob/master/pkg/manifest/manifest.go#L400-L413

akun1012: There is no `manifest.Marshal(config, err)` there is just manifest.Marshal() ([]byte, err) which cannot be used in this case, as manifest is manifest, we want to marshal `config`.

The reason why we need to marshal and unmarshal twice is we need to convert `config` to `preparerConfig` which is unique to have hooksmanifest. 

The only way to avoid this is to add `manfiest.getPreparerConfig()` method, which is not optimal as normal manifest does not preparer config, and this is not frequently used.

PR #1134, opened 2019-09-30
Changes to cpu period and cpu quota

akun1012:
set cpu period back to default, and added safe layer for changing cpu period to avoid invalid input holding back pods to be launched.

### DIFF
--- a/pkg/cgroups/cgroups.go
+++ b/pkg/cgroups/cgroups.go
@@ -11,7 +11,7 @@ import (
 )
 
 const (
-	CPUPeriod = 100000 // one hundred thousand microseconds
+	CPUPeriod = 1000000 // one hundred thousand microseconds
 )
 
 // maps cgroup subsystems to their respective paths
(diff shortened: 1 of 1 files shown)


arepraneeth:
looks good to me

pkg/cgroups/cgroups.go
```diff
@@ -49,17 +49,23 @@ func (subsys Subsystems) SetCPU(name CgroupID, cpus int) error {
 		return UnsupportedError("cpu")
 	}
 
-	quota := cpus * CPUPeriod
-	if cpus == 0 {
-		// setting -1 here will unrestrict the cgroup, so the period won't matter
-		quota = -1
```
ausmith: So we were setting it to negative 1 in the code but not actually writing it to the file system? Oops.

akun1012: no, we were writing to the file system as -1 for quota if there was no cgroup for cpu

esuen:
Why does setting cpu quota to -1 first prevent the invalid argument error? What was p2-preparer trying to set the cpu quota as?

ausmith:
@esuen setting the quota to -1 is the equivalent of saying "unlimited". This is the equivalent for our purposes of resetting the CPU quota before modifying the CPU period. Assuming I understand how this was behaving before and is behaving now:

Before:
Set the local quota variable, if no CPU quota is provided, set it to -1 to represent unlimited
Ensure that the cgroup folder exists and create it if not
Write the CPU period cgroup information if value is different from what is on the filesystem
Write the CPU quota cgroup information if value is different from what is on the filesystem
done

After:
Ensure that the cgroup folder exists and create it if not
Write -1 as the CPU quota cgroup information if not already unlimited
Modify the local quota variable if no CPU quota provided
Write the CPU period cgroup information if value is different from what is on the filesystem
Write the CPU quota cgroup information if value is different from what is on the filesystem
done

This theoretically avoids a situation where the period can be a different order of magnitude from the quota, causing the invalid argument errors we saw. If quota is -1 when changing the CPU period though, the system will happily let the period be whatever we want. Then we can change the quota to be the actual `period * CPUs desired`.

PR #1135, opened 2019-10-01
corrected the annotations

akun1012:


pkg/cgroups/cgroups.go
```diff
@@ -11,7 +11,7 @@ import (
 )
 
 const (
-	CPUPeriod = 1000000 // one hundred thousand microseconds
+	CPUPeriod = 1000000 // one second
```
esuen: Is there reason why we don't want to specify the unit is in microseconds?

akun1012: nope, the change is not about changing unit from microsecond to second, it's about correcting it from "one hundred thousand microseconds" to "one million microseconds" which is one second.

esuen: Maybe put "one million microseconds which equals 1 second".

ausmith: +1 to @esuen's suggestion of making it explicit.

akun1012: polished

PR #1137, opened 2019-11-13
refresh some checked fields in addPods

zapman449:


pkg/rc/replication_controller.go
```diff
@@ -330,6 +330,15 @@ func (rc *replicationController) addPods(rcFields fields.RC, current types.PodLo
...
 		}
+
+		// since significant time may have passed since these values were instantiated,
+		// get updated values each iteration, and leverage those
+		eligible, err := rc.eligibleNodes(rcFields)
+		if err != nil {
+			return err
+		}
+		possible = types.NewNodeSet(eligible...).Difference(types.NewNodeSet(currentNodes...))
+		possibleSorted = possible.ListNodes()
```
ausmith: Why are we repeating lines 300 and 304 here?

PR #1138, opened 2019-11-13
Working to fix Travis builds

zapman449:
debug info, prepurge runit as cargocult, and then install its systemd unit file too

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,7 @@
 language: go
 
 go:
-  - 1.12
+  - "1.12"
 
 env:
   - RACE=-race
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -1,22 +1,26 @@
 language: go
 
+dist: trusty
```
zapman449: Lock the build to trusty aka ubuntu 14.04.  Technically went EOL on 2019-04, but Travis supports it and 12.04 for the time being.

.travis.yml
```diff
@@ -1,22 +1,26 @@
...
+
 go:
-  - 1.12
+  - "1.12"
```
zapman449: Travis suggests forcing this to be a string, else you use golang 1.1 (via float interpolation to a string)

PR #1141, opened 2020-01-09
Remove old bin

esuen:


ausmith (approved):
We don't use these bin files from here anymore, nuke 'em.

bwester:
I'm curious, what prompted this removal of functionality?

esuen:
@bwester Everything here is being moved internally.

